PR #208, opened 2016-02-15
Fix #200: Dont end up in a locked ui state when somebody leaves.

### LINKED ISSUE #200: Time compression locked after contract expiry dialog
Originally reported by: **moorkai (Bitbucket: [moorkai](http://bitbucket.org/moorkai), GitHub: [moorkai](http://github.com/moorkai))**

Hello everyone.

So, I've freed Alma, and just went to visit Tixa. It would be all fine and dandy if Conrad wouldn't complain that his contact is expiring. I can renew contract for other chaps, but it lets me only to dismiss Conrad ... which I don't want to. He's an invaluable addition to my small team ;P.
(plus, his rant breaks time compression, which results in an unusable game afterwards - locked options, laptop and tactical button)

Running v0.12.352 Strac on Windows 7 x86 with JA2 Gold 1.12 files.

Also, I've attached the game save. Just ramp the time compression to trigger the second bug mentioned.

Best regards,
moorkaï

- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/200

selaux:
Bitbucket Comments:

@lynxlynxlynx:
While the usefulness of locking is questionable there, wouldn't it be better to just unlock it afterwards? Or does the execution of the dialog happen in the next tick?

@selaux:
I couldn't dig anything particularly insightful from the refactoring history.

### DIFF
--- a/Build/Strategic/Strategic_Merc_Handler.cc
+++ b/Build/Strategic/Strategic_Merc_Handler.cc
@@ -385,9 +385,8 @@ void MercsContractIsFinished(SOLDIERTYPE* const pSoldier)
 		{
 			//Send the merc home
 
-			InterruptTime( );
+			InterruptTime();
 			PauseGame();
-			LockPauseState(LOCK_PAUSE_09);
 
 			// Say quote for wishing to leave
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
I have no objections to merging this. :+1:

Issue #179, opened 2015-10-04
Strange corspes in strange places after auto-resolve


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

----------------------------------------

Enemy attacked the SAM site in D2. They came from D3. Biff and Flo retreated to let the enemy duke it out with the militia. After the battle, corpses ended up in strange places.
Here:
![SCREEN001.jpg](https://bitbucket.org/repo/6zkzG5/images/1943756554-SCREEN001.jpg)
And here:
![SCREEN002.jpg](https://bitbucket.org/repo/6zkzG5/images/261349096-SCREEN002.jpg)

It's a minor thing, but I guess that any items on enemy corpses are also lost if the corpses is "underwater"...

----------------------------------------
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/179

derSteFfi:
shouldn't we first merge this? :)

lynxlynxlynx:
Oops, you just ff-ed it to master. Done.

PR #209, opened 2016-02-15
Fix corpses and items appearing in water.

selaux:
Bitbucket Comments:

@lynxlynxlynx:
Judging from the diff, this is a prime candidate for refactoring — both chunks are the same (the same stop-spot typo suggest c/p).

@selaux:
True, I'll check what I can do.


derSteFfi:
i guess this would fix #179 ?

selaux:
Yes. Sorry for not refactoring. I tried twice but actually the two functions do two slighly different things, so I was not able to.

PR #212, opened 2016-02-15
fixed #199 (rare boxing crash)

### LINKED ISSUE #199: 2nd boxfight - segfault on darren giving money
Originally reported by: **misanthropos (Bitbucket: [misanthropos](http://bitbucket.org/misanthropos), GitHub: [misanthropos](http://github.com/misanthropos))**

This savegame is just before giving money and having a SECOND fight on the same day. If Bull wins the fight, Darren hands out the money and the game segfaults. Just to make sure the bug is can be found if something is amiss before the fight.
The first fight per day works - only the second never does.

- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/199

lynxlynxlynx:
I've traced the variable and nothing particularly interesting happens. The money transaction works normally and is dumped to the ground just fine with a full inventory.

### DIFF
--- a/Build/Tactical/Handle_Items.cc
+++ b/Build/Tactical/Handle_Items.cc
@@ -2308,6 +2308,10 @@ void SoldierGiveItemFromAnimation( SOLDIERTYPE *pSoldier )
 	// Get items from pending data
 
 	// Get objectype and delete
+	if (!pSoldier->pTempObject) {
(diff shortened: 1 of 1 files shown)


selaux:
Could you find out why `pTempObject` is null at this spot? Is the function called twice or something like that?

lynxlynxlynx:
It looked ok to be NULL, since it is just a _temporary_ object for the move. It is more likely the call was extraneous or broken. The only difference I can see between the second and third run is Kingpin's interjection though.

Issue #206, opened 2016-02-08
Proposal: Migrate to SDL2


lynxlynxlynx:
Originally reported by: **Stefan Lau (Bitbucket: [selaux](http://bitbucket.org/selaux), GitHub: [selaux](http://github.com/selaux))**

----------------------------------------

Hi,

I have made efforts to migrate ja2-straciattella to SDL2 here: https://github.com/ja2-stracciatella/ja2-stracciatella/tree/sdl2. It's not done yet but most of it works.

## Rationale ##
* SDL 1.2 is outdated
* SDL2 brings an accellerated 2D render API (which we would then have to use)
* SDL2 has better fullscreen support (scaling!)
* SDL2 allows compilation for Android and iOS

## Current Status ##
* Builds with SDL2
* Rendering works
* Audio works
* Mouse works
* Gameplay works (minus undiscovered bugs)

## Issues ##
* [x] Does not build on OS X (because of `SDL_putenv`)
  - no3btnmouse is not necessary anymore on mac, ALT + LMB works fine, CMD + LMB is used as right-click alternative
* [x] Keyboard works only partially
* [x] Graphical corruption when scrolling upwards on tactical map
* [x] Mouse moves outside of fullscreen
* [ ] Not using dirtyRectangles anymore but doing a full screen refresh every time (not sure if that is an issue)

## Moving forward ##
Please feel free to test this branch and tell me your oppinions about the migration and if you want squish the bugs listed above.

----------------------------------------
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/206

lynxlynxlynx:
It seems the packaging stuff (_build) currently relies on SDL1, so that would have to be changed too. I think SDL2 is common enough now that we don't need to support both versions.

derSteFfi:
+1
builds fine here, is playable. the upscrolling in tactical is quite unfancy indeed. didn't test keyboard but audio and video work as described, mouse too.
I also like that mouse moves out of fullscreen, i am on a multiscreen setup and this way i can switch to other windows. With master (SDL1) i can't even alt+tab out of the game, now i can.

derSteFfi:
[...]

PR #216, opened 2016-02-16
Update to SDL2 

selaux:
Issue: #206


sgp/Input.cc
```diff
@@ -184,11 +190,23 @@ void MouseButtonUp(const SDL_MouseButtonEvent* BtnEv)
 	}
 }
 
+void MouseWheelScroll(const SDL_MouseWheelEvent* WheelEv) {
```
SergeL: :nail_care: New line before `{` missing

sgp/Video.cc
```diff
@@ -142,27 +124,64 @@ void InitializeVideoManager(void)
...
+	           SCREEN_HEIGHT,
+	           PIXEL_DEPTH,
+               Rmask,
+               Gmask,
+               Bmask,
+               Amask
+    );
+
+    if (ScreenBuffer == NULL) {
+        SLOGD("Topic", "SDL_CreateRGBSurface for ScreenBuffer failed: %s\n", SDL_GetError())
```
derSteFfi: i see that a ';' is not neccessary here, but every other place in the codebase puts semicolons there.
The #define for SLOGD adds it, but i would prefer consistency across the code

selaux: :+1: the topic seems wrong to me as well...

sgp/Video.cc
```diff
@@ -136,8 +136,9 @@ void InitializeVideoManager(void)
                               g_window_flags);
 	GameRenderer = SDL_CreateRenderer(g_game_window, -1, 0);
 	SDL_RenderSetLogicalSize(GameRenderer, SCREEN_WIDTH, SCREEN_HEIGHT);
+  SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "linear");
 
-		SDL_Surface* windowIcon = SDL_CreateRGBSurfaceFrom(
+  SDL_Surface* windowIcon = SDL_CreateRGBSurfaceFrom(
```
derSteFfi: this looks like you introduced spaces here instead of tabs...

selaux:
Allright now I'm able at least to build this branch on every OS. Currently I'm working on how the SDL2 framework needs to be distributed for OS X (that seems to have changed)

selaux:
Finally everything seems to be working now. Please test. Attached are the releases I built with the SDL2 dependencies. I split the update of the dependencies to a separate PR, so actual changes that needed to be done are more emphasized here (otherwise GitHub wouldn't render the full diff).

[sdl2-releases.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/454044/sdl2-releases.zip)

misanthropos:
I just tried the amd64 version on Ubuntu. It seems libsdl2 has a different name but still got installed and I could start it. However I cant continue the SaveGame from before the lawfabric.

error [queen command] assertion failed in campaign_types.h line 11

has anybody checked if one can play through a sector?

lynxlynxlynx:
That would mean an invalid sector number, but it's likely to just be the aggressive asserts thing we have now in master (they're not ifdefed out anymore). So the question is, do you have the same problem with sdl1 or not?

selaux:
We really need either to playtest a lot for the asserts, or disable them before the next release 😉

lynxlynxlynx:
I'd just disable them for non-debug builds. Too many permutations for it to be live for the masses.

lynxlynxlynx:
I've built it and tested a bit all around — seems to work fine. :)

lynxlynxlynx:
Yay! Good job. :)

derSteFfi:
Agree! Great one :)

PR #223, opened 2016-02-19
Fix222: Render Credits Screen correctly

derSteFfi:


selaux:
I would not close PRs like this. You can do fixups to the commits by doing an [interactive rebase](https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase-i) As long as nobody branches away from your branch and you notify everyone about the force push this should be allright.

derSteFfi:
interactive rebase is nice, i'll try to remember this :)

PR #224, opened 2016-02-19
Fixed #222: render the credits Screen correctly

### LINKED ISSUE #222: Credit Screen: double misplacement
Credits background is not centered but top left, text starts running from bottom left to top left see screenshot.

bildschirmfoto von 2016-02-19 12-36-13

derSteFfi:


selaux:
Codewise this looks good to me :+1: . Two things I noticed though when trying the branch:
- The copyright notice is still rendered in higher resolutions but is not there in 640x480
- If you move the mouse below where the scrolling text appears, artifacts of it (the text) are rendered below the mouse cursor

derSteFfi:
I noticed the copyright notice too. Question is how to handle it... I think we could just remove it? Maybe this is a seperate issue...

The mouse cursor thing is tricky. The cursor expands the rendering area and the 'hidden' text is revealed... Thinking into it, but I don't have an obvious solution.

selaux:
We could try to do a full-screen update when entering the credits screen, which will probably remove the copyright notice.

derSteFfi:
question is whether we remove the notice at all?

selaux:
I would add this to 0.15 since it basically only improves stuff (except for the minor glitches). The copyright can also stay imho.

PR #229, opened 2016-02-22
Externalize lib boost

derSteFfi:
This makes the use of boost dependant on LOCAL_BOOST_LIB, like for SDL.
Beforehand i did some formating, sorting and whitespace changes to the Makefile, purely cosmetic and split into a sperate commit.
Tested the build with both settings, works fine. boost 1.60 actually saves us around 100kb in the application, which is quite nice i think :)

We should discuss which version shall be the default, local or host. (+1 for host ;) )

Two (minor?) glitches:
1. 'make clean' doesn't clean libboost now. 'make clean LOCAL_BOOST_LIB=1' does however. Since I see no reason to patch the old, local version of libboost and hence rebuild it, i think this is OK.

2. I have no idea how this is supposed to work on Windows. I wasn't able to find a nice command line to print the paths we need to include (like for sdl, 'sdl-config') so I hardcoded them. If anyone has a solution for this, I am happy to hear about it.


derSteFfi:
cool, travis hasn't got boost installed :)

selaux:
You can add it to the `apt-get install` line in `.travis.yml`.

selaux:
:+1: For this, not sure why we didn't use the systems boost library before.

derSteFfi:
the ubuntu host we are running travis on installes boost 1.46, for some components 1.48, which is even older than the version we deliver with the repo...

derSteFfi:
travis is running on Ubuntu 12.04 LTS... for 14.04 LTS they moved to boost 1.54. Our repo delivers 1.53, upstream is 1.60.
How do we solve this?

lynxlynxlynx:
Do we depend on anything only present above 1.46? If not, then we don't need to care about the version mismatches.

selaux:
I agree, lets not care for that. It should run with any relatively current boost version.

derSteFfi:
seems so. one of the unit tests fails because of an old bug fixed in boost 1.50
http://www.boost.org/users/history/version_1_50_0.html
(Filesystem -> 3rd Point, Fix 5118)

selaux:
We could also use ubuntu 14.04 on travis by adding

```
sudo: required
dist: trusty
```

to the `.travis.yml`. But then we would need to think about what to do when releasing a new version.

derSteFfi:
how does trusty change that?

lynxlynxlynx:
We could also move to the container-based recipe (faster), but I'd wait for all the build system work, since you can't use sudo there and are therefore limited in the types of packages you can get. Pretty sure it uses trusty by default too, so that one unit test wouldn't be problematic.

.travis.yml
```diff
@@ -1,11 +1,15 @@
 language: cpp
 
+sudo: required
```
selaux: Is this necessary? I think installing packages also works without sudo.

selaux:
I think we also have to adapt the release builds. Can you try running `make build-releases` to verify that this still works?

derSteFfi:
well, needs adjustments, currently on it. Problem is, the boxes are built around 12_04 and boost is present in version 1.46 then again. So, one unit test would fail and we would end up with an older version of boost than the one we currently ship with the repo...

derSteFfi:
so, unit tests do not run when building releases so it at least doesn't fail (and porting the buildmachines to 1404 is not much effort).
But building the win release on linux fails for me (currently it's running through with vagrant but that takes a while... if I build manually with 'make build-win-release-on-linux' it fails with
```
===> CXX Build/Credits.cc
In file included from /usr/i686-w64-mingw32/include/c++/6.1.1/string:52:0,
                 from src/ContentManager.h:4,
                 from Build/Credits.cc:24:
/usr/i686-w64-mingw32/include/c++/6.1.1/bits/basic_string.h: In Funktion »std::__cxx11::wstring std::__cxx11::to_wstring(int)«:
/usr/i686-w64-mingw32/include/c++/6.1.1/bits/basic_string.h:5559:46: Fehler: »WINvswprintf« ist kein Element von »std«
   { return __gnu_cxx::__to_xstring<wstring>(&std::vswprintf, 4 * sizeof(int),
                                              ^~~
/usr/i686-w64-mingw32/include/c++/6.1.1/bits/basic_string.h:5559:46: Anmerkung: empfohlene Alternative:
In file included from sgp/Types.h:8:0,
                 from Build/Credits.h:5,
                 from Build/Credits.cc:1:
sgp/SGPStrings.h:26:5: Anmerkung:   »WINvswprintf«
 int WINvswprintf(wchar_t* s, size_t n, const wchar_t* fmt, va_list arg);
     ^~~~~~~~~~~~
```

not really sure what this is :)

selaux:
I've rebased with current master and updated the buildboxes. They build fine. Maybe your issue had to do with a non-clean build.

derSteFfi:
seems so. thans for the updated buildboxes :)

lynxlynxlynx:
On a separate note, how much do we really need boost? It's a heavy dependency ...

selaux:
Currently we mostly use boost for the cross-platform filesystem implementation. Not sure if there is a well-maintained alternative though.

derSteFfi:
we could look into that, but boost would also be a candidate for a cross-platform wchar_t alternative, i think.

PR #232, opened 2016-02-24
Fix #78: Double sample rate of sound samples if necessary.

### LINKED ISSUE #78: Some sounds sound wrong.

After having played v0.12.189 for some time, I realized, that some sounds sound different than they used to in vanilla JA2 or v1.13.
Prominent examples are: "walking on dirt" and "opening laptop".

This is an issue I already discussed with BirdFlu, almost 2 years ago when he was working on SMP (which was based on Stracciatella).

The discussion thread on BP forums can be found here: [http://www.ja-galaxy-forum.com/board/ubbthreads.php/topics/291952/1.html](http://www.ja-galaxy-forum.com/board/ubbthreads.php/topics/291952/1.html)

I don't expect anyone to read the whole thing. Here's my initial report:

*"1) The sound when opening/closing the laptop sounds wrong somehow. Much brighter and "cheaper" then I'm used to.
2) The same is true for other sounds: starting of players turn (interrupt), target out of range, walking on dirt (roads), opening doors."*

After some discussion, BirdFlu wrote: 

*"I stopped the time for Hitman walking from one side to the otherside of the MAP (a9):
- 56 seconds for Ja2 Gold
- 47 seconds for Ja2 Stracciatella*

*Either this speed difference is the cause for 1) or it is the SDL library. Since this effect also occurs on Stracciatella, I'm quite sure that it is nothing that i could have introduced."*

Imo, this is a severe issue!

- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/78

selaux:
This is a proposed fix for #78. Since `44100hz` waves sample rates were halved when loaded, I doubled the sample rates of the `11025hz` samples. This seems to fix the affected sounds. A few notes though:

- I could not test with any stereo samples since i could find none. If anybody could point me to some I could test if they work as well (8 and 16bit)
- There is a lot of  `.WAV` file parsing going on in `SoundMan.cc` not sure if we want to keep this in the long run. SDL also has functions to load WAV files, which I tried to use but in the end failed. But this might be a better approach.

### DIFF
--- a/sgp/SoundMan.cc
+++ b/sgp/SoundMan.cc
@@ -623,6 +623,84 @@ static BOOLEAN HalfSampleRate(SAMPLETAG* const s)
 	return TRUE;
 }
 
+static BOOLEAN DoubleSampleRate(SAMPLETAG* const s)
+{
+	UINT8 bitcount = s->uiFlags & SAMPLE_16BIT ? 16 : 8;
+
+	SNDDBG("SMPL \"%s\" %dbit from %uHz to %uHz\n", s->pName, bitcount, s->uiSpeed, s->uiSpeed * 2);
+
+	UINT32 const n_samples = s->n_samples * 2;
+	void*  const ndata     = malloc(n_samples * GetSampleSize(s));
+	if (ndata == NULL) return FALSE;
+	void*  const odata     = s->pData;
+	if (bitcount == 16)
+	{
+		INT16*       const dst = (INT16*)ndata;
+		const INT16* const src = (const INT16*)odata;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Looks fine to me. 
I also agree SDL{,_mixer} would be a cleaner solution eventually. I doubt the original code is endian-safe, which could be problematic for portability. On the other hand SDL surely handles it fine.

derSteFfi:
Code looks good. Sound sounds fine :)
Maybe jasmine can retest with this patch to have a direct comparison again?

PR #240, opened 2016-02-28
Fix `make build-releases`

selaux:
`make build-releases` referenced old paths. Now it works again. I tested the resulting builds on the following configurations and they seem to run fine. Building the Mac release still needs to be tested though and we could consider creating an .app directory structure instead of a common build so it integrates better for OS X users. Then we are ready to do a release.

- Ubuntu 12.04 i386
- Ubuntu 14.04 i386
- Ubuntu 14.04 x64
- OS X 10.11
- OS X 10.9
- Windows 10 x64

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -690,7 +690,7 @@ build-win-release-on-linux:
 # end users to setup the library.  LGPL allows static linking if we
 # provide source codes for our application.
 MACOS_SDL_STATIC=./_build/lib-SDL-devel-1.2.15-macos-i386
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
:+1:

derSteFfi:
can someone confirm confirm the mac build so we can merge this?

selaux:
I also built a release for mac now, but could only test it on the same mac I built it on. So if anybody could test the attached zip it would be great.

[ja2-v0.14.xx-macos.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/153383/ja2-v0.14.xx-macos.zip)
.

selaux:
Allright tested it on another mac and found and fixed an issue with older OS X versions.

PR #248, opened 2016-03-07
Remove unused functions

derSteFfi:
part of some cleanups. Also adds the -Wunused-function flag as default for compilation.


selaux:
Nice work cleaning up :+1:. I double checked all the functions, some of them are still used either in comments or in some special setup.

- `SmoothUndoMapTileTerrain`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Editor/Editor_Undo.cc#L415)
- `DoesCharacterHaveAnAttitude`, `DoesCharacterHaveAPersonality`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Laptop/IMP_Attribute_Selection.cc#L107) we have to check what is vanilla behavior
- `HandleCurrentMercDistortion`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Laptop/Mercs.cc#L775)
- `EliminateAllMercs`, `EliminateAllFriendlies`: Still used [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Strategic/Auto_Resolve.cc#L2578) and [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Strategic/Auto_Resolve.cc#L2565)?
- `ScheduleAirRaid`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/remove_unused_functions/Build/GameScreen.cc#L833)
- `DoneFadeOutActionSex`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Tactical/Interface_Dialogue.cc#L2177)
- `MagazineClassIndexToItemType`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Tactical/Items.cc#L2944) can probably be deleted
- `BillyBlocksDoorCallback`, `HookerInRoom`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/TileEngine/Explosion_Control.cc#L2120), maybe it's possible to reenable this functionality :wink: 
- `CalcTranslucentWalls`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Tactical/Overhead.cc#L2138)
- `TrashMapTile`: Referenced in FIXME and in code [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Editor/EditScreen.cc#L2601)
- `ExecuteCopyCommand`, `ExecutePasteCommand`, `ExecuteCutCommand`: Referenced in code [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Utils/Text_Input.cc#L726)
- `findObjectCaseInsensitive`: Can this really be removed if you have a case sensitive fs (didnt test yet)
- `GetFreeSpaceOnHardDrive`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/sgp/FileMan.cc#L496)
- `convertDialogQuotesToJson`: Referenced in comment [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/sgp/SGP.cc#L492)
- `joinInRectangle`: Is used when `EXPENSIVE_SDL_UPDATE_RECT` is set

About some of the commented out lines I'm not sure if that is not missing functionality from vanilla. So we have to make sure to keep track of that before removing them. Please double check.

derSteFfi:
You are right of course :)
i did some research:
- SmoothUndoMapTileTerrain was already commented out on tron's initial import and SmoothAllTerrainTypeRadius looks like its replacement that when undoing Terrain changes not only changes the tile that was changed but also smoothes the direct environment. So I think this (including the comment) can be safely removed.
- DoesCharacterHave: can someone test on a vanilla version? I can't run one here on Linux...
- HandleCurrentMercDistortion was also commented out on initial import. Its purpose looks like adding random distortion effects on mercs pics during the phone call for recruiting. This also happens on current master, so there seems to be a replacement for this.
- EliminateAll*: it's hidden behind a JA2BETAVERSION and has been since initial import. If we ever want to use the functionality of this again (some of this looks helpful for developing, much like a very precise and powerful cheat mode) then we should probably keeps those 2.
- ScheduleAirRaid is tricky. It was used there to trigger the arrival of mercs in Omerta at game start. It is not exported in Air_Raid.h so only visible in Air_Raid.cc. It's also static. I tested making it non-static, promoting it in Air_Raid.h and calling it in GameScreen.cc, nothing changed when arriving in Omerta (no crash, mercs get dropped off, everything normal). So I guess this is an artifact an can be deleted (together with the comment).
- DoneFadeOutActionSex: it is assigned to a variable where it is referenced (one line above from where you linked) but is a void function. I'm not much into SDL but SetPendingNextScreen could be the replacement for the other calls commented out there, at least in my imagination :). Anyway, it is commented out since the initial import. I would tend to deleting both function and comment
- MagazinClassIndexToItemType: agree. we should delete the comment too. gennady put this there in 1aa7ef5e with log entry 'Expernalize all hardcoded weapon information'
- BillyBlocks.., HookerInRoom: Yes, I also read that somewhere else. Tried around a bit, it's doable (at least compiles with ~5 minutes of work), I'll open an Issue for this (if not already present) and put in the place to find it.
- CalcTranslucentWalls was also commented out in the initial import. Since Walls are translucent if your see stuff behind it, i think this one got replaced and can be safely deleted.
- TrashMapTile: Seems like someone was trying to enhance the editor but didn't finish testing. If someones into that, this is good knowledge. I don't use the editor, so I can't advise on whether this is useful. Anyway, I'd restore this function for now.
- Execute*Command look like gennady tried to implement copy and paste for text. Does that make sense in JA2? He deactivate it himself when he commited it. How is this going to be with SDL2?
- findObjectCaseInsensitive is in all occurences hidden behind ```#ifdef CASE_SENSITIVE_FS``` and this is not defined (wouldn't be reported unused otherwise). Since I myself and some others here are working with sensitive filesystems and don't experience problems, I think we can actually remove this.
- getFreeSpaceOnHardDrive: looks like some logic trying to prevent errors when the hard drive runs full. But i actually don't get why to check that, fprintfs will fail when hard drive is full, isn't that kind of enough? Or do we actually want to check for at least x MB of free space for temp and log files?
- convertDialogueQuotesToJson: looks strange. converting the proprietary data file to json at the start of the game seems like the wrong direction. wouldn' it be more useful to read from json files here? Anyway, a function converting from and/or to json could be helpful for externalizing resources, so I'll restore this one.
- joinInRectangle: Right. I'll hide this behind the switch than, at least for now. Maybe you SDL guys take a look at it when we are moving to SDL 2.

selaux:
Looks good now, can we merge?

Issue #164, opened 2015-09-26
Change name of the user folder?


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

At the moment, relevant data like savegames etc. is stored in a folder called "JA2". Maybe, it would be more appropriate to rename the folder to "JA2 Stracciatella", since the folder is exclusively related to JA2 Stracciatella?
Of course, this is only a trivial aspect...

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/164

selaux:
Maybe we could also make `ja2.ini` a json-file so we can drop the ini parsing library.

derSteFfi:
That sounds great. We should tackle this as soon as v0.15 is out.

lynxlynxlynx:
As for the folder name, I'd keep it, except perhaps on windows. On linux it should also be moved eventually, to respect XDG paths.

Issue #60, opened 2013-07-27
Rework debug logging


lynxlynxlynx:
Originally reported by: **Gennady Trafimenkov (Bitbucket: [gennady](http://bitbucket.org/gennady), GitHub: [gennady](http://github.com/gennady))**

----------------------------------------

There is a lot of code like this littered all over the place:

```
			#ifdef JA2BETAVERSION
				ScreenMsg( FONT_MCOLOR_LTYELLOW, MSG_ERROR, L"Turning off AI...
```

```
				if (gfTurnBasedAI)
				{
					DebugAI( String("Ending turn for %d ...
```

```
					DebugMsg( TOPIC_JA2, DBG_LEVEL_3, String("Attack busy count ...
```

It would be great to develop a unifed scheme of dealing with such debug messages.

Requirements:
- logging to console
- logging to a file
- ? logging to the screen
- ability to turn logging off and on at run time, not compile time

----------------------------------------
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/60

derSteFfi:
so, as for now, we use SLOG (which was implemented by gennady). We discussed this topic in the meeting too. We agreed we have 2 partly seperated issues: rework debugging logs to use a unified scheme and decide which scheme it should be.

For now, we agreed SLOG offers everything we want to have. So whereever we touch files/functions, we should switch the logging to SLOG. I will start to convert the debug messages behind JA2[TEST,BETA]VERSION to SLOG calls with an appropriate level. This way we also solve part of #10 without actually hurting anything.

selaux:
This can be closed now right?

PR #250, opened 2016-03-07
Logging rework

derSteFfi:
still to go: a ton of fprintf's and all the DebugMsg calls.

As always, comments welcome. :)


derSteFfi:
OK, the next load of work will be the functions NumMessage, AINumMessage, NameMessage and PopMessage. None of those is implemented somewhere, they are all hidden behind config switches or commented out.
Question is: Shall we (I :) ) delete those or port them to slog? I do have the feeling we are spamming around a bit if we port all logs that are currently present to slog. Most of them are much too detailed, I think, which is probably why they're commented out or switched off...
If we decide for deletion, I would apply this to AIPopMessage and AINameMessage too. (Already ported them...)

lynxlynxlynx:
The new on-by-default calls could even be a performance problem if we don't find a way to noop them (just speculating though). We could mitigate it by defining the topics globally, having filtering capabilities and outputing only a few by default (the rest if specified via a -param).

derSteFfi:
Well, IF this becomes a problem, we can still get this done by renaming SLOG* functions to something like internalSLOG* and do one or both of the following steps:
- have something like
```C
#ifndef DEBUG_LEVEL_FOO
     #define SLOG(foo) (0)
#else
     #define SLOG(foo) internalSLOG(foo)
#endif
```
- and/or filter by topic during runtime (this then atleast is a jump to slog, a check, and a jump back)

I still don't  think this will be problem.

derSteFfi:
Anyway, I'm gonna delete PopMessages, NumMessages, NameMessages and their AI realtives. Since they are dead anyway and they're debugging really really detailed.

lynxlynxlynx:
d0f4cd76 is a bit harsh. There's no problem with debugging aides being hidden by switches (hard or soft) and I bet AI debugging would be easier with the messages in place. Am I missing something?

derSteFfi:
yes, it's harsh. You're not really missing something, but the functions are not implemented anywhere anymore. And there is/was the fear that converting all of these detailed messages spam SLOG quite much...
In the original Issue (#60), gennady stated that runtime switching of debug levels and/or topics is preferable over compile-switching. That's why I mostly removed the #ifdef's.

lynxlynxlynx:
But we're losing information this way. We could create a wrapper that uses a define (otherwise a noop), until such a time comes, that we have runtime toggling. In general I of course agree with Gennady, since excessive ifdefs make too much code not be compiled by default and we know how fast that can bitrot.

derSteFfi:
Well, can we do something like this:
```C
#ifdef DEBUG_AI
   #define SLOGD(DEBUG_TAG_AI,x)    internalSLOGD(DEBUG_TAG_AI, x)
   #define SLOGE(DEBUG_TAG_AI,x)    internalSLOGE(DEBUG_TAG_AI, x)
   ...
#else
   #define SLOGD(DEBUG_TAG_AI,x)    0
   #define SLOGE(DEBUG_TAG_AI,x)    0
   ...
#endif
```
I don't wether this works, but it would solve everything (with one a bit complicated looking header)

lynxlynxlynx:
Sure, it's a tiny bit prettier that way instead of having it in the body.

derSteFfi:
So, current process is as follows:
- There is still a lot of logging and messaging scattered around the code. For example, there still is LOSDebug, Assert(Msg), ScreenMsg. For Assert I am not sure how to handle it, it seems pretty nice the way it is and is defined to 0 if debugging is disabled. For the other two, I'm in the process of replacing them with SLOG calls. This also hits ScreenMsg where i could imagine some want to keep those. For me, on-screen-debugging doesn't seem desirable but I am open to your opinions.
- The second big problem is how to handle the massive amount of SLOG calls. defining them out depending on the TAG doesn't work (I tried). Right now, my idea would be to move all the ```#define DEBUG_TAG_X``` to slog.h, put in Numbers not Strings and have a table that stores whether for tag x on level y logging is to be done to both file and temrinal(3), only file(2), only terminal(1) or not at all(0). This way we could also switch on/off logging very precisely at runtime, since we can alter the table of course. A call to SLOG which should not be logged could immediately return like this:
```C
if (!gPriorityTable[x][y])   return;
```
Overhead should be pretty small then. Any suggestions?

lynxlynxlynx:
I'd do it this way:
a. enum for all the topics, but as bits, not natural numbers for values. The slog calls would use these.
b. list for strings (their topic names)
c. enum for logging type bits as you suggested
d. one variable to store both a&c (eg. a<<8|c), which is then exposed to the config and can easily be compared to everywhere

derSteFfi:
I'm not sure I understand. what information is kept in the respective bits? Can you be more verbose? :)

lynxlynxlynx:
a and c above refer to the enumeration of my list. So there would be no per-stream choice, just dumping everything selected to file, only to console, nothing or both. With fixed topics files can easily be filtered with a tool of one's choice, so I think we can skip the extra complexity of per-topic choice of destination (leaving on/off).

derSteFfi:
ok, I get it this far. :)
I still don't see how to press this into an enum. We have x different debug 'topics' and 4 'severity' levels for each of them. Assuming a 'fixed' (or maximum) number of topics, I could imagine having a flag variable that contains one bit for each topic saying 'logging on/off'. We could also have 4 of these bits per topic to switch the different levels on or off. Your c than contains the global information for the stream to log to (which would only take 2 bit). My concern right there is that I do not see the performance improvement made by this (probably my fault :) ) and that this inherits a limit to the number of topics we can support.
I'm still trying to think of a solution that allows a fast return from SLOG if this call is 'unwanted'. Maybe I just don't see where you are headed :)

lynxlynxlynx:
Aren't the severity levels shared, the same? I was not trying to improve performance over your solution, just make it easier for the user to set (one special variable instead of a whole array of them). The SLOG call passes the topic, so you can quickly check against it like with your array — shift away the severity bits and you can do a normal bitwise AND.

The enum for bits is just a way to avoid #defines, so it's easier to work with them while in gdb. It would look like TOPIC_X = 0, TOPIC_Y = 1, TOPIC_Z = 2, TOPIC_W = 4 and so on. The other one for severity would look like you suggested. And a separate list for the strings. I don't think merging would make much sense, at best the topic bits and names in a std::map.

derSteFfi:
ah, OK, I get it now. like enable the topic independently and enable the different levels. A log message will then appear when level AND topic are set.

OK, makes sense to me. (At least right now i can not think of a situation where i want our logging to be different) I'll start implementing it tomorrow to have a proposal we can discuss.

The other Topic: What about ScreenMsgs and/or Asserts?

lynxlynxlynx:
ScreenMsg is used for ingame stuff, so can't be removed. If you meant just the debug calls, I guess they could be changed to slog and their own topic, but would it be worth the effort? Assert on the other hand looks like a prime candidate for slog.

derSteFfi:
for ScreenMsg i mean stuff like this:
```C
#ifdef JA2BETAVERSION
				ScreenMsg(FONT_MCOLOR_LTYELLOW, MSG_ERROR, L"AI %d got error code %d from HandleItem, doing action %d... aborting deadlock!", pSoldier->ubID, iRetCode, pSoldier->bAction);
#endif
```
Some of them are more complex, like this:
```C
#ifdef JA2TESTVERSION
		if (!q)
		{
			if (!gfTriedToLoadQuoteInfoArray[ubNPC]) // don't report the error a second time
			{

				ScreenMsg( MSG_FONT_RED, MSG_DEBUG, L"ERROR: NPC.C - NPC needs NPC file: %d.", ubNPC );
				gfTriedToLoadQuoteInfoArray[ubNPC] = TRUE;
			}
			// error message at this point!
		}
#endif
```
But it still is easy to port them. When I talk about ScreenMsg in logging context, I of course refer to debug output, not in-game messages :)

For Assert, it's defined as:
```C
#define Assert(a)       (a) ? (void)0 : _FailMessage(NULL, __LINE__, __FILE__)
#define AssertMsg(a, b) (a) ? (void)0 : _FailMessage(b   , __LINE__, __FILE__)
```
(or 0'ed out if SGP_DEBUG and FORCE_ASSETS_ON are 0) we should easily be able to port this to SLOG.

derSteFfi:
Sooo... for now this works. And is very verbose :)
Tell me what you think. By the way, even with full Debugging on (both console and file set to DEBUG) I can't feel a difference in performance.

sgp/Debug.h
```diff
@@ -3,57 +3,17 @@
...
-
-#define Assert(a)       ((void)0)
-#define AssertMsg(a, b) ((void)0)
-
-#endif
-
-
-#ifdef SGP_DEBUG
+#define Assert(a)       (a) ? (void)0 : SLOGW(DEBUG_TAG_ASSERTS, "Assertion failed in %s, line %d", __FILE__, __LINE__)
+#define AssertMsg(a, b) (a) ? (void)0 : SLOGW(DEBUG_TAG_ASSERTS, "Assertion failed in %s, line %d:\n%s", __FILE__, __LINE__, b)
```
lynxlynxlynx: This doesn't abort anymore, so it's not really an assert. Why not just add "; abort()" to the macros?

derSteFfi: You're right, it did before. I'll add it.

derSteFfi: well, not that easy. putting a ';' into the macro makes some asserts fail that are used within for or if loops. do we need this abort?
If so, I think adding a flag to SLOG calls like ```fAbortAfter``` is the easiest way to go.

lynxlynxlynx: Yeah, needs braces or a normal if-clause.

derSteFfi: braces don't work. At least not the way i tried. I don't see an if working either when Assert is inside an if clause... would end up like this:
```C
if(if(!ptr) SLOGW...)
```
I don't think this works...

derSteFfi: found a solution. see aaaa5fb

lynxlynxlynx: yeah that's fine, but I'd blame the initial poor usage.

derSteFfi:
sooo, the list to solve is getting shorter here. actually right now, there is nothing on it. Any opinions or demands? :) What would need to be added to merge this?

Build/Strategic/Strategic_Movement.cc
```diff
@@ -3178,13 +3178,6 @@ BOOLEAN GroupWillMoveThroughSector( GROUP *pGroup, UINT8 ubSectorX, UINT8 ubSect
 			{
 				dy = -1;
 			}
-			else
-			{
-				Assert( 0 );
-				pGroup->ubSectorX = ubOrigX;
-				pGroup->ubSectorY = ubOrigY;
-				return TRUE;
-			}
```
lynxlynxlynx: This checks out, but please split such commits in the future.

lynxlynxlynx:
Aren't we missing logging to file now?

Build/Strategic/Strategic_Movement.cc
derSteFfi: Yeah, I thought it was somehow related and was trying to prevent ugly merge conflicts...

derSteFfi:
no, we don't. ja2.log gets placed in the directory where the game is started from and contains the same messages that are shown on the console (Warnings and Errors).
Two Issues for that: Maybe we should set the logging to file to Info (and above) per default. Second is we should maybe place the logfile somewhere else (~/.ja2/ or similar).
For the standard log levels, its just the init call to changeso pretty trivial. For the place of the logfile, FileMan offers a lot but we should decide where we want to have it. For developers like us, it could be enerving to have the log somewhere else than the source. For unexperienced users, it might be better to have the logfile where the executable is so they do not need to search for it.

lynxlynxlynx:
I'm fine with it being in the game dir, but of course you can unhardcode it and add it to game.json.

derSteFfi:
@selaux: your opinion? For me, less complexity is the greater benefit here, and we can retackle externalization when we get to #164 
If you both are ok with the way it is now, of course :)

selaux:
I aggree, lets keep it in the game directory, we can think about changing the location in #164.

derSteFfi:
then i guess, we can merge this. or shall we wait until after 0.15?

lynxlynxlynx:
Let's wait, it's a big change.

lynxlynxlynx:
Shall we just merge this now? It's too big for a second look from me.

Delaying too much will only cause more conflicts ...

derSteFfi:
as soon as 0.15 is out like really out, we should imho. same with sdl2 and maybe externalize libboost etc.  Bigger stuff should go in early to have a reasonable testing phase after the merge and before release.

I'm going to make this mergable until today afternoon.

lynxlynxlynx:
0.15 is out already. :)

PR #252, opened 2016-03-08
Mem man cleanup

derSteFfi:
Stumbled across this when trying to rework logging. Situtation on Master right now is we have 3 possible Memory-Backends:

- "for release versions" we use a pretty standard wrapper around stdlib
- "_DEBUG" switches on an old implementation that Tron originally marked as to do. It logs some stuff but is broken right now (semms he added the release version to not depend on this one)
- "EXTREME_MEMORY_DEBUGGING" turns on a beast that logs everything that happens with Memory

In this PR, I deleted the _DEBUG variant. The extreme version still compiles but question is whether we need it. I think the widely used stdlib implementation is the way we should go. In the wrapper, we throw exceptions on bad allocations and for me this seems reasonable and enough.

In conclusion, I would vote for removing the extreme version too. We can add more logging to the 'release' variant of course, if anyone wants to. Did anyone ever use the extreme version to debug memory issues?

### DIFF
--- a/sgp/MemMan.cc
+++ b/sgp/MemMan.cc
@@ -14,13 +14,6 @@
 #include "Debug.h"
 #include "slog/slog.h"
 
-#undef _DEBUG // XXX TODO
-
-#ifdef _DEBUG
-//#	define DEBUG_MEM_LEAKS // turns on tracking of every MemAlloc and MemFree!
-#endif
-
-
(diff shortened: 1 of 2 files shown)


selaux:
Looks good to me, keeping broken code seems unneccessary.

PR #256, opened 2016-03-11
Fix #76 by reenabling the brothel

### LINKED ISSUE #76: Behavior in San Mona's brothel

The brothel doesn't seem to work as intended. When talking to Madame for the first time, she does nothing but give you access to her establishment, allowing you to go to Maria and rescue her without paying any money.

It's not possible to "use" the service.

I'm not quite sure, if the brothel was somehow censored in English JA2 Gold?
I guess, that this is an issue not introduced by Stracciatella. Maybe, Stracciatella could revert this back to normal (=working) behavior, though?

I've attached a savegame just before talking to Madame for the first time.

- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/76

derSteFfi:
only little adjustments besides reenabling the code. Tested, worked.

### DIFF
--- a/Build/JAScreens.cc
+++ b/Build/JAScreens.cc
@@ -474,6 +474,7 @@ ScreenID SexScreenHandle(void)
 	{
 		// Load face....
 		guiSMILY = AddVideoObjectFromFile(INTERFACEDIR "/luckysmile.sti");
+		PlayJA2StreamingSampleFromFile(SOUNDSDIR "/sex.wav", HIGHVOLUME, 1, MIDDLEPAN, NULL);
 
 		// Init screen
 		bCurFrame = 0;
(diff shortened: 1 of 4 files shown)


selaux:
Works for me as well. One glitch I noticed is that the sound file wasnt playing properly. I could fix that by moving the `PlayJA2StreamingSampleFromFile` call [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/72b4f78f387da327b29f50a736aa67cc50570cad/Build/JAScreens.cc#L476). But we have to check whether everything regarding the brothel still works properly, plus the missing resources mentioned in the bug.

derSteFfi:
Codewise, the proposed patch does the same. The npc file differs from mine (english gold version), but that could be localisation specific. Looking into this right now.

derSteFfi:
so, I tried the npc file proposed with the patch, it didn't change anything. Sound is still distorted though. But moving it as you proposed actually solves it. Commit incoming.

lynxlynxlynx:
I guess that's it then. :)

PR #259, opened 2016-03-12
Fix remaining credits issues

selaux:
- Fixed glitches by restoring background rectangle below scrolling text after rendering text
- Credits screen now has a background mouse region with the default cursor


lynxlynxlynx:
Looks good.

derSteFfi:
Mouse cursor works fine. No artifacts while hovering over mentioned region.

derSteFfi:
works fine.

Issue #258, opened 2016-03-12
What to do with `EXTREME_MEMORY_DEBUGGING`


selaux:
This is a follow-up for #252.

"EXTREME_MEMORY_DEBUGGING" turns on a beast that logs everything that happens with Memory.

Lets decide if we want to keep it, replace it or remove it.

lynxlynxlynx:
I'm sure this was helpful when developing the game, but we have better tools if we ever need to lower the memory footprint or debug leaks (eg. from valgrind). So I'm for removal. Unlike AI, this area has little potential for changes too.

derSteFfi:
I'm also pro delition. Looks like an overkill and very unlikely to be useful

PR #262, opened 2016-03-16
Remove extreme memory debugging

derSteFfi:
discussion: #258 
this PR is for the case we agree on deleting it.


selaux:
Already agreed I think. BTW thanks for doing all these cleanup PRs :+1:.

PR #263, opened 2016-03-16
Remove dead inline assembly

derSteFfi:
this gets rid of old inline assembly that was ported by tron (maybe others too). All of this code was translated to C and wasn't touched for very long. Still, there is assembly remaining that is deactivated but unported. For now, clean this up.


selaux:
+1 for removing the assembly stuff. The changes look good to me. But for the next time it would be nice splitting the code reformatting from the deletion of lines so it's easier to review. Github cant even display the files :smile:.

derSteFfi:
You're right :)
I will try to split commits a bit more so they don't move above the threshold for displaying them.

lynxlynxlynx:
go for it.

derSteFfi:
only things left are blitters and stuff in libs. so blitters :).

PR #264, opened 2016-03-16
Implement uninplemented stuff

derSteFfi:
I started to tackle the last ~20 UNIMPLEMENTED functions. Most of them are blitters, but surprinsingly we do not seem to use them. Whenever an unimplemented function is called and actually has the define UNIMPLEMENTED there, the game should chrash with an error about the unimplemented function. I never saw that happen. Any of you ever experienced this?

However, implemtenting them doesn't hurt for now, i think. Those functions that I deleted because they are unused actually appear nowhere across the codebase (I grep'ed).


lynxlynxlynx:
I didn't look into the blitter, but the general idea is fine. Anything TODO0000 related going away is good.

derSteFfi:
So, there are 6 blitters left: ...BufferIntensify['','Z','ZNB','Clip','Zclip'] and ...BufferOutlineShadowClip.
The last one has a similiar one already implemented and should be easy to do, the intensify family has none of it's members implemented and I need a rest from asm :)

derSteFfi:
I've been thinking and actually would like to leave this as it is right now. The stuff in sgp/ is an actual improvement and could be useful for the 0.15 release, the remaining blitters are marked as UNIMPLEMENTED and can stay like this until the next of us (or me again) has motivation to disassemble again. I'm still tired of the blitters for now so this won't be resolved by me in the next time.

lynxlynxlynx:
@derSteFfi are you willing to fix the conflicts (I guess from logging) and then merge this?

derSteFfi:
sure. I'll get on it tomorrow.

PR #265, opened 2016-03-20
Rename Font to SGPFont

selaux:
- Font is very generic
- Font is prone to naming conflicts


derSteFfi:
makes sense, looks good to me.

PR #266, opened 2016-03-21
Fix #253 by eventually invalidating gpBattleGroup in RemoveGroup

### LINKED ISSUE #253: Crash after AI vs AI battles.
Getting a random crash after AI vs. AI battles.  Not sector specific and occurs maybe once every 10-20 battles.  Crashes cleanly to desktop without anything being logged or error.sav created.  So far unable to reproduce from savegame.  Last entry in log is always the radarmap for the sector the battle is occuring in. 

Battle resolves without issue.  After clicking the 'Done' button the screen goes completely black for a split second and then crashes to desktop.  Sorry I can't be any more vague.

derSteFfi:
This is more of a workaround, but it removes SEGFAULTs so I think it's worth it.

### DIFF
--- a/Build/Strategic/Auto_Resolve.cc
+++ b/Build/Strategic/Auto_Resolve.cc
@@ -67,6 +67,7 @@
 #include "Debug.h"
 #include "UILayout.h"
 #include "WeaponModels.h"
+#include "slog/slog.h"
 
 #ifdef JA2BETAVERSION
 #	include "Cheats.h"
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Looks good. You can now also remove the same code at:
Build/Strategic/Auto_Resolve.cc:400

derSteFfi:
Yes, we could. Question is, can we eliminate one of two functions that call RemoveGroup at all? Anyway, invalidating gpBattleGroup will never ever take place where you pointed it out and thus can be removed.

lynxlynxlynx:
I don't follow, one of two? I see many more callers.

derSteFfi:
I meant the two conccurring i mentioned in the corresponding issue

derSteFfi:
So, I added another check and logging. We should keep an eye on this, but I think this is the best we can do for now.

LonCJ:
Thanks for this fix.  It's working perfectly for me.

PR #267, opened 2016-03-22
resolve TODO0011

derSteFfi:
The proposed solution is the only one that makes sense. The only use of this in Soldier_Control.cc shows a lot of similarity to gubAnimSwatSpeeds which looks quite the same


selaux:
:+1:

PR #268, opened 2016-03-22
Resolve todo0017

derSteFfi:
Both calls do exactly the same. I tested by printing the towns loyalty with both functions to check for differences:
dirtywrite


selaux:
:+1: Looks good.

PR #270, opened 2016-03-25
[WIP] Stracciatella launcher

selaux:
Implements the launcher using fltk
- [x] Launches game correctly on linux
- [ ] Launches game correctly on OS X
- [ ] Launches game correctly on Windows
- [ ] Has Tests :smile:


selaux:
This now also includes a local dependency in `_build/lib-fltk` to be able to build on OS X / Windows. It adds another 300K LOC though (does not show correctly in github, only in git). Another approach would be to build it from a tarball using cmake's `ExternalProject`. Could not make that work yet though... Do you think it would be the better option?

lynxlynxlynx:
Indeed, that is huge and would just pollute our repo. If ExternalProject ends up irredeemable, how about just spawning a subprocess to do the dirty work?

Ideally the launcher would be written with sgp, but that'll probably remain unfeasible for a long time.

Manuel-K:
Should we create a new repo for the launcher until we can create on using our own toolkit?

Manuel-K:
We would have to remove the FileMan dependency, but I think this would be a step in the right direction. I don't see any chance to finish this in the near future if we do not make the launcher completely standalone.

selaux:
FYI: I think we could even continue having it here. The issue I had (and why I didn't continue) is that there are no prebuild versions of fltk for Windows, macOS and I didn't want to add the full fltk source code to the repo...

Manuel-K:
Ok. I've rebased this branch. Now I have to replace the MicroIni stuff with corresponding rapidjson calls.

I think I'll push my changes to a different branch first.

selaux:
Superseeded by #548

PR #271, opened 2016-03-27
Fix #261: Helicopter Shadow stays visible until helicopter has left t…

### LINKED ISSUE #261: Helicopter shadow disappears too early on high resolutions
At the start of the game in Omerta, the shadow of the helicopter disappears midflight if a higher resolution than 640x480 is chosen.

RinneT:
Fix #261: Helicopter Shadow stays visible until helicopter has left the sector map


RinneT:
Hi all, first of all I think it's awesome that you continue Gennady's (equally awesome) work on ja2-stracciatella. I was really thrilled to see it being alive again.

Regarding this patch:
This is the simple solution that it is in line with the current implementation of the helicopter animation.
The helicopter shadow does now stay visible until it leaves the sector (tested with Resolution that shows the whole sector).

I find it a bit ugly though. Every animation step is saved in an array. It might be an option to refactor it to get a more generic approach. E.g. the array only defining the single steps and controlling the animation steps with counters. That would also allow for dynamic adaption to the screen size.

That would mean quite a lot of work for a component that just works as is with the maximum sensible resolution already, though.

Input welcome.

derSteFfi:
Welcome! :) We're all quite happy, this is still alive.
I'd say there is no need for refactoring since this is only done once per game, but the way it is now, the helicopter takes a fixed time to disappear even if not visible (on smaller resolutions). This feels like a needless wait. You can interrupt this with ESC, but I don't think newbies will know that. So maybe we should find a way to model this depending on the screen size.

lynxlynxlynx:
Ship it. :+1:

PR #272, opened 2016-03-27
Issue#194 running without moving

RinneT:
Changed behaviour to the following:
Condition:
Merc is running at the beginning of a turn and has enough AP to continue his path.
Player doesn't order merc to continue to move.

In that case the Merc goes to an idle stance as soon as the user ends the turn.


RinneT:
I think this is actually the cleanest way to implement it.
There's also a StopSoldier function in Soldier_Control.cc, however using that allows you to move your merc during the ai turn.

Build/Tactical/TeamTurns.cc
```diff
@@ -202,6 +202,11 @@ void EndTurn( UINT8 ubNextTeam )
 		FOR_EACH_IN_TEAM(s, gTacticalStatus.ubCurrentTeam)
 		{
 			s->bMoved = TRUE;
+			// Cancel merc movement if continue path was not used
+			if ( CheckForMercContMove(s)) {
```
derSteFfi: I think (at least it looks like) we agreed on newlines before '{'.

derSteFfi:
Hey. Glad you found us. Currently testing, but looks fine at first glance.

derSteFfi:
Sooo, this works if the Mercs is selected when you end the turn. Reason is this line in ```CheckForMercContMove```:
```C
	if (s != GetSelectedMan()) return FALSE;
```
Maybe we can extract the checks we need from CheckForMercContMove and do them directly?

Build/Tactical/TeamTurns.cc
lynxlynxlynx: I prefer the terser variant, but we should just follow whatever is there already or it will get messy.

RinneT:
Thanks a lot for the feedback. That happens when I don't test my changes properly.
Moved the relevant code directly into the function and changed the bracket for style compliance.

derSteFfi:
Much better :) Still, I observed this only working for certain situations. I could not figure out the pattern, but the source showed you're using guiPathingData to obtain the mercs direction. Thats the global copy for painting the path under the mouse cursor etc.
If you use s->usPathingData, which is the copy of the mercs actual path, it works fine (tested here).

RinneT:
Very well noticed, thanks for the hint! :+1:  I was briefly wondering why I'd use a global variable for individual mercs but then didn't think about it anymore when it just worked.
Thinking about that I found out when it fails:
Merc A has two AP left and is moved first. To do his next step he needs 3 AP.
Merc B has one AP left and needs two AP to do his next step. He is moved last.
Now all mercs will be tested against the 2AP from pathing data.

It's more obvious when mercs are having different stances / going stealth.

What I'm wondering:
The function PtsToMoveDirection asks for an INT8.
usPathingData[x] is INT16
guiPathingData[x] is INT32

derSteFfi:
well, you are right (of course :) ). It doesn't seem senseful to have more than 8 bit there. PtsToMoveDirection calls DirectionInc and this checks for the value to be between 0 and 7. I'm trying around with this, but i'd say it is a seperate thing, for now, we should merge the bugfix.

PR #273, opened 2016-03-28
Issue#261 helicopter shadow

RinneT:
Thanks a lot for the feedback and the merge.

Very good point regarding the duration of the intro scene.
I reworked the solution to take that into account and think that resulted in nicer code as well.

I defined a counter that repeats the animation for a defined set of pixels.
STD_SCREEN_X is the (Screen Size - 640) / 2. The divisor by 7 seems to work well and adds a new frame roughly every 15 pixels above 640.

Only width matters here, I can't think of any situation where the height plays a significant role.

Duration of the scene changes as following, in comparison to 640 x 480:
3300 x y: + 25s <- take with a grain of salt as the game slowed down already due to system load
1920 x y: + 11s
1024 x y: + 3s
800 x y : + 1s


derSteFfi:
looks fine. whenever we feel like we have nothing t do, we should come back to this heli animation and imporve it, but the speed is fine now.

PR #276, opened 2016-03-29
remove an old switch that deactivated some ScreenMsgs

derSteFfi:
it certainly was hacked into the code for one the devs. We now have other ways of switching this on/off anyway and its control was hidden behind JA2BETAVERSION anyway so it was unreachable atm.


derSteFfi:
moved this into LoggingRework, fits better there.

Issue #194, opened 2015-10-24
Running w/out moving


lynxlynxlynx:
Originally reported by: **Agatae Tsag (Bitbucket: [Agatae](http://bitbucket.org/Agatae), GitHub: Unknown)**

----------------------------------------

In tactical / turn mode, if my merc ends their turn running, they'll start the next one in motion.
Good.
However, they can finish the next turn w/out moving at all. They would still be 'running'.
This alone is a glitch, it's weird that the time passes and the merc stays in a running position.
However, 'running' like that drains energy. That's a bug. And I think that fixing the original glitch is worthwhile.
I suggest that merc that ends their turn running, yet still with enouth AP to move in any direction should just stop and stay in place.

----------------------------------------
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/194

PR #277, opened 2016-03-30
Direction type fixes

derSteFfi:
as discovered when solving #194 we have unconsistent handling of pathingData and direction types. This solves all of this I could find for now. Should reduce memory usage by about 800 Bytes ;)


selaux:
:+1: for the find. Using this branch, I can unfortunately not load games anymore and issuing commands on the tactical map results in my MERC running steadily in one direction :smile:.

Backtrace for the crash:
```
0x00000000006514df in LightSpriteDestroy (l=0x400040000000000) at Build/TileEngine/Lighting.cc:2206
2206		if (l->uiFlags & LIGHT_SPR_ACTIVE)
(gdb) backtrace
#0  0x00000000006514df in LightSpriteDestroy (l=0x400040000000000) at Build/TileEngine/Lighting.cc:2206
#1  0x00000000005e4c3c in DeleteSoldierLight (s=0xc9cbe0 <Menptr>) at Build/Tactical/Soldier_Control.cc:765
#2  0x00000000005e4bf6 in ReCreateSoldierLight (s=0xc9cbe0 <Menptr>) at Build/Tactical/Soldier_Control.cc:756
#3  0x000000000064fb16 in LightSetBaseLevel (iIntensity=3 '\003') at Build/TileEngine/Lighting.cc:1518
#4  0x000000000049e3cb in LoadSavedGame (save_slot_id=0 '\000') at Build/SaveLoadGame.cc:1125
#5  0x00000000004a3ff5 in DoneFadeOutForSaveLoadScreen () at Build/SaveLoadScreen.cc:1332
#6  0x00000000004466df in HandleFadeOutCallback () at Build/Fade_Screen.cc:98
#7  0x00000000004a211f in SaveLoadScreenHandle () at Build/SaveLoadScreen.cc:267
#8  0x0000000000448317 in GameLoop () at Build/GameLoop.cc:299
#9  0x00000000006adafd in MainLoop (msPerGameCycle=25) at sgp/SGP.cc:247
#10 0x00000000006ae2b3 in main (argc=1, argv=0x7fffffffe5e8) at sgp/SGP.cc:511
```

There are still some occurences of `INT8 bDirection` left, but even fixing these did not resolve the issues. So there must be more.

derSteFfi:
for the savegame: makes sense, SOLDIERTYPE has a different size now, this should be solvable by a savegame version bump and a compability for all older savegames that loads with the old sizes. Of course I need to check the saving routine too. I will do this after:
for the error in tactical: I remeber a spot where a plotted guiPathingData path was copied over to the soldier but I do not remember changing sizes there. They might used different names there. I'll look into this.

derSteFfi:
should work now, tested loading an old game, saving new ones and loading them and save a loaded old one into new version, works fine as far as I can tell.

selaux:
Wouldn't it be easier to do type conversion when loading/saving? Otherwise these save games will not be compatible at all anymore with vanilla. Currently I think loading stracciatella save games still works (did not test though).

derSteFfi:
that's possible, but i can't confirm too.
We could also convert when loading/saving of course. Depends on whether we want compability for our savegames in Vanilla JA2. (the other way around it's obviously yes, I didn't think of this direction)

derSteFfi:
doing conversion on loading/saving now. also rebumped savegame version to 99.

selaux:
Thanks for the fix. But still a merc that is ordered to move somewhere continues to run in this direction endlessly.

derSteFfi:
should be fixed. Could you retest too please? If that's OK, I'd like to merge this to get it off my list.

selaux:
Retested, and still have the same issue, Merc can even run through walls.

derSteFfi:
oh, ok. I guess, I'm back into. Should really take testing more seriously...

derSteFfi:
Hm, can you tell me how to reproduce these errors? I can't get any right now here, though i saw a data corruption in one case but can't reproduce it. With this one exception, everything works fine.

selaux:
Yes, maybe I didn't explain thoroughly.

- load or start a game
- go into tactical mode
- order merc to run somewhere

After that Merc runs ito the correct direction but does not stop neither at walls nor at anything. He just runs like Forrest Gump.

derSteFfi:
Should be corrected with this.

selaux:
Looks good to me now, thanks for the refactoring :+1:

Issue #251, opened 2016-03-07
Crash (reproducible)


stativ:
The game crashes with the attached save game.

Steps to reproduce:
1. load the save game
2. unpause the game

[error.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/161842/error.sav.zip)

lynxlynxlynx:
Is this with a released version or something from git?

stativ:
Oh, sorry I forgot to mention the version. It happens with the current git. I think I encountered the crash last December with 97316b5. I forgot about it until now, when I tried it with the current git (5135575 at the time of writing)

derSteFfi:
here is the backtrace:
```
Thread 1 "ja2" received signal SIGSEGV, Segmentation fault.
0x00000000004f50a3 in NumEnemiesInSector (sSectorX=0, sSectorY=0) at Build/Strategic/Queen_Command.cc:175
175		ubNumTroops = (UINT8)(pSector->ubNumAdmins + pSector->ubNumTroops + pSector->ubNumElites);
(gdb) backtrace
#0  0x00000000004f50a3 in NumEnemiesInSector (sSectorX=0, sSectorY=0) at Build/Strategic/Queen_Command.cc:175
#1  0x00000000004d2ae9 in HandleHeliEnteringSector (sX=0, sY=0) at Build/Strategic/Map_Screen_Helicopter.cc:218
#2  0x000000000051ad00 in GroupArrivedAtSector (g=..., check_for_battle=1 '\001', never_left=0 '\000') at Build/Strategic/Strategic_Movement.cc:1384
#3  0x00000000004c28fd in ExecuteStrategicEvent (pEvent=0x2663880) at Build/Strategic/Game_Event_Hook.cc:165
#4  0x00000000004c2e35 in ProcessPendingGameEvents (uiAdjustment=60, ubWarpCode=1 '\001') at Build/Strategic/Game_Events.cc:179
#5  0x00000000004c1312 in AdvanceClock (ubWarpCode=1 '\001') at Build/Strategic/Game_Clock.cc:192
#6  0x00000000004c12c3 in WarpGameTime (uiAdjustment=60, ubWarpCode=1 '\001') at Build/Strategic/Game_Clock.cc:173
#7  0x00000000004c1cd2 in UpdateClock () at Build/Strategic/Game_Clock.cc:742
#8  0x000000000044829f in GameLoop () at Build/GameLoop.cc:312
[...]

PR #279, opened 2016-03-30
Debug screens

selaux:
I enabled some debug screens to be accessible when using the `-debug` parameter.

Tactical Screen
- `F11` for Quest Debug Screen
- `CTRL+f` for FPS Display
- `CTRL+z` to toggle z-buffer
- `ALT+m` for level node debug mode
- `ALT+n` to play quotes of hovered merc
- `q` for soldier and land debug mode
- `y` for struct debug mode

To exit any screen press `q`. To cycle through different pages use pageup and pagedown.

I'm in the progress of making everything look pretty and usable. Progress:
- [x] Quest Debug Screen
- [x] FPS Display
- [x] z-Buffer
- [x] Level Node
- [x] Quotes
- [x] Soldier and Land

Examples:
screenshot from 2016-03-31 00-15-28
screenshot from 2016-03-31 00-15-39
screenshot from 2016-03-31 00-18-59

### DIFF
--- a/Build/Cheats.h
+++ b/Build/Cheats.h
@@ -2,6 +2,7 @@
 #define _CHEATS__H_
 
 #include "Types.h"
+#include "GameState.h"
 
 extern	UINT8			gubCheatLevel;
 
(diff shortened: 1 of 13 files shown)


derSteFfi:
wow. this is awesome. tested what you described. works fine. FPS display seems quite unresponsive, but maybe just because it is actually on the max value all the time. I'm still baffled. this is awesome! :)

derSteFfi:
one thing i noticed: the text isn't properly erased unless you scroll. don't know if this is fixable though...

selaux:
This is what you find when you dig around in this mess :smile:. 

The text not scrolling correctly is an issue, hiding with `q` works fine for me though. Also the fps overlay sometimes does not draw correctly yet. We could also use the empty pages to display some additional data in the future.

derSteFfi:
yes, i'd very much like info about a tiles flags for #251 :)

derSteFfi:
looks much better now. fps is also nice and visible. scrolling through land/soldier debug pages works fine for now.
fps overlaps with the debug pages and if any page is shown, i can't exit tactical.

derSteFfi:
I discovered this one switch to toggle scrolling mode and activated it. Doesn't help right now, but could be useful...

selaux:
- @derSteFfi I wrapped the scrolling mode toggle into the debug cheat mode
- with `y` you can prettily debug structs (and their flags :wink:) now

This was force pushed so you need to remove your local branch before pulling.

derSteFfi:
works smoothly. I'd vote for merging.

lynxlynxlynx:
Please move the first post somewhere on the website — this is also useful for modders.

Issue #195, opened 2015-10-25
Enemy in the northmost sector on the map can't be shot in the head


lynxlynxlynx:
Originally reported by: **Agatae Tsag (Bitbucket: [Agatae](http://bitbucket.org/Agatae), GitHub: Unknown)**

---

Depending on his direction, enemy may or may not be headshottable because the head may be out of screen.
In the attached save, Johny sees an enemy who can't be shot in the head. If he shots elsewhere, the enemy turns and his head can be targeted.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/195

lynxlynxlynx:
Just a small update — haven't digged in much yet. From testing it seems these problems were there from svn times.

For the top, 61 is exactly enough of an offset there, while for the bottom, 20 is enough (not exact). The way the whole thing is set up, it would be hard to come up with a non-hardcoded solution (vs magic constant), so I believe the problem is elsewhere. GridNoOnScreen also uses some padding of its own.

lynxlynxlynx:
This was a problem in the original too, so it's not a regression nor tied to the other bugs tracked in the master bug.

There are some better candidate constants by name (eg. PRONE_HEIGHT), but they're used in a different context and the standing one is too high.

lynxlynxlynx:
Just to keep it all here: the fix was reverted because showing more tiles exposed bad map design. So currently this is more of a WONTFIX situation.

We could potentially show a popup akin to the door/container one, to choose the targeted body part.

selaux:
So should we move it out of the `0.16.0` milestone? As we basically have currenly no concrete plan on how to fix?

lynxlynxlynx:
Yep, done.

maxrd2:
This one is related to #171
Commit aeba89a (#733) fixes this issue.

PR #281, opened 2016-04-02
ApplyScrolling: buffer the top border, so npcs there can be shot in t…

lynxlynxlynx:
…he head #195

roof level automatically offsets, so such cases don't need to be handled separately

### DIFF
--- a/Build/TileEngine/RenderWorld.cc
+++ b/Build/TileEngine/RenderWorld.cc
@@ -2322,7 +2322,7 @@ static BOOLEAN ApplyScrolling(INT16 sTempRenderCenterX, INT16 sTempRenderCenterY
   // Checking if screen shows areas outside of the map
 	BOOLEAN fOutLeft   = (gsTLX > sTopLeftWorldX);
 	BOOLEAN fOutRight  = (gsTRX < sTopRightWorldX);
-	BOOLEAN fOutTop    = (gsTLY >= sTopLeftWorldY);            /* top of the screen is above top of the map */
+	BOOLEAN fOutTop    = (gsTLY >= sTopLeftWorldY + 62);       /* top of the screen is above top of the map (with padding for headshots) */
 	BOOLEAN fOutBottom = (gsBLY < sBottomLeftWorldY);          /* bottom of the screen is below bottom if the map */
 
   int mapHeight = gsBLY - gsTLY;


selaux:
This offset seems a little high to me and causes some corruptions (e.g. suddenly ending fences). Shouldn't a lower value suffice?

lynxlynxlynx:
Which sector are you looking at? I can see tree trunks get drawn on top of canopies in Meduna, but I don't have any saves with mercs in sectors with border fences. Since we need to draw a full extra grid row, I don't think we can change much here. The minimum value needed is 62.

selaux:
Looking at I13: Alma:

screenshot from 2016-04-02 17-20-18

lynxlynxlynx:
Ok, so it's not corruption, just level design issues. Looks better with the minimum though, so I'll push that.

selaux:
Looks better now, butI can still see the broken level design. My suggestion would be to merge and revisit these issues after the release. 

Question: Why do we need to use a minimum value here?

lynxlynxlynx:
We don't, but the minimum value minimizes the potential to see broken data, since less of it is exposed.

derSteFfi:
looks good to me too.

LonCJ:
Thought I'd mention there's also going to be a few places where you can see NPC that follows a schedule, like shopkeepers, before they become active.  You can't talk to them but they're still visible.   Suppose I should see if they can be attacked. 

For example Keith waiting for the pawn shop to open.
npc_waiting

derSteFfi:
hm. yes, we should investigate this. I'd still say it's data bug (or a hack in the original to place npcs on not visible tiles) but this is more serious than a fence that ends or similiar stuff.

lynxlynxlynx:
The original has the headshot bug, so it's immune. We could make NPCs outside the bbox invisible — we just did the reverse recently by showing anyone on inaccessible tiles.

LonCJ:
Looks like they can be killed or something happens anyway.  Sprayed Keith with bullets for awhile and there's no indication of hits, as in blood or damage text, but after awhile he stuck in what looks a death animation frame.  Nothing showed in history log(if Keith even has a death entry) but he never reappeared in the store and just stayed in the same location in this pose.
keith

selaux:
@LonCJ Can you provide a savegame for this, I'm to lazy to play til cambria :wink:

selaux:
@lynxlynxlynx Any idea on how to fix this? There are two things I would like to check:

- Does #282 allow NPCs to move out of this area
- Is it possible to hide them (and are they invincible when hidden)

I tried finding the place where it sais that enemy soldiers outside the visible tiles dont render but I could not...

LonCJ:
Sorry for the delay selaux don't know if you still need it.  Two savegames in zip, think the low number is Keith alive and high number is most likely dead Keith.

[Inactive_NPC_killable.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/296620/Inactive_NPC_killable.zip)

PR #282, opened 2016-04-02
Fix #125: Dont consider destinations outside of visible world as valid when finding paths.

### LINKED ISSUE #125: Enemy trapped on border of the tactical map

---

Sector A9 was attacked by 8 soldiers from the south. I killed 4 soldiers and searched for the others. This happened:

The enemy soldier is trapped in the inaccessible southernmost tile of the map. He can't be seen but he can see me. He can't be shot and can't hit anything either.
He can't be revealed with ALT+E cheat but can be killed (only) with ALT+O cheat.

I have seen this kind of bug years ago in v1.13. I can remember how it was reported back then and fixed somehow.

I have noticed before, that in Stracciatella, all tactical maps have inaccessible tiles in the south and don't show all accessible tiles in the north.
It feels, like the player can observe too much of a map in the south but not enough in the north.
I hope, you can understand what I'm trying to express...

Savegame attached.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/125

selaux:
This is my go at fixing #125. It considers tiles outside the visible world as an invalid move target. Unfortunately I have no save game where a sector is attacked from south to test properly. Additionally we could remove checking wether the origin tile is on the visible grid, which would allow enemies which are stuck there to move out (tested with the save game attached to the original issue on bitbucket).

### DIFF
--- a/Build/Tactical/PathAI.cc
+++ b/Build/Tactical/PathAI.cc
@@ -35,6 +35,8 @@
 #include "GameSettings.h"
 #include "Buildings.h"
 
+#include "slog/slog.h"
+
 
 // skiplist has extra level of pointers every 4 elements, so a level 5is optimized for
 // 4 to the power of 5 elements, or 2 to the power of 10, 1024
@@ -50,6 +52,8 @@ extern INT16 gsCoverValue[WORLD_MAX];
 BOOLEAN gfDisplayCoverValues = TRUE;
 static BOOLEAN gfDrawPathPoints = FALSE;
 #endif
+
+#define DEBUG_TAG_PATH_AI "PathAI"
 
 BOOLEAN gfPlotPathToExitGrid = FALSE;
 BOOLEAN gfRecalculatingExistingPathCost = FALSE;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Looks fine to me. The second suggestion can't hurt either, but I'd keep the logging call.

selaux:
Done. Do you have any save game that could be used to test?

lynxlynxlynx:
Nope. I think the fastest way is to conquer Drassen and wait for the counterattack. I guess I experienced similar situations with the SAM sites, but I never hit the bug in my playthrough.

derSteFfi:
maybe we should log these as warnings?
Also i think to remember that the drassen SAM is always attacked from the south...

selaux:
Just played through till Drassen SAM on easy difficulty and noticed that only 4 Elites are counter-attacking it, which always get placed correctly, even on master. Will have to try again with a higher difficulty to test properly :frowning:. Here is the save game anyways, if you want to have a try.
[SaveGame09.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/201235/SaveGame09.sav.zip)

selaux:
I tried about 20 times now but I just cannot reproduce this on current master, any ideas? A savegame right before Drassen is attacked is attached.

[QuickSave.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/208894/QuickSave.sav.zip)

selaux:
After unsuccessfully trying again to reproduce this for 20 times or so, I would suggest keeping the "Enemies are allowed to move away from invisible tiles"-part and discard the rest. Or should we keep this altogether and just assume this fix is ok as it is?

lynxlynxlynx:
I like the sanity checking, but now that we know some npcs want to be on invisible squares (see discussion of the toprow headshot bug), it would need extra testing. Should be fine if they get teleported there, but if they walk ...

selaux:
I adapted the PR and changed the logging to be warnings, so we are notified when such things happen.

lynxlynxlynx:
besides the whitespace foo, it looks fine to me.

derSteFfi:
I'd agree to monitoring this with the warnings. In the end, this betters the situation a lot.

selaux:
Allright I tested the behavior of Keith with this PR with somewhat mixed results. So the good part is that Keith stays at the Tile that he occupies when waiting for his store to open. The bad part is that

- The tile that is right beneath Keith on the same y-level is accessible to me
- Waiting for Keith to move, after about 20 minutes the game crashes, my assumption is when Keith wants to move. The stack trace includes the adapted function

The only solutions I can think of is check for the team and only allow enemies to move out or revisiting the spawning issue. Thoughts?

lynxlynxlynx:
Yeah, sounds like some lookahead would be needed. NPCs could jump/skip over full tiles. Same for enemies, since while it's probably more logical to move them sideways first, that'd be more work and we'd still have no guarantee that an empty space would be available.

PR #286, opened 2016-04-07
Fix 2 Coverity issues in Email.cc

derSteFfi:
1: HandleIMPCharProfileResultsMessage: adjust the size of a w_char_t array to the actual maximum size fo a mercs name

2: AddEmailRecordToList: add a 0-termination to a string at the end. all callers currently don't exceed its length but it's safer to ensure this here.

### DIFF
--- a/Build/Laptop/EMail.cc
+++ b/Build/Laptop/EMail.cc
@@ -53,7 +53,7 @@ struct Page
 
 struct Record
 {
-	wchar_t pRecord[640];
+	wchar_t pRecord[MAIL_STRING_SIZE + 1];
 	Record* Next;
 };
 
@@ -1612,8 +1612,9 @@ static void ClearOutEmailMessageRecordsList(void)
 }
 
 
-static void AddEmailRecordToList(const wchar_t* const text)
+static void AddEmailRecordToList(wchar_t* const text)
 {
+	text[MAIL_STRING_SIZE] = 0;
 	Record* const e = MALLOC(Record);
 	e->Next = NULL;
 	wcscpy(e->pRecord, text);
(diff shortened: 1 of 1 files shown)


Build/Laptop/EMail.cc
```diff
@@ -1916,7 +1917,7 @@ static void HandleIMPCharProfileResultsMessage(void)
 		// have to place players name into string for first record
 		if (i == 0)
 		{
-			wchar_t zTemp[512];
+			wchar_t zTemp[NAME_LENGTH];
```
lynxlynxlynx: what did it warn about here? That it's never fully filled?

derSteFfi: no, that the destination in the following copy (wcscat) is smaller and could be overrun.

lynxlynxlynx: Oh, right. Isn't it too low now though? NAME_LENGTH is 30, zName 32 (if I read it correctly) and the format string contains an extra space. So it could cut the string too short now.

The other question is, why is doing the catting separately?

derSteFfi: zName has the length of NAME_LENGTH too, that's why I chose it. So, if zName overflows, we will too, but it's probably good to see it when it happens so we notice and fix it.

lynxlynxlynx: Ah, I looked at the wrong one. Still, the name can get cut off if it is full length.

derSteFfi: If it contains 30 chars, yes. But if that happens, we'll at least see it instead of it only overflowing within the structure. We can adjust NAME_LENGTH easily whenever this should happen

lynxlynxlynx: There are two things here. You're talking about zName potentially overflowing somewhere else, which is not something this commit addresses (nor should it imo). The other thing is that the format specifier for swprintf contains a space, so the effective size will be one less for a fully packed (but not overflown) zName. So zTemp should be at least NAME_LENGTH+1 for the last char not to become null. Right?

derSteFfi: omg, I totally not saw that. Sorry, it took much to long for me to understand. You're right of course.

PR #287, opened 2016-04-07
fix a coverity issue in Points.cc

derSteFfi:
this looks like a typo

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -969,6 +969,10 @@ bool DefaultContentManager::loadAllDealersInventory()
 
 const ItemModel* DefaultContentManager::getItem(uint16_t itemIndex) const
 {
+	if(itemIndex >= m_items.size())
+	{
+		return nullptr;
+	}
 	return m_items[itemIndex];
 }
 
--- a/src/game/Tactical/Points.cc
+++ b/src/game/Tactical/Points.cc
@@ -1682,15 +1682,14 @@ INT16 MinAPsToThrow(SOLDIERTYPE const& s, GridNo gridno, bool const add_turning_
 	// Make sure the guy's actually got a throwable item in his hand
 	UINT16 const in_hand = s.inv[HANDPOS].usItem;
 	const ItemModel *item = GCM->getItem(in_hand);
-	// Gennady: This is a very strange piece of code.
-	//          Be very careful with it.
-	//          The added parenthesis is just to silence a compiler warning (https://github.com/ja2-stracciatella/ja2-stracciatella/issues/791),
-	//          it does not necessarily indicate desired behavior - it just conserves present behavior.
-	//          For more discussion around it, see https://github.com/ja2-stracciatella/ja2-stracciatella/pull/287.
-	if ((!item->getItemClass()) & IC_GRENADE)
+	if (!item)
 	{
-		SLOGI("MinAPsToThrow - Called when in-hand item is %s",
-					item->getInternalName().c_str());
+		SLOGW("MinAPsToThrow - in-hand item is missing");
+		return 0;
+	}
+	else if (!(item->getItemClass() & (IC_GRENADE | IC_THROWN))) // match MinAPsToAttack
+	{
+		SLOGW("MinAPsToThrow - in-hand item '%s' has unexpected item class 0x%x", item->getInternalName().c_str(), item->getItemClass());
 		return 0;
 	}
 


lynxlynxlynx:
just due to the comment, does ap for throwing stuff still work fine?

derSteFfi:
i have no idea :)

RinneT:
Is it bad that - despite debugging it - I have no idea what this code is supposed to do?

I can now happily throw all of my grenades without losing AP though. Stance changes necessary for throwing don't cost AP either. Sorry :(

derSteFfi:
Hm, i can confirm this. PathAI displays a reasonable AP cost but when executing the action, 0 AP are taken and the log behind this if is issued even though the item in hand clearly is a grenade. so the check is buggy. Looking into this.

derSteFfi:
So, the issue is that the grenade is already removed from the slot when the AP are about to be accounted. Backtrace for the removal:
```
(gdb) backtrace
#0  0x00007ffff6d6704c in __memset_sse2 () from /usr/lib/libc.so.6
#1  0x0000000000589f0a in DeleteObj (pObj=0xc99cbc <Menptr+188>) at Build/Tactical/Items.cc:1256
#2  0x000000000058a07a in RemoveObjs (pObj=0xc99cbc <Menptr+188>, ubNumberToRemove=1 '\001') at Build/Tactical/Items.cc:1308
#3  0x00000000006189f2 in UseThrown (pSoldier=0xc99c00 <Menptr>, sTargetGridNo=14340) at Build/Tactical/Weapons.cc:1433
#4  0x0000000000616604 in FireWeapon (pSoldier=0xc99c00 <Menptr>, sTargetGridNo=14340) at Build/Tactical/Weapons.cc:533
#5  0x000000000054da58 in HandleItem (s=0xc99c00 <Menptr>, usGridNo=14340, bLevel=0 '\000', usHandItem=135, fFromUI=1 '\001') at Build/Tactical/Handle_Items.cc:934
#6  0x0000000000555f85 in UIHandleMercAttack (pSoldier=0xc99c00 <Menptr>, pTargetSoldier=0x0, usMapPos=14340) at Build/Tactical/Handle_UI.cc:2147
#7  0x0000000000556201 in UIHandleCAMercShoot (pUIEvent=0xba1678 <gEvents+1240>) at Build/Tactical/Handle_UI.cc:2215
#8  0x00000000005537ba in HandleTacticalUI () at Build/Tactical/Handle_UI.cc:530
#9  0x00000000004499a7 in MainGameScreenHandle () at Build/GameScreen.cc:525
#10 0x0000000000448077 in GameLoop () at Build/GameLoop.cc:299
#11 0x00000000006ab7e5 in MainLoop (msPerGameCycle=25) at sgp/SGP.cc:247
#12 0x00000000006abf9b in main (argc=4, argv=0x7fffffffe028) at sgp/SGP.cc:511
```
Backtrace for the moment APs are supposed to be taken but aren't:
```
(gdb) backtrace
#0  MinAPsToThrow (s=..., gridno=14340, add_turning_cost=false) at Build/Tactical/Points.cc:1623
#1  0x00000000005dd485 in AdjustToNextAnimationFrame (pSoldier=0xc99c00 <Menptr>) at Build/Tactical/Soldier_Ani.cc:866
#2  0x00000000005b85aa in ExecuteOverhead () at Build/Tactical/Overhead.cc:770
#3  0x0000000000449992 in MainGameScreenHandle () at Build/GameScreen.cc:516
#4  0x0000000000448077 in GameLoop () at Build/GameLoop.cc:299
#5  0x00000000006ab7e5 in MainLoop (msPerGameCycle=25) at sgp/SGP.cc:247
#6  0x00000000006abf9b in main (argc=4, argv=0x7fffffffe028) at sgp/SGP.cc:511
```
We need to adjust the order of execution here or find another solution.

Manuel-K:
`!item->getItemClass() & IC_GRENADE == (!item->getItemClass()) & IC_GRENADE`

`!` is the logical not and not the bitwise not (`~`). So it should evaluate to `false` if `item->getItemClass()` is anything other than `0` or `false`. So it should be `false & IC_GRENADE == false`.

I don't think that there is any relevant item without a class, so the whole statement should always be false. 

I'm confused as well.

lynxlynxlynx:
To me, looking at the surrounding code, the goal was to exclude non-throwables. Do knives have the same flag? Are they excluded elsewhere? Do grenades do have that bit set or is only in the launcher ... Plenty data to look into.

flaviojs:

flaviojs:
From the placement and the debug message, it's probably meant to be "if the item can't be thrown, log the problem and return 0".

The check is only for grenades. Probably you could only throw grenades in an early version, then came new items that could be thrown. The check never worked (didn't trigger on the new items) so it survived unmodified.

There is no special AP number for "impossible", so returning 0 is appropriate.

flaviojs:
The normal code path never returns 0, so it's probably better to log a warning and continue the normal path instead of returning 0 (or just abort).
Do all AIs have APs? Those that have 0 probably won't be able to throw anymore.

As for the test, I would match `MinAPsToAttack`.

flaviojs:
Proposal (diff against master): [pr-287.txt](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3894001/pr-287.txt)

lynxlynxlynx:
Does it really make sense to continue there? If you hack an item to hit this check, does the merc still move closer/change stance & reduce the ap appropriately, does anything break later?

Maybe it should just be an assert.

flaviojs:
I'm taking over this PR.

flaviojs:
Now it matches MinAPsToAttack.

There might be other scenarios where an item can be thrown.
It is unclear because there is too much spaghetti code.
They will show a warning and have an AP cost of 0.

lynxlynxlynx:
The code looks fine, but what's the end behaviour in game? Throwing knives and grenades still work fine? I think you can also dump inventory items in general, but don't remember how AP was handled there (rare need).

flaviojs:
I tried throwing a grenade (attack), throwing a knife (attack) and throwing an item (drop).
The warning didn't show up so nothing changed for these cases.

PR #293, opened 2016-04-18
Add logo to releases

### LINKED ISSUE #246: Stracciatella logo
stracc
Since selaux mentioned a logo in #244, I had this idea.
I'm not much of a painter + this was a quick job done in ms paint.

I wanted to do something that is more like a pictogram or a symbol. A very simple shape with few colors. The red from the background is taken directly from the original JA2 logo, as is the color of the cone.
logo-orig

Now, I do not think that this should be the official logo for Stracciatella. But what do you think of the general style?

selaux:
- I added some rounded edges for the icon
- Closes #246
- Packages logo into releases
- Packages .desktop file


derSteFfi:
looks good to me. only thing not working: gnome doesn't seem to recognize the logo when alt+tab'ing. for the 'all applications' menu its fine however.

selaux:
I will look into setting the applications icon, should have done this in the first place...

selaux:
Should work now...

derSteFfi:
it does. I'm pro merging.

PR #299, opened 2016-05-03
Revert "ApplyScrolling: buffer the top border, so npcs there can be shot in t…"

selaux:
Reverts ja2-stracciatella/ja2-stracciatella#281


JAsmine-ja2:
So you _did _came up with a solution for this issue? Great stuff!

lynxlynxlynx:
Well, partly, since it shows the original wasn't prepared for such a simple fix. This here is temporary for release purposes.

PR #303, opened 2016-05-06
Iqha mini mod

JAsmine-ja2:
This is mainly a test. It's my first pull request and I might have done something wrong.
However, the mini-mod itself is legit.
It simply shows the effect of the different answers in the IMP quiz.


lynxlynxlynx:
cool! While I like the absurdity of the quiz, not getting predictable results is a pita for replays.

lynxlynxlynx:
You've actually only uploaded the readme and an empty test file — the data override is missing.

JAsmine-ja2:
[IMP quiz honest answers mod.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/282833/IMP.quiz.honest.answers.mod.zip)

I will try to make a proper pull request for the next project :)
Does it help if I just attach the file here?

lynxlynxlynx:
I've pushed a cleaned up version. The mod works fine. Thanks. :)

For any further mods: use a lowercase "data" dir and no spaces in the name.

nemuc:
> For any further mods: use a lowercase "data" dir and no spaces in the name.

@lynxlynxlynx 
For which reason should "data" dir be lowercase in this case?

On debian this happens:
My Jagged Alliance 2 from GOG names the directory `/Data/BinaryData/` with the D's in **Data uppercase**.
The v0.16.1 (2018-04-08) of the IMP quiz honest answers mod names the directory 
 `/data/Binarydata/`with the D's in **data lowercase**.
Thus two different directories are created.

Shouldn't it be `/Data/BinaryData/` in the mod as well, so that this installation instruction of the mod will be correct?

> Just copy the data-folder into your JA2 folder

lynxlynxlynx:
It should go to the datadir, sure. It didnt work for me without the change, but it could be my dir is locase too. More likely that we a hardcoded lookup. Worth a retedt, since there was a rewrite. Does the mod work for you after the move?

nemuc:
The mod worked fine after the move. Let me know if I can test something, so that others can also easily install this great mini-mod.

lynxlynxlynx:
Please open a bug report, I'm still stuck with phone only access.

Issue #308, opened 2016-05-15
New Game Mode Proposal: Dead is Dead


RinneT:
Dear all,

I really enjoy playing the game and think that Iron Man mode adds nicely to the difficulty.
I have, however, often caught myself reloading and retrying a sector, especially when someone dies at the beginning of a fight.

To make the game even more of a challenge, I thought about a new game mode for a while, that doesn't allow just reloading after someone died. A sort of "Dead is Dead" game mode.

I've occasionally been playing around with the code to get the experience into the game and am wondering if there is general interest in taking such a mode into the main game.

What I have so far is quite little as I'm making slow progress, but as of now some features are present.

What is working:
- Select the new game mode: Dead is Dead from the init screen
- The game automatically saves:
 + On Exit to Desktop (Alt +x)
 + On loading of another savegame
 + On Exit to Main Menu
- Doing a Quick Save is redirected to the Dead Is Dead Save
- Savegame compatibility with Vanilla (Converts to Iron Man, thanks to the way booleans are handled in c++)

What I still have planned:
- [x] Save on "hard" Exit (Alt + F4 / Window closing)
- [x] Disable the regular Save Screen for Dead is Dead
- [x] New tab in the Save / Load Screen for Dead is Dead games
- [x] Allow the user to enter a Savegame name in the beginning of the game (Currently hardcoded as DiD)
- [x] New texts and translations for the mode (No clue how that works as of yet)

Also, as expected from my rather low experience, so far code quality is not the best.
I try to keep things mostly easily readable as I'm constantly changing stuff as I learn new mechanics of the game. Also I'm very likely doing a lot of stuff that works in the few scenarios I tested, but probably isn't a good way to do it. This is especially true for the whole Screen handling. (Which screen are we in etc.)

Open Questions:

What happens if you exit during:
- [x] an enemy turn
 + the game loads as "Enemy turn", but the player actually has control.
 + should be fixed to let the enemy turn continue. Make sure all APs etc. are correct.
- [x] a dialogue
 + The game seems to be stuck in an infinite loop (the 'waiting' clock is displayed)
 + My preference would be to start the dialogue from the beginning.
- [x] a cut scene
 + The cut scene continues, but is flawed. E.g. in Deidrannas palace you can not see into the room anymore (The roof is rendered).

JAsmine-ja2:
I know that something like this was once mentionend by someone in the old BitBucket repository...

Ah, here it is: Take a look at this project!
[...]

PR #309, opened 2016-05-15
 New Game Mode Proposal: Dead is Dead #308 

RinneT:
The code for a new game Mode "Dead is Dead" as described in Issue #308 

Beware low code quality.
I'm also not sure if I'm doing it the right way by requesting a pull to master.
I'd be interested to have it as a new branch, if you are interested.

Best regards,
Thomas


Build/GameInitOptionsScreen.cc
```diff
@@ -489,10 +497,11 @@ static void RenderGIOScreen(void)
 	usPosY += GIO_GAP_BN_SETTINGS;
 
 	DisplayWrappedString(GIO_IRON_MAN_SETTING_X + GIO_OFFSET_TO_TEXT, usPosY, GIO_DIF_SETTINGS_WIDTH, 2, GIO_TOGGLE_TEXT_FONT, GIO_TOGGLE_TEXT_COLOR, gzGIOScreenText[GIO_IRON_MAN_TEXT], FONT_MCOLOR_BLACK, LEFT_JUSTIFIED);
-
-	usPosY += 20;
-	DisplayWrappedString(GIO_IRON_MAN_SETTING_X + GIO_OFFSET_TO_TEXT, usPosY, 220, 2, FONT12ARIAL, GIO_TOGGLE_TEXT_COLOR, zNewTacticalMessages[TCTL_MSG__CANNOT_SAVE_DURING_COMBAT], FONT_MCOLOR_BLACK, LEFT_JUSTIFIED);
-
+        DisplayWrappedString(GIO_IRON_MAN_SETTING_X + GIO_OFFSET_TO_TEXT, usPosY+20, 220, 2, FONT12ARIAL, GIO_TOGGLE_TEXT_COLOR, zNewTacticalMessages[TCTL_MSG__CANNOT_SAVE_DURING_COMBAT], FONT_MCOLOR_BLACK, LEFT_JUSTIFIED);
```
derSteFfi: you introduced a small whitespace error here :)

Build/SaveLoadScreen.cc
```diff
@@ -1473,7 +1474,23 @@ static void FailedLoadingGameCallBack(MessageBoxReturnValue const bExitValue)
 
 void DoQuickSave()
 {
-	if (SaveGame(0, L"")) return;
+        // Use the Dead is Dead function if we are in DiD
```
derSteFfi: this block's whitespace looks a bit meesed up, too.

RinneT:
Thanks SteFfi :)
I think I fixed the whitespace issues. Changed my IDE settings to meet the indent conventions recently though, and hope it doesn't screw it up further. (2 Whitespaces instead of tab).

I made further changes to files that are so far quite ugly, so won't commit them yet.

RinneT:
So, I'm a bit lost right now.

I want to add an additional "Tab" to the Load Screen that displays only Dead is Dead games.
At the moment it's just a button that switches between Dead is Dead and normal.
I handle them by just adding NUM_SAVED_GAMES to counters / ids where it is necessary.
That way all savegames SaveGame01.sav - SaveGame10.sav are handled as normal games, and SaveGame11.sav - SaveGame 22.sav are only displayed on the new "Tab".

I originally implemented it so that I have a static variable gbSaveGameOffset. That variable is 0 when the normal Tab is shown and 11 when we switch.

That way I could switch tabs ok, but it would not work correctly.
I found the reason for that was that in between, during the loading of Display names of the savegames (Savegame slot 3 to be precise) the variable would just reset to 0 for absolutely no reason. I still don't get it.
That lead to all succeeding actions to fail, as we suddenly miss the offset.

What I did then was replace that variable and rely entirely on the boolean gfDiDTab. That boolean is set to true if we are in the Dead is Dead tab and false if we are in normal.
Unlike the integer value this variable is not reset and acts as expected, however now I have an awful lot of ugly if-statements.

If we now want to change to the DiD tab, the game crashes with a SigSev during the rerender of the screen. Specifically, it crashes in VObject_Blitters.cc in the function BOOLEAN BltIsClipped(const SGPVObject* const hSrcVObject, const INT32 iX, const INT32 iY, const UINT16 usIndex, const SGPRect* const clipregion)
it seems that "pTrav" isn't initialized correctly.

I see absolutely no connection between the two and can, as of now, only imagine some pointer error.

What's interesting as well: If we start with gfDiDTab=TRUE, it works as expected. We can also switch to normal. However if we then want to switch back to DiD, it crashes.

Any idea as to what I'm doing wrong?
Attached the diff below (as txt as github doesn't allow diff), that should be applied to the last committed state.

[DiDTab_SigSev.txt](https://github.com/ja2-stracciatella/ja2-stracciatella/files/288408/DiDTab_SigSev.txt)

RinneT:
Found the issue. I initialized the savegame array every time I switched tabs. That was not only unnecessary, but also caused pointer errors, as I accidentally resized the array.

RinneT:
Finished (from a functional point of view) the Gui Work.
It is now possible to start a new game, and select a slot for that game.
The slot is dedicated to that game and can't be changed (from in game).
Saving / loading works.
Up next is refining the Save mechanism.

RinneT:
Started refining the save mechanism.
The game checks for current events in the game and does not save during cut scenes, dialogues, message boxes (Otherwise the messagebox input is lost), or non-player turns.

Additionally the game saves automatically at the end and start of each player turn and before a cutscene is triggered.

RinneT:
Fixed the aforementioned error of losing control after loading a game with mercs as prisoners.
It seems to have been connected to loading with the wrong screen.
I have also completely redone the screen selection for saving. It should always give correct results now. In case it fails it redirects to the Map Screen.

RinneT:
Added a savegame backup system. The mode should now be playable rather safely.

The game creates two rolling backups in a subfolder "Backups" in the savegame directory.
Named as "savegamename.backupnumber".

Currently the backup mechanism is linked to the game modes specific saving function.
However it might make sense to hook it into the SaveGame function and enable it for all saves. Or via parameter (boolean createBackup) or sth.
All it requires is the savegame number.

RinneT:
Added Proper tabs to the Loadscreen.
Thanks to Stefan Lau's opentoolset I could take a closer look at the existing game files.
I unpacked the STI to check the size of the Loadscreen header and then created my own buttons.
I packed the images with STI-Edit from [http://www.jaggedalliance.de/download/filearchiv_list.php?group=2](http://www.jaggedalliance.de/download/filearchiv_list.php?group=2), though. Since animated STIs only support a limited image palette (8bit) and I didn't fine tune anything, they look extremely bad at the moment.

The STI is placed in a new folder externalized->sti->interface.
I have placed the higher quality source file in docs->custom_artworks_source->interface. It is a Gimp xcf file.

I wish to do three more things now:
- Raise the visuals of the new buttons to the other Interface compontents standard
- Correct the behaviour of the Load button being disabled even if a save is available
- Add a save on strategic events, such as a new sector being reached or tasks being completed.

Especially the last one has proven very tricky as it can put the savegame in a broken state. I.e. if enemies are present in the new sector. I'll have to see if I can find a viable way to set that up. Otherwise it won't be too dramatic, as the game saves on exit anyways and autosaving via Alt+s is still allowed.

JAsmine-ja2:
Nice to see progress here. I'm really looking forward to this mod/game-mode!

selaux:
I could attach some builds here to make it testable more easily for people.

RinneT:
Thanks. That would be awesome :+1:

selaux:
Here you go:
[ja2-v0.14.xx-win.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/359936/ja2-v0.14.xx-win.zip)

RinneT:
The Load screen selects the right tab depending on the last selected save.
It also deactivates tabs that do not contain saved slots.
I have also rewritten a few things to make introduction of a system with more tabs possible/easier.
Load game button is also fixed.

The tab images look much better now, though anyone with a bit of skill is invited to improve them.
Btw. personally I'd put them under an open licence, though I'm not sure what I'd have to for that.

The game is now saved shortly before a group arrives in a (non-town) sector.

Except for code improvement and documentation I consider this done.

Build/GameInitOptionsScreen.cc
```diff
@@ -309,7 +316,13 @@ static void EnterGIOScreen()
...
 		INT16  const x   = GIO_IRON_MAN_SETTING_X + GIO_OFFSET_TO_TOGGLE_BOX;
 		INT16  const y   = GIO_IRON_MAN_SETTING_Y - GIO_OFFSET_TO_TOGGLE_BOX_Y;
-		size_t const def = gGameOptions.fIronManMode ? GIO_IRON_MAN : GIO_CAN_SAVE;
+		size_t def;
+		switch (gGameOptions.ubGameSaveMode) {
+		  	case GIO_CAN_SAVE:   	def = GIO_CAN_SAVE; break;
+			default:
+			case GIO_IRON_MAN: 	def = GIO_IRON_MAN;  break;
+			case GIO_DEAD_IS_DEAD:  def = GIO_DEAD_IS_DEAD; break;
+		}
```
lynxlynxlynx: Why not just:
size_t const def = gGameOptions.ubGameSaveMode ? gGameOptions.ubGameSaveMode : GIO_CAN_SAVE;

... or not allowing for it to ever be unset?

Build/GameInitOptionsScreen.cc
```diff
@@ -735,10 +761,16 @@ static void DoneFadeOutForExitGameInitOptionScreen(void)
 	gGameOptions.fTurnTimeLimit = GetCurrentTimedTurnsButtonSetting();
 #endif
 	// JA2Gold: iron man
-	gGameOptions.fIronManMode = GetCurrentGameSaveButtonSetting();
-
-	gubGIOExitScreen = INTRO_SCREEN;
+	gGameOptions.ubGameSaveMode = GetCurrentGameSaveButtonSetting();
 
+	if (GetCurrentGameSaveButtonSetting() == GIO_DEAD_IS_DEAD)
```
lynxlynxlynx: No need to call it twice, it was just saved into gGameOptions.ubGameSaveMode

Build/GameSettings.cc
```diff
@@ -186,7 +186,7 @@ void InitGameOptions()
 	gGameOptions.fSciFi						 = TRUE;
 	gGameOptions.ubDifficultyLevel = DIF_LEVEL_EASY;
 	//gGameOptions.fTurnTimeLimit		 = FALSE;
-	gGameOptions.fIronManMode			 = FALSE;
+	gGameOptions.ubGameSaveMode			 = DIF_CAN_SAVE;
```
lynxlynxlynx: above you're checking it against GIO_* constants. Are both sets needed, is there a bug somewhere?

Build/SaveLoadGame.cc
```diff
@@ -1466,6 +1479,33 @@ static void LoadSoldierStructure(HWFILE const f, UINT32 savegame_version, bool s
...
+	sprintf(zSourceSaveGameName, "%s%02d.%s", g_savegame_name, ubSaveGameID, g_savegame_ext);
+	for (int i = NUM_SAVE_GAME_BACKUPS - 1; i >= 0; i--)
+	{
+		if (i==0)
+		{
+			strcpy(zSourceBackupSaveGameName, zSourceSaveGameName);
+		}
+		else
+		{
+			sprintf(zSourceBackupSaveGameName, "%s.%01d", zSourceSaveGameName, i);
```
lynxlynxlynx: This can overflow if the source name is at full array length. Same for the sprintf below.

Build/SaveLoadGame.cc
```diff
@@ -2095,6 +2135,9 @@ static void LoadGeneralInfo(HWFILE const f, UINT32 const savegame_version)
 	EXTR_BOOL( d, gubPlayerProgressSkyriderLastCommentedOn)
 	EXTR_BOOL( d, gfMeanwhileTryingToStart)
 	EXTR_BOOL( d, gfInMeanwhile)
+	// Always set gfInMeanwhile to false for Dead is Dead. This must be done because it is saved as true if a Meanwhile event is in the event pipe
+	// Preventing the value to be saved in the first place leads to odd behaviour during the commencing cutscene
```
lynxlynxlynx: Is this comment still true — any "odd behaviour" remaining?

Build/SaveLoadScreen.cc
```diff
@@ -135,18 +147,23 @@ static ScreenID guiSaveLoadExitScreen = SAVE_LOAD_SCREEN;
...
 BOOLEAN		gfSaveGame=TRUE;
+static BOOLEAN   gfDiDTab=FALSE;
 
 static BOOLEAN gfSaveLoadScreenButtonsCreated = FALSE;
 
 static INT8 gbSelectedSaveLocation = -1;
 static INT8 gbHighLightedLocation  = -1;
+// We use this offset to differentiate between normal games and Dead is Dead. If we are in Dead is Dead, this offset is set to NUM_SAVE_GAMES
+// and simply added to counters etc. Dead is Dead games run in Slots NUM_SAVE_GAMES+1 to NUM_SAVE_GAMES+NUM_SAVE_GAMES (12 to 23 by default)
+//static INT8 gbSaveGameOffset = 0;
```
lynxlynxlynx: remove if it remains unused.

Build/SaveLoadScreen.cc
```diff
@@ -327,7 +349,8 @@ static void LeaveSaveLoadScreen()
 	ScreenID const exit_screen =
 		gfCameDirectlyFromGame                     ? guiPreviousOptionScreen :
 		guiPreviousOptionScreen == MAINMENU_SCREEN ? MAINMENU_SCREEN         :
-		OPTIONS_SCREEN;
+		guiPreviousOptionScreen == GAME_INIT_OPTIONS_SCREEN ? GAME_INIT_OPTIONS_SCREEN : // Added to handle interaction during Dead is Dead game start
+		guiPreviousOptionScreen == INTRO_SCREEN ? INTRO_SCREEN : OPTIONS_SCREEN;
```
lynxlynxlynx: What a monster statement! This would benefit from a refactor to a set + a conditional switch on guiPreviousOptionScreen.

Build/SaveLoadScreen.cc
```diff
@@ -348,9 +381,89 @@ static void SelectedSaveRegionCallBack(MOUSE_REGION* pRegion, INT32 iReason);
...
+	}
+}
+// This function determines which tab to activate in the load Screen.
+// Depending on:
+// In which tab is the last save
+// Are there any saves in the tab
+// It also deactivates tabs with no saves
+static void selectActiveTab()
+{
+	INT8 const lastSaveInTab = (INT8) gGameSettings.bLastSavedGameSlot / NUM_SAVE_GAMES;
```
lynxlynxlynx: Any chances for rounding errors here?

Build/SaveLoadScreen.cc
```diff
@@ -851,7 +1026,8 @@ static BOOLEAN DisplaySaveGameEntry(INT8 const entry_idx)
 			swprintf(options, lengthof(options), L"%20ls     %22ls     %22ls     %22ls",
 				difficulty,
 				/*gzGIOScreenText[GIO_TIMED_TURN_TITLE_TEXT + header.sInitialGameOptions.fTurnTimeLimit + 1],*/
-				header.sInitialGameOptions.fIronManMode ? gzGIOScreenText[GIO_IRON_MAN_TEXT] : gzGIOScreenText[GIO_SAVE_ANYWHERE_TEXT],
+				(header.sInitialGameOptions.ubGameSaveMode == 1) ? gzGIOScreenText[GIO_IRON_MAN_TEXT] : 
+				  (header.sInitialGameOptions.ubGameSaveMode == 2) ? gzGIOScreenText[GIO_DEAD_IS_DEAD_TEXT] : gzGIOScreenText[GIO_SAVE_ANYWHERE_TEXT],
```
lynxlynxlynx: Are the GIO/DEF mode variables not available here?

Build/SaveLoadScreen.cc
```diff
@@ -910,7 +1086,7 @@ static BOOLEAN DisplaySaveGameEntry(INT8 const entry_idx)
 	else
 	{
 		// If this is the quick save slot
-		wchar_t const* const txt = entry_idx == 0 ?
+		wchar_t const* const txt = entry_idx == (-1 + !gfDiDTab) ? // This is a beyond-ugly hack to remove the quicksave slot from the Dead is Dead Tab
```
lynxlynxlynx: Is the const that important? I'd just set it to the empty slot and check the idx if not in DiD mode (in a separate statement).

Or perhaps create a struct for all game mode data, so also the gzGIOScreenText tests can be removed.

Build/SaveLoadScreen.cc
```diff
@@ -1473,12 +1675,60 @@ static void FailedLoadingGameCallBack(MessageBoxReturnValue const bExitValue)
 
 void DoQuickSave()
 {
-	if (SaveGame(0, L"")) return;
+  // Use the Dead is Dead function if we are in DiD
+  if (gGameOptions.ubGameSaveMode == DIF_DEAD_IS_DEAD)
+    DoDeadIsDeadSave();
+  else
```
lynxlynxlynx: Please add braces, this is just asking for trouble. I'm surprised it compiled.

Build/SaveLoadScreen.cc
```diff
@@ -1473,12 +1675,60 @@ static void FailedLoadingGameCallBack(MessageBoxReturnValue const bExitValue)
...
+}
 
-	if (guiPreviousOptionScreen == MAP_SCREEN)
-		DoMapMessageBox(MSG_BOX_BASIC_STYLE, zSaveLoadText[SLG_SAVE_GAME_ERROR], MAP_SCREEN, MSG_BOX_FLAG_OK, NULL);
-	else
-		DoMessageBox(MSG_BOX_BASIC_STYLE, zSaveLoadText[SLG_SAVE_GAME_ERROR], GAME_SCREEN, MSG_BOX_FLAG_OK, NULL, NULL);
+// Save function for Dead Is Dead
+void DoDeadIsDeadSave()
+{
+ // Check if we are in a sane! Do not save if:
```
lynxlynxlynx: a sane what?

Build/SaveLoadScreen.cc
```diff
@@ -1498,7 +1748,7 @@ void DoQuickLoad()
 
 bool AreThereAnySavedGameFiles()
 {
-	for (INT8 i = 0; i != NUM_SAVE_GAMES; ++i)
+	for (INT8 i = 0; i != (2 * NUM_SAVE_GAMES); ++i)
```
lynxlynxlynx: All these 2s should either be a define similar to NUM_SAVE_GAME_BACKUPS or computed, if the way we track modes changes. Otherwise they'll be annoying to track down if anyone ever wants to add a third tab.

lynxlynxlynx: Ah, you even define further down: NUM_SAVE_GAMES_TABS

Build/SaveLoadScreen.cc
```diff
@@ -1580,8 +1830,16 @@ static void SaveGameToSlotNum(void)
...
-
-	if( !SaveGame( gbSelectedSaveLocation, gzGameDescTextField ) )
+  
+  // If we are selecting the Dead is Dead Savegame slot, only remember the slot, do not save
+  // Also set the INTRO_SCREEN as previous options screen. This is a hack to get the game started
+  if (guiPreviousOptionScreen == GAME_INIT_OPTIONS_SCREEN)
+  {
+    guiPreviousOptionScreen = INTRO_SCREEN;
+    gGameSettings.bLastSavedGameSlot = (gbSelectedSaveLocation + NUM_SAVE_GAMES);
+    wcscpy(gGameSettings.sCurrentSavedGameName, gzGameDescTextField);
```
lynxlynxlynx: Is this always safe?

Build/Tactical/TeamTurns.cc
```diff
@@ -146,6 +148,9 @@ void StartPlayerTeamTurn( BOOLEAN fDoBattleSnd, BOOLEAN fEnteringCombatMode )
   /// that a baddie can die and still maintain it's attacker ID
   gTacticalStatus.fKilledEnemyOnAttack = FALSE;
 
+  // Save if we are in Dead is Dead mode
+  if (gGameOptions.ubGameSaveMode == DIF_DEAD_IS_DEAD)   DoDeadIsDeadSave();
```
lynxlynxlynx: Up for debate: would it be cleaner to inline the check into the function? One included header less, perhaps in some other file too.

Build/Utils/_FrenchText.cc
```diff
@@ -2980,7 +2980,8 @@ static const wchar_t *s_fr_gzGIOScreenText[gzGIOScreenText_SIZE] =
 	L"Annuler",
 	L"En combat",
 	L"Temps illimité",
-	L"Temps limité"
+	L"Temps limité",
+	L"Dead is Dead"
```
lynxlynxlynx: Mort est mort (if the name stays)

lynxlynxlynx:
Hope that helps. :)
The name is fine with me and the same with "No mercy". The other two don't seem appropriate.

Build/SaveLoadGame.cc
```diff
@@ -160,7 +160,7 @@ static BYTE const* ExtractGameOptions(BYTE const* const data, GAME_OPTIONS& g)
 	EXTR_BOOL( d, g.fSciFi)
 	EXTR_U8(   d, g.ubDifficultyLevel)
 	EXTR_BOOL( d, g.fTurnTimeLimit)
-	EXTR_BOOL( d, g.fIronManMode)
+	EXTR_BOOL( d, g.ubGameSaveMode)
```
selaux: Shouldn't this be `EXTR_U8`?

Build/SaveLoadGame.cc
```diff
@@ -174,7 +174,7 @@ static BYTE* InjectGameOptions(BYTE* const data, GAME_OPTIONS const& g)
 	INJ_BOOL( d, g.fSciFi)
 	INJ_U8(   d, g.ubDifficultyLevel)
 	INJ_BOOL( d, g.fTurnTimeLimit)
-	INJ_BOOL( d, g.fIronManMode)
+	INJ_BOOL( d, g.ubGameSaveMode)
```
selaux: And here as well.

Build/SaveLoadGame.cc
```diff
@@ -219,7 +219,9 @@ BOOLEAN SaveGame(UINT8 const ubSaveGameID, wchar_t const* GameDesc)
 	InitShutDownMapTempFileTest(TRUE, "SaveMapTempFile", ubSaveGameID);
 
 	// Place a message on the screen telling the user that we are saving the game
-	{ UINT16 actual_w;
+	if (gGameOptions.ubGameSaveMode != DIF_DEAD_IS_DEAD)
```
selaux: Is there any reason not to display the message for dead is dead?

Build/GameSettings.h
```diff
@@ -57,6 +58,7 @@ enum
 struct GAME_SETTINGS
 {
 	INT8				bLastSavedGameSlot;							// The last saved game number goes in here
+	wchar_t     sCurrentSavedGameName[SIZE_OF_SAVE_GAME_DESC];                     // The name of the current Savegame goes here. Only relevant for Dead is Dead games
```
selaux: Can we use an UTF8 string for that instead? I would rather not introduce new wchar_t strings.

Build/SaveLoadScreen.cc
```diff
@@ -348,9 +381,89 @@ static void SelectedSaveRegionCallBack(MOUSE_REGION* pRegion, INT32 iReason);
...
+// Depending on:
+// In which tab is the last save
+// Are there any saves in the tab
+// It also deactivates tabs with no saves
+static void selectActiveTab()
+{
+	INT8 const lastSaveInTab = (INT8) gGameSettings.bLastSavedGameSlot / NUM_SAVE_GAMES;
+	for (INT8 i = 0; i < NUM_SAVE_GAMES_TABS; i++)
+	{
+		gfDiDTab = i; // Not strictly correct, as gfDiDTab is a boolean. This prepares the function for possible additional tabs without breaking the current Tab implementation.
```
selaux: Why not make it an `UINT8` then? Also I would rename it to `selectedTab` or something like that.

Build/SaveLoadScreen.cc
```diff
@@ -348,9 +381,89 @@ static void SelectedSaveRegionCallBack(MOUSE_REGION* pRegion, INT32 iReason);
...
+		bool tabHasSaves = FALSE;
+		for (INT8 j = 0; j != NUM_SAVE_GAMES; ++j)
+		{
+			if (gbSaveGameArray[j])
+			{
+				tabHasSaves = TRUE;
+				break;
+			}
+		}
+		if (tabHasSaves)
```
selaux: What about `gbActiveSaveGameTabs[i] = tabHasSaves`?

Build/SaveLoadScreen.cc
selaux: Also: This is again a two line ternary operator which is discouraged.

Build/SaveLoadScreen.cc
```diff
@@ -1473,12 +1675,60 @@ static void FailedLoadingGameCallBack(MessageBoxReturnValue const bExitValue)
...
+ // - we are in a Dialogue
+ // - we are in Meanwhile.....
+ // - we are currently in a message box - The Messagebox would be gone without selection after loading
+ if (gTacticalStatus.ubCurrentTeam == OUR_TEAM && !gfInTalkPanel && !gfInMeanwhile && !gfPreBattleInterfaceActive && guiPreviousOptionScreen != MSG_BOX_SCREEN) {
+			// Backup old saves
+			BackupSavedGame(gGameSettings.bLastSavedGameSlot);
+			// Save the previous option screen State to reset it after saving
+			ScreenID tmpGuiPreviousOptionScreen = guiPreviousOptionScreen;
+			// We want to save the current screen we are in. Unless we are in Options, Laptop, or others
+			guiPreviousOptionScreen = guiCurrentScreen;
```
selaux: As I see it this will always be overwritten inside the if so it does not make sense to set it here. It's also confusing since in the ifs `guiPreviousOptionScreen` is actually `guiCurrentScreen` and `tmpGuiPreviousOptionScreen` is the `guiPreviousOptionScreen`, so it would also make sense to change it in the ifs and set `tmpGuiPreviousOptionScreen` and `guiPreviousOptionScreen` as late as possible.

Build/Strategic/Meanwhile.cc
```diff
@@ -438,6 +442,7 @@ static void BeginMeanwhileCallBack(MessageBoxReturnValue const bExitValue)
 	else
 	{
 		// skipped scene!
+		gfInMeanwhile = FALSE;
```
selaux: Why was this added?

Build/Tactical/TeamTurns.cc
selaux: Or create a new function that wraps it and inline it there: `DoDeadIsDeadSaveIfNecessary`

Build/SaveLoadGame.cc
```diff
@@ -1466,6 +1479,33 @@ static void LoadSoldierStructure(HWFILE const f, UINT32 savegame_version, bool s
...
+		}
+		else
+		{
+			sprintf(zSourceBackupSaveGameName, "%s.%01d", zSourceSaveGameName, i);
+		}
+		sprintf(zTargetSaveGameName, "%s.%01d", zSourceSaveGameName, i+1);
+		// Only backup existing savegames
+		if (FileMan::checkFileExistance(i==0 ? GCM->getSavedGamesFolder().c_str() : backupdir.c_str(), zSourceBackupSaveGameName))
+		{
+			FileMan::copyFile(FileMan::joinPaths(i==0 ? GCM->getSavedGamesFolder().c_str() : backupdir, zSourceBackupSaveGameName).c_str(),
```
selaux: Wouldn't it be better to move the file instead of copying it? The old one does not need to stay intact.

selaux:
A short general question that comes quite late: Why the backup save games? Doesn't that somehow defeat the purpose of the DiD mode since you can basically load the backup if something went wrong?

selaux:
Another question: Did you think about introducing some abstraction for the save/load game tabs? Because there are a lot of ifs in that area.

RinneT:
Hi,
thanks for the plenty of feedback, I'm a bit thrilled :)
I'm having a very busy week and weekend, but hope I'll get to it next week.

The Save game backups: It is still possible that a savegame might end up corrupt and for that I prefer to have a backup. Imagine making it toward the end game and then the save breaks. Especially since gameplay with this mode is (at least for me) much much slower.
The frequency of saving (start and end of each turn) usually means you won't get out of lethal situations anyways.

I did also definitely think about abstraction for the save/load tabs. I was considering to push it to a new issue since I saw it was discussed elsewhere and properly implemented seems to be a bigger change. I'll see if I find very quick and fortunate ways while fixing.

Build/GameSettings.cc
RinneT: That had me a bit confused as well, however is done in the other Options as well. Basically GIO are the definitions for the Game Init Option Screen buttons and DIF are the saved difficulty variables. I remember finding a reason it was actually done that way but don't remember what it was.

Build/SaveLoadGame.cc
RinneT: Due to the frequency of saving, the popup message is getting very annoying very quickly.
It is still displayed in the message log.

Build/Strategic/Meanwhile.cc
RinneT: Saving while in a cutscene ("Meanwhile") causes errors in the save, so I check against it to avoid saving.
gfInMeanwhile is true when the game is in a cutscene. It is not set back to false if the player skips a cutscene. Thus the game always thinks it is in a cutscene, and would never save if a player skips one.

This is basically a vanilla error that had no effect previously, as the variable is never used alone.

Build/SaveLoadGame.cc
RinneT: Good find.

Build/Strategic/Meanwhile.cc
derSteFfi: Not setting it back sounds like a problem... I'll look into this, I used to know where the skipping happens :)

Build/SaveLoadScreen.cc
RinneT: As far as I know the INT8 conversion cuts off the positions after the decimal point, which always returns the right tab.

If there are possible compiler differences, I can make sure by using modf.

Build/GameSettings.cc
RinneT: So, the difference between DIF and GIO:
DIF is a GameSettings option and is universally used everywhere the game mode is checked.
GIO contains all values DIF contains, but in addition has the value NUM_SAVE_OPTIONS. This value is used in some loops used in the GameInitOptionsScreen.
This is true for all other options (NUM_GUN_OPTIONS, NUM_GAME_STYLES, NUM_DIFF_SETTINGS) and enhances readability a bit in these instances. I would leave it that way.

Build/SaveLoadGame.cc
RinneT: Any advice on how to prevent it?
This is the same way it is handled at other places.
Most other filepaths work with 256 char length.

E.g.:
	char zFileName1[256];
	sprintf(zFileName1, "%s/Auto%02d.%s", GCM->getSavedGamesFolder().c_str(), 0, g_savegame_ext);

Build/SaveLoadGame.cc
RinneT: Yes. At the moment saving within a cutscene causes the game to bug out if loading again.
E.g. in Deidrannas palace the roofs get opaque and you cannot see inside.
In the worst case the whole cutscene gets stuck forever.
This is why I prevent saving during a meanwhile at the moment, but the gfInMeanwhile variable is behaving quite odd in many cases. We should investigate further what the variable is actually doing and how it is used.
For the moment I am too afraid to break the rest of the game though and this workaround seems to have no side for the mode yet.

Build/SaveLoadScreen.cc
RinneT: state :)

Build/Utils/_FrenchText.cc
RinneT: thanks :)

Build/GameSettings.h
RinneT: Can I safely use UTF8 here?
This is a temporary variable that it used to copy data from the savegame header, which is wchar_t at the moment. I fear to break stuff here.

Build/SaveLoadScreen.cc
lynxlynxlynx: But won't it just convert the nominator, before the actual division?

Build/GameSettings.cc
lynxlynxlynx: Yes, I'm all for that, since the code reads odd now. However it doesn't need to be taken care of in this PR. I'll open a new bug as a reminder.

Build/SaveLoadGame.cc
lynxlynxlynx: Aren't you removing this function altogether?

Build/SaveLoadScreen.cc
RinneT: oh :)

Build/SaveLoadGame.cc
RinneT: You mean the backup function?
I believe it is necessary until at least all game breaking bugs are sorted out.
E.g. infinite loops that sometimes happen due to AI errors would cause this gamemode to break.
I would leave it.

lynxlynxlynx: ok. Since truncating the name is a bad idea, the target buffer has to be large enough to accommodate all the extra chars for the extension. So just increase that 512 and make sure the string is null terminated.

RinneT: Oh, I actually only just understand what the problem is here.
Got you totally wrong before.

According to the [sprintf reference](http://www.cplusplus.com/reference/cstdio/sprintf/) 
> A terminating null character is automatically appended after the content.

So I guess increasing the array size should suffice?

Build/SaveLoadScreen.cc
RinneT: It is.
The condition is true only if we switch from the Game Init Options Screen to the Save screen.
This is only possible with the new game mode selected in the Game Init Options Screen.
There is no other way of accessing the Save Screen from within the GIO Screen, and guiPreviousOptionScreen can never be GAME_INIT_OPTIONS_SCREEN in any other instance.

This was the most elegant solution I could find since the GameSettings cannot yet be read at this point.
Also tried a lot and could not get it to fail.

Build/SaveLoadGame.cc
lynxlynxlynx: For sprintf, but not strcpy. A large enough size + strncpy would take care of that.

Build/SaveLoadScreen.cc
lynxlynxlynx: The question was more about potential buffer overflows.

Build/SaveLoadGame.cc
Manuel-K: I don't see the problem, if the buffer is large enough.

strcpy keeps the terminating null character, which the source string contains, because it was createdd by sprintf.

lynxlynxlynx: indeed. :)

Build/SaveLoadScreen.cc
RinneT: oops,
but yes. both gGameSettings.sCurrentSavedGameName and gzGameDescTextField are initialized with SIZE_OF_SAVE_GAME_DESC

Build/SaveLoadScreen.cc
RinneT: I like the name :)

Build/SaveLoadScreen.cc
RinneT: Thanks.
This was indeed very complex and unclean code that took me a long time to understand again.
I simplified it and hope I didn't break anything in refactoring.

Figuring out what has to happen here in the first place was the biggest headache of the entire gamemode, btw.

Build/Tactical/TeamTurns.cc
RinneT: Used the wrapper function in GameSettings.h. It didn't change too much, but some headers could be removed. Also fits there as CanGameBeSaved() has it's home there as well.

Also reduces the risk of an accidental DiD Save without polling the mode before (which I found in one instance).

RinneT:
Replaced the Boolean gfDiDTab with a proper INT8 Tabnumber.

Since most of you already read into the code and gave very good feedback, is further code documentation still needed?

Build/SaveLoadGame.cc
```diff
@@ -160,7 +160,7 @@ static BYTE const* ExtractGameOptions(BYTE const* const data, GAME_OPTIONS& g)
 	EXTR_BOOL( d, g.fSciFi)
 	EXTR_U8(   d, g.ubDifficultyLevel)
 	EXTR_BOOL( d, g.fTurnTimeLimit)
-	EXTR_BOOL( d, g.fIronManMode)
+	EXTR_U8( d, g.ubGameSaveMode)
```
selaux: Does this work correctly for existing save games?

Build/SaveLoadScreen.cc
```diff
@@ -135,13 +147,15 @@ static ScreenID guiSaveLoadExitScreen = SAVE_LOAD_SCREEN;
...
 static BOOLEAN gbSaveGameArray[NUM_SAVE_GAMES];
+static BOOLEAN gbActiveSaveGameTabs[NUM_SAVE_GAMES_TABS];
 
 static BOOLEAN gfDoingQuickLoad = FALSE;
 
 //This flag is used to diferentiate between loading a game and saveing a game.
 // gfSaveGame=TRUE		For saving a game
 // gfSaveGame=FALSE		For loading a game
 BOOLEAN		gfSaveGame=TRUE;
+static INT8   gfActiveTab=FALSE;
```
selaux: INT8 = FALSE?

Build/SaveLoadScreen.cc
```diff
@@ -910,9 +1104,14 @@ static BOOLEAN DisplaySaveGameEntry(INT8 const entry_idx)
 	else
 	{
 		// If this is the quick save slot
-		wchar_t const* const txt = entry_idx == 0 ?
-			pMessageStrings[MSG_EMPTY_QUICK_SAVE_SLOT] :
-			pMessageStrings[MSG_EMPTYSLOT];
+		wchar_t txt[64];
+		if (entry_idx == 0 && gfActiveTab == 0)
```
selaux: Maybe `IsDeadIsDeadTab(gfActiveTab)`

Build/SaveLoadScreen.cc
```diff
@@ -994,7 +1219,7 @@ static void SelectedSaveRegionCallBack(MOUSE_REGION* pRegion, INT32 iReason)
 */
 
 		//If we are saving and this is the quick save slot
-		if( gfSaveGame && bSelected == 0 )
+		if( gfSaveGame && bSelected == 0 && gfActiveTab == 0)
```
selaux: Maybe `IsDeadIsDeadTab(gfActiveTab)`

Build/SaveLoadGame.cc
RinneT: Tested it on several saves, also ones I found in issues posted here. Everything worked correctly.
Since IronManMode was Boolean=TRUE before and saved as a "1", I don't see problems arising here either.

Build/SaveLoadScreen.cc
RinneT: Thanks, fixed it.

Build/SaveLoadScreen.cc
RinneT: That would actually be wrong here, since this is a check for the QuickSave slot which can only appear on the first page. So a check for the first page is more appropriate.

Build/SaveLoadScreen.cc
RinneT: That would actually be wrong here, since this is a check for the QuickSave slot which can only appear on the first page. So a check for the first page is more appropriate.

RinneT:
Thanks Stefan.
I think I have fixed the whitespace errors. This is always a bit confusing since they are handled differently throughout the project (e.g. most files use tabs, but FileMan uses whitespaces). Sometimes the formatting even changes throughout the file (First half of GameInitOptionsScreen is formatted differently than the last half).

I'm starting to rebase now, but guess it won't go very smoothly.

RinneT:
Done

lynxlynxlynx:
When in doubt, just use tabs, since there's more of them. We have an issue open to fix the stragglers, but it was pointless to do it before all the big PRs got merged.

PR #310, opened 2016-05-17
Update smacker.c

greg-kennedy:
Palette fixes from upstream (r26).

This fix from Bianca van Schaik will correct some odd palette artifacts in smacker videos.


selaux:
Looks fine to me. Thanks for the PR :+1:.

PR #311, opened 2016-05-20
Fixed stack corruption

vancv:
1, Function "AddEmailRecordToList" is called with parameter "text" which is always an array of length MAIL_STRING_SIZE.
therefore text[MAIL_STRING_SIZE] = 0; causes accessing an array out of bounds.
2, MAIL_STRING_SIZE length of pRecord should be sufficient. Maximum length that is read from data files is MAIL_STRING_SIZE (including trailing zero).

### DIFF
--- a/Build/Laptop/EMail.cc
+++ b/Build/Laptop/EMail.cc
@@ -53,7 +53,7 @@ struct Page
 
 struct Record
 {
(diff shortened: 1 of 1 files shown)


wendles:
Verified fixed, VS2015. Nice work.

selaux:
:+1: Works as expected.

PR #316, opened 2016-05-23
File endings fixes + changelog merge Fixes #313

### LINKED ISSUE #313: Some general stuff in the Stracciatella folder
I downloaded v0.15 and here are some small things I noticed (These issues were ofc present before the new version already).

- README.ENG and README.RUS still point to the old BB repo. Should be changed. Maybe include both the development page and the homepage?
- We have two changelogs in the Stracciatella folder: changelog.txt (old changelog) and changes.md (Stracciatella Continued). Maybe merge them into one file? Or make it clear which file is the right one?
- Why are there two .bat files for russian AND russian Gold version? Why is there only one .bat for english version? Confusing...
- Changelog.txt and \externalized\readme.txt are a mess when viewed in standard MS Notepad. Better programs like Notepad++ can view them without problems. Maybe change encoding of the files?

derSteFfi:


lynxlynxlynx:
Looks fine. The changelog merge is a bit pointless, but we do then have one file less.

selaux:
I added a commit that removes references to changelog from the builds. Please review and merge. Then I will rebuild the releases and push the tag forward.

PR #318, opened 2016-05-23
Sector exit

### LINKED ISSUE #314: Windowed mode: move to the next sector in tactical view.
When playing the game in windowed mode, I find it very hard to enter the next sector in tactical view. You need to move your mouse cursor to the edge of the screen. However, the it's extremely hard to get the "leave sector" cursor, because the mouse ends up leaving the JA2-window.
There is only a very, very small region in the window where you can get the "leave sector" cursor before actually going out of the window.
If we had an option to trap the mouse cursor inside the window when playing in windowed mode, it would be a big help!

derSteFfi:
fixes #314

### DIFF
--- a/Build/Tactical/Handle_UI.cc
+++ b/Build/Tactical/Handle_UI.cc
@@ -653,7 +653,7 @@ static void SetUIMouseCursor(void)
 				guiNewUICursor = NOEXIT_EAST_UICURSOR;
 			}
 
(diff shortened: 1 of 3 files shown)


selaux:
Tested and works fine. Just a question: Why were the `fOutRight` etc. checks removed?

derSteFfi:
not on purpose. I'll fix that tomorrow.

PR #323, opened 2016-05-25
Fix #322

### LINKED ISSUE #322: Any objection to removing WITH_MODS?
We manually set the define, so it has no effect. The code it supposedly toggles enables the use of the -mod switch to use a mod and nothing else. So it doesn't affect vanilla behaviour if you don't want any mods.

lynxlynxlynx:
fixes #322


lynxlynxlynx:
ups, forgot to use another branch for the manual mod merge, so all commits got pushed. No real harm.

PR #325, opened 2016-05-26
Update boost to 1.61.

selaux:
This updates the integrated boost dependency. While we might not have any immediate advantages (except for possible bug fixes), this needs to be done to eventually build on Android.


lynxlynxlynx:
yikes, 60k files.

selaux:
Wow, I didn't realize the modules were cherry-picked before. Maybe we can hard reset master and do the cherry-picking again?

lynxlynxlynx:
Not a bad idea — I didn't notice the whatchanged count is so asymmetrical.

PR #326, opened 2016-05-26
Update boost to 1.61.

selaux:
Still a lot of changes, but should be better than last one.


lynxlynxlynx:
go for it, once travis is happy. I guess the project just grew so much in the meanwhile, that it's still almost a doubling of included code?

selaux:
Seems like it, it's also hard to find out which headers actually are used, so there are probably more headers included than necessary.

lynxlynxlynx:
I think I used cppclean for that before, but it was pretty noisy. Doing a quick search this popped up, but I haven't used it, so no idea if it's any better or worse: https://sourceforge.net/projects/chksem/

selaux:
I could not find anything that does what I want so I created a small python script: https://gist.github.com/selaux/f1eb93fde855ed12ac6973171da1ae56

The added/removed lines ratio looks much better now.

lynxlynxlynx:
Nice, now it's a significant reduction.

PR #339, opened 2016-06-06
integrate gui_extras

derSteFfi:
as previously done by vxx and 1.13, switchable via gui_extras in game.json

### DIFF
--- a/Build/Strategic/MapScreen.cc
+++ b/Build/Strategic/MapScreen.cc
@@ -692,7 +692,6 @@ static void PrintStat(UINT32 change_time, UINT16 const stat_gone_up_bit, INT8 st
 
 	if(GCM->getGamePolicy()->gui_extras)
 	{
-		SGPVSurface::Lock l(guiSAVEBUFFER);
 		ProgressBarBackgroundRect(x+1,y-2,15*progress/100,10,0x514A05,progress);
 	}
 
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
It would be cleaner if everything used the progress variant and then it checked for the presence of the flag.

Also, looks like a candidate for further refactoring, since the stat blocks are practically identical.

RinneT:
Looks interesting, do you have more specific info on what it does?

Also: I may be stupid but I can't seem to find it on my git. git fetch origin doesn't add it to my branch list.
Any input on what I'm doing wrong? Kind of new to all the git stuff (though I love it).

derSteFfi:
@lynxlynxlynx you're right. I'll change that as soon as I get ot it.

lynxlynxlynx:
What about the same in Build/Tactical/Interface_Panels.cc? It also looks like the PrintStat function could be shared.

derSteFfi:
i refactored it for the start. I'll try to use the One from Tactical in Map_screen now

derSteFfi:
So, not that easy. The handling of the coordinates is different. In Tactical, the coordinates point to the start of the short string for each stat (STR, AGI, ...) in Strategic it's the start of the actual value. Cleaning this up would be quite messy I guess because one would need to adapt the drawing of all strings for one of both...

PR #349, opened 2016-06-20
Fixme remove mem total

derSteFfi:
its always 0.


selaux:
LGTM. By the way: If you want to change your last commit, you could also use `git commit --amend`.

PR #354, opened 2016-06-26
Introduce cmake build system

selaux:
Should be on par with the current build system, some additional perks:

- Windows release now also builds a installer package
- macOS release can be built in a VM just as the other releases (only on Apple machines though due to licensing)
- Can create XCode and MSVC solutions
- Packaging is automatically done using cpack

There are still some remaining issues, but feel free to play around. Cmake is building out-of-source so a separate directory needs to be created in which the build will reside:

```
mkdir _cmake_build && cd _cmake_build && cmake .. && make
```

Switches  can be passed  using `-D`:

```
cmake -DLOCAL_BOOST_LIB=ON ..
```


lynxlynxlynx:
cmake supports globbing. Since the file structure is unlikely to change much, it could be used to cut the source lists short.

Manuel-K:
Using glob can cause problems if the file structure changes at any time. It's very likely that someone will forget to call cmake after a `git pull` that adds new files and will encounter linker problems.


The cmake manual warns against using glob for source files: 
> Note:  We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

selaux:
Everything should be done and working now. Please test, I will also add some releases built with cmake which should be checked.

[ja2-stracciatella_0.15.x_amd64.deb.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/380489/ja2-stracciatella_0.15.x_amd64.deb.zip)
[ja2-stracciatella_0.15.x_i386.deb.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/380488/ja2-stracciatella_0.15.x_i386.deb.zip)
[ja2-stracciatella_0.15.x_macos.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/380487/ja2-stracciatella_0.15.x_macos.zip)
[ja2-stracciatella_0.15.x_win-mingw.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/380486/ja2-stracciatella_0.15.x_win-mingw.zip)

_build/buildboxes/common/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+[ -d /tmp/build ] && rm -rf /tmp/build
+mkdir /tmp/build
+cd /tmp/build && cmake -DCMAKE_BUILD_TYPE=Release $2 /tmp/ja2-stracciatella
+cd /tmp/build && make $1 -j2
```
lynxlynxlynx: one cd is enough

_build/buildboxes/openbsd-5.9/bootstrap.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# installing missing xbase55
+test -f /usr/X11R6/README || (cd /tmp && curl -O http://ftp.openbsd.org/pub/OpenBSD/5.9/amd64/xbase59.tgz)
+test -f /usr/X11R6/README || (cd /tmp && tar -C / -xzphf xbase59.tgz)
```
lynxlynxlynx: this would also be better in one line, in case the download fails.

lynxlynxlynx:
Looks good to me, but I haven't tested.

The only odd thing is _bin. Why not "build" like everyone else uses? The two existing ones will be confusing either way and it's less a hassle to type.

selaux:
That is exactly why :wink:. Having `_build`, `Build` and `build` wont work on Windows since `build` and `Build` is the same. Also it looks way too confusing to me. So I chose `_bin`, which is only a suggestion since you could also use any other directory, but I won't be included in the gitignore. We could also think about seizing the opportunity and getting rid of the `Build`/`_build` confusion.

selaux:
A few this I wanted to note but forgot:
- deb packages are built using local boost since we would need to build multiple packages for each Debian/Ubuntu otherwise because they boost is versioned by package name in Debian
- Headers are included via globbing since they are only needed for MSVC and XCode and wont break the build otherwise when not added to the sources

lynxlynxlynx:
Yes, that would be my suggestion. To make it worse, we already have a separate src/ dir (only new stuff?). How about:
_build -> release
src -> Build/externalised
Build -> src

selaux:
I like the idea :+1: and we also should not forget about `sgp`. What about:

- `_build/lib-*` -> `dependencies`
- `_build/buildboxes` -> `release`
-  the rest of `_build` -> `assets`
- `externalized` -> `assets/externalized`
- `Build` -> `src/game`
- `src` -> `src/externalized`
- `sgp` -> `src/sgp`

Maybe we could also rethink the name `externalized`, maybe `content` is better?

lynxlynxlynx:
I agree and maybe "data" instead of assets. And maybe someday game.json out of it to toplevel or config/ or something.

Manuel-K:
Sane folder names? Yes, please.

derSteFfi:
big plus here, too. I would also be fine to do that in a new pr, and starting about 3 days from now, i will finally have some time to spend on ja2 again, so I could do that, if @selaux doesn't have the time for it...

selaux:
Lets do that in a new PR. Large other merges should be done first, otherwise it might get complicated.

selaux:
I'm getting some segfaults and build errors randomly since I added the headers to the sources. Can anybody reproduce or does anybody have an idea where this might come from?

selaux:
Found some remaining issues, now this is good to merge from my side. I cannot reproduce the issues I had with compilation before, so it might have been some temporary issue on my machine.

lynxlynxlynx:
We didn't deprecate the internal boost yet, did we? Neither master nor the cmake branch builds for me. master fails due to an include using the <system> syntax, cmake because FindBoost doesn't find everything it wants (I don't have all of boost installed).

selaux:
`boost-system` and `boost-filesystem` should be the only two boost dependencies that are required. What exactly does cmake not find?

lynxlynxlynx:
This is from the old system, since I'd have to dig more to find it for cmake:
boost/foreach.hpp which is from the main lib.
boost/shared_ptr.hpp
boost/filesystem.hpp
boost/smart_ptr.hpp

All exist in _build.

selaux:
With cmake the default is to use system boost. To make it local boost you have to specify it in the cmake command: `cmake -DLOCAL_BOOST_LIB=ON ..`

lynxlynxlynx:
ah, right. It builds fine then and ja2 runs ok. So for me, an extra reason for a quicker merge.

Manuel-K:
It builds and runs on Arch Linux x86_64. Disclaimer: I only did a quick test.

selaux:
*Closing my eyes hovering over the merge button*

lynxlynxlynx:
Woohoo, thanks for all the work!

PR #359, opened 2016-08-02
Fix #300: Show version number in main menu

### LINKED ISSUE #300: Idea - version no in main menu?
Another cheap mockup from me :)
main

Stracciatella could display the current version no in the game menu.
Of course, this would be purely eye-"candy" and had no real use, but I imagine it would make the Stracciatella-experience feel a bit more legit right from the start.

It's just an idea. What do you think?

selaux:
Removes it from the splash instead otherwise it looks weird with the copyright fading in.

### DIFF
--- a/Build/JAScreens.cc
+++ b/Build/JAScreens.cc
@@ -99,7 +99,6 @@ void DisplayFrameRate( )
 	}
 }
 
-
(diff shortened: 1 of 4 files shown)


selaux:
Made the functions local to `MainMenuScreen.cc`

PR #362, opened 2016-08-07
[RFC] External dependencies

selaux:
This PR builds lib boost from a tarball. This should theoretically be possible for all our dependencies (might need some patching of the sources since they do not build and install as easily as boost).

Advantages:
- Less clutter in our code from dependencies
- Less incentive to change anything in the libraries

Disadvantages:
- IDEs might not pick up in the included files
- Without modifying tarballs they can be huge (e.g. for boost 80M)

Things to note:
- We could also pull the files from the internet during the build, cmake supports that


lynxlynxlynx:
Unless the rest are pure, any time we update them, we'll have extra work anyway. Boost subsetting should be eventually scripted to be faster, if we ever need to update it again. Then it doesn't matter anymore how it is stored. FLTK is another big one.

Why wouldn't IDE's pick them up? They can be extracted to the same locations.

selaux:
They are not simply extracted but build separately and then installed into `${CMAKE_BUILD_DIR}/deps`. So this build needs to be run before the IDE can find the includes.

lynxlynxlynx:
But don't build target dependencies take care of that, so it would be all ok after the first build?

selaux:
Yep, it would. Do you think we could use `git lfs` for build dependencies?

lynxlynxlynx:
Don't know what that is, but I'm sure it's not Linux from scratch.

selaux:
[git lfs](https://git-lfs.github.com/) is a git extension that handles large binary files efficiently. This extension is not standard git but should be available for most systems and the backend is provided by github. Generally we should only use it for large dependencies (e.g. boost or gtest), so they are only pulled when needed (e.g. on build systems or Windows)

Manuel-K:
I think we should bump this to the release after the next one.

PR #369, opened 2016-09-03
Fix definitions for XCode

selaux:
XCode generation resulted in broken XCode file otherwise

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,28 +42,25 @@ option(WITH_MAEMO "Build with right click mapped to F4 (menu button)" OFF)
 ## Build
 
 set(BUILD_SHARED_LIBS OFF CACHE BOOL "")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -42,28 +42,25 @@ option(WITH_MAEMO "Build with right click mapped to F4 (menu button)" OFF)
 ## Build
 
 set(BUILD_SHARED_LIBS OFF CACHE BOOL "")
-set(CFLAGS "${CFLAGS} -DJA2 -DGAME_VERSION=\\\"${GAME_VERSION}\\\" -DMICROINI_STATIC")
+add_definitions(-DJA2)
+add_definitions(-DGAME_VERSION="\\"${GAME_VERSION}\\"")
```
lynxlynxlynx: you lost -DMICROINI_STATIC

selaux: Yes, that should not be necessary anymore since it is built in a different project...

lynxlynxlynx: Ok, then it looks good to me.

PR #372, opened 2016-09-04
Update build dependencies for SDL2

selaux:
Separate from the other PR (#216) to emphasize on actual changes.


lynxlynxlynx:
So the other PR is a blocker for this one, right? Especially wrt 0.15.1.

selaux:
Indeed.

PR #373, opened 2016-09-05
enable assertive asserts only on debug builds

lynxlynxlynx:
to do away with all the hiccups, but Debug users should be fine with it.


selaux:
LGTM.

Issue #363, opened 2016-08-14
consolidate GIO_ and DEF_ constants


lynxlynxlynx:
From the Dead is Dead review:
> So, the difference between DIF and GIO:
> DIF is a GameSettings option and is universally used everywhere the game mode is checked.
> GIO contains all values DIF contains, but in addition has the value NUM_SAVE_OPTIONS. This value is used in some loops used in the GameInitOptionsScreen.
> This is true for all other options (NUM_GUN_OPTIONS, NUM_GAME_STYLES, NUM_DIFF_SETTINGS) and enhances readability a bit in these instances. I would leave it that way. We could debate if the NUM_ values should be taken over to the GameSettings.

So merge the two, so the code ambiguity and duplication can go away. Rename if needed.

PR #385, opened 2016-09-24
Consolidate difficulty constants

geca:
As per #363.


Build/GameInitOptionsScreen.cc
```diff
@@ -288,14 +277,7 @@ static void EnterGIOScreen()
...
 		INT16 const y = GIO_DIF_SETTINGS_Y - GIO_OFFSET_TO_TOGGLE_BOX_Y;
-		size_t def;
-		switch (gGameOptions.ubDifficultyLevel)
-		{
-			case DIF_LEVEL_EASY:   def = GIO_DIFF_EASY; break;
-			default:
-			case DIF_LEVEL_MEDIUM: def = GIO_DIFF_MED;  break;
-			case DIF_LEVEL_HARD:   def = GIO_DIFF_HARD; break;
-		}
+		size_t def = gGameOptions.ubDifficultyLevel - 1;
```
selaux: Why is it -1 here?

Build/GameSettings.h
```diff
@@ -73,11 +73,10 @@ struct GAME_SETTINGS
 //Enums for the difficulty levels
 enum
 {
-	DIF_LEVEL_ZERO,
-	DIF_LEVEL_EASY,
-	DIF_LEVEL_MEDIUM,
-	DIF_LEVEL_HARD,
-	DIF_LEVEL_FOUR,
+	DIF_LEVEL_EASY = 1,
```
selaux: Can't it be starting at 0?

geca: It's for backward compatibility. These values are being assigned to `ubDifficultyLevel ` which is then used in couple of computations and array accesses that assume 1-based difficulty. For example
`curr->ubNumCreatures = (UINT8)(2 + gGameOptions.ubDifficultyLevel*2 + Random( 2 ))`. So I decided to stay conservative instead of risking "off by one" bugs.

Build/GameInitOptionsScreen.cc
geca: Previously, 1-based difficulty level was converted to 0-based (1-0, 2-1, 3-2). We can achieve the same thing by just subtracting 1 from 1-based difficulty.

Build/GameSettings.h
selaux: Where does this assignment happen, because there seems to be no change for that and before both enums were `0`-based.

geca: What I meant with zero/one-based difficulty (not enum) is that 'easy' equals 0 with GIO_ enum and 1 with DIF_ enum. I've decided to keep it that way with DIF_ enum to ensure backward compatibility. I removed DIF_LEVEL_ZERO (wasn't used anywhere) and started numbering DIF_ enum with with 1 instead.

selaux: 👍

Issue #386, opened 2016-09-25
Request: more music variety


JAsmine-ja2:
Some events in the game (turn-based battle [battle], real-time battle [tensor] and peacefully exploring [nothing]) have several music tracks, which are played at random selection.

It is possible to add new music to JA2 by replacing existing files in data\music. It is however **not** possible to increase the number of tracks for specific events by this method.

It seems like BP user Taro M made some changes to the sourcecode which allow for more different tracks for the turn-based battle event.
http://thepit.ja-galaxy-forum.com/index.php?t=msg&th=22790&goto=343506&#msg_343506

His changed source is here:
[JA2 more battle music source files.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/491567/JA2.more.battle.music.source.files.zip)

This feature would make a nice addition to Stracciatella. The feature itself doesn't change the game at all. The player won't even notice that it's there. Yet, it allows for modders to be more flexible since they don't have to overwrite stuff

From looking at the code, I can't tell how elaborated this is and how easy/hard it is to implent this for Stracciatella.

remaing questions/issues:
- what happens if not all music slots are filled?
- more slots needed for "nothing" and "tensor", too. Maybe 10 or so slots for each?

PR #387, opened 2016-09-25
Moddable music

selaux:
Based on #386 and should fix it.

- Music resources are defined in `music.json`
- There are different music modes which can have a list of music files attached to them
- No limiting on the number of possible music pieces
- Pieces are chosen by a uniform distribution


JAsmine-ja2:
Nice effort! Looks very nice, especially with the control via json.

lynxlynxlynx:
LGTM.

derSteFfi:
lgtm too.

PR #389, opened 2016-09-25
Use integer abs function where necessary

selaux:
Follow up of  #388 i replaced all `abs()` that were not explicitly `fabs()` with a macro that does a integer abs.

This seems to fix (at least) the AI deadlocks occuring during enemy turns.

### DIFF
--- a/Build/TileEngine/Lighting.cc
+++ b/Build/TileEngine/Lighting.cc
@@ -537,9 +537,9 @@ static DOUBLE LinearDistanceDouble(INT16 iX1, INT16 iY1, INT16 iX2, INT16 iY2)
 {
 INT32 iDx, iDy;
 
-	iDx=abs(iX1-iX2);
+	iDx=ABS(iX1-iX2);
 	iDx*=iDx;
-	iDy=abs(iY1-iY2);
+	iDy=ABS(iY1-iY2);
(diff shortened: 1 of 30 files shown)


sgp/Types.h
```diff
@@ -24,6 +24,7 @@
 #define endof(a) ((a) + lengthof(a))
 
 
+#define ABS(a) (UINT16)abs((int)a)
```
lynxlynxlynx: Should be (int)(a) or simply int(a) so the result is cast, not only the first argument in non-trivial expressions.

And while the change seems automated, more whitespace errors crept up.

selaux: Should both be fixed. Thanks for the review.

Build/TileEngine/Lighting.cc
```diff
@@ -537,9 +537,9 @@ static DOUBLE LinearDistanceDouble(INT16 iX1, INT16 iY1, INT16 iX2, INT16 iY2)
 {
 INT32 iDx, iDy;
```
derSteFfi: maybe we should make those unsigned... the parsing might give errors.

derSteFfi: of course not really part of this PR

PR #390, opened 2016-09-26
Fix manpage installation

AMDmi3:
- DESTINATION share/man/man6/ja2.6 is incorrect, it creates named
  directory instead of renaming the file. Use RENAME instead
- Introduce condition for FreeBSD-specific manpage directory. Other
  *BSDs may use it as well


drm00:
While at it, could you add *OpenBSD* to the list of systems? I'm porting ja2 to openbsd and currently patch the man path manually.  Thanks!

AMDmi3:
Done!

lynxlynxlynx:
Ideally we would check for "*BSD", as there are some others, but I'm fine with incremental improvements.
Thanks!

PR #391, opened 2016-09-26
Unbundle gtest

AMDmi3:
On FreeBSD, ja2-stracciatella build fails somewhere in gtest. I haven't investigated thoughtfully, but since FreeBSD port of gtest includes some patches, I assume these are needed here as well. However, the correct solution would be to unbundle gtest and always use system one. With this change (use system gtest through detection script which comes with CMake), the project builds fine out of box on FreeBSD. The change also removes bundled gtest as it is no longer needed.


AMDmi3:
Hmm, not sure why it can't find the dependency in Travis.

lynxlynxlynx:
This should not remove the bundled gtest, just prefer the system one (it's used also for other closed oses). You can look at how we handle boost.

Travis is using ununtu 14.04 iirc, so perhaps the name was different then or wasn't yet packaged.

selaux:
Thanks for the PR,  I'm pro having the possibility to use external dependencies and that is another step into that direction. As @lynxlynxlynx mentioned we unfortunately need to bundle some dependencies for other OSs than Linux/BSD, especially Windows 😉.

Not sure about the Travis stuff either. From the logs it installs the dependency just fine...

Manuel-K:
Google warns about using a system-wide gtest installation: https://github.com/google/googletest/blob/master/googletest/docs/FAQ.md#why-is-it-not-recommended-to-install-a-pre-compiled-copy-of-google-test-for-example-into-usrlocal

lynxlynxlynx:
But distributions do just that. Source based ones have to deal with such things by default, so I don't see a problem.

Manuel-K:
The option to use a system-wide installation of gtest is a good idea, but the article I linked to gives the impression that for this library the bundled version should be preferred (same for e.g. libsmacker).

lynxlynxlynx:
Yeah, but it's not such a big issue, since you'd need to switch compilers to potentially get in trouble. That's what my previous post was about (compiler homogeneity), not the option.

janisozaur:
The linked FAQ is missing one point: there's `gtest-config` which should let you use whatever configuration gtest was built with as a package. At least on Arch there is.

lynxlynxlynx:
@Manuel-K externalised the choice without removing the bundled version, closing.

PR #396, opened 2016-09-29
Properly rerender subtotal when changed

### LINKED ISSUE #394: Shipping cost BR not redrawn properly
bobby1
bobby2

Notice how the numbers are somehow screwed up when changing the cheaper shipping cost. This can be triggered as well when changing around the destination (different destinations have diferent prices).

Interesting: This bug can be observed in vanilla JA2, too!

selaux:
Fixes #394.

Increase dirty rectangle size


lynxlynxlynx:
Eh, I was looking at the same thing. 2 is enough (1 isn't).

selaux:
Damit 😄 , changed it to 2.

selaux:
I'm trying now to implement some debug mode where one can see the outlines of blitted areas.

PR #403, opened 2016-09-30
Missing money description box fix

### LINKED ISSUE #379: Description box missing when opening "Deposit/Withdraw money" widget in Tactical screen
Load this [savefile](https://github.com/ja2-stracciatella/ja2-stracciatella/files/473470/money-glitch.sav.zip) using 0.15.1 and bring up "Deposit/Withdraw money" widget (dollar icon). Description box is missing, only the buttons are displayed as shown on the screenshot bellow.

This looked like a perfect trivial bug for my first pull request. Boy was I wrong. :) What follows is my analysis of the bug with couple of suggestions/questions on how to fix it. I'd really appreciate your input.

jagged alliance 2_003

It turns out the description box renders properly if you enter and exit strategic screen before clicking on the dollar icon in tactical. Let's call this "Scenario 2", "Scenario 1" is when we skip strategic screen and just open the money widget after loading the game.

If we want to render the entire description box, `fInterfacePanelDirty` must be set to `DIRTYLEVEL2` (so we re-render the whole interface as opposed to `DIRTYLEVEL1` which only re-renders bullets and status). 

Bellow code snippet from `Strategic_Turns.cc/HandleStrategicTurn() ` that eventually decides whether dirty level is set to 1 or stays 2:

```
if ( ( uiTime - guiLastTacticalRealTime ) > uiCheckTime ) 
{ 
  HandleTacticalEndTurn( ); 
  guiLastTacticalRealTime = uiTime; 
}
```
In Scenario 2 (where we toggle strategic screen before we open money widget) this evaluates to false, with Scenario 1 the condition is met however. These functions are then called in sequence: 

```
HandleTacticalEndTurn (Tactical_Turns.cc) 
EVENT_BeginMercTurn (Soldier_Control.cc) 
HandleEndTurnDrugAdjustments (Drugs_And_Alcohol.cc) 
```
Yep, all the way to drugs'n'alco. :) The following condition is met inside `HandleEndTurnDrugAdjustments` that sets `fInterfacePanelDirty` to DL1, meaning the rendering of the description box will be skipped later on. 

```
if ( pSoldier->bDrugSideEffect[ cnt ] <= 0 ) 
{ 
  pSoldier->bDrugSideEffect[ cnt ] = 0; 
  fInterfacePanelDirty = DIRTYLEVEL1; 
} 
```
The quick fix would be changing  `fInterfacePanelDirty` in the snippet above to `DIRTYLEVEL2` and calling it a day.

On the other hand, maybe we should be skipping `HandleTacticalEndTurn` altogether as we do in the second test scenario. Circling back to the first snippet --

```
if ( ( uiTime - guiLastTacticalRealTime ) > uiCheckTime ) 
{ 
  HandleTacticalEndTurn( ); 
  guiLastTacticalRealTime = uiTime; 
}
```
-- the reason why Scenario 1 executes `HandleTacticalEndTurn` and Scenario 2 does not is `giTimeCompressMode`. After loading the game, compression mode is set to `TIME_COMPRESS_X1`, later ensuring the condition `( uiTime - guiLastTacticalRealTime ) > uiCheckTime` is met. When we enter/exit map screen, compression mode is set to `TIME_COMPRESS_60MINS`, and therefore later skipping  `HandleTacticalEndTurn`.

What do you guys think should be the right approach in dealing with this bug? What is expected behavior (i.e. time compression mode) when we are in tactical screen and is it supposed to change if you enter/exit map screen which explicitly says it is being paused? Now that I think of it: none of the mercs in the savegame is "under influence" so `HandleEndTurnDrugAdjustments` should probably return immediately anyway. However, we would still probably get only partial money widget when bringing it up on drugged merc ...

geca:
Fixes #379.

### DIFF
--- a/Build/Tactical/Drugs_And_Alcohol.cc
+++ b/Build/Tactical/Drugs_And_Alcohol.cc
@@ -205,9 +205,9 @@ BOOLEAN ApplyDrugs( SOLDIERTYPE *pSoldier, OBJECTTYPE *pObject )
 		ScreenMsg( FONT_MCOLOR_LTYELLOW, MSG_INTERFACE, pMessageStrings[ MSG_MERC_TOOK_DRUG ], pSoldier->name );
 	}
 
-  // Dirty panel
-  fInterfacePanelDirty = DIRTYLEVEL2;
-
+        // Dirty panel
+        DirtyMercPanelInterface(pSoldier, DIRTYLEVEL2);
+       
 	return( TRUE );
 }
 
@@ -229,7 +229,7 @@ void HandleEndTurnDrugAdjustments( SOLDIERTYPE *pSoldier )
 			if ( pSoldier->bDrugSideEffect[ cnt ] <= 0 )
 			{
 				pSoldier->bDrugSideEffect[ cnt ] = 0;
-				fInterfacePanelDirty	= DIRTYLEVEL1;
+                                DirtyMercPanelInterface(pSoldier, DIRTYLEVEL1);
 			}
 		}
 
@@ -243,8 +243,8 @@ void HandleEndTurnDrugAdjustments( SOLDIERTYPE *pSoldier )
 			{
 				pSoldier->bDrugEffect[ cnt ] = 0;
 
-        // Dirty panel
-        fInterfacePanelDirty = DIRTYLEVEL2;
+                                // Dirty panel
+                                DirtyMercPanelInterface(pSoldier, DIRTYLEVEL2);                                
 
 				// Start the bad news!
 				pSoldier->bDrugSideEffectRate[ cnt ] = ubDrugSideEffectRate[ cnt ];
--- a/Build/Tactical/Drugs_And_Alcohol.cc
+++ b/Build/Tactical/Drugs_And_Alcohol.cc
@@ -11,17 +11,17 @@
 #include "Interface.h"
 #include "Timer_Control.h"
 
-UINT8 ubDrugTravelRate[] = {4, 2};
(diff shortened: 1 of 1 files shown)


Build/Tactical/Drugs_And_Alcohol.cc
```diff
@@ -243,8 +243,8 @@ void HandleEndTurnDrugAdjustments( SOLDIERTYPE *pSoldier )
 			{
 				pSoldier->bDrugEffect[ cnt ] = 0;
 
-        // Dirty panel
-        fInterfacePanelDirty = DIRTYLEVEL2;
+                                // Dirty panel
+                                DirtyMercPanelInterface(pSoldier, DIRTYLEVEL2);
```
derSteFfi: you left some trailing whitespaces in here.

geca: [3rd commit](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/403/commits/decfafd6dc7d576b9f166317dd2ab51d8e1f3334) fixed that.

derSteFfi: yep, i was confused.

geca:
Should I ditch this pull request and make another without formatting? The fix is really simple, since the setter (see #403) is already provided in form of `DirtyMercPanelInterface`. So it's just replacing `fInterfacePanelDirty` assignments with this function call.

lynxlynxlynx:
Ideally yes and with the last two squashed (one part into the first commit).

geca:
Will do.

PR #411, opened 2016-10-05
Update rapidjson to 1.1.0 and add comments to game.json

selaux:


lynxlynxlynx:
Looks good, but I can't test it. So if it works, I'd just merge it.

lynxlynxlynx:
I'm back and this works fine. :)
Note to people running it from the build dir - edit the game.json in this one's externalised dir, not the source one.

selaux:
Hm, it should copy the externalized dir on build but maybe that does not work as expected.

lynxlynxlynx:
it does, but if you edit the source after that, it can't have an effect.

PR #414, opened 2016-10-06
Fixes dead is dead saves when exiting game

selaux:
The addiotional checks dont really make sense here since the second branch does not depend on tmpGuiPreviousOptionScreen.

@RinneT Can you have a look at this? Currently most of my saves get corrupted.

### DIFF
--- a/Build/SaveLoadScreen.cc
+++ b/Build/SaveLoadScreen.cc
@@ -1731,13 +1731,13 @@ void DoDeadIsDeadSave()
 		ScreenID tmpGuiPreviousOptionScreen = guiPreviousOptionScreen;
 		// We want to save the current screen we are in. Unless we are in Options, Laptop, or others
 		// Make sure we are always in a sane screen.
-		if (guiCurrentScreen != MAP_SCREEN && guiCurrentScreen != GAME_SCREEN && tmpGuiPreviousOptionScreen != MAP_SCREEN && tmpGuiPreviousOptionScreen != GAME_SCREEN)
-		{
-			// If all fails, go to the map screen, this (almost) guarantees the game will start
-			guiPreviousOptionScreen = MAP_SCREEN;
-		} else
-		{
-			guiPreviousOptionScreen = guiCurrentScreen;
+		if (tmpGuiPreviousOptionScreen != MAP_SCREEN && tmpGuiPreviousOptionScreen != GAME_SCREEN) {
+			if (guiCurrentScreen != MAP_SCREEN && guiCurrentScreen != GAME_SCREEN) {
+				// If all fails, go to the map screen, this (almost) guarantees the game will start
+				guiPreviousOptionScreen = MAP_SCREEN;
+			} else {
+				guiPreviousOptionScreen = guiCurrentScreen;
+			}
 		}
 
 		BOOLEAN tmpSuccess = SaveGame(gGameSettings.bLastSavedGameSlot, gGameSettings.sCurrentSavedGameName);


RinneT:
Oh there we have my least favourite piece of code again.
I don't remember the whole thing, but the checks were vital.
The save function decides which screen to use (current or previous options) depending on where you are and what you did (e.g. if you are in the options save load screen, it cannot take the currentScreen). I don't really know how it's done exactly.

Since we can now save in screens we couldn't before (e.g. laptop screen) by leaving the game, it can quickly end in an invalid screen, which usually means you are taken to the main menu after loading.
This is especially prominent immediately after starting a new game.
The additional checks make sure all screens that could be used are valid which means either Map Screen or Game Screen. I think this piece of code was discussed before in the PR.

What happens when you try to load the corrupted saves? Could you attach one?

selaux:
Steps to reproduce save game 11:
- Start deadisdead game with current master
- Immediately exit game
- Load Game again
- Load Screen Repeats on Loading

Steps to reproduce save game 12:
- Load Game
- Immediately exit game
- Load Game again
- Game crashes

[SavedGames.tar.gz](https://github.com/ja2-stracciatella/ja2-stracciatella/files/514760/SavedGames.tar.gz)

RinneT:
So apparently I've been using a wrong executable ever since the merge (the old make executable, not the cmake one). Can reproduce it now and it seems somewhat familiar. Especially with the flickering buttons after loading save game 11. I'll check if I missed something or actually messed up during the rebase.

Build/SaveLoadScreen.cc
```diff
@@ -1731,12 +1731,11 @@ void DoDeadIsDeadSave()
 		ScreenID tmpGuiPreviousOptionScreen = guiPreviousOptionScreen;
 		// We want to save the current screen we are in. Unless we are in Options, Laptop, or others
 		// Make sure we are always in a sane screen.
-		if (guiCurrentScreen != MAP_SCREEN && guiCurrentScreen != GAME_SCREEN && tmpGuiPreviousOptionScreen != MAP_SCREEN && tmpGuiPreviousOptionScreen != GAME_SCREEN)
+		if (guiCurrentScreen != MAP_SCREEN && guiCurrentScreen != GAME_SCREEN)
```
RinneT: What you actually want to do here is:
`if (tmpGuiPreviousOptionScreen != MAP_SCREEN && tmpGuiPreviousOptionScreen != GAME_SCREEN)
{
	if(guiCurrentScreen != MAP_SCREEN && guiCurrentScreen != GAME_SCREEN )
	{
		// If all fails, go to the map screen, this (almost) guarantees the game will start
		guiPreviousOptionScreen = MAP_SCREEN;
	} else {
		guiPreviousOptionScreen = guiCurrentScreen;
	}
}`

The reason is that guiPreviousOptionsScreen is the one that is written to the savegame. By checking guiCurrentScreen as well (which can contain a valid Screen in situations where guiPreviousOptionsScreen does not) you increase the chance of getting the player back to where he/she was.

RinneT:
Added a comment to fix the problem with save game 12.

Save game 11 I can reproduce by using your save. However I can't reproduce it myself. When exactly do you exit the game to the main menu and how?
I.e. I don't know of any shortcut. After the cutscene I get to the laptop screen. From there the problem of save game 12 applies.

Did you restart the game after testing save game 12?
If not it has left the game in a broken state even after quitting to the main menu. No variables are flushed.

selaux:
About save game 11: Just start a game and without doing anything, go to options screen and quit again.

RinneT:
Did you close the laptop before going to the options screen? (Pressing o in the laptop screen doesn't work). If so, it doesn't happen anymore with the changes posted above.

selaux:
Yes, closing is necessary of course, sorry 😄

RinneT:
Then the above applies :)
Is there any way to include proper formatting (linebreaks and indents) in the github code comment?
As the compressed version looks messy.

Basically just put:

if (tmpGuiPreviousOptionScreen != MAP_SCREEN && tmpGuiPreviousOptionScreen != GAME_SCREEN)
{

before 

if(guiCurrentScreen != MAP_SCREEN && guiCurrentScreen != GAME_SCREEN )
{

and add an extra bracket after
guiPreviousOptionScreen = guiCurrentScreen;

RinneT:
A case where the additional check for tmpGuiPreviousOptionScreen is necessary is if you are in the Game Screen (Tactical / Real Time) and then quit via the options menu. In that case tmpGuiPreviousOptionScreen would carry the correct screen while guiCurrentScreen carries an invalid one.
Without the additional check the game is saved in the Map Screen.
Strictly speaking the  checks for MAP_SCREEN are not necessary, as it is set anyways if GAME_SCREEN is not selected. However I would leave it as it is a bit more explanatory of the valid results.

selaux:
Like this? Works for me as well...

RinneT:
Exactly :)

Issue #355, opened 2016-07-21
Write new stuff in Rust


selaux:
I'm interested in Rust and just wanted to ask if we might switch to the rust language for new stuff. Rust is  interoperable with C/C++ and we get some of the improvements of the rust language:
- Memory safety
- An easier syntax
- Easy testing

But there are also drawbacks:
- Unfamiliar to some developers

But imho by using rust we could attract developers that are somehow detered by the current status of the codebase (a big ball of C-like C++). We could create a library (e.g. `libstracciatella`) that is completely written in rust, use it in C++ and link it statically/dynamically.

BTW: This could also be a step to a new license which all the rust code will live under (*GPL)

lynxlynxlynx:
How dependency and setup heavy is that?

selaux:
I think the only dependency in the beginning should be the rust compiler. Maybe I will create a small prototype where I migrate a single function or so, so we can see the impact.

RinneT:
I'm quite interested in the idea. Had similar thoughts regarding Go, though it would be ill-suited - I think - and rust definitely sounds worth to learn.

Just thinking wether it might be a good idea to wait for the SDL2 branch to be merged, as that goes with large changes (I think).

lynxlynxlynx:
We're not talking about a complete rewrite — that'd be mad.

selaux:
It think even more important than SDL is the cmake PR, since we need to see how it integrates with the build system. Perspectively (my most favorite word) we could look at a complete rewrite, but since rust is mostly interoperable with C++ this can be done over months or rather years...

derSteFfi:
Definitly interested here. I'm keen on the prototype...
A rewrite most of the time is a good way to find a lot of old bugs and straighten out a few things...

lynxlynxlynx:
A rewrite most of the time is a good way to introduce a lot of new bugs. :P Usually regressions, as old complexity is confused by overcomplexity.

But in general, sure, the project is in flux. It started as a pure C app, then got partially plusplused, but there's still lots of cleanup to be done.

selaux:
So I played around a little bit with rust and c++ and it's not as simple as I thought. Basically all memory that is allocated in rust needs to be freed by rust again. So returning anything else than simple types (ints, doubles, etc) gets pretty tedious.

[...]

PR #420, opened 2016-10-09
Implement command line parsing in rust

selaux:
So here comes the rust prototype, regarding #355. Notes:
- The library implements a `CommandLineArgs` struct in rust ad exports utility functions to create it and get to its contents.
- Command line args currently need to be used with two dashes, e.g. `--fullscreen`

Discovered Pros:
- Code looks more understandable
- Tests are easy to write
- Dependencies can be loaded very easily
- Can be linked statically

Discovered Cons:
- Only simple structs may be used in rust as well as C++
- Memory management might be tedious sometimes, e.g. passing strings from rust to C++
- Kind of a weird interface

Builds 🙂 : http://builds-ja2.stefanlau.com/pull-requests/420/


RinneT:
Very nice to see this takeoff, very good place to start as well.

My main downside so far is that I have to compile rust myself on my distro and it took literally two hours, far longer than any other package.

Works nicely, though.
What I don't understand is where the actual link to rust is in sgp.cc.
I see there are some calls, however how does the application know in which files to look for them?

lynxlynxlynx:
From what I saw it's done during link time, so it's kind of a library interface.

selaux:
Yes, it simply looks for the according symbols in the built static rust library. It's the `external` definitions.

selaux:
So do you think we should go further into this direction? Then I might make the PR somehow bigger to encompass the `ja2.json` changes...

lynxlynxlynx:
How easy is to set up rust on windows? Personally I don't care much either way, but it'd be a shame if we created a barrier for contribution.

Re --: args could be fixed before parsing, so we maintain backward compatibility. Either with a regex crate or manually.

selaux:
Some more progress here. `ja2.json` parsing is implemented in rust now, including setting command line options with it. The MicroIni dependency has been removed. One drawback is that the rust serialization library does not support comments in json though. But I could also remove the MicroIni dependency this way.

Command line compatibility is fixed, we now use the current master of the `getops` crate which supports this. Stuff to optimize

- Tests for home dir creation and better error handling
- Having a better and safer interface to get rust strings in c++
- Testing `create_engine_options`

@RinneT You could probably also use http://rustup.rs/ to install the current rust version in your home dir.

lynxlynxlynx:
Is it lacking only serialisation or also deserialisation support for comments?

Also you mistyped something about Microini.

selaux:
Both currently. Not sure if they are wanting to implement that because it's not part of the spec. I will inquire about it.

Yes, it was already late in the evening 😉 . But MicroInI has been removed.

selaux:
From my side this is now done. Two differences to before remain:

- CLI switches can now be used with double dashes as well as singe so `--fullscreen` works as well as `-fullscreen`
- No comments in ja2.json allowed

Featureset:
- Command line switches are evaluated in rust
- Finding home dir is done in rust
- ja2.json is evaluted in rust (and created if necessary)
- ja2.json allows for CLI switches to be specified, e.g. `{ ... "res": "1024x768" ... }`

All releases work for me (some fixups were needed for the nightly/debug builds anyways). Only "initial" dependency is rust and cargo that can be installed via [rustup]() on most systems. Rust dependencies will be pulled on demand.

lynxlynxlynx:
For -options, I'd just document only the normal --options instead. It's good that the old ones work, but they're not standard, so I'd discourage them.

Comments we don't use yet, but it's a shame. Can we use placeholder fields for it instead?

The string cleanup makes everything look less scary.

selaux:
I added a comment into the default json. Unfortunately it's not possible to document the `--` options with the argument parser library. We might remove the `-` ones when we include the launcher though, that will fix it.

selaux:
So what about renaming `~/.ja2` to `~/.ja2-stracciatella`? Should we do it? And should we put in some migration logic so the user does not have to do it manually (we could do the same for the ini)?

lynxlynxlynx:
I see no benefit. Also, what was the reason behing removing the shorthand options?

selaux:
That they weren't there before. Imho there is not a lot of benefit in having them...

selaux:
BTW: You see no benefit in changing or in adding migration? Because imho we could also leave the directory as it is and close the  ticket 🙂

lynxlynxlynx:
No benefit in renaming.

PR #421, opened 2016-10-09
Add missing debug string

selaux:
Not sure when this went missing...


lynxlynxlynx:
Looks ok. NUMBER_OF_TOPICS can actually be decreased by one, since the last slot is unused even with this change.

selaux:
Done.

Issue #423, opened 2016-10-14
Selling attachments to Tony - strange prices


JAsmine-ja2:
Selling some stuff to Tony, I noticed a strange thing:
Tony pays different prices for attachments, if they are attached to a gun.
Look at the three pictures.
1) Selling a fully loaded Minimi with attached Talon - 2.946$
2) Selling a fully loaded Minimi with the same Talon extra - 2.914$
3) Selling an unloaded Minimi with the Mag and the Talon extra - 2.914$
![att1](https://cloud.githubusercontent.com/assets/17213795/19393547/66e33fbc-9235-11e6-90f1-b72644055279.jpg)
![att2](https://cloud.githubusercontent.com/assets/17213795/19393546/66e11da4-9235-11e6-9872-28a3762df10c.jpg)
![att3](https://cloud.githubusercontent.com/assets/17213795/19393548/66e9e420-9235-11e6-84ab-44ae8705b776.jpg)

So, the gun with attachment is more expensive than both items seperately. This is true for other attachments as well (can be tested in the savegame).

I was wondering if maybe the game doesn't take the status % for attached items into account when calculating their value? So, attached Talon is always treated as 100% Talon, even though it has less status % ??

Savegame:
[attachment_prices_tony.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/529886/attachment_prices_tony.zip)

lynxlynxlynx:
A different option that comes to mind is that perhaps not everyone can attach a talon and so the price reflects it. Does it require a mechanics check like the merges?

JAsmine-ja2:
No, everyone can attach a Talon. No skill check (afaik).

JAsmine-ja2:
I just tested it with gun and attachment both at 100% status - the same phenomena can be observed.
Gun+attached add-on is more expensive than gun+seperate add-on.

selaux:
You're finding the most edgy cases of all 😉 . It seems like this has to do with Flo doing the actual dealing. Flo (according to code and merc profile) gets discounts for weapons and it seems that the Talon attachment is not recognized as a weapon. Can you try if you have the same issue with a different merc?

JAsmine-ja2:
Further testing shows:
Other mercs than Flo can trade the Talon attached or unattached with no change in price.
This is true in both vanilla (1.12) and Stracciatella.
So, it's indeed Flo, who makes a difference here.

[...]

PR #425, opened 2016-10-16
Fix issues with weapons dealers

### LINKED ISSUE #424: Yellow or green star for attached Talon
This is a really minor thing (nitpicking) I noticed in connection to issue #423 

Usually, a weapon with an attachment gets a green star in inventory.
A weapon with attached Talon however gets a yellow star in inventory.

When selling a weapon with attached Talon, the yellow star disappears and is replaced by a green star as can be seen in this screenshot:
yellow-green-star

As I said, it's a very minor thing. Maybe, it can also be fixed with minor effort?

selaux:
Fixes #424 #423.

- Flo now also gets discounts for `IC_WEAPON` class items (according to her merc description)
- Extracted attachment star coloring into own function and check for `IC_LAUNCHER` class instead of talon launcher
- Use extracted function in dealer inventory

### DIFF
--- a/Build/Tactical/ShopKeeper_Interface.cc
+++ b/Build/Tactical/ShopKeeper_Interface.cc
@@ -2160,7 +2160,8 @@ static UINT32 DisplayInvSlot(UINT8 const slot_num, UINT16 const item_idx, UINT16
 
 	if (ItemHasAttachments(item_o))
 	{ // Display the '*' in the bottom right corner of the square
-		DrawTextToScreen(L"*", x + SKI_ATTACHMENT_SYMBOL_X_OFFSET, y + SKI_ATTACHMENT_SYMBOL_Y_OFFSET, 0, TINYFONT1, FONT_GREEN, FONT_MCOLOR_BLACK, LEFT_JUSTIFIED);
+		UINT8 attachmentHintColor = GetAttachmentHintColor(&item_o);
+		DrawTextToScreen(L"*", x + SKI_ATTACHMENT_SYMBOL_X_OFFSET, y + SKI_ATTACHMENT_SYMBOL_Y_OFFSET, 0, TINYFONT1, attachmentHintColor, FONT_MCOLOR_BLACK, LEFT_JUSTIFIED);
 	}
 
 	{ // Display 'JAMMED' or 'REPAIRED', if appropriate
--- a/Build/Tactical/ShopKeeper_Interface.cc
+++ b/Build/Tactical/ShopKeeper_Interface.cc
@@ -2578,7 +2578,7 @@ static UINT32 CalcShopKeeperItemPrice(BOOLEAN fDealerSelling, BOOLEAN fUnitPrice
 		// if it's a GUN or AMMO (but not Launchers, and all attachments and payload is included)
 		switch ( GCM->getItem(usItemID)->getItemClass() )
 		{
-			case IC_GUN:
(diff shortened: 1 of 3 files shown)


JAsmine-ja2:
Nice fixes!
Imo, other item classes should be added to Flo's discount.

In vanilla JA2, Tony doesn't accept grenades!
However, I usually mod him to accept grenades as well, since it makes sense. He sells 40mm grenades - so why shouldn't he also trade with handgrenades?
Hard to say, if it's an oversight by the devs or a 'real' decision.
But adding grenades (and other classes) to Flo's discount has (almost) no effect on a game as long, as the dealer isn't modded.
So, I'm pro adding more classes.

Build/Tactical/ShopKeeper_Interface.cc
```diff
@@ -2578,8 +2578,7 @@ static UINT32 CalcShopKeeperItemPrice(BOOLEAN fDealerSelling, BOOLEAN fUnitPrice
 		// if it's a GUN or AMMO (but not Launchers, and all attachments and payload is included)
 		switch ( GCM->getItem(usItemID)->getItemClass() )
 		{
-			case IC_GUN:
-			case IC_AMMO:
+			case IC_WEAPON:
```
lynxlynxlynx: Is the removal of IC_AMMO on purpose? IC_WEAPON doesn't include it. Considering what it includes, the change seems excessive:
`define IC_WEAPON         ( IC_GUN | IC_BLADE | IC_THROWING_KNIFE | IC_LAUNCHER | IC_TENTACLES )`
So I'd prefer to just add IC_LAUNCHER to the cases.

chmallon:
> * Flo now also gets discounts for IC_WEAPON class items (according to her merc description)
This change is wrong.
No item has class IC_WEAPON.
IC_WEAPON is defined as IC_GUN | IC_BLADE | IC_THROWING_KNIFE | IC_LAUNCHER | IC_TENTACLES.
Items only have a single bit set and the test is done with 'case IC_WEAPON:'
So now Flo only gets a discount for ammunition ('case IC_AMMO:') but for no weapon anymore.

> * Extracted attachment star coloring into own function and check for IC_LAUNCHER class instead of talon launcher
> * Use extracted function in dealer inventory
This change misses the case in RenderItemPickupMenu().

chmallon:
> * Flo now also gets discounts for IC_WEAPON class items (according to her merc description)

This change is wrong.
No item hass class IC_WEAPON.
The test is 'case IC_WEAPON:', IC_WEAPON has multiple bits set and items only have a single bit set.
So now Flo gets no discount for weapons anymore, only for ammunition ('case IC_AMMO:').

> * Extracted attachment star coloring into own function and check for IC_LAUNCHER class instead of talon launcher
> * Use extracted function in dealer inventory

This change misses the case in RenderItemPickupMenu().

Manuel-K:
@chmallon You're right. Switch/Case do not work like that.

chmallon:
I think the changes regarding Flo are against the gist of her biography.
It clearly states firearms and also the comments in the code say guns and explicitly not launchers.
Now for example it also includes knives.
Also I think this price discrepancy also appears with other gun attachments like the sniper scope (an IC_MISC).

I should be reverted back to the original state.
It seems this change also was never tested, because it did not work, i.e. Flo got no discount anymore (except for ammunition).

PR #429, opened 2016-10-18
Fix issues with text inputs

### LINKED ISSUE #419: Can't type IMP code after saving via Save screen
If you save the game via the save screen and hence select a text input field, you can't type the IMP access code on the IMP homepage.
This affects new Dead is Dead games as you are forced to select a savegame slot, but also works on normal games if you save before entering the code.
Other fields seem to be unaffected, like the florists message cards or shortcuts.

Only workaround is to completely close the game, restart and load the save.

[Edit]: Additionally if you put a text in the flower cards the IMP textbox doesn't work either.
You can select it and the blinking '|' will appear, but you can't type. So it seems to get the focus but somehow doesn't manage to bind the keyboard.

selaux:
Closes #419 

This ended up in an almost complete rewrite of the `IMP_HomePage` and `IMP_Begin_Screen` Screens. Basically they had own custom code to handle input fields and keyboard input. I replaced it with inputs from `Text_Input`. Additionally I had to fix that keydown events were not marked as handled anymore, so e.g. pressing one inside an input on the save screen afterwards selected the first save slot. Implications from all of this:
- If a text input is focused, almost all key events are ignored by the respective screens (except for some "chosen" ones)
- The cursors now blink differently on the two mentioned screens than before
- Checkboxes on `Imp_Begin_Screen` now dont get focus anymore when they are hovered over
- Removed ifs that mentioned visiting the `IMP_Begin_Screen` after character creation was finished because I couldn't get to it after doing the questionnaire anymore

I also had a fun idea: Should we add contributors as easter eggs using their github names? We could even create portraits n stuff 😄

### DIFF
--- a/Build/Laptop/IMP_Begin_Screen.cc
+++ b/Build/Laptop/IMP_Begin_Screen.cc
@@ -1,3 +1,4 @@
+#include <sgp/VObject.h>
 #include "CharProfile.h"
 #include "Directories.h"
 #include "Font.h"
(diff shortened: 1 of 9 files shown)


JAsmine-ja2:
[ja2aepatch.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/538270/ja2aepatch.zip)
Don't know either, if it's relevant. But here it is, anyway ;)
AFAIR, there were severeal patches/tools to allow for more than one IMP. This is one of them.

Concerning adding contributors as easter eggs:
I'm 100% pro this.
On the one hand, changing the content of the game is generally not a good idea, because it doesn't go along with the idea behind Stracciatella.
On the other hand, a harmless change like this (that will remain unnoticed by most players anyway) adds to the whole "Stracciatella experience".
Another thing that I once brought up to gennady was, to include new information in the in-game help, to make it more useful. Maybe as a mod?
Another idea that is similar is to add short helpful texts to the loading screens with Stracciatella-specific information (e.g. "use ALT+RETURN to change between full screen and windowed mode" - this questions was brought up at BP's forum recently).

selaux:
Can anybody confirm that the text input works as expected with this?

selaux:
Builds 🙂 : http://builds-ja2.stefanlau.com/pull-requests/429/

JAsmine-ja2:
Downloaded and tested win-64 build. Textinput works great.

selaux:
Thank you very much 👍. Merging.

Issue #307, opened 2016-05-12
integrate vxx's work


lynxlynxlynx:
In the old bb repo, wvx was working on porting minor improvements from 1.13:
https://github.com/wvx/ja2-stracciatella/commits/20160205

I remember there were some problems with the warning work, but the featurettes mostly look fine. We need to decide what to accept at some point, hence this reminder.

MOVED:
- mod: Import your old IMP merc into a new game; related/tracked in #32 
- recheck and extract extender fix #624

IGNORED:
WIP: extend pathing (needs more work)
WIP: skip sleep explanation (needs more work)
WIP: vehicle inventory (probably needs more work)
externalise: perpetual crepitus queen corpse (cheating)
mod: recruit militia to squad (scope)
Inventory Management Extras (several things at once, some wip, some odd; too much work)
mod: soldiers can have all skill traits (get *all* traits if you meet stat criteria)
mod: accurate aim circle (unreliable also from my experience)
allow realtime tactical saving in ironman (it was optional, but contradictory to the difficulty setting's purpose)
Wdelete-non-virtual-dtor (unresolved leaks)
Warray (looks pointless)

derSteFfi:
most of it looks good at first glance. I'd definitly cherry pick the warning fixes and stuff like auto reload on alt+r (although I don't see the necessity to hide minimal, optional stuff like this behind a json switch).
What I also like is the enhancement to strategic view ('gui_extras') even though this changes the look obviously. vehicle inventory, modded ai and multiple interrupts are things i don't care for too much but I'm also not opposed to integrating them.

JAsmine-ja2:
My humble opinion on this is:
As long as new features remain optional, everything is cool.
One additional aspect is that all the new stuff desperately needs to be properly documented.
Again, I draw a line to v1.13: TONS of new features, but the documentation is rather poor. Most documentation can even be found exclusively in the depths of the Bear's Pit. I'd go as far and say that there's probably only about two handful of people out there that know all the features of v1.13 and how to properly use them.
Of course, we are FAR from that situation and will propably never even get close.
But the core of the problem remains the same. If the game.json (or ini or whatever) has more than a few options, we need some kind of documentation. When we get to that point, I'll gladly help, of course.

lynxlynxlynx:
my favourite is the multiple interrupt thing and not because I wrote it. It just didn't make any sense before.

Do you think just having lengthy-enough comments in the game.json is not enough for documentation?

[...]

PR #436, opened 2016-10-26
Fix 307 p3

lynxlynxlynx:
Final part of #307, before deciding on the stragglers (please take a look at the lists).


externalized/game.json
```diff
@@ -55,6 +55,12 @@
     //  vanilla value: 0
     "enemy_weapon_minimal_status": 0,
 
+    //  Make the AI a bit smarter and let it target heads and legs sometimes
+    //  vanilla values: false, 0, 0
+    "mod_ai_a": false,
```
selaux: Can we call this differently? Maybe `ai_targets_body_parts`? Or more generic (so we can add stuff later) `mod_ai_aiming`?

Build/Tactical/Soldier_Create.cc
```diff
@@ -1254,6 +1255,8 @@ void CreateDetailedPlacementGivenBasicPlacementInfo( SOLDIERCREATE_STRUCT *pp, B
 			break;
 		case SOLDIER_CLASS_ELITE:
 			pp->bExpLevel = (INT8) 6 + bExpLevelModifier;
+			if(pp->bExpLevel<GCM->getGamePolicy()->enemy_elite_minimum_level) pp->bExpLevel=GCM->getGamePolicy()->enemy_elite_minimum_level;
```
selaux: Isn't there a a comparison missing here?

Build/Tactical/Soldier_Create.cc
```diff
@@ -1336,6 +1339,7 @@ void CreateDetailedPlacementGivenBasicPlacementInfo( SOLDIERCREATE_STRUCT *pp, B
 	pp->bWisdom				= (INT8)(bBaseAttribute + Random( 9 ) + Random( 8 ));
 	pp->bMorale				= (INT8)(bBaseAttribute + Random( 9 ) + Random( 8 ));
 
+	if(ubSoldierClass==SOLDIER_CLASS_ELITE && pp->bExpLevel>GCM->getGamePolicy()->enemy_elite_maximum_level)
```
selaux: Body of the if statement is missing?

Build/Tactical/Soldier_Create.cc
lynxlynxlynx: No, it's just bad style. In one of the following commits I cleaned it up (the abstraction one).

Build/Tactical/Soldier_Create.cc
lynxlynxlynx: Looking at it, it's more correct to only run the next line if these two conditions are true.

lynxlynxlynx:
Cleaned it up further and reverted a bit.

PR #442, opened 2016-11-01
Automated nightly builds

selaux:
This enables us to build nightlies using travis and appveyor, this way engaged people can test our changes earlier. The nighlies are currently uploaded to a personal server of mine available here:
http://builds-ja2.stefanlau.com/nightlies/

It builds most of our available target configurations: Windows 32/64 bit MSVC builds, OS X build and the 64bit debian package.

Currently the builds need to be triggered externally, so I setup a Jenkins task on a personal Jenkins instance that just does that (making a request to the travis and appveyor APIs). I can also share the Jenkinsfile if requested, so this setup can be reproduced without hassle on any Jenkins instance.

@JAsmine-ja2 Can you maybe test both Windows builds? They both work for me, but I have a "development-tools-infested" machine...


lynxlynxlynx:
I think we should keep them in debug mode. They're for avid testers and we'd want the capable to be able to provide us backtraces, core dumps.

selaux:
You're probably right, I put them into Release mode because of the frequent crashes due to asserts. But we probably want to know about them as well...

lynxlynxlynx:
Some of those bugs are still open iirc - I'll go add the one I remember to the milestone, just so we don't forget.

lynxlynxlynx:
If you're done, go ahead and merge it. LGTM

selaux:
@JAsmine-ja2 Can you test if the latest nightly build works on your machine? Then I can merge...

JAsmine-ja2:
Sorry, I must've overlooked this.
Gave both windows builds a quick test, and they work.

However:
I hear a BIG difference in the audio quality of the main menu music. To me (as a musician) it is very obvious, that the quality of the 0.15.x builds is inferior to 0.15.1 (true for both 32 and 64 bit).
I could do a recording for comparison, but when you listen to the beginning of the main theme and listen to the sound of the cymbals, it should be obvious to everyone, that the 0.15.x-cymbal sounds very "noisy, trashy, dirty" compared to the clean sound of 0.15.1 (and vanilla).

Could this be because of some drastic changes like sdl 2?

I'm on Windows 8.1 64 bit.

PR #445, opened 2016-11-10
Build pull requests

selaux:
This builds packges for pull requests so changes can be tested more easily also by non-developers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,22 @@ dist: trusty
 sudo: required
 
 before_install:
+  - export PUBLISH_BINARY="false"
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
temporary problem? 

LGTM

janisozaur:
FTP 500 looks to be:

> Permanent Negative Completion reply

> The command was not accepted and the requested action did not take place. The User-process is discouraged from repeating the exact request (in the same sequence).

Most likely failed auth. Do note Travis will not deploy your `secure` token on PRs, which may hold your `$SFTP_USER` and `$SFTP_PASSWORD` variables (I do not see them defined elsewhere in the job).

If you just want a quick and easy solution for temporarily storing build artifacts, I recommend https://transfer.sh

selaux:
Yes, it's missing the credentials (they were set by Jenkins before) also appveyor does not build PR packages yet as it seems. Still some work to do 🙂

Issue #106, opened 2013-11-11
Graphical glitch in main menu after quitting a game


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

There is a graphical glitch on the main menu screen:

![http://img.xrmb2.net/images/142535.jpeg](http://img.xrmb2.net/images/142535.jpeg)

This is how to reproduce:
1. Start a new game or load a game
2. End the current game and go back to main menu
3. The glitch will be there

More info:
- This does not work with standard resolution (640x480)
- I can reproduce this in both 800x600 and 1024x768
- This issue is not present in vanilla JA2 (because you can't change res there)
- The screen looks fine when starting JA2. It gets messed up only after exiting a game
- This issueis present in v0.12.258-win, but was already there before

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/106

derSteFfi:
still present. 640x480 works fine, i tested 800x600 and 1920x980 and the glitch is present.

JAsmine-ja2:
Still present in v0.15 as described in initial post.

Issue #87, opened 2013-09-01
Main menu music doesn't restart


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

When I start the game, the main menu music plays fine.
But when I load or start a game and then end the game and return to the main menu, it remains silent. No music starts.
I think the main menu music should always play when inside the main menu.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/87

derSteFfi:
probably related to #106 or at least can be solved at the same spot

JAsmine-ja2:
How to easily reproduce this:
1) Start a new game.
2) End the current game.
-> Game takes you back to the main menu. With no music playing.

PR #446, opened 2016-11-14
Blit main menu backgrounds to frame buffer

### LINKED ISSUE #72: menu bug when ending a game
Originally reported by: **Anonymous**

---

When you end a current game and go back to the main menu, this is what it looks like:
ending a game, going back to main menu

the "old game" is still visible around the main menu. I guess, that most graphics (especially backgrounds) have to be provided to the game in stretched resolutions, as the game won't stretch the graphics itself???

I'm still playing v0.12.189 on english ja2 gold with ja2-ENGLISH-1024x768.bat

btw, sorry for spamming al these issues as a random anonymous nobody. Maybe I should register before posting more issues?!

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/72

selaux:
- Save buffer is restored by the button system, so first we need to blit to frame_buffer (because it is cleared) and then to the save buffer (to be able to restore). Otherwise we would restore from a non-cleared save buffer
- Fixes #72, #106 

Builds 🙂 : http://builds-ja2.stefanlau.com/pull-requests/446/

### DIFF
--- a/Build/MainMenuScreen.cc
+++ b/Build/MainMenuScreen.cc
@@ -359,10 +359,9 @@ static void CreateDestroyMainMenuButtons(BOOLEAN fCreate)
 
 static void RenderMainMenu(void)
 {
-	BltVideoObject(guiSAVEBUFFER, guiMainMenuBackGroundImage, 0, STD_SCREEN_X,       STD_SCREEN_Y     );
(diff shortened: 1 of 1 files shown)


JAsmine-ja2:
Tested win-64 build: Works! Nice job! 👍 

Another issue in this regard is #87 .
Ofc, #87 is audio and not graphics, but it is triggered in the same way as #72 and #106 used to be triggered.

selaux:
Thanks. I had a look into #87 as well, but no idea what is wrong there currently...

PR #448, opened 2016-11-27
fixed regression from 989f2d9, fixes #447

### LINKED ISSUE #447: No parallel movement of several mercs in real time
With the current build I get the problem that I can't move several mercs at once in real time.
E.g. I have several mercs selected and give an order to move or change position.
The game then unselects all mercs other than the "primary" selected and only he/she moves.

In addition to that I get a stopwatch until the movement is done. Essentially the game behaves as if it were in round based mode.

lynxlynxlynx:
== has precedence over &
and constants need to be resolved first


lynxlynxlynx:
Manually.

selaux:
Aren't there two lines missing in `Turn_Based_Input.cc`? If not 👍 and feel free to merge.

lynxlynxlynx:
Nope, one is just a setter, the other was a different composite check before.

chmallon:
I see that you combined the flags `INCOMBAT` and `TURNBASED` in the macro constant `IN_TB_COMBAT`.
You can drastically simplify this:
The flag `TURNBASED` is useless, because it is *always* set in `gTacticalStatus.uiFlags`, see `InitOverhead()`.
I.e. remove `TURNBASED` and the new `IN_TB_COMBAT` and just check `INCOMBAT`.

Similarily `REALTIME` is *never* set.

lynxlynxlynx:
Yeah, it was left in on purpose, as we haven't gone through all community contributions by then and it wasn't clear if anyone bolted anything on top. The situation is much clearer now, so it can go.

Issue #360, opened 2016-08-05
Folder renames


derSteFfi:
So, as discussed in the cmake PR, we all would like to change a few things regarding the folder structure. I copied Selaux' last list here as a starting point. I'm highly unsure about how git works with renamed files and rebasing changes onto that, so if we are unsure, we should do this as soon as nearly no PR is open....
- [ ]     _build/lib-\* -> dependencies
- [ ]     _build/buildboxes -> release
- [ ]     the rest of _build -> assets
- [ ]     externalized -> assets/externalized
- [ ]     Build -> src/game
- [ ]     src -> src/externalized
- [ ]     sgp -> src/sgp

lynxlynxlynx:
I don't think it has much problems, though digging through logs becomes more tedius (--follow and -M). But yes, something for after the big PRs.

selaux:
What do you think about removing the buildboxes after building releases through travis?

Advantages:
- Less code clutter
- Less maintenance effort

Disadvantages:
- No i386 deb can be built anymore
- We cannot ensure builds on *BSD anymore

Manuel-K:
Another question: Is the examples subfolder in _build/lib-slog necessary? I does not provide more information than the example in the readme.

selaux:
Agreed. Also slog is less of a dependency than tightly coupled with the project itself through tags...

Manuel-K:
How about _build/lib-slog -> src/slog?

Manuel-K:
I've tested the proposed changes in #455. It does look nicer, but I'm not completely happy. I've kept lib-slog in dependencies for now. 

I'd propose to move rust->src/rust and move the contents of its src subfolder one level up (i.e. to src/rust as well). Also _unittests->unittests.

selaux:
I would keep the rust folder separate since it's an own project with its own build system. The rest looks fine. Only for `_unittests` I would rather use `assets/unittests`.

derSteFfi:
agreed. for the buildboxes I'd keep those we can't replace with a travis build for now. i386 sounds like something we could set up in travis and for *BSD we could check whether anyone uses it. We currently do not offer a *BSD tarball anyway but we could and see whether anyone downloads it.

Manuel-K:
`assets/unittests` sounds very good and seems to be the obvious choice. After you wrote it I'm not sure how I could have missed that.

[...]

PR #455, opened 2016-12-07
[RFC] Implementation of #360 (Folder renames)

Manuel-K:
This shows the source tree structure as suggested by @derSteFfi and without the samples subfolder of lib-slog.

I've probably missed something in CMakeLists.txt – I only tested building/linking/running on Linux.


Manuel-K:
Hm. AppVeyor complains about SDL2…

selaux:
Yes you have to adapte the toolchain files inside the cmake directory.

selaux:
Imho we could merge, you just have to fix the conflicts...

Manuel-K:
Self-created merge conflicts. That was very smart. ;-)

Rebased on master.

PR #463, opened 2016-12-15
[RFC] Cmake cleanup

Manuel-K:
I wanted to add the options to use system installed versions of GTest and RapidJSON and do some minor cleanups of `CMakeLists.txt`. This therefore includes @AMDmi3's #391 in an optional form.

While cleaning the includes array I've stumbled upon the line that includes every directory of the `src` folder. I went on to investigate what changes are necessary to remove it and the **horror** began. ;-)

It turns out that the includes are even messier and over the place than I feared. I wrote a python script to change the include directive to a full path for all header files that aren't in the same directory. I think this is only a *slight improvement*, but it does highlight that some real refactoring is necessary.

Note: One file got completely rewritten by my script. I assume that it used to have Windows line endings. If the changes are too much, I'm happy to restore that or even trash the path changes and only keep the other `CMakeLists.txt` improvements.


Manuel-K:
Okay, I've missed some paths inside `#ifdef _DEBUG` blocks that uses other formatting than the rest of the code. My bad for only testing with `-DCMAKE_BUILD_TYPE=release`…


```
#ifdef X
#    include "Y.h"
#endif
```
vs.
```
#ifdef X
  #include "Y.h"
#endif
```

While I rectify the issue, should I equalize the formatting? If yes, which style should I use?

selaux:
Will close as the cmake cleanup #482 was already merged.

Issue #10, opened 2013-03-07
Think about removing JA2BETAVERSION code


lynxlynxlynx:
Originally reported by: **Gennady Trafimenkov (Bitbucket: [gennady](http://bitbucket.org/gennady), GitHub: [gennady](http://github.com/gennady))**

---

Think about removing JA2BETAVERSION code.
This code probably may not work anymore, since nobody tests it or even build.
I can't see value of keeping this code.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/10

derSteFfi:
```
~/ja2-strat (git)-[master] % grep -r JA2BETA * | wc -l
319
```
For non-compiling code with 319 occurences and for the sake of maintainability i would vote for removal, simply because no one needs it anymore. The only Occurences i came across didn't have anything to do with current issues i am aware of.

lynxlynxlynx:
Originally, there was another JA2DEMO define, but all that code got cut out.

Manuel-K:
@lynxlynxlynx Wasn't this solved by #470?

lynxlynxlynx:
Indeed, closing.

PR #470, opened 2016-12-30
JA 10 - Think about removing JA2BETAVERSION code

edwbuck:
This series of commits demonstrates what the ja2-stracciatella code would look like with JA 10 implemented.  

After removing the JA2BETAVERSION pre-processor directive, it was obvious that some of the code
was misleading, in the sense that empty function calls existed within the code base.  For clarity, each
of the empty functions were removed, with a separate commit (for ease of review).

Also, after removing the JA2BETAVERSION pre-processor directive, it was obvious that one function 
returned a constant BOOLEAN value of false.  Refactoring was done to remove unreachable code 
blocks and to simplify the program logic in light of the function's inability to return True.  For clarity, this
was done in a separate commit for ease of review.

Finally, a number pre-processor directives lost the ability to be set dynamically.  Previously, these mapped function names to implementations that only existed when JA2BETAVERSION was enabled.  When not enabled, they mapped to ((void)0) which the compiler would optimize out as an expression 
statement with  no side-effect.  For clarity, these pre-processor directives were removed, as was the
text that expanded to the NOOP expression.  Again, a separate commit was done for ease of review.

Please consider this code submission for the closing of #10 .  I believe the code will be easier to read and maintain without the misleading text. 

Thank you.
Edwin Buck


lynxlynxlynx:
LGTM. Maybe the cheats could be salvaged — some sound potentially useful for development purposes.

& wow, welcome back Edwin! One of the first porters to gcc and make.

src/game/SaveLoadScreen.cc
```diff
@@ -48,10 +48,6 @@
 #include "Meanwhile.h"
 #include "PreBattle_Interface.h"
 
-#if defined JA2BETAVERSION
```
derSteFfi: this leaves a newline in line 50 that's unneccessary

src/game/Strategic/Auto_Resolve.cc
```diff
@@ -71,10 +71,6 @@
 #include "slog/slog.h"
 #include "Easings.h"
 
-#ifdef JA2BETAVERSION
```
derSteFfi: again, empty newline here.

src/game/Strategic/MapScreen.cc
```diff
@@ -273,9 +273,6 @@ enum MapEvent
 	MAP_EVENT_PLOT_PATH,
 	MAP_EVENT_CANCEL_PATH,
```
derSteFfi: again, empty newline

src/game/Tactical/Soldier_Init_List.cc
```diff
@@ -33,12 +33,6 @@
 #include "MemMan.h"
 #include "FileMan.h"
```
derSteFfi: here again :)

src/game/Tactical/Soldier_Init_List.h
```diff
@@ -70,10 +70,6 @@ void AddProfilesNotUsingProfileInsertionData(void);
 
 void StripEnemyDetailedPlacementsIfSectorWasPlayerLiberated();
```
derSteFfi: and here

src/game/Tactical/Turn_Based_Input.cc
```diff
@@ -2413,12 +2395,6 @@ void GetKeyboardInput(UIEventKind* const puiNewEvent)
 			}
 			else if ( InputEvent.usKeyState & CTRL_DOWN )
       {
```
derSteFfi: this leaves the else if empty and ready for removal

src/game/Tactical/World_Items.cc
```diff
@@ -23,10 +23,6 @@
 #include "MemMan.h"
 #include "FileMan.h"
```
derSteFfi: here is a newline again

src/game/TacticalAI/NPC.h
```diff
@@ -96,10 +96,6 @@ BOOLEAN NPCWillingToAcceptItem( UINT8 ubNPC, UINT8 ubMerc, OBJECTTYPE * pObj );
 void SaveBackupNPCInfoToSaveGameFile(HWFILE);
 void LoadBackupNPCInfoFromSavedGameFile(HWFILE);
```
derSteFfi: newline :)

src/game/SaveLoadScreen.cc
edwbuck: Don't mind fixing white space issues.  Look for another commit removing them.

src/game/Strategic/Auto_Resolve.cc
edwbuck: Will be fixed in next commit on JA-10 branch

src/game/Strategic/MapScreen.cc
edwbuck: Will be fixed in next commit on JA-10 branch.

src/game/Tactical/Soldier_Init_List.cc
edwbuck: Will be fixd in next commit on JA-10 branch.

src/game/Tactical/Soldier_Init_List.h
edwbuck: Fixed in next commit on JA2 branch.

src/game/Tactical/Turn_Based_Input.cc
edwbuck: Removal won't be trivial to review, or implement.  Cutting the block puts more keystrokes in the else block which appears three lines below (due to poor formatting and poor newline / brace style choices, it appears to be an else block for a different else statement).

I don't mind cleaning up the flock-of-seagulls nesting smell, but I haven't determined if I'll be doing that in this branch.  Currently I'm leaning against it, but if it's the only item left, please consider merging without repair in this case.

The empty block is the safest, even if it is less clear.

src/game/Tactical/World_Items.cc
edwbuck: Fixed in next commit on JA2 branch.

src/game/TacticalAI/NPC.h
edwbuck: Fixed in next commit on JA2 branch.

edwbuck:
Lynxlynxlynx,

Thank you for the kind words.  It is heartwarming to know that a few people remember me, even from programming work done a decade or so ago.

In a few short words, I cannot summarize all that's happened in the last decade.  I hope that time has treated all my JA2 programming friends well.  Life is very different for me now, and in general I'm enjoying it.

Cheers,
Edwin

edwbuck:
I wouldn't worry about salvaging the cheats until we can implement them with polymorphisim.  Currently we lack the policy object to be replaced with the cheating policy.  The structure of cheating by "if statement" makes the code harder to read.

src/game/Tactical/Turn_Based_Input.cc
edwbuck: I'm open to putting this as a bug on the backlog, and merging what's done into the mainline.  That way we don't lose the task, and we can shut the branch down quickly.

derSteFfi: ah, ok. I didn't notice the tabbing here. In this case I agree, let's solve that somewhere else.

lynxlynxlynx:
It's easy to remember, since I think I was the only one helping you back then, with most players being on windows. Looking at the old wiki, I see I wrote a bunch of regexy fixers for includes and leading whitespace (again a problem). Old days of CVS. :)

Thanks for you work!

PR #474, opened 2016-12-31
remove JA2 preprocessor define

### LINKED ISSUE #471: Remove #ifdef JA2
We are not on-the-hook to maintain the sgp code that doesn't apply to Jagged Alliance 2.

We should go through the code removing the JA2 pre-processor directives as if they were unconditionally set.

This would decrease the amount of conditional source code required to be processed by future maintainers, leading to a lighter mental carrying capacity.  Currently there are 12 instances of the JA2 pre-processor define.

derSteFfi:
fixes #471


lynxlynxlynx:
yep, looks good. Just wait for the bots to do their thing.

PR #482, opened 2017-01-21
CMake cleanup (minimal version)

Manuel-K:
This is a minimal version of #463 without the nightmarish `#include` path commit.

It allows to use system wide installations of RapidJSON and GTest and cleans up some minor things.


lynxlynxlynx:
Thanks, LGTM!

Manuel-K:
This should also solve #491.

Issue #483, opened 2017-01-21
Determine minimal rust version


Manuel-K:
We need to figure out the minimal rust version if we want to add a check to https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/CMakeLists.txt#L3. #481 shows that this would be a good idea.

http://doc.crates.io/specifying-dependencies.html mentions cargo version 0.9.0 for the cfg, but it is not clear that this version can build the library.

I have rust 1.14.0 (with cargo 0.15.0) installed which is (hopefully) complete overkill.

Manuel-K:
@johndoe31415's latest problem comes from serde_json. I cannot find a minimal required rust version in its documentation.

lynxlynxlynx:
Why another report, now we have two about the same thing. :confused: 

0.9 was released just over 3 years ago. But yes, it would need testing if the parser upgrade is the only needed change.

Manuel-K:
Not rust 0.9. We need at least cargo 0.9 which corresponds with rust 0.18. That was released in April.

Manuel-K:
That raises another problem. It is possible to install a newer version of cargo without installing the corresponding version of rust. So we should check both versions.

Manuel-K:
Obviously it should be rust 1.8 instead of 0.18.

johndoe31415:
Just for reference, the default Ubuntu versions of rust/cargo are 1.7.0 and 0.8.0, respectively, at the moment (xenial).

Manuel-K:
Okay. I was just compiling rust 1.8.0 on my Arch Linux system and after a long time the build aborted because of the stupid `-Werror=misleading-indentation` flag. Because I was using my default setup, I don't have the compiled object files anymore. 

Maybe someone else could try to find a binary package for another distribution...

Manuel-K:
Okay. I keep forgetting about ALA. 

It works on Arch Linux with rust downgraded to 1.8.0 and cargo downgraded to 0.9.0.

PR #484, opened 2017-01-21
Added checks for rustc and cargo versions

Manuel-K:
Both minimal versions might have to be increased (see #483).


lynxlynxlynx:
Did you try building with these versions?

Manuel-K:
No, that's why I created the ticket. I just wanted to add the cmake code to check the rustc version.

lynxlynxlynx:
It's better than what we have now anyway. RUSTC_EXECUTABLE could be used instead of rustc now, but then appveyor would need a lot more work for negligable gain.

PR #488, opened 2017-02-02
Fix AI hang during turn based combat

### LINKED ISSUE #466: AI deadlocks
Version 409ae2a2. Throughout my game, every battle has been slowed down by multiple AI deadlocks per turn. Example attached; end the turn, and you should get several AI deadlocks:

[SaveGame07.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/662343/SaveGame07.sav.zip)

2016/11/19 19:55:46 ERROR   [AI              ] Aborting AI deadlock for 26. Please sent LOG file and SAVE.
2016/11/19 19:56:10 ERROR   [AI              ] Aborting AI deadlock for 34. Please sent LOG file and SAVE.
2016/11/19 19:56:28 ERROR   [AI              ] Aborting AI deadlock for 36. Please sent LOG file and SAVE.

momoko-h:
The AI actions CHANGE_FACING and CHANGE_STANCE must be included in the
movement actions otherwise the soldiers can never reach their target. Closes #466


lynxlynxlynx:
Thanks, sounds sensible. How much testing/theorizing did you do?

momoko-h:
I started with selaux' observation that it only happened when pSoldier->bAction is  AI_ACTION_CHANGE_STANCE, so I watched for that and then quickly noticed that the code at AIMain.cc line 418 where the comment says "an old action was in progress; continue it" was reached again and again without actually doing anything because it immediately failed the if check in line 419.

Testing: I observed the hang happening with Reaper against a single soldier in Omerta and verified that the patch fixes the hang. Then I played a couple of minutes with a 5 mercs against 6 soldiers situation and didn't notice any problems.

lynxlynxlynx:
Perfect, thanks again!

RinneT:
Really awesome you could fix it. Thanks so much!

andreas12345:
Awesome, thanks a lot!

chmallon:
Did you consider the implications of this change on MovementMode[][]?
(which already misses one initializer for vanilla)

Manuel-K:
Interesting. `AI_ACTION_MOVE_TO_CLIMB`, `AI_ACTION_CHANGE_FACING` and `AI_ACTION_CHANGE_STANCE` are uninitialized. `{WALKING,  WALKING,  WALKING}` or `{WALKING,  RUNNING,  RUNNING}` seem to be the most obvious choices.

Manuel-K:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/game/TacticalAI/AIUtils.cc#L42

chmallon:
It's a global array, so it is zero-intialized, i.e. in this case WALKING.

Manuel-K:
That is not an answer to the question what the values should be, but you obviously know that.

lynxlynxlynx:
While it doesn't seem to matter much (most code treats both walking and running equally), I agree that {WALKING, RUNNING, RUNNING } looks like a good choice for the missing two.

PR #490, opened 2017-02-07
Build fails if WITH_UNITTESTS is off

momoko-h:
main() contains two calls to functions that are not present when you try to build without unit tests. Additionally it would be possible to completely remove the -unittests command line option in such builds but IMO options that are sometimes present, sometimes not kind of suck.


lynxlynxlynx:
It's unfortunate that it can't be done in should_run_unittests.

Issue #68, opened 2013-08-31
Crash in laptop AND huge memory footprint


lynxlynxlynx:
Originally reported by: **Anonymous**

---

There is a crash in the laptop which I can reproduce all the time:
- start a new game
- close the help screen, close the message from your mailbox
- click on the buttons to the left (E-mail, Web, Files, History, Personnel, Financial) **quickly one after another**.
## \- the game crashes.
- works also in running campaigns
- this does not work (= not crash) when you do it slowly
- this does not work (= not crash) when you've already opened the websites once before in a running laptop-session

i'm running v0.12.189 w/english ja2 gold, ja2-ENGLISH-1024x768.bat

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/68

derSteFfi:
still present on master:
```
Program received signal SIGSEGV, Segmentation fault.
0x00000000006bbef2 in SGPVObject::~SGPVObject (this=0x2209fb0, __in_chrg=<optimized out>) at sgp/VObject.cc:80
80			if (*anchor != this) continue;
(gdb) backtrace
#0  0x00000000006bbef2 in SGPVObject::~SGPVObject (this=0x2209fb0, __in_chrg=<optimized out>) at sgp/VObject.cc:80
#1  0x000000000049014e in DeleteVideoObject (vo=0x2209fb0) at sgp/VObject.h:121
#2  0x000000000049033a in RemovePersonnelGraphics () at Build/Laptop/Personnel.cc:402
#3  0x000000000049029b in ExitPersonnel () at Build/Laptop/Personnel.cc:350
#4  0x000000000048675c in ExitLaptopMode (uiMode=2) at Build/Laptop/Laptop.cc:1302
#5  0x0000000000485d13 in EnterNewLaptopMode () at Build/Laptop/Laptop.cc:825
#6  0x000000000048882c in HandleSlidingTitleBar () at Build/Laptop/Laptop.cc:2389
#7  0x00000000004865de in LaptopScreenHandle () at Build/Laptop/Laptop.cc:1228
#8  0x0000000000448189 in GameLoop () at Build/GameLoop.cc:299
#9  0x00000000006ada13 in MainLoop (msPerGameCycle=25) at sgp/SGP.cc:252
#10 0x00000000006ae1c9 in main (argc=3, argv=0x7fffffffe088) at sgp/SGP.cc:516
(gdb) 
```

[...]

PR #496, opened 2017-02-12
Fix laptop crash

momoko-h:
This is the moral equivalent of 'delete p; p = nullptr;' : it does not really address the underlying logic problem but it is an effective band-aid to prevent the crash. Fixes issue #68.


lynxlynxlynx:
Thanks, I manually cherry-picked it in (and fixed the indentation).

momoko-h:
I apologize for the inconvenience but this must be backed out, it breaks the web browser. Fragile little beast...

lynxlynxlynx:
But it looks correct. Does it just need extra checks?

PR #506, opened 2017-02-25
[RFC] Shrink gfKeyState

Manuel-K:
By using SDL_Scancode values instead of SDL_Keycode values to access elements of gfKeyState. Everything else does still use SDL_Keycodes.

Memory usage in main menu:
```
master:             RSS 63 MB, VIRT 1395 MB
shrink_gfKeyState2: RES 58 MB, VIRT  371 MB
```

This has the disadvantage that all _KeyDown() checks have to use SDL_Scancodes, which means that they check the physical position of the key instead of the localized value. On the plus all _KeyDown()s seem to be used for keys that are the same for all (most?) layouts.

**NOTE: I'm not sure if it works on all systems and I haven't even tested very thoroughly on ArchLinux.**

### DIFF
--- a/src/game/Tactical/Merc_Entering.cc
+++ b/src/game/Tactical/Merc_Entering.cc
@@ -452,7 +452,7 @@ void HandleHeliDrop( )
 			guiPendingOverrideEvent = LU_BEGINUILOCK;
 		}
 
-		if (_KeyDown(SDLK_ESCAPE))
+		if (_KeyDown(SDL_SCANCODE_ESCAPE))
 		{
 			// Loop through all mercs not yet placed
 			for ( cnt = gbCurDrop; cnt < gbNumHeliSeatsOccupied; cnt++ )
--- a/src/game/Tactical/Soldier_Control.cc
+++ b/src/game/Tactical/Soldier_Control.cc
@@ -4920,7 +4920,7 @@ void SetSoldierAniSpeed(SOLDIERTYPE* pSoldier)
 
 	AdjustAniSpeed( pSoldier );
 
-	if (_KeyDown(SDLK_SPACE))
+	if (_KeyDown(SDL_SCANCODE_SPACE))
 	{
 		//pSoldier->sAniDelay = 1000;
 	}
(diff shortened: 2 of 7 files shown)


lynxlynxlynx:
Thanks! Hardcoding a layout sounds like a bad plan. I forgot which bug this was discussed in, but wasn't the problem just from the mod keys being masked to higher bits? In that case I would prefer a smarter getter/setter, so the array can remain small without switching from keycodes.

momoko-h:
In that case, commit a3c7a5ae101b3a0810146e75f6b3acdd4bef757b from my experiments branch might be what you are looking for. I've played through a half campaign without noticing any problems but I can only test it under Fedora, so it would definitely require some additional testing with other OSs.

lynxlynxlynx:
Looks good to me. bitset is from c++11, so sure, it could be missing from some old platforms, but I think we burned that bridge long ago (if not, then rust did). Not your fault, but it also shows how _KeyDown is a silly name.

@selaux any thoughts?

Manuel-K:
I agree, it does look good.

I've abandoned my first idea to fix this issue (that was similar to this) because I could not find any information why `SDLK_SCANCODE_MASK` is so big and and if there really aren't any keycodes outside of [0... SDL_NUM_SCANCODES] and [SDL_SCANCODE_TO_KEYCODE(SDL_NUM_SCANCODES)…SDL_SCANCODE_TO_KEYCODE(SDL_NUM_SCANCODES)+SDL_NUM_SCANCODES].

momoko-h:
According to the [history of C++](http://en.cppreference.com/w/cpp/language/history), bitset became part of the standard with C++98, any compiler still not supporting it really isn't worth our time. As to the reason why SDL chose bit 30 as their mask bit: at least theoretically a keycode could be any Unicode character so you have to choose a bit that won't get in your way. Fortunately at the moment the set of keycodes generated by SDL is limited to the ones described in [SDLKeycodeLookup](https://wiki.libsdl.org/SDLKeycodeLookup) and AFAIK there aren't any plans to change that.

selaux:
https://github.com/ja2-stracciatella/ja2-stracciatella/commit/a3c7a5ae101b3a0810146e75f6b3acdd4bef757b looks good. We specified our standard as C++98 [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/CMakeLists.txt#L19), so we should be fine.

Manuel-K:
Used a3c7a5a in #523 instead.

Issue #497, opened 2017-02-13
Crash when entering single acute into IMP website password


Manuel-K:
The game segfaults after printing the messages
```
terminate called after throwing an instance of 'InvalidEncodingException'
  what():  Invalid UTF-8 encoding
```

This happens on Arch Linux with XkbLayout = de and XkbVariant = nodeadkeys.

JAsmine-ja2:
In vanilla, we can't enter a single acute (in fact, we can't enter anything special besides underlines and periods), so the bug can't be reproduced there.

I also can't reproduce the crash with Stracciatella 0.15.1 on Windows. I can enter single acutes without crashing.

Manuel-K:
The problem is more general than I thought.  If one of the problematic keys (not only ´, also ü, ö, ä and many other keys) is pressed, the game crashes anywhere. Even in the main menu. At first I suspected `nodeadkeys` to be the source of the problem, but that was clearly wrong.


Note: I'm using master, i.e. db7fa39972e61cbb58a39e6e210c23220cdb7c0b.

Manuel-K:
I'm still not sure why it does crash on my system and not on @JAsmine-ja2's, but the problem is in `void TextInput(const SDL_TextInputEvent* TextEv)` in `src/sgp/Input.cc`.

Manuel-K:
Should we add an `try{}`, `catch(){}` around the functions body? That would ignore invalid inputs.

JAsmine-ja2:
I can't help with a solution, but, for the record, this is how it looks here:

![lap1](https://cloud.githubusercontent.com/assets/17213795/23338391/00b4dc48-fc0a-11e6-8c57-56a1e8bd1695.jpg)
![lap2](https://cloud.githubusercontent.com/assets/17213795/23338392/00d7f61a-fc0a-11e6-8f5a-a43d440e5941.jpg)

Manuel-K:
That's very strange.

The problem is either in `src/sgp/UTF8String.cc` or it is in the generation of the `SDL_TextInputEvent`. Either way, adding a try, catch ignores the problem and I think that is a good idea, because pressing a wrong key should not crash the game on any platform.

selaux:
Yes, but entering `ö` should enter `ö` into the input 🙂 . You shouldn't be able to enter a character that is not UTF16-compatible. I have almost the same configuration and pressing these buttons does not crash for me. What is the output of `locale` for you?

[...]

PR #507, opened 2017-02-26
catch InvalidEncodingException& in TextInput(...)

Manuel-K:
This fixes the crash described in #497.

Although it does not fix the underlying issue, I think it is a good idea to add the `try{} catch{}` because it is very easy to hit a wrong key and lose all game progress.

### DIFF
--- a/src/sgp/Input.cc
+++ b/src/sgp/Input.cc
@@ -355,8 +355,20 @@ void KeyUp(const SDL_Keysym* KeySym)
 }
 
 void TextInput(const SDL_TextInputEvent* TextEv) {
-	UTF8String utf8String = UTF8String(TextEv->text);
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
It doesn't hurt, but I'd like to see why only you're having those problems so far.

selaux:
Maybe a warning would be nice in this case.

Manuel-K:
Yes, a warning would be nice…

`SLOGD(DEBUG_TAG_SGP, 'Received invalid utf-8 character.');`
on the first exception?

lynxlynxlynx:
I don't know if you get notifications about emoji responses, so: yes, please go ahead.

Manuel-K:
Done.

@lynxlynxlynx There are no notifications about emoji responses.

selaux:
LGTM

lynxlynxlynx:
Thanks!

PR #509, opened 2017-03-23
Deal more and explosive damage to enemies when killing everyone

### LINKED ISSUE #503: ALT+O cheat can't damage tanks
ALT+O cheat can't damage tanks - this is how it always worked. I guess, we'll never know if this was intended or if it is an oversight.
I think, this has some relevance, because the tanks are the most dangerous enemies, the player faces in the game (maybe together with stange biological experiment creatures and bloodcats). If the cheat is needed somewhere, then here, right?
What do you think - should the option for the ALT+O cheat to damage tanks be added to Stracciatella?

selaux:
Deal substantially more and explosiver damage to kill tanks and creatures.

Fixes #503 

NB: Not sure if appveyor issue needs fixing as it will probably be resolved with https://github.com/appveyor/ci/issues/1431

### DIFF
--- a/src/game/Tactical/Turn_Based_Input.cc
+++ b/src/game/Tactical/Turn_Based_Input.cc
@@ -2760,7 +2760,7 @@ static void ObliterateSector()
 	FOR_EACH_NON_PLAYER_SOLDIER(s)
 	{
 		if (s->bNeutral || s->bSide == OUR_TEAM) continue;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
What about travis? Similar curl problems.

Otherwise looks good; that function is also only used by the cheat key.

selaux:
So, the issue on travis is that cmake cannot locate cargo, the curl thingy is just a result from that. Maybe something changed with rustup... I will check.

Issue #444, opened 2016-11-09
Audio quality in master


JAsmine-ja2:
Just tested 0.15.x (#442)

I hear a BIG difference in the audio quality of the main menu music.
To me (as a musician) it is very obvious, that the quality of the 0.15.x builds is inferior to 0.15.1 (true for both 32 and 64 bit).
I could do a recording for comparison, but when you listen to the beginning of the main theme and listen to the sound of the cymbals, it should be obvious to everyone, that the 0.15.x-cymbal sounds very "noisy, trashy, dirty" compared to the clean sound of 0.15.1 (and vanilla).

Could this be because of some drastic changes like sdl 2?

I'm on Windows 8.1 64 bit.

lynxlynxlynx:
By 0.15.x you mean what we have in git, latest development stuff? Which revisions did you test (so it's easier to find potential culprits)?

JAsmine-ja2:
I tested all the windows releases from nighlies:
http://builds-ja2.stefanlau.com/nightlies/
I tested both versions (20161103 and 20161109) in both, 32 and 64 bit.

selaux:
While I can reproduce this on windows, on linux everything seems fine to me. I will have a look how a mingw build behaves, maybe it is a compiler specific issue.

selaux:
Nope, also there with a MinGW build, seems to be a Windows specific issue though...can anybody confirm?

JAsmine-ja2:
Just tested it. Can confirm that bad sound quality is present on MinGW builds.

selaux:
So setting the output sample rate to 44100 seems to fix it but when the sounds are upsampled they consume significantly more ram, so we have to do that on the fly... I will try to implement that.

selaux:
This is hard, after a few attempts I still cannot get audio streaming right... Take a look at [the fix-sdl2-audio branch](https://github.com/ja2-stracciatella/ja2-stracciatella/tree/fix-sdl2-audio), which is the one that consumes more memory right now. Suggestions are very welcome on how to continue 😉

PR #513, opened 2017-03-30
Fix sdl2 audio

selaux:
This sets 44100Hz 16bit as the target audio spec and converts all loaded files (currently in-memory) into this format.

- Makes output rendering easier
- Fixes bad audio on sdl2 (#444)

Issues:
- [ ] Reduce memory footprint by streaming audio
- [ ] Implement streaming from file which should be possible now

Both points proove quite hard to implement. Streaming is hard and SDL2 does not have a function to read only wav spec.

### DIFF
--- a/src/sgp/FileMan.cc
+++ b/src/sgp/FileMan.cc
@@ -21,6 +21,7 @@
 #endif
 
 #include "PlatformIO.h"
+#include "Debug.h"
 
 #if MACOS_USE_RESOURCES_FROM_BUNDLE && defined __APPLE__  && defined __MACH__
 #include <CoreFoundation/CFBundle.h>
@@ -28,11 +29,14 @@
 
 #if CASE_SENSITIVE_FS
 #include <dirent.h>
+#include <SDL_rwops.h>
+
 #endif
 
 // XXX: remove FileMan class and make it into a namespace
 
 #define LOCAL_CURRENT_DIR "tmp"
+#define SDL_RWOPS_SGP 222
 
 enum FileOpenFlags
 {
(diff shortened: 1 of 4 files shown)


Manuel-K:
Well… hard is not the right word… streaming from the disc is very **tedious** if you have to implement the caching yourself.


How much more memory does it require?

selaux:
@Manuel-K Looking at it it does not seem to change the max memory size that much, maybe 1MiB. Can you confirm?

@JAsmine-ja2 Could you please test if [this build](http://builds-ja2.stefanlau.com/pull-requests/513/) fixes the issue in general.

misanthropos:
Considering the to days availability of RAM (even for SmartPhones) the temporary overhead is likely , negligible, isn't it?
Besides: We already have a Sound System which we feed with Data (Buffer) and we would have to implement another one and would have to make sure they don't collide - correct me if I am mistaken.

JAsmine-ja2:
@selaux 
Tested your build. It fixes the bad audio.

Manuel-K:
I had to increase `SOUND_DEFAULT_MEMORY` in `src/sgp/SoundMan.cc` from 16 MB to get around a huge number of dropped sounds during the game. It's difficult to determine the exact amount of additional memory that is needed because of the caching mechanism. 

We should at least double the amount of memory.

Btw. the size of the wav file is used to calculate if a new sound can be played without breaking the limit, so the check is very inaccurate if we convert it afterwards.

lynxlynxlynx:
There's also the option to introduce another dependency and rip more code out. SDL_mixer comes to mind, but I don't know if has what you're looking for.

selaux:
I will try to increase the memory limits for now (and setting the used memory correctly). I the long term, we probably should still aim at streaming audio, as it will be vital for longer music tracks for example. Currently the sounds we have with vanilla ja2 are quite short, even the soundtrack.

@misanthropos True, but in general we should not use any resources lightly. Currently basically every sound is fully loaded into memory and "streamed" from there, so long tracks will have a big memory footprint.

@lynxlynxlynx I actually considered three approaches:
- Implementing Streaming in C++ (tedious as @Manuel-K already noted)
- Rewriting SoundMan in Rust (tedious as well and would require a lot of "plumbing" to be done for rust to access data, but personally more interesting 😉 )
- Using something like SDL_mixer (would introduce another big dependency, with multiple built versions to include in the repo, which with the current state of depdendency manangement I didn't want to do)

selaux:
So should we just merge this?

Manuel-K:
Yes. I've noticed no problems and the code seems fine.

I'm not completely happy with the whitespace. You've used spaces (for example in the lines with the `DecreaseSoundMemoryUsedBySample` calls), but the rest of `SoundMan.c` is indented with tabs.

lynxlynxlynx:
Can't test right now, so if it works for everyone that did, just go ahead.

Issue #510, opened 2017-03-24
can't load webpages


kavol:
Except for the I.M.P. web, I cannot get past the homepage of individual services - cannot buy weapons from Bobby, cannot hire mercs at A.I.M. ...
Clicking on some option from the respective web homepage, I get "Downloading ..." but it stays on the homepage, and on subsequent clicks on the same option it says "Reloading" and doesn't get me anywhere.
[noweb.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/869389/noweb.zip)

kavol:
reverting commit db7fa39972e61cbb58a39e6e210c23220cdb7c0b helps my issue

kavol:
hm, now I see RinneT already mentions that in the commit comments ...

lynxlynxlynx:
Yeah, pick one bug or the other.

Manuel-K:
I'm working on another fix for the laptop mode crash that outputs a warning at problematic places, but so far I haven't managed to trigger the bug. I've only seen it once and never managed to recreate it. 

Any tips?

Manuel-K:
It turns out that I had trouble to trigger the laptop crash because I was clicking too fast while thinking I was clicking too slow.

PR #514, opened 2017-04-05
Fix laptop page logic (issue #510)

Manuel-K:
Without reintroducing the laptop double free problem (issue #68).

The solution is based on the previous one by @momoko-h, but introduces a new
variable to store the information if a page has been exited.
Untangling the page transition mess seems highly complicated without rewriting
the whole laptop code.

This also changes a global variable to a local variable.


selaux:
Can you share how to reproduce #510 because it does not say how in the issue. Codewise LGTM, thanks for the fix.

Manuel-K:
Sure. 

How to reproduce #510: Open the A.I.M. on the browser. Click on members. It should display the downloading box and stay on the main A.I.M. page.

selaux:
Sweet, looks like it's fixed.

PR #515, opened 2017-04-06
Fix and enhance cli switch parsing

selaux:
Fixes:
- JSON options being prefered over cli switches

Adds:
- Datadir cli switch to set vanilla data directory

### DIFF
--- a/ja2_manpage
+++ b/ja2_manpage
@@ -1,6 +1,6 @@
 .\" Manpage for ja2.
-.\" Contact gennady.trafimenkov@gmail.com to correct errors or typos.
-.TH ja2 6 "04 May 2014" "1.0" "ja2 man page"
+.\" Goto https://github.com/ja2-stracciatella/ja2-stracciatella to report or correct errors or typos.
+.TH ja2 6 "30 March 2017" "1.0" "ja2 man page"
 .SH NAME
 ja2 \- JA2 Stracciatella
 .SH SYNOPSIS
@@ -11,6 +11,9 @@ This is only the game engine.  Game resources from the original JA2 are required
 .TP
 When you first start the game, file \fB~/.ja2/ja2.json\fR is created.  You should edit this file, so that \fBdata_dir\fR points to the directory with game resources from the original JA2 game.
 .SH OPTIONS
+.TP
+\fB\-datadir /opt/ja2\fR
+Path to vanilla data
 .TP
 \fB\-res WxH\fR
 Screen resolution, e.g. 800x600. Default value is 640x480
@@ -52,4 +55,4 @@ Turn the sound and music off
 \fB\-window\fR
 Start the game in a window
 .SH AUTHOR
-Gennady Trafimenkov (gennady.trafimenkov@gmail.com)
+The JA2 Stracciatella Team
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
LGTM, but I can't test yet (will return in a week).

Manuel-K:
I've messed up the path in ja2.json and was able to override it with -datadir using an absolute path. Relative paths do not work.

selaux:
@Manuel-K Nice catch! I will try to add relative path support to the PR as it definitely makes sense.

Manuel-K:
Can we merge this and leave the support of relative paths for later?

lynxlynxlynx:
Sure, merge and open an issue for the rest, so it's not forgotten.

Manuel-K:
Okay. That wasn't a good idea.

Is seems to have caused a lot of CI issues, although the x64 platform on AppVeyor seems to be broken for everything at the moment. 

Travis reports a problem on Mac:
```
thread 'tests::parse_json_config_should_return_the_correct_canonical_data_dir_on_linux' panicked at 'assertion failed: `(left == right)` (left: `"/private/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/ja2-tests.7FhrfeFjMhqO"`, right: `"/var/folders/my/m6ynh3bn6tq06h7xr3js0z7r0000gn/T/ja2-tests.7FhrfeFjMhqO"`)', src/stracciatella.rs:650
```

Manuel-K:
Should I disable the test on Mac by changing the `#[cfg(not(windows))]` to `#[cfg(linux)]`?

Manuel-K:
Okay. It seems that the corresponding test on Windows fails as well.

Manuel-K:
@lynxlynxlynx I'm in favour of reverting this for now.

lynxlynxlynx:
The only difference in the paths is that left has /private prepended. Maybe a travis thing we're not accounting for, since this is a new test?

selaux:
That is why I didn't proceed right away, especially on windows this results in weird directories...

lynxlynxlynx:
Couldn't find any references to /private, but it's also impossible to search for. I guess it would be easiest to stick a bunch of prints in the test locally, as it appears it's actually the temp dir creation module where the problems start.

Manuel-K:
Similar issue on Windows:
```
 thread 'tests::parse_json_config_should_return_the_correct_canonical_data_dir_on_windows' panicked at 'assertion failed: `(left == right)` (left: `"\\\\?\\C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\ja2-tests.P2CY6hCvnVJT"`, right: `"C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\ja2-tests.P2CY6hCvnVJT"`)', src/stracciatella.rs:667
```
It adds `"\\\\?\\`.

Anyhow: Reverting the last commit seems to be the way to go. I've pushed a branch that does that. Let's see if the CI still complains.

selaux:
Relevant Rust issue for Windows: https://github.com/rust-lang/rust/issues/42869. So as I read this it might be ok to simply remove this prefix...

For macOS my guess is that this is a resolved symlink (as this is also done by canonicalize and the path looks almost correct).

Manuel-K:
It looks that way. This means that with the current code only the test fails on Mac while relative paths are completely broken on Windows.

selaux:
Well for Windows paths we need to strip the prefix (they might even work with them, we could check that) and for macOS only the tests are broken.

PR #517, opened 2017-04-08
"Fix" exception when attaching ceramic plates to armour

### LINKED ISSUE #516: Game crash when inserting ceramic plates
```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Opening file 'sounds/ceramic armour insert.wav' failed
```
Is the sound file is missing or incorrectly named?
Save game file: [error.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/907429/error.sav.zip)

elandesign:
Current master crashes with the following exception:
```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Opening file 'sounds/ceramic armour insert.wav' failed
```

Guessing it's a missing or misnamed sound file, but as I'm unfamiliar with the codebase it's easier to noop for now, unless someone can point me in the right direction?

Fixes #516


lynxlynxlynx:
Thanks! I'll close this in favour of #518 and we want to find the source of the problem too.

elandesign:
Agreed, definitely a better fix than mine. 👍

PR #519, opened 2017-04-10
Fix ceramics sound

### LINKED ISSUE #516: Game crash when inserting ceramic plates

selaux:
Should close #516. Also makes development builds throw again in development builds.

**I could not test this though as i have no save game with this situation**


COMPILATION.md
```diff
@@ -22,6 +22,16 @@ We suggest to install Rust and Cargo using [rustup](http://rustup.rs/). This way
 installed in your home directory. As rust is a rapidly developing language the binaries provided by your distribution
 might be too old to build ja2-stracciatella and its dependencies.
 
+## Building in release configuration
+
+If you want a **stable game experience**, we suggest to build in release mode. This will disable some runtime checks
+and will result in a lot less runtime crashes. This is especially cucial if you play in "Dead is Dead" mode, as runtime
```
lynxlynxlynx: *crucial

selaux: Fixed

COMPILATION.md
```diff
@@ -22,6 +22,16 @@ We suggest to install Rust and Cargo using [rustup](http://rustup.rs/). This way
...
 
+## Building in release configuration
+
+If you want a **stable game experience**, we suggest to build in release mode. This will disable some runtime checks
+and will result in a lot less runtime crashes. This is especially cucial if you play in "Dead is Dead" mode, as runtime
+crashes might result in loss of game progress. To enable release mode, simply add another cmake variable
+
+```
+cmake ...other switches... -DCMAKE_BUILD_TYPE=Release path/to/source
+```
```
lynxlynxlynx: Two things:
- IMO it's better to suggest RelWithDebInfo, since the cores are still useful there
- we could just override the default manually and if the majority doesn't have to specify extra flags, that's a good thing, plus the doc change becomes unneeded (or changed to something for developers).

selaux:
@lynxlynxlynx Good proposal 👍. I changed the default build type to `RelWithDebInfo`. Needed to up the rust version though to do the same for rust code.

Issue #521, opened 2017-04-15
Automatic resource tests


Manuel-K:
Like I mentioned in #516, I'm still convinced that we should add a tool that checks if all resource paths that are hardcoded (or externalized) exist.

I started looking into a sound tester and discovered that the sound file name definitions are **not** restricted to `game/Utils/Sound_Control.cc` but are in fact all over the place (`grep` for `SOUNDSDIR`). We might want to move them to one place as a first step.

Adding the check as an unit test does not make sense, because we would have to upload the resource files to AppVeyor and Travis, so I'm thinking about either a separate command line switch or even a separate executable.

Opinions, please…

Manuel-K:
The tester could be written in python and use ja2py and read the resource paths from the source files using regular expressions.

Manuel-K:
Obvious difficulties:
```
game/Tactical/Soldier_Init_List.cc:             sprintf(str, SOUNDSDIR "/cowmoo%d.wav", Random(3) + 1);
game/Tactical/Weapons.cc:                                       sprintf( zBurstString, SOUNDSDIR "/weapons/silencer burst %d.wav", pSoldier->bBulletsLeft );
game/Tactical/Weapons.cc:                                       sprintf( zBurstString, SOUNDSDIR "/weapons/%s%d.wav",
```

Manuel-K:
Okay. I've hacked a python script together that does:
- Search for lines containing `SOUNDSDIR` or `STSOUNDSDIR`
- Checks if the files that are referenced are in sounds.slf
- Checks if there are matching local files
- Uses difflib.SequenceMatcher to check if there are files with similar names/paths in sounds.slf

It needs a bit of a cleanup, but the output seems promising. It finds the problem mentioned in #520 and its solution.


[check_sounds_out.txt](https://github.com/ja2-stracciatella/ja2-stracciatella/files/923548/check_sounds_out.txt)

lynxlynxlynx:
That's good enough for me. Now that we have a baseline, it shouldn't happen again. As for the externalised ones, if someone changes them, it's their own responsibility.

Manuel-K:
I've extended the script to search all paths from `src/game/Directories.h`. 

Found some more (non-sound) issues: [check_resources_out6.txt](https://github.com/ja2-stracciatella/ja2-stracciatella/files/923599/check_resources_out6.txt). It looks like there are several missing animations.

[...]

PR #522, opened 2017-04-15
Change "/doorcr_b.wav" to "/doorcr.wav"

### LINKED ISSUE #520: Sound file exception: doorcr_b.wav
Another missing sound file. I've not updated my binary from master yet which would have caught and handled the exception, but I figured someone would want to know about this one to correct the filename.

```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Opening file 'sounds/doorcr_b.wav' failed
```

Save file: [error.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/923093/error.sav.zip)
(the exception occurs when a merc dies, so not sure if this save file will be useful)

Manuel-K:
This fixes #520 (and therefore a part of #521)


lynxlynxlynx:
Let's hope it's a good sound and it wasn't just an ugly way to remove it.

PR #523, opened 2017-04-15
Shrink gfKeyState (alternative)

Manuel-K:
Alternative to #506 using @momoko-h's code.

Shrink the size of gfKeyState from 1Gb + 512 to 128 bytes

Remap SDL keycodes with bit 30 set to make them usable as gfKeyState indices.

### DIFF
--- a/src/sgp/Input.cc
+++ b/src/sgp/Input.cc
@@ -1,3 +1,4 @@
+#include <bitset>
 #include "UTF8String.h"
 #include "Types.h"
 #include "Input.h"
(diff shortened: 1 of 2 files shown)


Manuel-K:
Please do not merge, yet – I want to run some more tests.

Manuel-K:
Memory usage in main menu:
```
Before: RSS 68.7 MB, VIRT 1434 GB
After:  RSS 69.4 MB, VIRT  398 MB
```


---

Quick reminder: This change was discussed in #506 as a3c7a5a and okayed by everyone.

PR #525, opened 2017-04-19
Replace reference to non-existing flame animation

Manuel-K:
with the missile smoke animation.
The missing animation could lead to crashes when accessing
the unfinished flame thrower weapon using cheats.

Resolves a part of the issues mentioned in #521


lynxlynxlynx:
probably looks horrible, but that's still better than a crash.

JAsmine-ja2:
IIRC, some early mods that enabled the flamethrower also used the smoke trail animation (from the rocket rifle) with the color palette swapped towards a yellowish color.
Does indeed not look that good 

> but that's still better than a crash.

The only other option would be to rip the animation from v1.13 - ofc with proper asking and crediting and stuff.

Issue #528, opened 2017-05-28
Getting "Reading from file failed" when fast forwarding time after savegame load


Tercus:
Whenever I load the savegame and then want to forward the time, the game crashes with this message:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Reading from file failed
Creating an emergency savegame succeeded (error.sav).
Please report this error with a description of the circumstances. Do not forget to attach the savegame.
Aborted
```

I attached the created savegame. I can also attach one of my own if needed.
[error.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1034563/error.zip)

lynxlynxlynx:
What version are you using?

Manuel-K:
Some information about your operation system might also be useful.

Please attach the other savegame as well.

Tercus:
I compiled the latest sourcecode under KDE neon (based on ubuntu 16.04). The compilation worked perfectly and the game ran smoothly for quite some time. No problems with saving or loading saves. It just started one day and affected all savegames.

I can still use older savegames I had around and fast-forward time there.

Tercus:
Here are all the Savegames from my playthrough so far.

[SavedGames.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1044908/SavedGames.zip)

Tercus:
I was able to avoid a crash by loading a savegame, going into the sector and leaving the sector "manually". Since I was in a town, the 5 minutes had passed and from then on I was able to use fast forward again.

Manuel-K:
I'll try to reproduce the crash tomorrow.

Which `CMAKE_BUILD_TYPE` did you use?

Tercus:
CMakeCache.txt tells me I used this:
`//Choose the type of build, options are: Debug RelWithDebInfo.
CMAKE_BUILD_TYPE:STRING=RelWithDebInfo`

I hope that's the right info you needed.

Manuel-K:
The crash is related to ambient sounds. Apparently `LoadAmbientControlFile()` is never called and therefore `gAmbData` is not initialized which in turn makes `SetupNewAmbientSound()` make a chain of calls that end with `SoundLoadDisk()` being called with an empty string (`""`) as argument.

Tercus:
may I ask how you managed to trace back the error? I'd like to be more helpful, but my knowledge is limited in that area...

Manuel-K:
[...]

PR #530, opened 2017-06-02
Check if argument of SoundLoadDisk() is an empty string.

Manuel-K:
Disregard the invalid sound on release builds. Throw on debug builds.

Superficial fix for #528. Obviously the underlying issue should be fixed as well.


lynxlynxlynx:
Just as a note, since you came into the project later. We decided to not push to master directly, but always use PRs, to force peer review. So next time please wait for at least one LGTM or actual comments.

Manuel-K:
@lynxlynxlynx Sorry. I thought this change trivial enough to skip that.

PR #532, opened 2017-06-02
Fix issue with spaces in build path

selaux:
I dont have a lot of time currenly for this project, but hopefully enough to fix #529 🙂 .

Issue #529, opened 2017-06-02
Build fails if there is a space in the directory name


Manuel-K:
@selaux: I just discovered this because Qt Creator's defaults folder name for a RelWithDebInfo build contains spaces.

The rust portion of the build process fails. 

```
[Manuel@K ja-stracciatella.git]$ mkdir "build with space"
[Manuel@K ja-stracciatella.git]$ cd build\ with\ space/
[Manuel@K build with space]$ cmake ..
-- The C compiler identification is GNU 7.1.1
-- The CXX compiler identification is GNU 7.1.1
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Setting extra data dir to
<FindSDL2.cmake>
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Looking for pthread_create
-- Looking for pthread_create - not found
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
</FindSDL2.cmake>
-- Found SDL2: /usr/lib/libSDL2main.a;/usr/lib/libSDL2.so;-lpthread  
-- Boost version: 1.64.0
-- Found the following Boost libraries:
--   filesystem
--   system
-- Compiling with local RapidJSON libraries from /home/Manuel/ja-stracciatella.git/dependencies/lib-rapidjson
-- Found Cargo: /usr/bin/cargo (found suitable version "0.18.0", minimum required is "0.9.0") 
-- Found Rustc: /usr/bin/rustc (found suitable version "1.17.0", minimum required is "1.15.0") 
-- Compiling with unittests
-- Compiling with local GTest libraries from /home/Manuel/ja-stracciatella.git/dependencies/lib-gtest
-- Found PythonInterp: /usr/bin/python (found version "3.6.1") 
-- Configuring done
-- Generating done
[...]

PR #535, opened 2017-06-04
Fix rust build issues with space in build path (second try)

Manuel-K:
Alternative to #532 

This version works as a solution to #529 for me.


selaux:
Thx, LGTM. I don't clearly remember why the native stuff was introduced, and it works without on all supported systems so it is probably not needed...

PR #544, opened 2017-06-07
Fix the issues mentioned by @chmallon in #425

Manuel-K:


Manuel-K:
Right. Github does not pick up references in title. So here we go again: #425

chmallon:
I think the changes regarding Flo are against the gist of her biography.
It clearly states firearms and also the comments in the code say guns and explicitly not launchers.
Now for example it also includes knives.
Also I think this price discrepancy also appears with other gun attachments like the sniper scope (an IC_MISC).

I would revert back to the original state.

lynxlynxlynx:
It's such a minor thing, but initially I also thought this was excessive, suggesting only ammo be added to the list. We can't ever make everyone happy unless we make the check configurable.

chmallon:
> suggesting only ammo be added to the list.

Ammunition is already handled in vanilla by 'case IC_AMMO:'.

> We can't ever make everyone happy unless we make the check configurable.

The proposed change goes against the information in the game (both data and source code).
Also it does not fix the problem with attachments completely.
For both points see me previous comment.
So it is neither correct nor complete. ):

Manuel-K:
> The proposed change goes against the information in the game (both data and source code).

No, it does not. We might be arguing semantics here, but this is obviously false.

The design change was made in #425. You might argue that that change conflicts with some of the game's information, but it does not concern this proposed change.

This proposed change fixes a bug that was introduced during implementation of the previous design change.

chmallon:
> > The proposed change goes against the information in the game (both data and source code).
> 
> No, it does not. We might be arguing semantics here, but this is obviously false.

Uh, plain no.
The behavior is clearly against both Flo's bio and the comments in the source code.
And yes, this is about semantics, i.e. meaning, what else should it be about?

> This proposed change fixes a bug that was introduced during implementation of the previous design change.

The previous change should have never been done, so this proposed "fix" just makes it worse.
The best course of action is to revert the first "fix".
"Correcting" the broken change is not a suitable solution.
This is obvious because it contains such non-sensical cases like 'IC_TENTACLE'!

Manuel-K:
> The previous change should have never been done, so this proposed "fix" just makes it worse.

You may argue with @selaux and @JAsmine-ja2 at #425 if you want, but not here. It was their decision to make the change and I this is about a technical change that fixes a mistake that was made during the implementation of that design decision. Personally I tend to agree with their decision as it fixes an annoying ambiguity mentioned in #423.

Manuel-K:
@lynxlynxlynx Merge?

lynxlynxlynx:
alien tentacles really don't make sense, otherwise it looks good.

Manuel-K:
> alien tentacles really don't make sense, otherwise it looks good.

Done. I only included it because it is a part of the `IC_WEAPON` definition.

chmallon:
> > alien tentacles really don't make sense, otherwise it looks good.
> 
> Done. I only included it because it is a part of the IC_WEAPON definition.

This is because the whole "fix" did not make sense in the first place.
Now it is still broken and additionally non-sensical.
Blades and knives are not "firearms" (which is what Flo's skill is about).
Also it excludes launchers, which is stated *right* before the switch:

> // if it's a GUN or AMMO (but not Launchers, and all attachments and payload is included)

This change just makes the broken "fix" worse, because it clearly contradicts Flo's bio and the comments in the code.
Do you understand that?

Manuel-K:
@chmallon I've told you before: This is the wrong pull request for such comments and I'm not the one to talk to.

derSteFfi:
@Manuel-K I don't think it makes sense to argue in two seperate PRs touching the same spot. We can just find a solution here.
In #425 , @selaux and @JAsmine-ja2 seem to have changed that by design, so the discussion has reasons. Let's please go through this like adults, and please do it here, since #425 is closed and we have this PR handling the code in question.

I'm also going to reopen the discussion (I see no bad language here, just two folks not finding a common ground)

Manuel-K:
> I'm also going to reopen the discussion (I see no bad language here, just two folks not finding a common ground)

No, not bad language as such. Only passive aggressive behaviour and statements clearly meant as insults.

Either way. I'm out of this discussion as I've said many times.

lynxlynxlynx:
I agree with Manuel — 425 is where the discussion should be held, if anyone still wants to have it. It's trivial to reopen tracker items. This one here is practically unrelated and would just mean jumping back and forth between both threads.

PR #546, opened 2017-06-08
Fix slog context numbering comments

Manuel-K:
I know it's a trivial change, but the wrong numbering has fooled me more than once.


derSteFfi:
Yeah, that wasn't indexing but counting... I'd approve nonetheless :)

Manuel-K:
Well, it was a mixture of both. It changed in the middle. ;-)

PR #547, opened 2017-06-08
Fix #540: always show militia dialogues in the same spot

### LINKED ISSUE #540: Maximum militia Message at the wrong position
This is probably also related to higher resolution but this just happened:
screenshot_20170606_031217

The savegame is [here](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1053547/QuickSave.zip). Just let the time continue until 6:00 AM and you get that error. What happens is that two teams finish their task of training militia, but one team cannot continue (max militia).

derSteFfi:
inline the remaining one line function for the sake of clarity

### DIFF
--- a/src/game/MessageBoxScreen.cc
+++ b/src/game/MessageBoxScreen.cc
@@ -582,27 +582,18 @@ void MessageBoxScreenShutdown()
 	gMsgBox.box = 0;
 }
 
-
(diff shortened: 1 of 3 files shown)


Manuel-K:
Echoing the issue number because github ignores the title: #540

lynxlynxlynx:
LGTM here too, but haven't tested. I'm fine with the design decision though.

Manuel-K:
I've tested the changes using the provided savegame.

lynxlynxlynx:
yay, team work. :)

PR #548, opened 2017-06-08
Initial Launcher Implementation

Manuel-K:
Rebased version of #270. Updated to RapidJSON.


Manuel-K:
CI will probably fail because I've nuked the FLTK source code.

lynxlynxlynx:
why a different branch though?

Manuel-K:
I wanted to keep the original for reference because I did have to do a lot of changes during rebase.

derSteFfi:
for the SLOG change:
you should also increase NUMBER_OF_TOPICS and introduce a TAG in the enum in slog.h

Manuel-K:
>  you should also increase NUMBER_OF_TOPICS and introduce a TAG in the enum in slog.h

I think @selaux's and my changes do cover everything necessary. We've increased `NUMBER_OF_TOPICS` and added an entry to `SLOGTags` in `slog.c` and to `SLOGTopics` in `slog.h`.

Is there anything else that we missed?

derSteFfi:
yes, only looked at your commit, sorry :)

Manuel-K:
I'm looking into adding fltk as an cmake external project.

Manuel-K:
FLTK's readme says that `fltk_wrap_ui` is broken. In any case it causes huge issues when trying to build FLTK via the external dependency mechanic. cmake builds all external dependencies during the build step (i.e. **after** the cmake step), but it requires the FLTK's `fluid` binary during the `cmake` step when the `fltk_wrap_ui` macro is used.

I've tried to switch to an `add_custom_command` based approach like the readme suggests, but it seems to cause the same issue: cmake still wants to generate the form files during the `cmake` step.

Manuel-K:
I don't see a way to cleanly build FLTK automatically.

I think we should try to include this in 0.16 without the external_project stuff, i.e. use a system wide installation of FLTK or make the user point to FLTK manually. Obviously building the launcher should be optional.

selaux:
I pushed some more commits that unify json serializing and unserializing by going through rust. Now the launcher only updates the json file and then runs ja2 without arguments.

A lot of changes were done in the rust library, moving deserialization to the serde library and getting serialization support through it. The launcher now is more like a thin wrapper around the rust library. There are still some references to `sgp` in `main.cc` though.

We could get around the fluid issue by comitting the generated `StracciatellaLauncher.h` and `StracciatellaLauncher.cc`, or even writing them by hand. This way we could include fltk as an external project.

I'm still not very happy with the look though. Not sure what we can do about that.

Manuel-K:
> We could get around the fluid issue by comitting the generated StracciatellaLauncher.h and StracciatellaLauncher.cc, or even writing them by hand. This way we could include fltk as an external project that we only need to link to. Would that work?

I'm not sure. We might have to patch FLTK's CMakeLists.txt or provide our own file as well.

selaux:
Allright. This now builds and run in all CI environments. I've also checked that it can actually run the game in both Windows and Linux.

fltk is now downloaded and built as an external project and linked statically, so there is no need to include the shared library.

Things to still do:
- Fix data directory not being applied
- Pass build type to fltk
- Clean up a little

selaux:
Allright, should now be mergeable.

PR #553, opened 2017-06-11
Fix #537: use dynamic values for screen height and width when showing…

### LINKED ISSUE #537: Position of Mercenary task orders in Tactical screen
I Tactical screen, right-clicking a mercenary let's you give them a task (training, repairing, etc). The position of that little window is a bit off. I guess it should position itself next to the mercenary, but instead it is somewhere else:
screenshot_20170604_175226
screenshot_20170604_175241
It appears that the further down to the visible screen a mercenary is, the more worse the position of the little window becomes.

I'm playing with 1680x1050 on KDE neon (based on ubuntu 16.04). Can anyone confirm that bug?

derSteFfi:
… assigment boxes

fixes #537

### DIFF
--- a/src/game/Strategic/Assignments.cc
+++ b/src/game/Strategic/Assignments.cc
@@ -5564,8 +5564,8 @@ void SetTacticalPopUpAssignmentBoxXY()
 
 	// ATE: Check if we are past tactical viewport....
 	// Use estimate widths/heights
-	if (sX > g_ui.m_screenWidth  - 100)	sX = g_ui.m_screenWidth  - 100;
-	if (sY > g_ui.m_screenHeight - 130)	sY = g_ui.m_screenHeight - 130;
+	if (sX > g_ui.m_screenWidth  - 50)	sX -= 90;
+	if (sY > g_ui.m_screenHeight)		sY = g_ui.m_screenHeight;
 
 	gsAssignmentBoxesX = sX;
 	gsAssignmentBoxesY = sY;
--- a/src/game/Strategic/Assignments.cc
+++ b/src/game/Strategic/Assignments.cc
@@ -5564,8 +5564,8 @@ void SetTacticalPopUpAssignmentBoxXY()
 
 	// ATE: Check if we are past tactical viewport....
 	// Use estimate widths/heights
-	if (sX > SCREEN_WIDTH - 100) sX = SCREEN_WIDTH - 100;
-	if (sY > 320 - 130)          sY = 190;
+	if (sX > g_ui.m_screenWidth  - 100)	sX = g_ui.m_screenWidth  - 100;
+	if (sY > g_ui.m_screenHeight - 130)	sY = g_ui.m_screenHeight - 130;
 
 	gsAssignmentBoxesX = sX;
 	gsAssignmentBoxesY = sY;


lynxlynxlynx:
does it still look good on the original resolution? Seems like the height threshold is now higher.

Manuel-K:
On the original resolution the box can open on top of the merc if he/she is on the lower right corner of the screen. Without the change the merc isn't hidden – the box opens higher just above the mercs head.

lynxlynxlynx:
Did you mess up that double negative or does it really make it worse for the original?

Manuel-K:
That is a matter of taste. I don't think it is worse, just different.

Manuel-K:
Just to make it clear: `g_ui.m_screenHeight` is 480 per default and the original code used 320.

Manuel-K:
@derSteFfi I've change the y-limit to the original value. Feel free to drop my commit if you don't like it.

derSteFfi:
I could not tell which one I like better. I think the additionally substracted 120 reason in the team panel on old resolutions, so the 320 would not be necessary, but it a) looks good, too and b) ist the original behaviour, so...

Manuel-K:
I prefer your version… but both are better than the current hard coded version.

Can we get some other opinions?

Manuel-K:
Forgot to mention why I like the first solution: The box is opened closer to the location of the cursor.

Manuel-K:
Screenshot time (at original resolution) – maybe this brings some movement to this PR:

Before the change:
original_bottom
original_bottomright
original_right

Version 1:
v1_bottom
v1_bottomright
v1_right

Version 2:
v2_bottom
v2_bottomright
v2_right


Version 2 recreates the original behaviour at the original resolution, but I prefer Version 1.

lynxlynxlynx:
do they work properly near map corners?

Manuel-K:
Good question… I haven't tested that. I assume that only the screen borders matter. 

I'm working through Stephan's comments in 556 right now. I'll take a look when I'm done.

**You haven't answered the important question: 1 or 2?**

lynxlynxlynx:
I don't care about cosmetics, it just has to work.

Manuel-K:
I have no detected any difference at the map borders/corners.

derSteFfi:
I'm pro version 1. The 320 accounted for the team panel in original resolution, which we already account in the constants anyway. It's closer tothe merc and looks better to me. I did not test, but the borders of the map _should_ have no influence at all. Like if they have, that is a new bug.

Manuel-K:
This is a screenshot of the issue with **Version 1** that I mentioned earlier. It is possible to hide the merc if he/she stands in one specific position relative to the screen borders:
v1_hidden

derSteFfi:
hm. I am playing around with a solution that always puts the box right of the merc, exxcept when there is no space, than put it left. The vertical position is simply clipped to the screen. I'm still tweaking the numbers here and there because I can not seem to find the width of a merc or the box in pixels but in pseudo, it looks like:
```if(sX > screenWidth - CONST_A) sX -= CONST_B;```
CONST_A acoounts for the merc itself and the message box, CONST_B shall be twice the width of the box + the width of the merc.

Manuel-K:
I've dropped the version 2 changes with a force push.

@derSteFfi: Do you want to continue playing around with your future version or should we merge this as is? The version on master is definitely worse than **version 1**.

derSteFfi:
I won't be able to have time the next two weeks, so we should merge this but leave the issue open maybe

derSteFfi:
force pushed this variant. imho, it is the cleanest solution. When we ever get back to smoothless scrolling, we might have to adjust the pixel counts, but we will see then.

Manuel-K:
Looks good.

Manuel-K:

Issue #62, opened 2013-08-05
UI stretching/scaling


lynxlynxlynx:
Originally reported by: **Anonymous**

---

Though the in-game resolution can be modified, this option may not be much of use for some configurations due to UI elements becoming too small and hard to use.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/62

lynxlynxlynx:
sdl2 natively supports scaling. It works in windowed mode and you just have to resize the window. Not sure if all these hardcoded coordinates would be problematic though, but from my experience it also scales event handling.

supercom32:
I assume this is the same issue, but I did notice that when I set my game resolution with the appropriate command line options, that the UI remains at 640x480 with no stretching. When I enter actual gameplay, the map expands to cover the full resolution specified, but everyone looks super small given the original game's 640x480 size. 

Judging from the previous post, the workaround is to simply set the game into Windowed mode, and then resize the window?

Manuel-K:
In principal - yes. But SDL2 does not allow window resizing per default and we do not set the option to explicitly allow it. You need to add the option `SDL_WINDOW_RESIZABLE` to the `SDL_CreateWindow` call in `src/sgp/Video.cc`, i.e. add the line `g_window_flags |= SDL_WINDOW_RESIZABLE;` before the `        g_game_window = SDL_CreateWindow(APPLICATION_NAME,` line.

Manuel-K:
Note: This might break something if you resize the window.

supercom32:
I see. I guess given the fact that it probably won't work correctly, I won't bother with it. However, it would be nice to have 2x, 3x, etc scaling options to take a lower resolution and just blow it up so it's easier to see (even if it looks more pixely - personally I like razor sharp pixels, but I digress. :-) ).

Manuel-K:
> I guess given the fact that it probably won't work correctly, I won't bother with it. 

I have only tested it briefly, but I wouldn't be so pessimistic. It seemed to work.

> . However, it would be nice to have 2x, 3x, etc scaling options to take a lower resolution and just blow it up so it's easier to see (even if it looks more pixely - personally I like razor sharp pixels, but I digress. :-) ).

Yes, that would be better. I'd like to have such scaling options myself, but there are so many (other) issues with the code base…

lynxlynxlynx:
[...]

PR #555, opened 2017-06-12
Allow resizing of game window

Manuel-K:
Partial fix of #62.


Manuel-K:
IMO this is not a fix, but a useful workaround.

PR #557, opened 2017-06-16
Minor cleanups

derSteFfi:
Stumbled upon this while reviewing the whitespace changes.You can force to trigger this code for testing by initialising ```gfCheckForFreeSpace...``` to TRUE. I tested the inner if by comparing the free space to ```(uintmax_t) -1```.


lynxlynxlynx:
Looks like DoesUserHaveEnoughHardDriveSpace could go now too.

derSteFfi:
very possible :) I'll look into it, as soon as I can (probably tonight or tomorrow evening). If that's the case, I would incude it in this PR.

derSteFfi:
works fine without it, removed it :)

Issue #559, opened 2017-06-26
The rust cli parser does not accept relative paths


Manuel-K:
Discovered during testing of #515.

Manuel-K:
Should be solved.

selaux:
Retested and works for me. I didn't test on Windows though...

Manuel-K:
Can someone test this one Windows, please?

selaux:
Tested on Windows. Works.

PR #561, opened 2017-07-02
Fix rust datadir parsing

Manuel-K:
This should (hopefully) fix the UNC prefix issue on Windows and the `/private` prefix issue on Mac.

Related to #515 and #559.

### DIFF
--- a/rust/src/stracciatella.rs
+++ b/rust/src/stracciatella.rs
@@ -645,8 +645,8 @@ mod tests {
     }
 
     #[test]
-    #[cfg(not(windows))]
-    fn parse_json_config_should_return_the_correct_canonical_data_dir_on_linux() {
+    #[cfg(target_os = "macos")]
+    fn parse_json_config_should_return_the_correct_canonical_data_dir_on_mac() {
         let mut engine_options: super::EngineOptions = Default::default();
         let temp_dir = tempdir::TempDir::new("ja2-tests").unwrap();
         let dir_path = temp_dir.path().join("foo");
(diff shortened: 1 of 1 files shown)


Manuel-K:
@selaux Even if it works it's very ugly. My knowledge of rust is barely minimal.

Manuel-K:
Okay. Finally all tests pass.

rust/src/stracciatella.rs
```diff
@@ -648,7 +657,15 @@ mod tests {
 
         assert_eq!(super::parse_args(&mut engine_options, input), None);
         unsafe {
-            assert_eq!(str::from_utf8(CStr::from_ptr(super::get_vanilla_data_dir(&engine_options)).to_bytes()).unwrap(), temp_dir.path().to_str().unwrap());
+            let comp = str::from_utf8(CStr::from_ptr(super::get_vanilla_data_dir(&engine_options)).to_bytes()).unwrap();
+            let mut base = String::new();
+
+            // allow /private prefix on mac:
+            if comp.starts_with("/private") {
```
selaux: Imho it might be better to call fs::canonicalize on `temp_dir.path()`, because that is what we are expecting (otherwise this might fail when travis changes something in their directory structure...

selaux:
So just get on the hypetrain 🙂 . LGTM, except for the comment.

rust/src/stracciatella.rs
Manuel-K: Ok. But I think we should separate the tests into linux and mac functions and do the caonicalized assert only on mac.

rust/src/stracciatella.rs
```diff
@@ -193,7 +193,16 @@ fn parse_args(engine_options: &mut EngineOptions, args: Vec<String>) -> Option<S
 
             if let Some(s) = m.opt_str("datadir") {
                 match fs::canonicalize(PathBuf::from(s)) {
-                    Ok(s) => engine_options.vanilla_data_dir = s,
+                    Ok(s) => {
+                        let mut temp = String::from(s.to_str().expect("Should not happen"));
+                        // remove UNC path prefix (Windows)
+                        if temp.starts_with("\\\\") {
+                            temp.drain(..2);
+                            let pos = temp.find("\\").unwrap() + 1;
```
selaux: Unwrap should be avoided as much as possible imho. We could do it another way with the `regex` crate, but that might be something for the future.

Manuel-K: What's the alternative? I don't know anything about rust…

selaux: Matching a regex to the path and only replace if found basically something like (completely untested):

```rust
let re = Regex::new(r"^\\\\[!\\]+\\").unwrap();
let after = re.replace_all(temp, "");

return PathBuf::from(after)
```

selaux: There is also an unwrap there, but it is way more likely that it wont cause errors (because the regex string wont change)

selaux: Basically the unwrap denotes: I as a developer am absolutely sure that there wont be an erronous/non-existing value in there so you can just assume the value is ok.

If an error then occurs the whole application panics and exits. If you want to handle the non-existing case (in this case if the `\\` is not found, you can use either `match` to match for both cases or `if let` to handle a single case.

Manuel-K: But when the path starts with `\\\\` we know that we got an UNC path and therefore there has to be a `\\` somewhere. Although they might change the `canonicalize` function…

I agree. A regex is probably safer…

selaux: Yeah, that is why I don't see it as a blocker for now 🙂

rust/src/stracciatella.rs
Manuel-K: Oh what fun. If you use `canonalize` you have to account for the `std::result::Result<std::path::PathBuf, std::io::Error>` as well before you can use `to_str`…

Manuel-K:
Is this OK?

Manuel-K:
It does compile the Linux test on Mac. Why!?

Issue #550, opened 2017-06-08
Area select rectangle not removed if interrupted


Tercus:
If you create an area of selection (left mouse-button + drag) and you get interrupted by an action of one of your mercenaries, that rectangle stays.
![screenshot_20170608_232530](https://user-images.githubusercontent.com/3085791/26952162-82bfcbd6-4ca4-11e7-8c52-07e644b0a34e.png)

To reproduce with [this savegame](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1062504/SaveGame10.zip):
1. Select Player Terces to open the last door
1. create a selection rectangle and hold it
1. be interrupted by the "how to open the door" window.

JAsmine-ja2:
I can confirm this bug on Vanilla and Stracciatella.
However, the rectangle can easily be "clicked" away and doesn't stay permanently - still a bit annoying.

PR #563, opened 2017-07-10
fixing area select not removed if interrupted

daviel:
fixing #550. 
I am not that happy with that solution though. I Tried to ignore the next Left-Mousebutton-Up so if you command a merc to open a door/talk to a person/... and hold the button down it would not close the menu after. I wasn't able to do that because the event system in the code is kind of unpredictable.
Maybe the way it works with my commit is enough and not a problem for the user-experience view.

### DIFF
--- a/src/game/Tactical/Handle_UI.cc
+++ b/src/game/Tactical/Handle_UI.cc
@@ -3244,7 +3244,8 @@ static INT8 DrawUIMovementPath(SOLDIERTYPE* const pSoldier, UINT16 usMapPos, Mov
 
 			 if ( sGotLocation != NOWHERE )
 			 {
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Thanks for your work! Please open separate pull requests for unrelated issues, so people don't get confused when reviewing and testing.

It looks good to me, don't have a design issue with it.

Manuel-K:
You do not have to close this PR!

Switch back to master. Create a new branch and cherry-pick the last two commits. Push that branch and create a new PR for that.

Then switch back to this branch and remove the last two commits via `git reset --hard HEAD~2`.

daviel:
Thanks for the advice. :) I'm still figuring out how to use git properly. I reopened two pull requests now. Sorry for that.

Manuel-K:
Git can be confusing.

Issue #549, opened 2017-06-08
HtH Combat AP calculation wrong in certain situations


Tercus:
When I'm in a Hand to Hand combat, it can happen that the first shown AP is not correct. Once I focus or when I select a target, the AP is updated.

In the Savegame, Igor can attack the enemy in HtH, but it shows that it only take him 4 AP. In reality, the AP needed is 9.

[SaveGame01.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1062233/SaveGame01.zip)

lynxlynxlynx:
I think this was also already reported, even before the reboot. Please check the (AP) bug list to verify.

Tercus:
All I could find was a bugreport for 0 AP during HtH. Also, the correct amount of AP is used, just not shown in the beginning.

PR #564, opened 2017-07-10
Fixing AP-calculations for hand to hand combat[WIP]

daviel:
fixing #549 
Weird thing about the AP-Calculation for hand to hand combat is that it expects the MERC to turn first and then switch stance. Which means you can save AP by manually standing up rather than doing it with one attack-click.
		// Would expect that merc is standing up(2AP) and then looking(1AP)
		// rather than looking(2AP) and then standing up(2AP)

### DIFF
--- a/src/game/Tactical/Points.cc
+++ b/src/game/Tactical/Points.cc
@@ -1105,7 +1105,7 @@ static UINT8 MinAPsToPunch(SOLDIERTYPE const& s, GridNo gridno, bool const add_t
 {
 	UINT8	ap = 4;
 
-	if (gridno == NOWHERE) return ap;
+	if (gridno == NOWHERE || s.sGridNo == gridno) return 0;
 
 	if (SOLDIERTYPE const* const tgt = WhoIsThere2(gridno, s.bTargetLevel))
 	{ // On a guy, get his gridno
@@ -1116,18 +1116,40 @@ static UINT8 MinAPsToPunch(SOLDIERTYPE const& s, GridNo gridno, bool const add_t
 		{
 			ap += GetAPsToChangeStance(&s, ANIM_CROUCH);
 		}
-		else if (s.sGridNo == gridno)
+		else
 		{
 			ap += GetAPsToChangeStance(&s, ANIM_STAND);
 		}
 	}
 
-	if (add_turning_cost && s.sGridNo == gridno)
+	if (add_turning_cost)
 	{ // Is it the same as he's facing?
 		UINT8 const direction = GetDirectionFromGridNo(gridno, &s);
-		if (direction != s.bDirection) ap += AP_LOOK_STANDING; // ATE: Use standing turn cost
-	}
 
+		// Would expect that merc is standing up(2AP) and then looking(1AP)
+		// rather than looking(2AP) and then standing up(2AP)
+		// same for proning
+		if (direction != s.bDirection)
+		{
+			// We only add turning costs for enemys in attack range. If we have to
+			// move first turning cost are considered within walking
+			if (s.sWalkToAttackWalkToCost == 0)
+			{
+				if (gAnimControl[ s.usAnimState ].ubEndHeight == ANIM_STAND)
+				{
+					ap += AP_LOOK_STANDING; // ATE: Use standing turn cost
+				}
+				else if (gAnimControl[ s.usAnimState ].ubEndHeight == ANIM_CROUCH)
+				{
+					ap += AP_LOOK_CROUCHED;
+				}
+				else if (gAnimControl[ s.usAnimState ].ubEndHeight == ANIM_PRONE)
+				{
+					ap += AP_LOOK_PRONE;
+				} 
+			}
+		}
+	}
 	return ap;
 }
 
--- a/src/game/Tactical/Handle_UI.cc
+++ b/src/game/Tactical/Handle_UI.cc
@@ -3244,9 +3244,11 @@ static INT8 DrawUIMovementPath(SOLDIERTYPE* const pSoldier, UINT16 usMapPos, Mov
 
 			 if ( sGotLocation != NOWHERE )
 			 {
-				 sAPCost += MinAPsToAttack( pSoldier, sAdjustedGridNo, TRUE );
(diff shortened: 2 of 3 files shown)


Manuel-K:
The easiest way to clean this up is
```
git checkout hth
git reset --soft HEAD~4
git commit
git push --force
```

This combines all four commits into one and lets you write a new commit message for all changes.

daviel:
Thanks again. :)

lynxlynxlynx:
looks good, but I haven't tested. Especially the first diff could affect other things.

lynxlynxlynx:
I retract that. You only change MinAPsToAttack and then not use it at all. Something is awry.

Two more things to consider:
- starting your attack from a prone position
- gridno checks: are two soldiers on the same gridno if one is above the other (roof)? More potential problems there

daviel:
MinAPsToAttack is used within CalcTotalAPsToAttack but it doesn't consider the whole ap calculation for attacks(leaving out path, direction).

I also added changes for proning and the grid_no check for soldiers standing on different levels in my local branch.

But I found another problem: Attacking an enemy standing a few grids away. The calculation is right for path etc. but rotation(looking) isn't considered making it cost 0ap to rotate. Now I'm unsure whether it's part of the gameplay or not. There is a initialmove parameter in the method for changing direction(EVENT_InternalSetSoldierDesiredDirection) which leads to 0 costs for rotation if its set to true. It is set to true for hth-attacks which need moving to reach enemys.

Should looking-costs be considered if the soldier has to move to reach the enemy?

JAsmine-ja2:
Correct me if I'm wrong, but I think turning around always costs AP when it's part of a ranged attack. So I think it should also cost AP when attacking unarmed.

daviel:
That's the point which confuses me. What was the point adding an initialmove-parameter which only affects turning ap? As far as I can remember there was an issue that throwing knives didn't add turning costs. Maybe it is actually part of the gameplay that some kind of attacks don't add turning costs.

Here examples of hth-attacks:
1. Enemy is in reach. Merc looks in wrong direction. -> Turning costs are added
2. Enemy is out of reach. Merc looks in wrong direction -> Moving costs are added but turning cost is not

Can't test vanilla so I don't know how it was solved there. But would like to hear some opinions on that. :)

daviel:
Okay, I found the reason for that. If you're moving your mercs in turn-based-mode there are no costs for turning if you just want to walk/crawl. You can try it by looking a the ap costs just for moving one grid around your merc, it's always the same value whether going a step back, to the left etc..

So if you want to attack an enemy with melee and you have to walk to him first there are no turning costs. See my previous comment.

I commited my actual code...

src/game/Tactical/Points.cc
```diff
@@ -1116,18 +1116,37 @@ static UINT8 MinAPsToPunch(SOLDIERTYPE const& s, GridNo gridno, bool const add_t
...
 		}
-		else if (s.sGridNo == gridno)
+		else
 		{
 			ap += GetAPsToChangeStance(&s, ANIM_STAND);
 		}
 	}
 
-	if (add_turning_cost && s.sGridNo == gridno)
+	if (add_turning_cost && s.sGridNo != gridno)
```
lynxlynxlynx: s.sGridNo != gridno
is always true by this point.

src/game/Tactical/Points.h
```diff
@@ -41,6 +41,10 @@
 #define AP_MOVEMENT_LAKE        9       // deep wade -> slowest
 #define AP_MOVEMENT_OCEAN       8       // swimming is faster than deep wade
 
+#define AP_MOVE_STANDING        2
+#define AP_MOVE_CROUCHING       3
+#define AP_MOVE_PRONING         4
```
lynxlynxlynx: Wasn't it 6 for prone movement? Also, if you run, standing movement cost is 1.

daviel: In turn-based modus it's as defined. The ap costs are calculated per tile.

src/game/Tactical/Points.cc
daviel: Will remove the second condition as it's not needed.

src/game/Tactical/Points.h
lynxlynxlynx: You're right about prone, but please double check running — the first tile costs the same (2), the next ones only 1.

daviel:
I would like to split the issue in two parts as it seems that there is a problem with knifes too. I already have a working fix for hand-to-hand combat but can't come up with a solution for the knife-calculation which seems to be more complicated.
Weird thing is that attacking with knives is kind of a hybrid solution as it has parts of hth in it and parts of shooting also.

src/game/Tactical/Points.h
daviel: In turn based there is a fixed ap price for walking over tiles depending on stance. But I don't need these defines anymore so I will stripe them out.

src/game/Tactical/Points.cc
```diff
@@ -1116,18 +1116,40 @@ static UINT8 MinAPsToPunch(SOLDIERTYPE const& s, GridNo gridno, bool const add_t
 		{
 			ap += GetAPsToChangeStance(&s, ANIM_CROUCH);
 		}
-		else if (s.sGridNo == gridno)
+		else
 		{
 			ap += GetAPsToChangeStance(&s, ANIM_STAND);
 		}
```
lynxlynxlynx: are we missing a case here, when there's no need for a stance change?

src/game/Tactical/Points.h
lynxlynxlynx: Cool, but what I wrote is also true — tested in the original.

src/game/Tactical/Points.cc
daviel: Only lying/proning enemys need to be attacked while crouching for every other case the merc would need to stand. If there is no need for a stance change 0 will be returned by the GetAPsToChangeStance function. 

I don't know what the else if statement 
`else if (s.sGridNo == gridno)`
was for because the only case in which your merc can have the same gridno like your enemy is when they're on different height levels(roof/ground). But even this wouldn't be possible as you would receive an error message.

lynxlynxlynx: Which reminds me, why is there a special variable for gridno in the outer block? In other words, is tgt->sGridNo ever different from the gridno passed to the function? The way the target is derived, I'd hope it wouldn't be possible.

PR #556, opened 2017-06-14
Whitespace normalization

Manuel-K:
See issue #227 

I've started to replace space with tabs. I've used a custom python script, looked over the changes and fixed some leftovers by hand.

PR #565, opened 2017-07-10
Selecting Area not resetting after being interrupted by menus

daviel:
fixing #550. 
I am not that happy with that solution though. I Tried to ignore the next Left-Mousebutton-Up so if you command a merc to open a door/talk to a person/... and hold the button down it would not close the menu after. I wasn't able to do that because the event system in the code is kind of unpredictable.
Maybe the way it works with my commit is enough and not a problem for the user-experience view.

### DIFF
--- a/src/game/Tactical/Interface_Control.cc
+++ b/src/game/Tactical/Interface_Control.cc
@@ -561,6 +561,10 @@ void EraseInterfaceMenus( BOOLEAN fIgnoreUIUnLock )
 	PopDownMovementMenu( );
 	PopDownOpenDoorMenu( );
 	DeleteTalkingMenu( );
+
+	// Stop Rubberbanding every time a menu is erased/opened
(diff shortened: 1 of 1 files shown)


Manuel-K:
Okay. Let's clean up this PR first. Do you use the command line git client?

Manuel-K:
This PR does replace #563.

Manuel-K:
To get rid of the wrong commits and the revert commits you can use

```
git checkout selectarea
git reset --hard HEAD~4
git push --force
```
if you use the command line client.

lynxlynxlynx:
Thanks!

Manuel-K:

Manuel-K:
@lynxlynxlynx If you merge any PRs that affect files in the `src/` directory, can you please add a note  mentioning 556?

lynxlynxlynx:
Ok, noted.

PR #570, opened 2017-07-11
Add cargo lockfile

### LINKED ISSUE #562: CI fails on rust test (master)
```
thread 'tests::parse_args_should_abort_on_unknown_switch' panicked at 'assertion failed: `(left == right)` (left: `"Unrecognized option: \'testunknown\'"`, right: `"Unrecognized option: \'testunknown\'."`)', src/stracciatella.rs:575
```

Affects Travis and AppVeyor.

Very strange. The left argument seem to be generated by `getopts::Options` method `parse`. On my system the test passes.

selaux:
This avoids unwanted rust dependency updates. As no other projects than ours depend on the library we can commit the lockfile safely.

The lockfile needs to be updated from time to time though.

Should fix #562 properly.

### DIFF
--- a/rust/.gitignore
+++ b/rust/.gitignore
@@ -1,2 +1,1 @@
 target
-Cargo.lock
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -1,0 +1,186 @@
+[root]
+name = "stracciatella"
+version = "0.1.0"
+dependencies = [
+ "getopts 0.2.14 (git+https://github.com/rust-lang-nursery/getopts)",
+ "libc 0.2.26 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.1.80 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 0.8.23 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 0.8.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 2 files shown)


lynxlynxlynx:
Is the version of getopts now ok — the dot removal was already merged?

selaux:
Should be, tests run fine for me locally. Lets see what CI says.

Manuel-K:
Won't this create a huge workload? It seems we would have to update the file every time one of the crates is updated…

selaux:
True. You could also live with the old versions until you encounter a bug or need a new feature.

selaux:
We could also refer to a specific ref for the getopts dependency, so we won't have this issue again. But then other issues like this could pop up from time to time.

Manuel-K:
I think we should not lock in specific version. I'd prefer to fix problems when they appear.

In this specific case we could make it work with older versions of rust by ignoring the left argument's ending dot.

selaux:
See #571

selaux:
I don't think it depends on the rust version but the version of the `getopts` dependency that is used.

lynxlynxlynx:
don't care either way. :)

Manuel-K:
> I don't think it depends on the rust version but the version of the getopts dependency that is used.

Correct. I've linked to the problematic commit in #562:

> Found the problem: rust-lang-nursery/getopts@1374750

selaux:
So lets go for #571, I will close this then 🙂

PR #571, opened 2017-07-11
Fixed getopts dependency rev

selaux:


Manuel-K:
Is it possible to set a minimal rev? Note: I've made the assumption that this fixes a specific rev.

selaux:
Here you go 🙂

Manuel-K:
Note: this is related to #562

Manuel-K:
It seems it is not possible. 

From http://doc.crates.io/specifying-dependencies.html:

> Since we haven’t specified any other information, Cargo assumes that we intend to use the latest commit on the master branch to build our project. You can combine the git key with the rev, tag, or branch keys to specify something else. Here's an example of specifying that you want to use the latest commit on a branch named next:

lynxlynxlynx:
we can bump it manually, as it's not likely that it will be needed often.

selaux:
@Manuel-K What seems to not work?

lynxlynxlynx:
I think he meant an open interval, a minimum version vs a fixed revision/tag.

selaux:
Ah ok, I overread the minimal part

PR #572, opened 2017-07-11
Be less restrictive about libc dependency

selaux:


Manuel-K:
Can we use newer versions of serde?

selaux:
Maybe. Feel free to open a PR.

PR #573, opened 2017-07-12
fixed wrong AP calculations while throwing knives[WIP]

### LINKED ISSUE #186: Error in calculation of AP needed to throw a knife

---

There's a funny bug when throwing knives. When targeting enemy, the engine says X AP is needed. If the user actually makes the throw, (X-1) is used.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/186

daviel:
fixes #186 
If the direction of the merc was not right it didn't calculated AP-costs for turning around leading to lower AP-Costs. This happened because in the event-queue for the animations it attacked first then turned around.

### DIFF
--- a/src/game/Tactical/Points.cc
+++ b/src/game/Tactical/Points.cc
@@ -1038,7 +1038,10 @@ UINT8 MinAPsToShootOrStab(SOLDIERTYPE& s, GridNo gridno, bool const add_turning_
 	{
 		if (GCM->getItem(item)->getItemClass() == IC_THROWING_KNIFE)
 		{
-			ap_cost += AP_LOOK_STANDING;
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
LGTM, thanks!

daviel:
Second throw on same target is still wrong. Showing 1AP less than real cost.

Manuel-K:
@lynxlynxlynx Do we really want to merge this now? If it really is only a partial fix, I'd leave it as it is for the moment.

lynxlynxlynx:
I'd need to check those three functions to evaluate it better. An incremental fix is still better than none, if it has no bad side-effects.

PR #574, opened 2017-07-13
Fixes Stealing AP Bug

### LINKED ISSUE #191: Mysteriously disappearing APs when stealing

---

In the attached save Johny can steal enemies weapon. It costs 10 APs to do so. However, after he does so he looses another 5 AP.
He can attack first and steal later. Then the mysterious loss drops him to 0, taking 'just' 3 APs.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/191

daviel:
Fixes #191 
For some reason someone put the condition in comments. Can't tell why. Normal Hand to hand still works as aspected with the condition.

### DIFF
--- a/src/game/Tactical/Weapons.cc
+++ b/src/game/Tactical/Weapons.cc
@@ -1182,8 +1182,7 @@ void UseHandToHand(SOLDIERTYPE* const pSoldier, INT16 const sTargetGridNo, BOOLE
 
 	// Deduct points!
 	// August 13 2002: unless stealing - APs already deducted elsewhere
-
-//	if (!fStealing)
+	if (!fStealing)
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
you can use `git blame filename` to see who last changed each line, with which commit (hash). Then `git show commit-hash`. Its message may have a reason (like a different bug) or you just may end up at the end of history — when the code was first released.

daviel:
It's telling me that this line was last changed on initial import(Sep 6 2004). So the reason stays unknown.

lynxlynxlynx:
Great, thanks!

Manuel-K:

PR #575, opened 2017-07-13
fixed wrong ap for attacking lying/proning enemys[Depends on #564]

### LINKED ISSUE #190: Mysteriously disappearing APs

---

In the attached save, Johny can attack an enemy with a crowbar for 4 AP. If he does so, 6 AP are taken actually. Looking closely at his AP during the animation, one can see that AP drops 9->7->3, so there's something that takes the 2 AP before the attack

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/190

daviel:
fixes #190, depends on #564
The calculation wasn't correct for melee attacks on lying/proning enemys. Lying enemys stand on 2 fields/grids and are attacked on the torso not legs.

### DIFF
--- a/src/game/Tactical/Points.cc
+++ b/src/game/Tactical/Points.cc
@@ -1116,17 +1116,51 @@ static UINT8 MinAPsToPunch(SOLDIERTYPE const& s, GridNo gridno, bool const add_t
 		{
 			ap += GetAPsToChangeStance(&s, ANIM_CROUCH);
 		}
-		else if (s.sGridNo == gridno)
+		else
 		{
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
This inadvertently has code from other commits in it. You can wait for those to be merged first or do some git magic to get only this fix into a commit (reset, git commit -p).

daviel:
This fix depends on the changes made by the other commit and won't work correctly otherwise. Therefore I will wait until the other commit is merged/declined. Should I mark this PR as pending or something like this?

Manuel-K:
Just write something like "Note: depends on XXX".

PR #577, opened 2017-07-18
Soldiers who change their team will now be correctily put in the z-level

### LINKED ISSUE #567: Attacking teammate(dialogue) on roof puts him through the roof
If you attack a merc on a roof and he/she is saying something the merc will be put on the ground level through the roof.
This does not happen with merc which just shoot back.

Save:
[QuickSave.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1138638/QuickSave.zip)

daviel:
fixes #567


src/game/Tactical/Soldier_Profile.cc
```diff
@@ -608,6 +608,8 @@ SOLDIERTYPE* ChangeSoldierTeam(SOLDIERTYPE* const old_s, UINT8 const team)
 	new_s->bLastRenderVisibleValue = old_s->bLastRenderVisibleValue;
 	new_s->bVisible                = old_s->bVisible;
 	new_s->bCamo                   = old_s->bCamo;
+	new_s->bLevel                  = old_s->bLevel;
+	if(old_s->bLevel == 1) SetSoldierHeight(new_s, 58.0);
```
lynxlynxlynx: where's the 58 from? Do we have a named constant for it?

daviel: There is no constant for this value by now. I took it from the Soldier_create.cc as it is used to put the right z-value. It is also used mulitple times in the whole code with a hardcoded 58.0 value.

lynxlynxlynx: Ok, can you then please make it all better an create a named constant and use it everywhere? No need to rewrite this commit, just push extra ones to the same branch.

daviel:
I added values for the offset as defines in the WorldDef.h.

lynxlynxlynx:
Awesome, thanks!

PR #578, opened 2017-07-20
Fix mouse moving out of the window when in fullscreen mode

palmer-eldritch:
When the engine is configured to be in fullscreen and the size of the window is smaller than the desktop resolution, the mouse moves out of the window when it goes over the edge. And, if your desktop is, like mine, configured to have focus follow mouse, the game window immediately loses focus, which makes the game pretty much unplayable (or at least, very unpleasant to play.

Now, to avoid this, it's possible to set the size of the window at the same resolution as the desktop but, with the size of today's screens, it makes the objects/characters pretty small.

So, I fixed this by adding the flag SDL_WINDOW_INPUT_GRABBED to the window when it's in fullscreen so the cursor doesn't wander out of the window. And, of course, when switching back and forth to windowed/fullscreen mode, I needed to add a call to SDL_SetWindowGrab to set/unset that flag: if the flag is set when in windowed mode, the cursor can't leave the window, which I think isn't a desirable behavior.

I have only tested that fix on linux (debian sid), by quickly looking around the net it seemed to me that SDL tends to behave differently on different system so I don't know if that patch might bring other unwanted side effects on other systems (and I don't know if the problem was present on other systems either). But since I made that patch for myself to make the engine usable, here it is for you to decide whether or not it should be integrated in the main code base.

Thanks,
Youen

### DIFF
--- a/src/sgp/Video.cc
+++ b/src/sgp/Video.cc
@@ -99,11 +99,11 @@ void VideoSetFullScreen(const BOOLEAN enable)
 {
 	if (enable)
 	{
-		g_window_flags |= SDL_WINDOW_FULLSCREEN;
+		g_window_flags |= SDL_WINDOW_FULLSCREEN | SDL_WINDOW_INPUT_GRABBED;
 	}
 	else
 	{
-		g_window_flags &= ~SDL_WINDOW_FULLSCREEN;
+		g_window_flags &= ~(SDL_WINDOW_FULLSCREEN | SDL_WINDOW_INPUT_GRABBED);
 	}
 }
 
@@ -113,10 +113,12 @@ void VideoToggleFullScreen(void)
     if (SDL_GetWindowFlags(g_game_window) & SDL_WINDOW_FULLSCREEN)
     {
         SDL_SetWindowFullscreen(g_game_window, 0);
+        SDL_SetWindowGrab(g_game_window, SDL_FALSE);
     }
     else
     {
         SDL_SetWindowFullscreen(g_game_window, SDL_WINDOW_FULLSCREEN);
+        SDL_SetWindowGrab(g_game_window, SDL_TRUE);
     }
 }
 


lynxlynxlynx:
Wait, why do we allow a hybrid of fullscreen and windowed mode in the first place? Sounds broken. Or is this about multimonitor or similar setups?

Manuel-K:
>Wait, why do we allow a hybrid of fullscreen and windowed mode in the first place? Sounds broken. Or is this about multimonitor or similar setups?

I'm not aware that we do.

We should have true fullscreen mode (with annoying resolution switching) and resizable window mode. At least that was what we had when I enabled window resizing. We talked about using fake fullscreen, but decided against it.

palmer-eldritch:
A hybrid of fullscreen and windowed mode? What do you mean?

The fullscreen mode was broken on my system (debian sid, with proprietary nvidia drivers) when I built the latest git version. If the fullscreen resolution is smaller than the resolution of the desktop, the mouse can move out of the window. When you hit the edge of the screen, the game's screen moves around showing part of the desktop (because even if the screen resolution has changed, the desktop's resolution is still the same, making it larger than the screen), Basically, the screen's resolution is changed to match the resolution configured in ja2.json. But the desktop's resolution stays the same, making it bigger than the screen. Since the mouse isn't locked to the game's window, it can move out of the game's window. When that happens, part of the game's window goes out of the screen while part of the desktop becomes visible. Which, in itself is already a pain and can make the game completely unplayable when your desktop is configure so that the focus follows the mouse: whenever you hit the edge of the screen, the game's window loses focus.

So, to be able to play the game, I went out to see if I could fix this (in the first place, for myself, since I haven't seen any report here of someone having the same problem). Basically, I saw two quick ways to do this: use the SDL_WINDOW_FULLSCREEN_DESKTOP flag instead of the SDL_WINDOW_FULLSCREEN flag. Using this flag, your fullscreen window isn't truly fullscreen: it's resized so that its size matches you screen resolution, removes all window decorations and make sure it stays on top. However, since the engine has a feature to switch between fullscreen and windowed mode by pressing alt-tab, using this flag posed a problem: when switching back to windowed mode, the window keeps the same size: that of your screen.
The other alternative, which is the one I chose was to add the SDL_WINDOW_INPUT_GRABBED flag to the window when in fullscreen mode so the mouse doesn't escape the game's window. If the mouse stays in the game's window, the problem is fixed since you can't scroll around the larger than the screen desktop.

Here's a picture where I try to explain the problem, sometimes, pictures are better than words, although, with my artistic talents, it might make things even more complicated to understand:
explain fullscreen mouse bug

Now, that bug I encountered on my system (debian sid, with the latest proprietary nvidia drivers, on a single screen with a desktop resolution of 1920x1080 and a game resolution set to 1280x720 using the openbox window manager) might not be present on every system. Maybe it's the OS, maybe it's the GPU and it's drivers, maybe it's a problem caused by the window manager.
But, that's a problem I encountered, so I guess other people might encounter it too if nothing is done to fix it. Since I made a fix for myself, I thought I could at least share it with you. Now, this fix might introduce some other problem down the road, on another system, with a different configuration. But it worked for me, so, it's better than nothing.

Youen

lynxlynxlynx:
Still sounds fishy, but it doesn't hurt. BTW, you can toggle mouse grab via the scroll lock key.

palmer-eldritch:
That's right, if I tried using the scroll lock key on a build without my patch and I have no more problems with the mouse going out of the screen and losing the game's window focus.

It's right there in changes.md : (r528) Use scroll lock to toggle input grabbing. This is useful for scrolling with the mouse in tactical mode when playing in windowed mode.

I guess that's a quick fix for the problem I'm encountering with the engine, although, it doesn't seem that's the reason why it was implemented. I guess to be helpful, I'd have to dig and find the reason why I have that behavior on my system when it seems like other people using the engine (although, how much is that when we're talking about the git version?) don't have this problem. And I assume I'm not the only on running it on linux so I doubt it's related to the OS. Same thing for the GPU, I'm using a Nvidia GPU (GTX 1050 Ti) with the proprietary drivers, but using a Nvidia GPU with the proprietary driver is pretty common among people gaming on linux, so it seems unlikely to be it. Although, it could be a driver bug affecting only a few specifics GPUs.
It could be related to my windows manager I guess, I'm using openbox which, if it's not some obscure piece of software only used by a few isn't among the most used (although, with lxde/lxqt it certainly gained some popularity but I'm using barebone openbox with a few other tools to have a functional desktop, I don't know what these destktop environments might add on top of the windows manager which might alter its behavior).

Most likely, it's a combination of several factors that could also include the version of the libraries installed on my system and whatever I haven't thought of.

One thing I know for sure is that I don't have this problem with the latest stable release (v0.15.1 I think - but, if I'm not wrong, that one still uses sdl 1.2, that migration could have introduced that little problem), although, with that one, my cursor keeps moving around without being told to which is also a pain. Also, when I diff that version against the current head, it seems that there was a lot of changes since then, including some major refactoring so it wouldn't be easy to find which commit(s) introduced that behavior.

I can definitely understand that you can be reticent to merge that fix without understanding exactly what it fixes. I'll see if I can find some time to dig and find the reason why I have this problem, however, it might prove difficult.

Anyway, it doesn't matter much to me whether or not this gets merged in the engine, I just encountered a bug when trying to use it and made a quick fix so it would behave the way I expected it to. In the open source spirit, I share that fix with you. Now, if I'm the only one to have had that problem yet, it's definitely not a critical issue, and that issue might be due to an improbable combination of factors that can't be easily reproduced. But, if when the next version is released some other people were to encounter the same problem, you could dust off that small patch and maybe improve it a little if it needs to rather than starting from scratch.

BTW, thanks for the scroll lock tip, I doubt I would have found it by myself, I don't think it's documented anywhere but in the changelog and without any reference to a similar problem. For now, I think I'd rather use my patched build, I prefer my games to work as intended out of the box rather than having to rely on some obscure undocumented feature. Maybe later, if the problem keeps occurring when I try building newer versions of the engine.

lynxlynxlynx:
Scrolllock is undocumented just like any other hotkey. This is tracked here https://github.com/ja2-stracciatella/ja2-stracciatella.github.io/issues/4

My cpu fan died, so I don't have access to my main box to test anything. And that's what's required, we can stop theorizing.

palmer-eldritch:
Well, after some more testing, the main root of the problem seems to be that I have openbox configured with focus follow mouse. When I disable that option, I can move the mouse out of the game's window without any problem. As long as I don't click out of the window and give focus to another one.

I also tested it on KDE an had the same results: as long as the game window doesn't lose focus, which can happen if you click outside of the game's window, there's no problem.

Now, I'd think the proper behavior for a game lauched in fullscreen would be to prevent the window from losing focus unless the user does something very specific, like alt-tabbing out of the game's window or changing the mode from fullscreen to windowed. Which can be done properly by setting mouse grab on when in fullscreen. Because even if it's harder to accidentally lose focus when your windows manager isn't configured with that focus follows mouse on, it's still possible for it to happen.

alanswanson:
Using focus follows mouse is not a bug! ;-) This pull request is still valid (albeit commit needing tweaked for code movement) and the correct solution when fullscreen (especially when at a lower resolution than desktop). Using this in my git build of ja2 on Linux.

selaux:
Is this still valid? I've just noticed that we actually have functionality to grab the cursor by pressing the scroll lock button, so I guess it can be achieved this way as well...

lynxlynxlynx:
We haven't heard back in two years, so yeah, let's close. It can always be reopened if interest returns.

Issue #289, opened 2016-04-09
Ideas on things to externalize


JAsmine-ja2:
Having certain aspects of the game externalized heavily increases moddability. Without externalized data, the possibilities are very limited (unless the modder touches the sourcecode).

In this issue, we can discuss what/why/how/... to externalize and create some sort of overview.

The first question should probably be if we think that externalizing data is a good idea (now or in the near future) or if we don't want to spend time on this.

selaux:
Imho externalizing should be an essential part of this project. A few things I would externalize to improve moddability:

- Merc profiles (for all mercs and npcs)
- Dialogues

Notably externalizing is not the only thing that needs to be done to ensure moddability. Eventually some of the game logic needs to be moved into scripts, especially the story and quest related stuff. But this is talk for the future...

JAsmine-ja2:
Other things that could be externalized are:

Arulcan Army:
- garrison groups (permanent presence in certain sectors)
- patrol groups (mobile squads moving around)
- army composition (details on the groups like number and probability of being replaced)

Mapping:
- sector names
- sector loadscreens
- cities (location, size, name ...)
- movement costs (moving from one sector to another)
- accessibility of setors ("borderland", ocean...)
- SAM sites (sector, location of the computer)
- air control table (the area that is defended by each SAM site)
- BR shipping destination
- Bloodcats (sectors, number, chance to encounter)
- Coolness (per sector)
- underground sectors (increase max. no.)

Items:
- all weapons (including "special" stuff like flamethrower) -> possibility to increase max. no. of weapons
- armor
- other items
- attachments (increase max. no.)

Misc. stuff:
- delivery methods of BR (cost, time)

Of course, the list could go on and on.

lynxlynxlynx:
- chopper refuelling stations / bases (it's annoying that he can't use the meduna airport)

derSteFfi:
Can we maybe make this a meta-issue and add achecklist with shared ideas to have a kind of todo-list?

JAsmine-ja2:
> Can we maybe make this a meta-issue and add achecklist with shared ideas to have a kind of todo-list?

Don't know how to do this.

Anyway, scrolling through https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/Tactical/Campaign.cc#L1042 gave me some more ideas on things to externalize:

- Possibility to increase stats with lvl 0 (zero) --- this may make the game too easy, but the original setting can be quite restricting
[...]

Issue #156, opened 2014-11-19
Externalize merc inventory


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

Since Proedit can not really handle the items when the newly externalizations are used, it is vital to externalize the merc inventory, if we want to make items fully moddable via json.

**_Example:**_ 
If you create a magazine in an item index that was previously empty, you won't be able to edit the quantity of this new magazine via proedit, since proedit still assumes, that 

"ubPerPocket": 0 .

So, in order to change the quantity of this new magazine, an externalized merc inventory would be the best solution.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/156

PR #580, opened 2017-07-20
making an externalized merc inventory file

Lexers615:
Making an externalized merc inventory file would be great... It would, for example, allow players to make Shadow start with a Silencer, which would totally makes sense...


JAsmine-ja2:
Externalizing stuff is always a very good idea! General externalization issue: #289 

The idea to externalize the merc inventory is especially appreciated, since it already came up once: #156

Are you aware that adding a silencer to Shadow's starting inventory (and much more!) is possible in Stracciatella and even in Vanilla JA2 with prof.dat and Proedit?

Manuel-K:
Why is this a PR? Seems to be a mistake.

PR #581, opened 2017-07-26
preinit dontchargeap to prevent false AP costs

### LINKED ISSUE #188: When enemy is noticed, no AP for action may be taken

---

A weird situation happens in attached save. I know of 2 enemies next to each other; one is seen by my people and one by militia. Hitman sees one. If I order him to shoot either of the two, he turns and sees both. The shot is canceled, but Hitman has turned and locked his weapon on an enemy for a total of 0 AP.
Nearly the same happens if Hitman is ordered to look towards both enemies - no gun locking at the target (obviously), but the turn takes 0 AP.

Another thing is that I'm not sure if locking the gun should happen. Normally, he would first see enemy and then aim. Interruption should happen in between, shouldn't it?

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/188

daviel:
fixes #188 
preinit a value to prevent false ap costs; actions which dont charge turning ap are coming after this line and put the value to true. The assumption should be that every tunring costs ap not the other way round.

### DIFF
--- a/src/game/Tactical/Handle_UI.cc
+++ b/src/game/Tactical/Handle_UI.cc
@@ -2181,6 +2181,9 @@ static ScreenID UIHandleCAMercShoot(UI_EVENT* pUIEvent)
 {
 	SOLDIERTYPE* const sel = GetSelectedMan();
 	if (sel == NULL) return GAME_SCREEN;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

PR #582, opened 2017-07-28
fixes helicopter collision bug by adding its grids right after the se…

### LINKED ISSUE #317: Helicopter is ignored by pathfinding
When starting a new game, the helicopter is present in Drassen but ignored when finding paths (moving a merc). Moving though it is not possible though. Screenshot will follow...

daviel:
fixes #317 
I recognized that the collision was correct if you saved a game in B13 and loaded it but if you leave the sector, enter an other one and then return to drassen the helicopter collision was false again.


lynxlynxlynx:
Still can't test, but looks good and harmless. Thanks!

Issue #126, opened 2014-02-28
Bloodcats don't attack can't be killed with ALT+O cheat without being detected first.


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

In this savegame, the squad got ambushed by Bloodcats.
There are two problems here.

No.1
The Bloodcats are in the top-right corner of the map. The squad in the middle. The bloodcats are passive and don't come running for my mercs like they should.

No.2
I can reveal the Bloodcats with ALT+E. However, I can't kill them with ALT+O unless one merc has visual contact before. This behavior is imo not right. ALT+O should kill all living enemies no matter if detected or not. (It works normally with enemy soldiers)

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/126

lynxlynxlynx:
I can confirm that they're pretty passive. I don't remember if they always ganged up on you immediately though. Clearing their den is now possible in several waves, much like with normal enemies.

JAsmine-ja2:
I just ran into the passive Bloodcats again in Stracciatella 0.16.0.
[SaveGame11.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1629224/SaveGame11.zip)

The squad got ambushed by Bloodcats... nothing happens. ALT+O cheat reveals a single Bloodcat, not too far from the squad. It remains passive.
There are several issues with Bloodcats - Maybe, they are connected on some level?

daviel:
@JAsmine-ja2 I did not entcounter passive bloodcats with your savegame. Do you have another one or could you try to recreate such a situation?

flaviojs:
I tried the save.
Sector B4 already has 1 bloodcat there, maybe a leftover from a previous ambush?
In this case there is a 5% chance that the existing bloodcats will ambush you again:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/5d7057eccd305dee3cae80da2057672c06261cd3/src/game/Strategic/Strategic_Movement.cc#L3394-L3397

This looks like intended behavior so mark the issue as `wontfix`?

lynxlynxlynx:
I don't know, it sounds more like the problem is they are (is) there and don't attack, rather than being there in general.

flaviojs:
[...]

PR #583, opened 2017-07-30
Cheat: alt+o damages bloodcats now

daviel:
fixes the second part of #126

### DIFF
--- a/src/game/Tactical/Turn_Based_Input.cc
+++ b/src/game/Tactical/Turn_Based_Input.cc
@@ -2759,8 +2759,11 @@ static void ObliterateSector()
 	SLOGD(DEBUG_TAG_INTERFACE, "Obliterating Sector!");
 	FOR_EACH_NON_PLAYER_SOLDIER(s)
 	{
-		if (s->bNeutral || s->bSide == OUR_TEAM) continue;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
The core bit looks fine, but the damage looks too low — can it still affect tanks? See where all this started: #503

daviel:
Works totally fine with tanks. Took the gamesaves out of #503. You will need to press alt+o up to 4 times though.

lynxlynxlynx:
So the damage should be higher. If it's causing healing somehow, that's a separate bug that should be fixed as well at some point.

daviel:
Okay, so it seems that in the code there is a 2-times multiplier for the damage value. This could lead to an overflow on a 16-bit value. This could be reason for the false calculations for the maximum number of 0xFFFF. 
I thought of two solutions for this:
1. We set the damage value to a number we know is the highest health-value in game(queen-bug as far as I know)
2. We set the damage value to 0xFFFF divided by 2
What would you recommend?

lynxlynxlynx:
I'd just set it to 10000, which is safe and much higher + a comment noting it should be high enough to kill tanks/queen and lower than 0xffff/2.

daviel:
Okay, done.

lynxlynxlynx:
Good enough, thanks!

PR #584, opened 2017-07-30
Passive bloodcat ambush fix

daviel:
fixes the first part of #126
I recognized that loading an ambush gamestate could lead to get the wrong statuscode for encounters because they were loaded later. I get around this by calling the enemies to ambush after a gamestate has been fully loaded.
Also I introduced a new method called CallAvailableTeamEnemiesToAmbush to prevent code duplication.


lynxlynxlynx:
Can you split this into two commits? First with just the method extraction and the second with the actual fix (minus the space-only changes — check your editor). You can use git commit -p to great effect when it comes to isolating changes.

daviel:
Okay, splitted it in two commits.

lynxlynxlynx:
Thanks, looks good!

PR #585, opened 2017-07-30
checking for existence of subject while climbing

### LINKED ISSUE #402: Climbing makes mercs visible at night
In this savegame I could capture a bug that I've already seen once or twice before.

First you see Barry in a dark place behind a building.
barry1

When he climbs the bulding, an enemy soldier gets an interrupt and shoots. Why/how can the soldier see Barry, when he's too far away and Barry isn't within a lightning source?
barry2

I suspect that either the jumping on roof makes Barry visible or maybe the lighting source within the building also lights the roof?
Anyway, something is wrong here, gameplay-wise.

Savegame before: Just select Barry and make him jump on the roof.
[QuickSave.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/503126/QuickSave.zip)

daviel:
fixes #402 
While soldiers/mercs were climbing there is a short time in which they have no fixed grid_no. If you use the WhosIsThere method to determine the soldier on that grid_no it returns 0(null). So by checking whether the value is 0 we can overcome the problem of false calculations of visibitily while climbing.

### DIFF
--- a/src/game/Tactical/OppList.cc
+++ b/src/game/Tactical/OppList.cc
@@ -877,6 +877,14 @@ INT16 DistanceVisible(const SOLDIERTYPE* pSoldier, INT8 bFacingDir, INT8 bSubjec
 	INT8	bLightLevel;
 
 	const SOLDIERTYPE* const pSubject = WhoIsThere2(sSubjectGridNo, bLevel);
+  
+       // While the merc is climbing up/down he is shortly on an 
+       // undefined grid_no(NOWHERE) we can't calc anything if this happens so
+       // we check the return of WhoIsThere2 for zero value
+       if (pSubject == 0) 
+       {
+           return( FALSE );
+       }
 
 	if (pSoldier->uiStatusFlags & SOLDIER_MONSTER)
 	{


lynxlynxlynx:
Man on a mission. 🥇 

It looks ok, here are two thoughts:
- now the merc will be invisible. Should we maybe treat him instead as standing up on the lower tile? I guess it would only create further problems (where does he end up in an interrupt? Meaning he could be invisible once turn-based starts) though, so it seems like a good middle ground
- it has bad indentation (nicely visible in the web view)

daviel:
Indentation and code style is now fixed.

He wouldn't be invisible. With this change there are only two moments of calculations of sight: Standing in front of the wall and standing on the roof. Before there was a calculation in which the grid_no NOWHERE was considered which doesn't make any sense because it would calculate the chance of being seen in perfect conditions(max range, max light etc.).

Furthermore I tested a few conditions like getting interrupted while climbing. Here some test cases and their result:
1. Not being in turn-based and climbing up while an enemy stands in front of the building -> getting an interrupt as soon as staying on the roof.
2. Not being in turn-based and climbing up while enemy stand in front of the building looking in a different direction -> turn-based starts and player is first to do actions

Maybe I should add that the change is in a method which just calculates the range of sight and nothing more. This gets calculated very frequently everytime something affects visibility on the whole map. This means that a calculations doens't affect the merc/soldiers visibility on a longer term.

lynxlynxlynx:
Exactly, so he's invisible during climbing, no? While before he was almost omnivisible. It's a matter of 1-2s, which is not negligible, but like I said, I can see how changing this could open worse cans of worms.

Btw, do you know if after climbing the stance is always standing first? Seems silly if true.

And please, start using spacing as was discussed and is used in all new code:
```
if (pSoldier->uiStatusFlags & SOLDIER_MONSTER) <<< GOOD
vs
if( pSoldier->uiStatusFlags & SOLDIER_MONSTER ) <<< BAD
```

daviel:
Okay, so I crunched some custom log files which gave me this insight:

1. Looking right at the wall you want to climb while standing doesn't produce visibility calculations while on the ground. But changing direction or stance to be able to climb does. No surprise here.
2. There are no mid-climb calculations! The problem is that there are visibility calculations for the merc on the roof before his grid is set to roof grid. This means it computes a value for a not-now-existent soldier(the reason why a null check is needed).

Here an example:
Situation: Merc in fron of a wall. Looking in different direction while crouching.
Standing up -> calc-sight -> rotate -> calc-sight -> climbup -> **NO calc** -> animation not done and **future position not set yet** -> calc-sight(max visibility; no soldier found yet on roof grid) -> animation done and position set -> calc-sight

Another note: 
If climbing is commanded the soldier receives a temporary grid_no(his future pos), the animation starts and sets his grid_no to the temporary grid_no after the animation is done. Can't tell what produces the too-early sight-calcs.. Investigating this would be overkill I think.


Climbing up needs the merc to stand while on ground and puts him in crouch after reaching the roof.
Climbing down needs the merc to crouch while on the roof and puts him to standing position after reaching the ground.

lynxlynxlynx:
Thanks! The other thing can be handled in a separate issue, if it ever becomes a noticeable problem.

chmallon:
This change seems wrong to me for multiple reasons.

First, the function returns a distance, not a BOOLEAN, so returning FALSE is questionable.
(Yes, there is one other place in the function, which also returns FALSE, but this is questionable, too.)

Second, the function now always returns 0, when there is no soldier.
This means the function does not work anymore, if there is no soldier at 'sSubjectGridNo'.
It looks like the function should work also, if there is no soldier, because there are multiple checks in the function whether 'pSubject' is a null pointer.
E.g. CalcIfSoldierCanSeeGridNo() uses DistanceVisible() and it seems perfectly fine for this caller that there is not soldier at this position.
Or in HandleDeidrannaDeath() there is certainly no soldier anymore where Deidranna died, so no soldier will make a quote about her death anymore.
Probably a better fix would be to test for 'sSubjectGridNo == NOWHERE'.
Or maybe even check at the caller site that 'sSubjectGridNo' is not NOWHERE.

daviel:
Thanks for your reply.

I tested this and you're right. It fails to return the right values for CalcIfSoldierCanSeeGridNo and HandleDeidrannaDeath. 
What bothers me is that why should I use the DistanceVisible method if it returns the highest value(MaxDistanceVisible-method) possible, everytime I calculate it for a grid_no where no soldier is standing at? Wouldn't it be better to use MaxDistanceVisible() right away?

We can't check for nowhere because climbing a roof isn't setting the soldiers grid_no to nowhere, it's setting it to his future position even if he's not yet standing on the roof. 

My new idea would be to exchange DistanceVisible everywhere where it's used to get the MaxDistanceVisible-value. But to do that we have to find out where it is used like that in the 37 occurences of the method in the whole code.

chmallon:
> What bothers me is that why should I use the DistanceVisible method if it returns the highest value(MaxDistanceVisible-method) possible, everytime I calculate it for a grid_no where no soldier is standing at?

I do not see how this always returns the same value as `MaxDistanceVisible()`.
There are cases, in which it returns the same value, but definitely not in all cases.
E.g. if the levels differ (`pSoldier->bLevel` and the parameter `bLevel`), then 2 is added.

> We can't check for nowhere because climbing a roof isn't setting the soldiers grid_no to nowhere

Then I misunderstood your earlier comment, in which you said something about `NOWHERE`.

I think the real culprit is this check:
```C
  if ( pSubject && !( pSubject->fMuzzleFlash && (bLightLevel > NORMAL_LIGHTLEVEL_DAY) ) )
  {
    // ATE: Made function to adjust light distence...
    sDistVisible = AdjustMaxSightRangeForEnvEffects(bLightLevel, sDistVisible);
  }
```

`sDistVisible` is only adjusted, if there is a soldier.
I.e. in this case there is no soldier (yet) on the roof, so the grid is treated as if it was at day (the default before adjustment).
Probably the check should be `bLightLevel > NORMAL_LIGHTLEVEL_DAY && (!pSubject || !pSubject->fMuzzleFlash)`.
I.e. adjust `sDistVisible` if it is darker than normal (yes, >, because higher numbers mean darker) and there is no soldier with a muzzle flash.

Please be aware, that even then climbing on a roof still can trigger that an enemy sees you, because (as mentioned above) a level difference increases the visibilty by 2.
So if you are very close to the visibility range, the additional 2 visibilty could be sufficient.

chmallon:
I thought about it again and I think this is a better condition:
```C
if (!pSubject || !pSubject->fMuzzleFlash || bLightLevel <= NORMAL_LIGHTLEVEL_DAY)
```
The second and third parts are logically equal to the original code. (just applied De Morgan`s rule for clarity)
Only the first part changed from `pSubject &&` to `!pSubject ||`.
I guess originally the null pointer check was added later and the one, who added it, botched it.
With the correction, `sDistVisible` is only used unaltered (i.e. treated as day) if there is a soldier, he has a muzzle flash and the surroundings are darker than day, i.e. the muzzle flash brightens the tile (again: higher light level is darker).
In particular now `sDistVisible` is adjusted when there is no soldier.
This is the only case, which changes with this correction and this is the important case for a climbing merc.

Further,  the test for the muzzle flash can be factored out, because it is used above, too:
```C
auto const hasMuzzleFlash = pSubject && pSubject->fMuzzleFlash;
// ...
if (!hasMuzzleFlash || bLightLevel <= NORMAL_LIGHTLEVEL_DAY)
```

Even further, if weather was implemented, then `AdjustMaxSightRangeForEnvEffects()` should be called nonetheless.
Otherwise a muzzle flash negates poor visibility due to rain.
In this case a better fix would be:
```C
if (hasMuzzleFlash)
  bLightLevel = MIN(bLightLevel, NORMAL_LIGHTLEVEL_DAY);
sDistVisible = AdjustMaxSightRangeForEnvEffects(bLightLevel, sDistVisible);
```

daviel:
Could you put this in a PR or a branch of your own? I would like to test this.

My commit has to be undone either your changes work or not. But I don't know how regressions are handled in this project.

lynxlynxlynx:
Nice analysis guys! I'll go check what happened in the interim.

PR #587, opened 2017-08-02
Moneycheat added (Laptopview and + key)

### LINKED ISSUE #426: Adding money with '+' key cheat not working?
I just tried to use the cheat and it doesn't work. I strongly suspect this to be a problem on my side. Can someone confirm the cheat to be working in Stracciatella?

EDIT: Strange thing is, that the "removing money cheat" with '-' key works.

daviel:
fixes #426 
In the code there wasn't any handling for the CTRL and + cheat. So added it.
The value for decreasing the money value was 10.000 therefore too low. I increased it to 100.000.

Disclaimer: I have a german keyboard layout.

### DIFF
--- a/src/game/Laptop/Laptop.cc
+++ b/src/game/Laptop/Laptop.cc
@@ -2840,6 +2840,7 @@ void HandleKeyBoardShortCutsForLapTop(UINT16 usEvent, UINT32 usParam, UINT16 usK
 			break;
 
 		case '=':
+		case '+':
 			if (CHEATER_CHEAT_LEVEL())
 			{
 				// adding money


lynxlynxlynx:
LGTM, but needs testing on another layout.

JAsmine-ja2:
Quote from the original JA2 manual: 

> Laptop
> + Increase funds by $100,000.
> - Decrease funds by $ 10,000.

Seems like the "decrease money cheat" is supposed to be -10.000 instead of -100.000 ?

daviel:
Okay. I took the information from different websites. Will undo the -100.000$ to -10.000$. 
Sorry, it just takes the + and - key for the cheat. There is no CTRL needed as I see.

daviel:
Okay, changed it! :)

lynxlynxlynx:
Thanks, finally got around to testing it. Wasn't working for my layout either.

PR #589, opened 2017-08-16
Climbingvisibility revert + new bugfix

daviel:
Reverts the change faulty change in #585 see the discussion for more information(thanks to chmallon)

new approach for fixing #402 

As the method distancevisible is also used on empty grids(no soldier on grid) we can't return a 0 value. Now we calculate the visibility correctly by night by using better conditions for lightlevel-calculation.

### DIFF
--- a/src/game/Tactical/OppList.cc
+++ b/src/game/Tactical/OppList.cc
@@ -876,7 +876,10 @@ INT16 DistanceVisible(const SOLDIERTYPE* pSoldier, INT8 bFacingDir, INT8 bSubjec
 	INT16 sDistVisible;
 	INT8	bLightLevel;
 
+	// IMPORTANT! WhoIsThere2 can return a null-pointer for grid calcs
 	const SOLDIERTYPE* const pSubject = WhoIsThere2(sSubjectGridNo, bLevel);
+	// muzzleflash can only considered if there is actually a soldier on the grid_no
+	BOOLEAN const hasMuzzleFlash = pSubject && pSubject->fMuzzleFlash;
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Looks good, also making weather count all the time (just add braces please). And no wonder 1.13 added flash suppressors ...

chmallon:
Adjusting `bLightLevel` is probably the right solution for the subsequent code, too.
E.g. if it is very dark (`bLightLevel > NORMAL_LIGHTLEVEL_DAY + 5`) then in the old code a soldier with muzzle flash was even more visible than at normal day to a soldier with `NIGHTOPS`.
Now `bLightLevel` is adjusted to `NORMAL_LIGHTLEVEL_DAY` due to the muzzle flash and so `NIGHTOPS` has no effect.
Something similar holds for all three kinds of goggles.

chmallon:
BTW: It seems that JA2-1.13 has my first suggestion (turning `pSoldier &&` into `!pSoldier ||`) since it was imported into version control.
This from r1 on Mon Apr 17 19:25:32 2006:
```C
// Snap: I think this was intended to give maximum visibility to targets with muzzle flash...
// Corrected accordingly:
//if ( pSubject && !( pSubject->fMuzzleFlash && (bLightLevel > NORMAL_LIGHTLEVEL_DAY) ) )
if ( !( pSubject && pSubject->fMuzzleFlash && (bLightLevel > NORMAL_LIGHTLEVEL_DAY) ) )
```
So there soldiers with muzzle flash can be "too much" visible.

lynxlynxlynx:
Sorry for the delay && thanks!

PR #592, opened 2017-09-27
travis: use the apt addon

lynxlynxlynx:
one sudo less (closer to container-based runs) and fixes this apt dependency problem that recently broke the build:

> The following packages have unmet dependencies:
>  libsdl2-dev : Depends: libegl1-mesa-dev
>                Depends: libgles2-mesa-dev
> E: Unable to correct problems, you have held broken packages.


lynxlynxlynx:
@selaux which also raises the question why we need root for installing. Is it just for debs? Can't it take a prefix into account?

selaux:
LGTM, we could remove the commented out line though.

@lynxlynxlynx We need root because when we do a nightly build we set `EXTRA_DATA_DIR=/usr/share/ja2`, which will make the unittest fail unless we actually install the necessary data files into that directory (which of course needs root). We could easily work around that by introducting a cli switch or env variable to be able to control where to look for data during runtime.

lynxlynxlynx:
Ok. It's not that important for us, but good to know.

PR #594, opened 2017-10-05
make INSTALL_LIB_DIR configurable

MeisterP:
in gentoo, the library should end up in /usr/lib64 otherwise there's a qa check failing. therefore, make the library path configurable.

```
making executable: usr/lib/libstracciatella.so
Files matching a file type that is not allowed:
   usr/lib/libstracciatella.so
 * ERROR: games-strategy/ja2-stracciatella-9999::poncho failed:
 *   multilib-strict check failed!
 * 
 * Call stack:
 *   misc-functions.sh, line 605:  Called install_qa_check
 *   misc-functions.sh, line 217:  Called source 'install_symlink_html_docs'
 *   80multilib-strict, line  46:  Called multilib_strict_check
 *   80multilib-strict, line  42:  Called die
 * The specific snippet of code:
 *   		[[ ${abort} == yes ]] && die "multilib-strict check failed!"
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,6 +36,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 set(VERSION_TAG "" CACHE STRING "Build date for nightly versioning")
 set(EXTRA_DATA_DIR "" CACHE STRING "Directory for externalized data")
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Looks fine, thanks.

selaux:
👍

PR #596, opened 2017-10-16
Check for blocking doors when attacking with HTH

### LINKED ISSUE #595: MERCS can punch through open doors
Igor is able to attack an enemy through a open door which should not be possible.

See screenshot and savegame:

[QuickSave.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1388731/QuickSave.zip)

daviel:
fix #595
Added a check for blocking doors when attacking with Hand-to-hand.


lynxlynxlynx:
Looks good, thanks!

PR #597, opened 2017-10-16
Fixed grenades getting blocked by open doors

### LINKED ISSUE #94: Open doors can block grenades

---

In this savegame, an enemy soldier is in prone position, inside a building, just behind the open front door. Throwing a stun grenade just in front of the soldier has absolutely no effect on the enemy. No reaction. It might be, because the enemy is inside the building and the door (although open) is somehow shielding him.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/94

daviel:
Fixes #94 
Some defines weren't used for explosion calculations, also the spread of grenades was stopped due to high structureheight which is weird because the default size for doors and walls is 4.

### DIFF
--- a/src/game/Tactical/PathAI.cc
+++ b/src/game/Tactical/PathAI.cc
@@ -2511,7 +2511,7 @@ UINT8 InternalDoorTravelCost(const SOLDIERTYPE* pSoldier, INT32 iGridNo, UINT8 u
 
 	if ( IS_TRAVELCOST_DOOR( ubMovementCost ) )
 	{
-		ubReplacementCost = TRAVELCOST_OBSTACLE;
+		ubReplacementCost = ubMovementCost;
 
 		switch( ubMovementCost )
 		{
@@ -2591,6 +2591,7 @@ UINT8 InternalDoorTravelCost(const SOLDIERTYPE* pSoldier, INT32 iGridNo, UINT8 u
 				iDoorGridNo = iGridNo + DirIncrementer[ NORTHWEST ] + DirIncrementer[ WEST ];
 				break;
 			default:
+				ubReplacementCost = TRAVELCOST_OBSTACLE;
 				break;
 		}
 
--- a/src/game/TileEngine/Explosion_Control.cc
+++ b/src/game/TileEngine/Explosion_Control.cc
@@ -1472,8 +1472,8 @@ static void GetRayStopInfo(UINT32 uiNewSpot, UINT8 ubDir, INT8 bLevel, BOOLEAN f
    // Have we hit things like furniture, etc?
 	 if ( Blocking != NOTHING_BLOCKING && !fTravelCostObs )
 	 {
-      // ATE: Tall things should blaock all
-      if ( bStructHeight == 4 )
+      // ATE: Tall things should blaock all; Default wall/door height is 4
+      if ( bStructHeight > 4 )
       {
 				 (*pubKeepGoing) = FALSE;
       }


src/game/Tactical/PathAI.cc
```diff
@@ -2533,62 +2533,77 @@ UINT8 InternalDoorTravelCost(const SOLDIERTYPE* pSoldier, INT32 iGridNo, UINT8 u
...
 				break;
 			case TRAVELCOST_DOOR_OPEN_SW_W:
 				fDoorIsObstacleIfClosed = FALSE;
 				iDoorGridNo = iGridNo + DirIncrementer[ SOUTHWEST ] + DirIncrementer[ WEST ];
+				ubReplacementCost = TRAVELCOST_DOOR_OPEN_SW_W;
 				break;
 			case TRAVELCOST_DOOR_OPEN_NW_W:
 				fDoorIsObstacleIfClosed = FALSE;
 				iDoorGridNo = iGridNo + DirIncrementer[ NORTHWEST ] + DirIncrementer[ WEST ];
+				ubReplacementCost = TRAVELCOST_DOOR_OPEN_NW_W;
```
lynxlynxlynx: why not just set `ubReplacementCost = ubMovementCost;` before the switch?

src/game/TileEngine/Explosion_Control.cc
```diff
@@ -1472,8 +1472,8 @@ static void GetRayStopInfo(UINT32 uiNewSpot, UINT8 ubDir, INT8 bLevel, BOOLEAN f
    // Have we hit things like furniture, etc?
 	 if ( Blocking != NOTHING_BLOCKING && !fTravelCostObs )
 	 {
-      // ATE: Tall things should blaock all
-      if ( bStructHeight == 4 )
+      // ATE: Tall things should blaock all; Default wall/door height is 4
+      if ( bStructHeight > 4 )
```
lynxlynxlynx: where did you get the default? Is it PROFILE_Z_SIZE or some other define?

src/game/Tactical/PathAI.cc
daviel: Before the switch case there is 
`ubReplacementCost = TRAVELCOST_OBSTACLE;`

Why would you want to override it with the movementcost?

src/game/TileEngine/Explosion_Control.cc
daviel: I printed the bStructHeight in the console and red the output. Strange thing is that bStructHeight isn't consistently used in the whole code there are calculations in LOS.cc and Weapons.cc looking like this
`if (bStructHeight > 0)`
`dEndZPos = ((FLOAT) (bStructHeight + bLevel * PROFILE_Z_SIZE) - 0.5f) * HEIGHT_UNITS_PER_INDEX;`

src/game/Tactical/PathAI.cc
lynxlynxlynx: So you don't need to reassign it for each case in the switch, just the special ones from before and TRAVELCOST_OBSTACLE for the default case. All the new additions are like:
```
case A:
  ...
  ubReplacementCost = A
```
which is not as sleek as it could be.

src/game/TileEngine/Explosion_Control.cc
lynxlynxlynx: ok, thanks. It's a thing to eventually clean up, but not a requirement for this PR.

daviel:
Okay changed it. I did not look into the IS_TRAVELCOST_DOOR function which just checks if ubMovementCost lies within an interval of the door-defines.

lynxlynxlynx:
You forgot to change it for the default case, where it should be the previous default.

daviel:
Okay, added the line.

PR #598, opened 2017-10-16
Soldiers with enough BP can now stand up properly after being hit by …

### LINKED ISSUE #247: Hit by grenade issues
I got a little stupid but found a bug.  Enemy threw grenade at Grunty.  Grunty is knocked prone but facing enemy.  MD and Barry remain crouched, MD facing enemy but Barry facing North.  Blood  is knocked back.

Grunty and MD are able to fire.

When attempting to fire with Barry I kept getting the stopwatch animation and nothing would happen.   During gameplay moving him one square North showed and used the correct AP for crouching but he then fell prone and fired at the enemy while still facing North(90 degree shot).  When loading this savegame he will fire as soon as you go into the tactical screen so you can't see the move, fall prone, shoot over the shoulder bit.   I'm guessing a problem with the game trying to place him in a prone position without the room to do so. 

Trying to change stance with Blood gives a 'Cannot change stance at this time' message which is weird as he still has almost full stamina.  Blood can also still be moved, just prevented from changing stance at his starting location.

screen005
screen006
[grenade_bug.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/160142/grenade_bug.zip)

daviel:
…a stun grenade

fixes #247
I'm not sure what exactly is the problem in this situation. I opened a pull request for fixing the problem with blood won't standing up. But I'm still unsure whether the issue-creator meant exactly this or whether it's part of the gameplay that someone being hit by a stun grenade can't stand up even with enough BP.

Would like to hear some feedback. :)

### DIFF
--- a/src/game/Tactical/Soldier_Control.cc
+++ b/src/game/Tactical/Soldier_Control.cc
@@ -7663,7 +7663,7 @@ BOOLEAN InternalIsValidStance(const SOLDIERTYPE* pSoldier, INT8 bDirection, INT8
 
 	if ( pSoldier->bCollapsed )
 	{
-		if ( bNewStance == ANIM_STAND || bNewStance == ANIM_CROUCH )
+		if ((bNewStance == ANIM_STAND || bNewStance == ANIM_CROUCH) && pSoldier->bBreath < OKBREATH)
 		{
 			return( FALSE );
 		}


lynxlynxlynx:
@JAsmine-ja2 can you chime in? You know the game best.

src/game/Tactical/Soldier_Control.cc
```diff
@@ -7663,7 +7663,7 @@ BOOLEAN InternalIsValidStance(const SOLDIERTYPE* pSoldier, INT8 bDirection, INT8
 
 	if ( pSoldier->bCollapsed )
 	{
-		if ( bNewStance == ANIM_STAND || bNewStance == ANIM_CROUCH )
+		if ( bNewStance == ANIM_STAND || bNewStance == ANIM_CROUCH && pSoldier->bBreath < OKBREATH)
```
lynxlynxlynx: I'm pretty sure you need to put parentheses around the original conditions, since && has precedence over ||. At the same time you can kill that starting space.

daviel:
Okay, changed it. 

Some additional thoughts:
Stun grenads have a specified radius within you can get hit by them. The further you are away from the "explosions" center the less BP your merc loses. So there are situations in which your merc gets hit and is standing up right after getting hit though. So the only situation in which your mercs can't stand up is not having enough BPs.

JAsmine-ja2:
> So the only situation in which your mercs can't stand up is not having enough BPs.

According to my experience with the game, this is 100% true.

lynxlynxlynx:
Thanks to both, merging!

PR #599, opened 2017-10-16
Fixed error with gas on roof hitting mercs below

### LINKED ISSUE #477: Gas on a roof can damage the merc underneath the roof
This one is rather difficult to reproduce, but I managed to make it happen three times.

Throwing a gas grenade on top of a roof works as it should. Throwing a gas grenade inside a building (with roof ;) works, too.

However, sometimes, a merc on ground level can be damaged by gas which is over him (on the roof). I think, this has something to do with swithcing between height levels with TAB. Maybe, if height level is toggled during the throwing animation, things get messy. Can't say for sure.

gas-effect-roof

Savedgame:
[SaveGame09.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/685591/SaveGame09.zip)
(let Stephen throw the grenade on the roof and make him run inside. Maybe, toggling the height level during the throwing animation triggers the glitch?).

daviel:
Fixes #477 

Preventing damage on the wrong level by rechecking whether there is gas on the grid and height-level, also considering gasmask etc..

### DIFF
--- a/src/game/TileEngine/Explosion_Control.cc
+++ b/src/game/TileEngine/Explosion_Control.cc
@@ -1172,7 +1172,11 @@ static BOOLEAN ExpAffect(const INT16 sBombGridNo, const INT16 sGridNo, const UIN
 			if (pSoldier == NULL) return fRecompileMovementCosts;
 			// someone is here, and they're gonna get hurt
 
-			fRecompileMovementCosts = DishOutGasDamage(pSoldier, pExplosive, sSubsequent, fRecompileMovementCosts, sWoundAmt, sBreathAmt, owner);
+			// only damage if there is smoke/gas on gridno and bLevel
+			if (gpWorldLevelData[pSoldier->sGridNo].ubExtFlags[pSoldier->bLevel] & ANY_SMOKE_EFFECT)
+			{
+			  fRecompileMovementCosts = DishOutGasDamage(pSoldier, pExplosive, sSubsequent, fRecompileMovementCosts, sWoundAmt, sBreathAmt, owner);
+			}
 /*
 		 if (!pSoldier->bActive || !pSoldier->bInSector || !pSoldier->bLife || AM_A_ROBOT( pSoldier ) )
 		 {


src/game/TileEngine/Explosion_Control.cc
```diff
@@ -1168,11 +1168,58 @@ static BOOLEAN ExpAffect(const INT16 sBombGridNo, const INT16 sGridNo, const UIN
...
-			fRecompileMovementCosts = DishOutGasDamage(pSoldier, pExplosive, sSubsequent, fRecompileMovementCosts, sWoundAmt, sBreathAmt, owner);
+		 // Look for gas
+		 EXPLOSIVETYPE const* pExplosive = 0;
+		 if (!AM_A_ROBOT(pSoldier))
+		 {
+			 if (gpWorldLevelData[pSoldier->sGridNo].ubExtFlags[pSoldier->bLevel] & MAPELEMENT_EXT_TEARGAS)
+			 {
+				 if (!(pSoldier->fHitByGasFlags & HIT_BY_TEARGAS) && bPosOfMask == NO_SLOT)
+				 {
+					 // check for gas mask
```
lynxlynxlynx: Shouldn't this comment be higher up?
Also, saving `gpWorldLevelData[pSoldier->sGridNo].ubExtFlags[pSoldier->bLevel]` to a temporary var would make this whole section more readable.

lynxlynxlynx:
Thanks!

daviel:
I changed the code. DishOutGasDamage already checks for a gas mask so there is no need to check before it.

src/game/TileEngine/Explosion_Control.cc
```diff
@@ -1168,11 +1168,33 @@ static BOOLEAN ExpAffect(const INT16 sBombGridNo, const INT16 sGridNo, const UIN
...
+		 // Look for gas
+		 if (!AM_A_ROBOT(pSoldier))
+		 {
+			 if (soldierbLevel & MAPELEMENT_EXT_TEARGAS)
+			 {
+				 pExplosive = &Explosive[GCM->getItem(TEARGAS_GRENADE)->getClassIndex()];
+			 }
+			 if (soldierbLevel & MAPELEMENT_EXT_MUSTARDGAS)
+			 {
+				 pExplosive = &Explosive[GCM->getItem(MUSTARD_GRENADE)->getClassIndex()];
```
lynxlynxlynx: Sorry, realised something else. If the two gases are both present, the latter can override pExplosive. So I guess it would be better to use |= for assignment?

daviel: As far as I tested there can't be 2 or more gases on one spot. There is kind of a priority for the gases leading to normal smoke getting overwritten by mustard gas which get's overwritten by creatures gas.

Here a screenshot in which i threw a tear gas grenade right on a mustard gas grenade

daviel: Okay, wrong. There can be 2 or more gases present but only 1 is visible at a time.

daviel: Okay, wrong again. There can be 2 gases on one spot but only one gas is affecting mercs at a time you can't stack grenade effects.
I saw mustard gas decreasing BP and thought the reason for this the underlying tear gas. But mustard gas is decreasing BP and HP.

lynxlynxlynx: So how do you know that both are not being applied?

daviel:
Okay, after testing with console output I recognized that gas affects stack. I came up with a much smaller solution now.

chmallon:
I think the actual problem is in DecaySmokeEffects():

    // damage anyone standing in cloud
    SpreadEffectSmoke(pSmoke, REDO_SPREAD_EFFECT, 0);

This line always affects the floor level and ignores where the gas/smoke effect actually is.
The `0` should be `bLevel` instead.
The wrong effect happens in real-time mode.
I corrected this problem some time ago in my development version.

I think the new check in ExpAffect() should be removed again.
The call chain is `DecaySmokeEffects() -> SpreadEffectSmoke() -> SpreadEffect() -> ExpAffect()` and with my proposed fix only happens in the correct case.
I'm not sure what side effects the current change has, but it seems wrong for `ExpAffect()` to inspect world data.

lynxlynxlynx:
Definitely cleaner.

PR #602, opened 2017-10-21
Gasroofbug

daviel:
See last comment #599


lynxlynxlynx:
Did you also test it in practice?

daviel:
I did. The bug didn't show up again and debris damage is still working as aspected.

PR #605, opened 2017-10-31
Update serde and serde_json to Version 1. Also be less restrictive.

Manuel-K:


Manuel-K:
Doesn't work because an error message was changed. That's bad.

Manuel-K:
The error message was changed before the release of version 0.9.7: https://github.com/serde-rs/json/commit/9b0010c5a324e592dfc7cd8d4adf40982282d083

lynxlynxlynx:
Any particular benefit of the bump? LGTM otherwise.

Manuel-K:
We were forcing rust to use specific outdated versions of both libraries.

I don't see a reason to do that unless there is a specific bug that forces us to do so.

Manuel-K:
These changes (or similar changes) are included in #548

PR #610, opened 2017-11-05
Copy audio into a buffer of the correct size

### LINKED ISSUE #609: Sound memory limit hit after upgrading to SDL 2.0.7
SDL2 2.0.7 seems to have changed something in the audio resampling code that makes the game hit the sound memory limit in the main menu:

```
2017/10/04 18:55:17 ERROR   [Sound           ] SoundLoadDisk: trying to play sounds/big switch 03 in.wav, not enough memory
Size: 40532, Used: 46464992, Max: 33554432
```

This does not happen with 2.0.5.

selaux:
This should fix #609. But it comes with a new problem: Performance. Basically everytime a music piece (aka. long sound) is loaded, this results in a tiny hang. Not sure if that was the same before and not sure if we can actually improve performance on this.

Imho it does too many big allocations (one for source wav, one for cvt buffer, one for destination wav). It would probably be better to stream into AudioCVT, with preallocated tiny buffers. I tried this approach already once, but failed implementing it. It's not as simple as it seems, as you have to ensure that you have enough content in the buffers, so the data can be pulled at any time.

### DIFF
--- a/src/sgp/SoundMan.cc
+++ b/src/sgp/SoundMan.cc
@@ -623,15 +623,20 @@ static SAMPLETAG* SoundLoadDisk(const char* pFilename)
     };
   }
 
+  UINT32 convertedSize = cvt.len * cvt.len_ratio;
+
   strcpy(s->pName, pFilename);
-  s->n_samples = UINT32(cvt.len * cvt.len_mult / (wavSpec.channels * 2));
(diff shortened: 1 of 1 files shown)


Manuel-K:
It's slower now because the conversion buffer is four times larger. 

Streaming the music is the only real solution.

Issue #607, opened 2017-11-04
Launcher: Improve GUI


Manuel-K:
Continuation of #548 

[ ] When switching from predefined to custom resolution, the custom resolution should be updated to the selected predefined resolution if it was 0 x 0.
[ ] The game does not support all possible custom resolutions. We should warn the user if the selected resolution is too low.
[ ] Make it prettier. How? We need mockups/suggestions.

JAsmine-ja2:
Some Mockups back from the Bitbucket days:
https://bitbucket.org/gennady/ja2-stracciatella/wiki/Drawing%20Board

Manuel-K:
@selaux I don't think it was a good idea to remove the fl file when you commited the generated cc and h files.

Manuel-K:
Current layout:
![launcher_0](https://user-images.githubusercontent.com/775439/32417541-776cc2ac-c25b-11e7-9d08-ddb0c40e211a.png)

Minor improvement:
![launcher_1](https://user-images.githubusercontent.com/775439/32417543-799c0b50-c25b-11e7-8f1e-e620d9fb91e2.png)

Manuel-K:
Any opinions?

JAsmine-ja2:
Does the user have to select the game version or does the launcher detect the game version?

Should the game version include the version no besides the language?
e.g. 1.12 "Gold" English
or 1.05 Deutsch

Manuel-K:
The user has to select the game version manually. For languages other than Russian which has the options RUSSIAN and RUSSIAN_GOLD, it's basically a language selector. 

At the moment the options are all caps i.e. they match the possible settings in the config (`ja2.json`) file. I'd prefer normal spelling.

JAsmine-ja2:
Does/Could the launcher use our custom logo as icon?

selaux:
I would like to add vanilla data directory and game version auto detection some time in the future, but for now I think that would be too complicated. For now it should reflect what we can actually set. 

Branding imho is a must (at least use the logo). Also currently resizing does not work properly.

@Manuel-K Minor improvements look good.

Also we have some mocks in a very old ticket: #19 or here https://bitbucket.org/gennady/ja2-stracciatella/wiki/Drawing%20Board.

lynxlynxlynx:
Can we adjust control padding? Looks too small now for the input boxes.

[...]

PR #613, opened 2017-11-07
Minor launcher improvements

Manuel-K:


Manuel-K:
Merged as discussed in #607

src/launcher/Launcher.cc
```diff
@@ -47,6 +49,8 @@ void Launcher::show() {
 	populateChoices();
 	initializeInputsFromDefaults();
 
+	const Fl_PNG_Image icon("logo32.png", logo32_png, 1374);
```
selaux: Should this rather be `sizeof(logo32_png)`?

Manuel-K: I don't think it really matters. It is supposed to be an upper limit to guard against corruption.

Issue #608, opened 2017-11-04
Crash in SoundLoadDisk() on Arch Linux with SDL2 2.0.6


Manuel-K:
It seems the call to `SDL_ConvertAudio` crashes for `music/menumix1.wav`.

I've never had that problem before.

lynxlynxlynx:
Is it as easy to reproduce ingame as the filename suggests?

Manuel-K:
At least on my system. I'm currently investigating if the problem is the version of SDL2.

Manuel-K:
Obviously I'm very good a triggering main menu segfaults.

Manuel-K:
Updating sdl2 from Arch Linux package version 2.0.6-2 to 2.0.7-1 fixed the problem. Downgrading to the previous version reintroduces the segfault.

Manuel-K:
It's probably not our fault.

lynxlynxlynx:
If that version is consistently bad, we should document it somewhere.

Manuel-K:
I've tested several other versions:
- official Arch Linux 2.0.6-1 package: segfault.
- rebuilt 2.0.6-2 package: segfault
- official Arch Linux 2.0.5-4 package: okay

Are there any problems with 2.0.6 on other distributions?

Manuel-K:
Multiplying the malloced amount (https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/sgp/SoundMan.cc#L614) by an arbitrary factor e.g. 3 fixes the segfault with 2.0.6.

selaux:
Maybe related: https://bugzilla.libsdl.org/show_bug.cgi?id=3858

Manuel-K:
Yes, it seems to be related. But it also seems that we are already doing it the correct way.

selaux:
Can we actually do something about this, or do we just assume `2.0.6` is the issue and we note that in the README. The crash actually occurs inside SDL, but I didn't debug further.

Manuel-K:
It seems that chocolate-doom had the same crash and it seems they managed to fix it for 2.0.6 by changing the malloc size to something that looks similar to ours.

I've already built a sdl2-2.0.6-2 package with debug symbols, but I was busy with the launcher and the white space stuff.

Manuel-K:
Ran the debugger and checked the error in the SDL2 code. Our buffer is large enough.

[...]

PR #615, opened 2017-11-08
SDL2 2.0.6 Warnings

Manuel-K:
Fix (or rather workaround) for #608 

- [x] Disable sound if version 2.0.6 is detected at runtime
- [x] Print error message to console and log file
- [x] Display warning in main menu
- [x] Add warning to `COMPILATION.md`

### DIFF
--- a/COMPILATION.md
+++ b/COMPILATION.md
@@ -1,6 +1,9 @@
 ## Dependencies
 
-- SDL 2.0.4 (included for Windows and OS X)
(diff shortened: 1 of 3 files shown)


selaux:
Added note to `COMPILATION.md`

Manuel-K:
It is definitely a run time issue not a compile time issue. Perhaps it would be better to change the `compile with a different version` to `run the game with a different version`.

selaux:
👍 Fixed, will merge when builds succeed...

Manuel-K:
Erster ;-)

PR #616, opened 2017-11-10
Launcher: Warn if the custom resolution is too low

Manuel-K:


selaux:
I would argue that we now have two places where the logic for the minimal resolution lives. Can we unify them somehow?

Manuel-K:
Inlined. Now all the logic is in one place.

PR #617, opened 2017-11-10
Add missing bracket around COUNTERDONE define

Manuel-K:
Found by @derSteFfi in #556


selaux:
👍 Does it fix a specific issue or is this more precautionary?

Manuel-K:
Precautionary: Don't play with fire unless it is absolutely necessary.

chmallon:
This is in the case '!defined CALLBACKTIMER', which is not used. (see the beginning of Timer_Control.h)
I deleted the whole case in my development version, because it won't work properly anyway.
Interestingly you did not add the parentheses in the case, which is actually used -- or even simpler: Remove the pointless and confusing '? TRUE : FALSE'.
In fact I turned all the macros into inline functions in my development version.

lynxlynxlynx:
Is there a particular reason you don't want to share your work more transparently any more?

PR #619, opened 2017-11-10
Increase C++ standard to C++11

selaux:
I got annoyed that we cannot use modern C++ features, so I updated the C++ standard we use to C++11.

I removed the `throw` annotations as they are deprecated with C++11. According to the [spec](http://en.cppreference.com/w/cpp/language/noexcept_spec) all the functions are potentially throwing functions so the annotations should not matter.

It now builds for me without warnings. Lets check if it is the same on macOS/Windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ endmacro()
 project(ja2-stracciatella)
 set(JA2_BINARY "ja2")
 set(LAUNCHER_BINARY "ja2-launcher")
-set(CMAKE_CXX_STANDARD 98)
+set(CMAKE_CXX_STANDARD 11)
 
 ## Versioning
 
--- a/src/sgp/UTF8String.cc
+++ b/src/sgp/UTF8String.cc
@@ -4,7 +4,7 @@
 // #include <stdio.h>
 
 /** Create object from existing UTF-8 encoded string. */
-UTF8String::UTF8String(const char *utf8Encoded) throw (InvalidEncodingException)
+UTF8String::UTF8String(const char *utf8Encoded)
 	:m_encoded("")
 {
 	std::string value(utf8Encoded);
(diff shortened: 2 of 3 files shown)


selaux:
Can I haz feedback? 😉

Manuel-K:
I'm not sure about this. 

We know it works on g++ shipped with Ubuntu >= Trusty, Visual Studio and whatever compiler macOS uses at the moment. 

MXE has g++ 5.4.0. That should also work.

It also works on Arch Linux.


What obscure setups/systems do we lose if we increase the standard?

selaux:
According to [this](http://en.cppreference.com/w/cpp/compiler_support) we would loose:
- MSVC < 2015
- GCC < 4.8 (~from 2013)
- CLang < 3.3 (~from 2013)

Some things from C++11 are also supported in some previous versions. So I think it also depends on what we use from C++11.

I'm also not sure what kind of obscure systems/setups we should think about in this case 😉 . What systems do we want to support?

Manuel-K:
If we increase the standard, we should definitely **wait until after the release of 0.16**.

lynxlynxlynx:
I agree on the last point. We can add a warning to the release notes, asking for people to contact us if they run this on odd stuff like amigas, suns, whatever.

janisozaur:
For c++11 you would be probably losing no support whatsoever. Even the rhel/sales and the like which hold off on GCC 4.9 due to some ABI breakage, should be able to compile c++11 code. Note that was mostly compiler/language thing, so as long as you have a compiler already, the platform should still be supported. Barring maybe some atomics and threading.

That said, we (OpenRCT2) are planning to move on to c++17 soon-ish, which poses more issues due to this being more library-focused version.

derSteFfi:
I honestly don't know enough about this to have an informed opinion. If all build systems mentioned by @Manuel-K remain working, I have no objections to merging, though. I would also agree that after releasing 0.16 is the better point in time

PR #623, opened 2017-11-18
extra item combinations

lynxlynxlynx:
place goggles on helmet
tiny items as legging attachments

game.json:
set "extra_attachments": true

based on vxx's patch, another part of #307


src/game/Tactical/Items.cc
```diff
@@ -336,6 +337,16 @@ static const AttachmentInfoStruct AttachmentInfo[] =
...
+
+	// extras
+	{NIGHTGOGGLES,				IC_ARMOUR,	NO_CHECK,								0},
+	{UVGOGGLES,					IC_ARMOUR,	NO_CHECK,								0},
+	{SUNGOGGLES,				IC_ARMOUR,	NO_CHECK,								0},
+
+	{ADRENALINE_BOOSTER,		IC_ARMOUR,	NO_CHECK,								0},
+	{REGEN_BOOSTER,				IC_ARMOUR,	NO_CHECK,								0},
+	{BREAK_LIGHT,				IC_ARMOUR,	NO_CHECK,								0},
+
```
lynxlynxlynx: these are aligned with the line of max length, if we ever want to make the whole thing look like a table.

JAsmine-ja2:
Maybe we could add the Robot Remot Control to the things that can be attached to a helmet. Just for general consistency.

assets/externalized/game.json
```diff
@@ -79,6 +79,9 @@
     //  vanilla_setting: false
     "gui_extras": true,
 
+    //  Enables extra attachments: boosters and break lights on leggings, goggles on helmets
```
Manuel-K: Slightly OT, but I think we should move the comments to a separate file. Although there are some libraries that support comments in JSON files,  they are not part of the standard.

lynxlynxlynx: that's exactly the reason we use one that does. This way is much more helpful for the user.

lynxlynxlynx:
@JAsmine-ja2 I guess so. I don't have it in any save game, but it would be easy to add.

Do note that this is just an inventory management improvement. Eg. goggles on a helmet aren't considered to be on the face (worn). It's more obvious with the legging attachments.

lynxlynxlynx:
One other thing, is it possible to apply compound18 (or whatwasit) or the alien goo too a steel helmet as well? If so, the two variants are now excluded.

assets/externalized/game.json
Manuel-K: We use a different library (`serde_json`) for the launcher. It doesn't look like it supports comments (Note: not tested, only skimmed the manual). If we want to include a `game.json` editor, it will become messy.

lynxlynxlynx: It seems silly to me to write an editor, when we can just launch some platform default (notepade, xdg-open and ?? on mac).

lynxlynxlynx: we have to move game.json first any way or it may get installed in a non-writable place. Uff, I can see a snowball nightmare ...

JAsmine-ja2:
It was never possible to upgrade the steel helmet to C18 or Coated version. I don't think, we even have specific graphic files for upgraded steel helmets.

Concerning head gear stuff on the helmet: Ofc they don't work when not in the head gear slot. Still, the advantage for inventory management is considerable, since inventory space is rather limited in vanilla JA2.

JAsmine-ja2:
[SaveGame03.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1484924/SaveGame03.zip)

Here's a savegame with robot remote control. I just spawned one in Sidney's inventory with good ol' cheating ;)

lynxlynxlynx:
ok, so everything is working as expected. :)

Manuel-K:
Code looks good. Not tested.

PR #627, opened 2017-11-18
Don't crash if we don't have a target for punching

lynxlynxlynx:


lynxlynxlynx:
it's trivial to reproduce, just unequip your weapon and try attacking an empty cell.

Manuel-K:
Finally got my power cord - my battery was running low.

PR #632, opened 2017-12-09
 Build release tags

selaux:
__Goal__: Build release packages when a git tag is created.

Also include some rust changes for the newest serde version that was released recently.

Contains a bunch of changes to CI, so I have to open a PR early again to ensure everything still works.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ sudo: required
 before_install:
   - export PUBLISH_BINARY="false"
   - export BUILD_SWITCHES=""
-  - export BUILD_TYPE="Debug";
+  - export BUILD_TYPE="RelWithDebInfo";
 
   - if [[ "$TRAVIS_PULL_REQUEST" != "false" ]]; then
       export PUBLISH_BINARY="true";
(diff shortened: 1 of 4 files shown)


selaux:
Should be working now. Test release is available here: http://builds-ja2.stefanlau.com/releases/

selaux:
We get some warnings when installing the deb though (ownership of files etc.). Will fix in another PR.

lynxlynxlynx:
Not that important, but the nigthlies should use the Debug BUILD_TYPE.

selaux:
True, I tried adapting configuration per build, but I figured it is unachievable with appveyor. This is why I settled with using `RelWithDebInfo` everywhere. The only disadvantage is assertions being disabled for nightlies.

Manuel-K:
I usually use `RelWithDebInfo` for testing. Too many weird bugs are caused by a higher optimization level which makes the `Debug` builds less useful.

Manuel-K:
Should the release builds contain the data used by the unit tests?

selaux:
They are built with unittest capabilities, so they also include the data. Not sure if it is worth to remove. Building with debug info also considerably increases the size of the releases.

selaux:
Feel free to have a try at building releases with different build types. After spending a lot of time waiting for the CI and searching the web I just thought it isnt worth investigating further...

selaux:
There is one more thing I will try...

So do we want to roughly double the release sizes by including debug info?

Manuel-K:
No

selaux:
Allright new releases have been built. Setting `BUILD_TYPE` now works nicely as well. Can be merged imho.

PR #633, opened 2017-12-13
Bump version to 0.16.0 and add changelog

selaux:
And so it begins...


Manuel-K:
I've updated the changelog with the three new maintenance issues.

selaux:
Updated Changelog.

PR #636, opened 2017-12-14
Cleanup turnbased

Manuel-K:
Remove `TURNBASED` checks.

Might be incomplete and have errors.

Went through the code once while waiting for dinner.


chmallon:
The check in `HandleMoveModeInteractiveClick()` was always false in vanilla.
With this change this is not the case anymore.

Manuel-K:
Fixed.

lynxlynxlynx:
Is this still WIP? Besides whitespace, it looks ok, but I haven't checked if you missed any remaining uses.

Manuel-K:
Good question. I remember `grep`ing for missed instances. I'll re-check tomorrow.

Manuel-K:
```
[src]$ grep TURNBASED -r --color .
./game/Tactical/Overhead_Types.h:#define TURNBASED                                      0x000000004
./game/Tactical/Overhead.cc:    t.uiFlags                 = TURNBASED;
./game/TacticalAI/AIList.cc: * (in TURNBASED COMBAT) keeps track of which AI guy should
```

Seems to be gone.

lynxlynxlynx:
So why the first two matches, savegame compatibility?

Manuel-K:
That was the initial idea, but I'm not sure if I checked if it gets saved.

Manuel-K:
Yes, it does get saved.

`INJ_U32(d, s->uiFlags)` in `InjectTacticalStatusTypeIntoFile(HWFILE const f)` in `src/game/LoadSaveTacticalStatusType.cc`.

lynxlynxlynx:
I guess it can't hurt and could be handy for bisecting with newer saves. Didn't check if the Overhead setting happens before loading a game, but it's likely.

Manuel-K:
We could inline it, e.g.:
`t.uiFlags                 = 0x000000004: // == TURNBASED for save game compatibility`

lynxlynxlynx:
Don't care either way. Ready for merge if you ask me.

lynxlynxlynx:
So ... want to do the same for REALTIME?

Manuel-K:
It seems that REALTIME is never set, correct? It is unset once in `game/Editor/EditScreen.cc`.

I've `grep -G`ed  for `uiFlags.*REALTIME` and for `uiFlags.*2`.

lynxlynxlynx:
Yes, useless. I was checking via:

> ggn =.*REALTIME[^_]
> ggn =.*IN_RT_COMBAT

Manuel-K:
I've already removed IN_RT_COMBAT because it wasn't used anywhere.

PR #658, opened 2018-01-03
Replace boost::shared_ptr with std::shared_ptr

selaux:


lynxlynxlynx:
beautiful. :+1:

janisozaur:
Missed one:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/a746bc61d9cc454fb670c84aac6b11e60bc2fab0/dependencies/lib-boost/boost/filesystem/operations.hpp#L31

selaux:
I cannot remove the headers though as they are referenced internally by boost...

Issue #650, opened 2017-12-22
Game is caught in infinite stopwatch


JAsmine-ja2:
My current playtest with DiD mode ended up in an infinite stopwatch. I was talking to Michy in the Bar in San Mona (C5). At the same time, I was sending some mercs around the map to investigate/loot the buildings Micky's speech was interrupted by a query for a locked door. Now, it is impossible to continue this game. Whatever happened, it is totally game-breaking.

[SaveGame11.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1583464/SaveGame11.zip)

lynxlynxlynx:
This sounds like something that might have been broken from the start. Is it easy for you to reproduce?

RinneT:
This bug was actually the reason why I implemented the backup function in the first place.
IIRC it would always happen when you talk to someone and another merc causes a popup screen in the meantime.
However I can't reproduce it anymore.

I'll see if I can at least put in a check for the stopwatch.

JAsmine-ja2:
How does the bbackup function work again? ^^'

RinneT:
It creates a "Backup" Folder in the savegames folder. That one contains the last two saves as savegamename.sav.1 and savegamename.sav.2
Copy them to the root SavedGames folder and remove the .1 or .2

Problem could be if you loaded and saved the game multiple times, it will overwrite them with the stopwatch game. Good idea to rename the backup folder to make sure they aren't overwritten.

lynxlynxlynx:
So now part of this is fixed, thanks Rinne!

lynxlynxlynx:
Closing as a dupe of #427.

PR #661, opened 2018-01-03
#650: Prevent save during stopwatch

RinneT:


RinneT:
That was extremely depressing. Now my repo is clean and let's hope I don't mess it up again.
Thanks for all the help.

RinneT:
Btw. this is not a complete fix for #650 
It just circumvents the nastiest effect in Dead is Dead mode.
It can still happen that an infinite stopwatch is triggered immediately after loading, or in other game modes.

PR #669, opened 2018-01-05
Publish AppVeyor-built packages as artifacts

janisozaur:


lynxlynxlynx:
So, what's the rationale?

janisozaur:
So that it's easier to find the prebuilt binaries for each of the commits – and to actually provide them for each build in the first place.

janisozaur:
Travis failed with:

> thread 'tests::build_engine_options_from_env_and_args_should_overwrite_json_with_command_line_args' panicked at 'called `Result::unwrap()` on an `Err` value: "Vanilla data directory has to be set either in config file or per command line switch"', /checkout/src/libcore/result.rs:906:4

This has nothing to do with this PR.

lynxlynxlynx:
Doesn't appveyor keep them only for a short time?

lynxlynxlynx:
Interesting!

I'm not convinced this is needed though. We save both pull request builds (meaning all pushes) and have nightlies on top of that, so it's really just about discoverability.
http://builds-ja2.stefanlau.com/pull-requests/

janisozaur:
> We save both pull request builds

No you don't.

If I made a pull request, the build wouldn't be available anywhere. Point in case, check this PR giving 404: http://builds-ja2.stefanlau.com/pull-requests/669/

lynxlynxlynx:
I guess it's limited to organisation members. Or maybe it happens only after merging.

Manuel-K:
> I guess it's limited to organisation members.

Correct. For security reasons.

janisozaur:
> I guess it's limited to organisation members.

Yes, although somewhat indirectly. The private keys required to decrypt secure env vars are only available when the job is scheduled on behalf of `origin` and not any of the external PRs.

The way I see you usually do PRs is to create a branch in the `origin` and also create a PR back into `origin`. That means you use double the resources on a public, free and contested resources the CIs are: one build for branch push and one build for PR. As a user of those same services I would urge you to stop doing that, especially when it gives you no benefit whatsoever.

E.g. you could configure your CIs to not build branches other than `master` or you could follow the most common practice of not abusing the main repository and doing PRs from your own forks instead.

Back to the topic of benefits of this PR:
* Right now there is no way for me (external contributor) to create a build with MSVC for reasons stated above
* Likewise, if I enable AppVeyor for my own fork, the builds won't ever get deployed
* I haven't seen how your solution handles multiple commits on the same PR, but I imagine they simple get overwritten. With this you can easily associate a build with particular commit (build job) that created it.

lynxlynxlynx:
I have no problem with this, but since @selaux wrote the whole thing, I'd like to hear his opinion before merging.

The double builds have meanwhile been fixed/removed.

Manuel-K:
I like the current way because we do not provide builds that have not been checked by one of us.

janisozaur:
@Manuel-K there's nothing in the way of having them both, although _I_ see little reason to manually copy the builds from PRs.

@lynxlynxlynx I personally would have liked if you chose the latter, more common approach of submitting PRs from your own forks.

The reasoning is simple: if anyone wants to improve the way you distribute prebuilts, they can do so in their own AppVeyor project or likewise, they can prove their work by submitting PR, which doesn't need to be a branch on this particular repository to be able to be inspected.

Manuel-K:
@janisozaur May I quote your own words?

>All that said I still miss your point – or perhaps you're missing mine: I don't care, as long as you and maintainers can compile, all is fine.

selaux:
From my side I dont see any harm in publishing the artifacts via appveyor as well. Still good to keep the ones uploaded via ftp as well, because imho they are way easier to find...

PR #672, opened 2018-01-06
[WIP] Launcher: Locate game.json

Manuel-K:
The start of the refactoring that is required to display the location of `game.json`.

I've started to clean my git stash. The changes to the launcher are still messy and depend on copy and pasted rust code.


flaviojs:
#833 shows the location of `game.json`, maybe this PR is no longer needed?

Technically the current code can read game.json from more locations (see `openGameResForReading`) but both PRs agree that `externalized/game.json` should either be in the EXTRA_DATA_DIR or next to the executable.

lynxlynxlynx:
Yes, I think this one has been superseded. I don't like the paths personally, since EXTRA_DATA_DIR will be undefined for most, so we'll default to the binary location. Which is not handy for installed or built-only versions: first has permission implications, while both suffer from losing the settings on a reinstall/rebuild.

So, I think we should first check if the file is in the home settings dir, alongside the saves and ja2.json.

flaviojs:
That seems like a discussion that should happen in a dedicated issue.
Searching led me to #707, which seems appropriate, can you copy your thoughts there?

lynxlynxlynx:
Sure, nice find.

PR #697, opened 2018-01-28
CMake: do not add the directory to the sources list

Manuel-K:
QtCreator displays the directory twice in the projects view if it is included
in the sources list. The working entry in the tree is generated automatically
by the inclusion of the source files. The second broken entry is caused by the
inclusion of the directory.


Manuel-K:
Before the change:
glob_master
After the change:
glob_fixed

lynxlynxlynx:
lgtm, I guess cmake has recursive globs; I don't use any ide to check if it has problems too before or after.

Manuel-K:
Recursive globs are a separate matter:

```
file(GLOB variable [RELATIVE path] [globbing expressions]...)
file(GLOB_RECURSE variable [RELATIVE path]
     [FOLLOW_SYMLINKS] [globbing expressions]...)
```
from https://cmake.org/cmake/help/v3.0/command/file.html

It seems that the `${CMAKE_CURRENT_SOURCE_DIR}` was intended as the optional `[RELATIVE path]` but it gets misinterpreted as a second globbing expression.

lynxlynxlynx:
Yeah, keywords aren't optional — if that was the case, RELATIVE should have been added before the dir.

PR #703, opened 2018-02-02
BUILD: fix install paths by utilizing GNUInstallDirs

hasufell:
This fixes proper distro support on linux. Paths
must be more configurable to accomodate various
filesystem layouts. GNUInstallDirs is the defacto
standard for that.


CMakeLists.txt
```diff
@@ -36,8 +41,16 @@ if (NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Choose the type of build, options are: Debug RelWithDebInfo." FORCE)
 endif()
 set(VERSION_TAG "" CACHE STRING "Build date for nightly versioning")
-set(EXTRA_DATA_DIR "" CACHE STRING "Directory for externalized data")
-set(INSTALL_LIB_DIR lib CACHE PATH "Installation directory for libraries")
+if (UNIX AND NOT MINGW AND NOT APPLE)
```
lynxlynxlynx: you can just check for ${CMAKE_INSTALL_FULL_DATADIR} to make it cleaner.

CMakeLists.txt
```diff
@@ -36,8 +41,16 @@ if (NOT CMAKE_BUILD_TYPE)
...
+	set(EXTRA_DATA_DIR "${CMAKE_INSTALL_FULL_DATADIR}/ja2" CACHE STRING "Directory for externalized data (usually just set CMAKE_INSTALL_DATADIR instead)")
+else()
+	set(EXTRA_DATA_DIR "" CACHE STRING "Directory for externalized data")
+endif()
+if (UNIX AND NOT MINGW AND NOT APPLE)
+	set(INSTALL_LIB_DIR ${CMAKE_INSTALL_FULL_LIBDIR})
```
lynxlynxlynx: why not cached?

CMakeLists.txt
```diff
@@ -316,22 +329,18 @@ set(CPACK_PACKAGE_FILE_NAME "${PROJECT_NAME}_${CPACK_PACKAGE_VERSION}_${PACKAGE_
...
+	install(TARGETS ${JA2_BINARY} RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
     if(BUILD_LAUNCHER)
-        install(TARGETS ${LAUNCHER_BINARY} RUNTIME DESTINATION bin)
+        install(TARGETS ${LAUNCHER_BINARY} RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
     endif()
-    install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/assets/externalized assets/mods assets/unittests DESTINATION share/ja2)
-    install(FILES ${STRACCIATELLA_SHARED_LIB} DESTINATION ${INSTALL_LIB_DIR})
-    install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/assets/distr-files-linux/ja2-stracciatella.desktop DESTINATION share/applications)
+	install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/assets/externalized assets/mods assets/unittests DESTINATION ${CMAKE_INSTALL_DATADIR}/ja2)
+    install(FILES ${STRACCIATELLA_SHARED_LIB} DESTINATION ${CMAKE_INSTALL_LIBDIR})
```
lynxlynxlynx: if you won't be using INSTALL_LIB_DIR, why even set it?

CMakeLists.txt
```diff
@@ -36,8 +41,16 @@ if (NOT CMAKE_BUILD_TYPE)
     set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "Choose the type of build, options are: Debug RelWithDebInfo." FORCE)
 endif()
 set(VERSION_TAG "" CACHE STRING "Build date for nightly versioning")
-set(EXTRA_DATA_DIR "" CACHE STRING "Directory for externalized data")
```
lynxlynxlynx: The old default in cmake may be "", but in code we then assume it's in the bin path — if the path is unset. This now changes ...

CMakeLists.txt
hasufell: It is used, see https://github.com/hasufell/ja2-stracciatella/blob/f4635ad2cf67a2b5611c09a1f1c8b44a1e5d0327/CMakeLists.txt#L80

CMakeLists.txt
hasufell: Because it's bad to be able to set both variables. That may accidentially lead to a broken setup. It's best to hide the internals as much as possible.

CMakeLists.txt
hasufell: I don't consider that cleaner. Especially not from a readability point of view.

CMakeLists.txt
hasufell: Yes, but only for linux and it's documented in the README.

CMakeLists.txt
lynxlynxlynx: It's inconsistent, please make it always or never cached.

CMakeLists.txt
lynxlynxlynx: In retrospect it's not clear if that's needed, since none of our code uses it. But more importantly, anyone relying on overriding INSTALL_LIB_DIR will now get files in a potentially different path, since there is no graceful migration set up.

CMakeLists.txt
lynxlynxlynx: Also for BSDs. But the main question is basically whether this still works for running the game from the build dir without setting EXTRA_DATA_DIR. The previous default was to look in the cwd and now it's a different path.

CMakeLists.txt
hasufell: I don't see a problem with that since INSTALL_LIB_DIR is not documented here: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/COMPILATION.md#additional-options

CMakeLists.txt
hasufell: I'm fine with making it never cached, but I don't know the impact on non-linux systems.

CMakeLists.txt
hasufell: > But the main question is basically whether this still works for running the game from the build dir without setting EXTRA_DATA_DIR. The previous default was to look in the cwd and now it's a different path.

Yes. It can still be set to empty so it looks in cwd. The default changed on linux and is documented in COMPILATION.md. The documentation could be expanded to explain that an empty variable means that cwd is checked (which it currently doesn't explain).

lynxlynxlynx: You miss the point. If I understand the change correctly, there is now an extra thing to remember when compiling — most of the developers are on linux and there is no need to pass any cmake options now. Changing that just makes contributing harder. It's the packagers that are the special case here.

hasufell: Well, the way it is now is an utter mess. If you want to make contributing easier, you better introduce a developer build option or utilize [CMAKE_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html) properly instead of having people micro-manage specific variables.

lynxlynxlynx: There's no requirement for micromanagement at all.

hasufell: Ok, we obviously won't agree on anything. The build system is in a state where it cannot be used by linux users or packagers properly, since it doesn't adhere to any standards. It hardcodes paths on cmale level and has broken path handling on code level. Good luck!

Issue #475, opened 2017-01-04
Glitchy opening of locked doors


JAsmine-ja2:
I stumbled upon this glitch/trick in the BP forum: http://thepit.ja-galaxy-forum.com/index.php?t=msg&th=23271&goto=348045&#msg_348045

Basically, you supposedly can open locked doors just by spamming "kick open". It doesn't matter, if your merc is strong enough and you can still kick the door, after the merc has collapsed from being out of energy.

Naturally, I had to test it immediately - **and it works.** I was able to kick open the locked door at the Drassen SAM site with Raven. Now, two problems are present here:

1) Raven with STR=68 shouldn't be able to open the door. It has a lock with lock ID 5 = metal door, which has a smashing difficulty of 80. 

2) (more importantly) A merc without any energy, who just collapsed, shouldn't be able to just get up and kick a door.

![door-test](https://cloud.githubusercontent.com/assets/17213795/21644820/b75c9502-d28f-11e6-9940-29caa2660b8d.jpg)

How to reproduce:
- position the cursor so that the hand cursor appears over the locked door (without pressing ctrl) and so that you can just left-click two times to use the "kick" option without having to move the cursor. There are only a few pixels where this works, but it does work. This makes the following process much easier.
- Now just spam "kick open".
- Raven complains, complains and complains even more until she eventually collapses. Everything okay so far. Keep spamming!!
- She gets up with 0 energy (!) and keeps kicking the door.
- Eventually (took me about 10-20 more kicks, the door opens !!!

The whole process should take less than 2 minutes.

Savegame:
[SaveGame10.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/684827/SaveGame10.zip)

lynxlynxlynx:
This is handled in AttemptToSmashDoor. Also lock damage comes into play, but more importantly, some randomness. So really, this could all be explained by just giving it enough tries.

JAsmine-ja2:
This might explain issue no.1 - being able to smash a door with insufficent strength seems not okay in an RPG-ish environment, but it can very well be an intended feature and not an oversight by the original devs.

But still the thing with being able to get up and keep on kicking the door although being totally out of energy remains. This is definitely a bug.

JAsmine-ja2:
Some more testing:
[...]

PR #705, opened 2018-02-15
Collapsing mercs can't kick doors anymore

### LINKED ISSUE #476: Collapsing merc can fall through closed door
While investigantin #475, I discovered this glitch.

When collapsing in front of a door, the merc automatically turns to a "free" direction to collapse. This is normal behavior and works as intended:
thrudoor1

BUT, if I quickly order the merc to smash the door while he is collapsing, he ends up falling through the door:
thrudoor2

This is 100% reproducible. It has no negative impact on gameplay, because the merc remains on the side of the door where he was before. It just looks weird.

Savegame:
[SaveGame10.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/685537/SaveGame10.zip)

daviel:
Fixes #476 
In the enoughPoints function there is now a check whether the merc is collapsed in realtime-mode. Weird thing about this function is that there is a sBPCost(BreathPoints) argument which isn't used at all.

### DIFF
--- a/src/game/Tactical/Points.cc
+++ b/src/game/Tactical/Points.cc
@@ -464,6 +464,8 @@ BOOLEAN EnoughPoints(const SOLDIERTYPE* pSoldier, INT16 sAPCost, INT16 sBPCost, 
 	if (!(gTacticalStatus.uiFlags & INCOMBAT))
 	{
 		sAPCost = 0;
+		if ( pSoldier->bCollapsed )
+		  return( FALSE );
 	}
 
 	// Get New points


lynxlynxlynx:
BPs were unused in this function already at the time of import, so the parameter can go.

daviel:
I think there needs to be another function for calculating bp costs. It's weird that you can kick doors with 0 BPs or climb roofs and collapsing on top/bottom... But as BP calculation is non-trivial see DeductPoints and AdjustBreathPts in Points.cc we should figure out which actions are depending on BP, realtime/non-realtime, etc. and whether it's worth it for gameplay-wise.

JAsmine-ja2:
Does this also fix this part of #475 ?

> This time, I created an IMP with STR=35. She shouldn't be able to ever open the locked door (again in D15). At least, if I understand SkillCheck.cc correctly (which might very well NOT be the case ;). It says that if a chance to succeed in smashing is <30%, then it should NEVER work.
> However, after several kicks, I was able to open the locked door again.

selaux:
@JAsmine-ja2 I dont think so, I specifically checks for the merc being collapsed, not for strength.

flaviojs:
Can a collapsed soldier do anything during combat?
If not then the that check should move out of the "in realtime" block.

lynxlynxlynx:
Moved to #1100, closing.

Issue #65, opened 2013-08-22
[RU] Game crashing when hiring when browsing M.E.R.C.


lynxlynxlynx:
Originally reported by: **Anonymous**

---

To reproduce:
1. Enter M.E.R.C.
2. Listen to founder's monologue and create an account
3. Start browsing mercenaries - game crashses when showing merc's stats

Always reproductable, in all versions of ja2 stracciatella with russian localisation. The game crashes without restoring original screen resolution.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/65

derSteFfi:
Also can't reproduce. Loaded the savegame, surfed MERC extensively and even hired some MERCs. No crash. Can anyone confirm?

lynxlynxlynx:
Probably requires someone with 0.12.352 Russian localization.

lynxlynxlynx:
Another Russian tester had no problems with this, closing.

PR #706, opened 2018-02-23
Allow to load any map version with RUSSIAN resources

RoddyAtHat:
Hello.
I did a try to start with mod 'from-russia-with-love' but game surprised me with runtime error in the process of loading map. Soon I realized that russian version is restricted to maps with v6.0. Map A9 from mod has v5.0.
From the other side, I think non-russian versions will fail to load v6.0. So, I made map loading compatibility patch - any map version will be loaded despite language version.

### DIFF
--- a/src/game/TileEngine/WorldDef.cc
+++ b/src/game/TileEngine/WorldDef.cc
@@ -2037,11 +2037,6 @@ try
 	FLOAT dMajorMapVersion;
 	FileRead(f, &dMajorMapVersion, sizeof(dMajorMapVersion));
 
-	if(isRussianVersion() && (dMajorMapVersion != 6.00))
-	{
-		throw std::runtime_error("Incompatible major map version");
-	}
-
 	UINT8 ubMinorMapVersion;
 	if (dMajorMapVersion >= 4.00)
 	{
@@ -2281,7 +2276,7 @@ try
 		}
 	}
 
-	if(isRussianVersion())
+	if (dMajorMapVersion == 6.00)
 	{
 		FileSeek(f, 148, FILE_SEEK_FROM_CURRENT);
 	}


lynxlynxlynx:
Thanks! This needs a thorough review, as it could break a lot. 

Btw, we are thinking of removing that mod, since generous-rebels supersedes it and they're practically identical. Can you try if this one has the expected area version?

Also, since we rarely hear from people with the Russian game, can you check #65 ?

RoddyAtHat:
> Can you try if this one has the expected area version?

I looked in hex-editor - the map from generous-rebels has the same version as from-russia-with-love. So i'm almost confident that it will fail for me too. Anyway, I'll check it in game to be sure.

> can you check #65 ?

Yes, I'll try to reproduce.

RoddyAtHat:
> Anyway, I'll check it in game to be sure.

Well, tested in game - the map does not load, runtime error again.

> Yes, I'll try to reproduce.

I couldn't reproduce. It just works, no crash while browsing MERCs stats. I was able to hire MERCs too.

lynxlynxlynx:
Thanks! Do you know if the file format versions are documented somewhere?

RoddyAtHat:
No, I haven't met map format documented on network. I think, the most accurate map format description is the source code.

flaviojs:
Those mysterious 148 bytes being skipped in WorldLoad are never written in WorldSave, so I doubt ja2-stracciatella loads properly in russian... (load->save->load skips non existing data)

My guess is that there is a flag value for that data so the FileSeek guard should look like:
`if (dMajorMapVersion >= 6.00 && (uiFlags & MAP_TODO_SAVED))`

A sample file with that mysterious data should give more info.

RoddyAtHat:
Here is map file from russian version:
[A9.DAT.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3248868/A9.DAT.zip)
Hope it helps to figure out.

flaviojs:
Compared your A9.DAT with mine (english 1.12 from GOG).
Mine also has major version 6.0 so your check is definitely wrong.
The flags are the same so they can't be used.
Mine has minor version 25 and yours has minor version 26, so you can probably guard that seek with:
`if (ubMinorMapVersion >= 26)`

`SaveWorld` writes minor version 25, so loading after save will work fine.
(in my case it will change the major version from 6.0 to 5.0, dunno if that's bad, but that's out of scope for this PR)

flaviojs:
It seems I made a mistake in in my earlier analysis.
My A9.DAT is actually 5.25.

I got the steam version now, which includes a russian version so I could finally compare different languages (and wildfire).
This PR is superseeded by #941, which was made based on what I learned.

PR #710, opened 2018-04-06
Add freenode note to README

selaux:


lynxlynxlynx:
Link to http://webchat.freenode.net/?channels=ja2-stracciatella instead, so people can directly join via the browser.

selaux:
:+1:. I also made the text less "bossy".

PR #711, opened 2018-04-07
[WIP] External project all the things

selaux:
Goal is to create a "superbuild" that builds everything using external project, so dependencies can be located via `find_package` everywhere and dont have to be included as source. Superseedes #362.


selaux:
Again, too big and too many changes. I will try to do this more incrementally...

PR #715, opened 2018-05-04
Integer scaling for pixel perfect image

simonmarcel:
Added support for scaling to the nearest integer scale that will fit your monitors resolution while using nearest neighbor scaling function. This way the image will always look crisp when being upscaled instead of the blurry look of linear interpolation. 

So 640x480 will be upscaled to 1280x960 on my full HD monitor with black bars all around but a very crisp image like it originally looked.

SDL 2 supports this out of the box but the minimum requried SDL version that does this is 2.0.5
see: 
https://wiki.libsdl.org/SDL_RenderSetIntegerScale

I made the scaling optional with the option "integer_scale" in the json config. If set to false linear interpolation is used just like before.

I have not implemented the option within the launcher though since i couldnt be arsed to download and compile fuse on my windows installation.


JAsmine-ja2:
This sounds very good! Hopefully, we can get this feature as an option into the launcher!

lynxlynxlynx:
Interesting, I haven't noticed artifacts, but it's also true I'm on a CRT. :place_of_worship:

CMakeLists.txt
```diff
@@ -38,13 +38,14 @@ endif()
...
-option(LOCAL_BOOST_LIB "Build with local boost lib" OFF)
+set(LOCAL_SDL_LIB "dependencies/lib-SDL2-2.0.8-VC" CACHE STRING "Use local SDL library from this directory")
+set(LIBSTRACCIATELLA_TARGET "x86_64-pc-windows-msvc" CACHE STRING "Rust target architecture for libstracciatella")
+
+option(LOCAL_FLTK_LIB "Build with fltk lib" ON)
+option(LOCAL_BOOST_LIB "Build with local boost lib" ON)
 option(LOCAL_RAPIDJSON_LIB "Build with local rapidjson lib" ON)
 option(LOCAL_GTEST_LIB "Build with local gtest lib" ON)
-option(WITH_UNITTESTS "Build with unittests" ON)
+option(WITH_UNITTESTS "Build with unittests" OFF)
```
lynxlynxlynx: All but the first change look like mistakes. Options are cmake configure-time switches — why change the defaults?

CMakeLists.txt
```diff
@@ -59,6 +60,7 @@ if (NOT (VERSION_TAG STREQUAL ""))
 endif()
 
 add_definitions(-DGAME_VERSION="\\"${GAME_VERSION}\\"")
+add_definitions("-DBOOST_ALL_NO_LIB")
```
lynxlynxlynx: Why is this needed?

CMakeSettings.json
```diff
@@ -0,0 +1,49 @@
...
+      "configurationType": "RelWithDebInfo",
+      "inheritEnvironments": [ "msvc_x64_x64" ],
+      "buildRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\build\\${name}",
+      "installRoot": "${env.USERPROFILE}\\CMakeBuilds\\${workspaceHash}\\install\\${name}",
+      "cmakeCommandArgs": "",
+      "buildCommandArgs": "-v",
+      "ctestCommandArgs": ""
+    }
+  ]
+}
```
lynxlynxlynx: isn't this something cmake generates for you and it shouldn't be tracked (better put it into .gitignore)?

rust/src/stracciatella.rs
```diff
@@ -544,6 +547,11 @@ pub fn should_start_in_fullscreen(ptr: *const EngineOptions) -> bool {
     unsafe_from_ptr!(ptr).start_in_fullscreen
 }
 
+#[no_mangle]
+pub fn should_use_integer_scaling(ptr: *const EngineOptions) -> bool {
+	unsafe_from_ptr!(ptr).use_integer_scaling
+}
+
```
lynxlynxlynx: The indentation of new code doesn't match the rest of this file.

src/sgp/Video.cc
```diff
@@ -99,24 +100,28 @@ void VideoSetFullScreen(const BOOLEAN enable)
 {
 	if (enable)
 	{
-		g_window_flags |= SDL_WINDOW_FULLSCREEN;
+		g_window_flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
 	}
 	else
 	{
-		g_window_flags &= ~SDL_WINDOW_FULLSCREEN;
+		g_window_flags &= ~SDL_WINDOW_FULLSCREEN_DESKTOP;
```
lynxlynxlynx: Why the change of type? If the nearest scaling requires the fake fullscreen, and can't work with the mode change, I think it would be better for that to swap out the flags and leave normal fullscreen for the rest. Two things are in my head; is it a performance hit and would it work or nicely degrade on android.

CMakeLists.txt
simonmarcel: Yes sorry, I accidently commited the chanes I did to the build configuration, to make the build work with VS 2017, which supports cmake nativly. I'll revert them.

CMakeLists.txt
simonmarcel: Same here, see above.

CMakeSettings.json
simonmarcel: It's generated by Visual studio's cmake tools to configure the build. Should probably be on the ignore list though.

src/sgp/Video.cc
simonmarcel: The renderer needs to be set to the Desktop resolution to let SDL do the scaling. If we don't set integer scaling, SDL still scales up the image to fullscreen with linear interpolation (like the monitor would otherwise), so I did not see any reason to not use SDL_WINDOW_FULLSCREEN_DESKTOP all the time, so to keep the code simple. It then work similar to the fix in issue #62, when the window is resized, only for fullscreen this time. 
I guess it would not be a performance problem on anything that is not ancient. 

Although as a CRT user you would probably be better served by SDL_WINDOW_FULLSCREEN. So I could change it as proposed.

simonmarcel:
Well it is different with a CRT of course. It only affects LCD's, because they must do linear interpolation for anything that is not their native resolution. It's not really rendering artifacts but a generally blurry, smoothed out look that is created by the linear interpolation. Some even prever it this way. But I prefer my old games to look authentic and I like the more blocky look, especially with games like JA2 that make heavy use of sprites and pixel art.

janisozaur:
Hi, just browsing through the code, I have some suggestions
* you could separate the SDL update into its own commit, so that it's easier to spot actual changes. Why is there an external library bundled in here in the first place is beyond my comprehension.
* In OpenRCT2 we used (I think) a similar approach, but relied on integer scales being detected automatically (it was implemented prior to SDL 2.0.5), if I understand the merit of this PR. Then we had a PR (https://github.com/OpenRCT2/OpenRCT2/pull/6588) that implemented "smoothed NN", which first upscaled to `⌈scale⌉` then downscaled to exact `scale`

simonmarcel:
The external library was in the repo previously for mac and windows. I guess to have an out of the box compile under these platforms. I have no idea how i can change the commit retroactivly though.

The way I implemented integer scaling was the most easy to accomplish as it is effectively a one liner if you ignore the configuration switch in the json. What you mention from OpenRTC is actually not Pixel perfect. It blows the picture up beyond the display resolution and scales it down. It creates a sharper image than straight upscaling with linear interpolation but not a pixel perfect image of the original pixel art. It's better though as straight upscaling and would probably also be a great option to have. 
The pixel perfect patch for dosbox does basically the same with its "near pixel perfect" mode (https://www.vogons.org/viewtopic.php?f=41&t=49160)

selaux:
@janisozaur SDL2 is bundled for Windows and macOS builds. This should not be necessary anymore once I can finally finish https://github.com/ja2-stracciatella/ja2-stracciatella/pull/711, which should download these dependencies at build time. I would be interested how you handle this with openrtc though...

janisozaur:
> I would be interested how you handle this with openrtc though

Sure. Please excuse the rather-lengthy post, intercepting the discussion here is not my purpose. (It's Open**RCT** though)

For platforms with sane package managers (aka Linux: Ubuntu and Arch) we use them. All the libraries get found via `cmake`+`pkg-config` duo. Whilst it would be cleaner to use `find_package`, the approach we have now is good enough and the PR addressing this shortcoming hasn't received any work lately. Installing the packages in each job is slooooow and Travis CI offers mostly obsolete software stacks, for those reasons I've migrated the environment setup into docker and we use Docker Hub to produce an image that we simply pull into relevant job. We test with 4 configs: Arch (for the most current packages and my current distro of choice), Ubuntu AMD64 and i686 (still Xenial, move to Bionic is a bit overdue, but coming; we have some explicit need for i686 tests), Fedora (for it's mingw packages; it used to be Arch, until Arch decided to drop them). This has the drawback we cannot add and use a package in a single PR/commit, but that's rarely an issue.

For Windows we use AppVeyor with MSVC, there is a MSVC solution in the repository, which is the recommended way to use the project by our Windows people. We use vcpkg for compiling the packages and have another repository that handles most of that: https://github.com/OpenRCT2/Dependencies it has a simple script that pulls latest packages versions, compiles them, puts into an archive and we do a release, with precompiled binaries. vcpkg integrates well with cmake as well as MSVC's native solutions and as somewhat heavy consumers, we've submitted many fixes to vcpkg itself or some of the packages. Right now our cmake setup should handle MSVC, but it's not really tested thoroughly.

For macOS we have _yet another_ project ~~file~~ directory, Xcode's `OpenRCT2.xcodeproj` and relevant `pbxproj` and it's the platform I have least knowledge of, but I believe the process is similar to the one on Windows: we use prebuilt packages based on `homebrew`. cmake handles that platform well.

To sum up: where possible, we use native package managers. Where we can't we use our own pre-built binaries using platform's common tools, which live outside of the main repo and we actively maintain state of those packages. This particular process suits _us_, much more active project; I'm also sure it can be further improved upon.

src/sgp/Video.cc
lynxlynxlynx: For me it probably makes no difference or is a slight improvement, in case a crash somehow prevents proper video mode restoration. All I could find on the net were reports of multiscreen issues, but for both modes, so meh.

So unless @selaux, @janisozaur or anyone else has any further complaints, this can stay. It would be better in a separate commit though.

simonmarcel:
Did some rebasing and split the commits as requested.

rust/src/stracciatella.rs
simonmarcel: What is the correct formating though? The readme says tab with 8 and to intend new code with tabs. But the current formatting suggests using spaces with a width of 4.

lynxlynxlynx: Yes, tabs are fine, it's this file that's using spaces erroneously. But consistency is always better. There's ongoing work to harmonize everything to tabs (eg. #556 ).

simonmarcel:
Played around a bit more with different scaling methods and implemented the smoothed NN method of scaling that @janisozaur mentioned, calling it "near perfect" scaling. Attached 3 different screenshots of the game running in its original 640x480 resolution upscaled with the different scaling techniques to fullhd, showing the differences between straight linear upscaling, upscaling with NN and then downscaling (smoothed NN, near perfect) and pixel perfect straight upscaling to the nearest integer fit.
Make sure to look at the pictures unscaled. 

It's not completly ready yet, as the scale value for near perfect scaling is currently hardcoded.

laptop view linear
laptop view near perfect
laptop view pixel perfect

lynxlynxlynx:
I'm not sure if it's worth the time to unhardcode that 4. Why would anyone want to change it? If it's for performance reasons, then near perfect mode is already a bad choice.

simonmarcel:
Well, for the scaling technique to work we have to first upscale to a resolution that is at least larger than the monitor and then scale down from there. Because linear interpolation creates a less blurred picture that way as it is easier to remove information than adding it.

lynxlynxlynx:
Sure, but why wouldn't 4 be a universally good choice for upscaling?

lynxlynxlynx:
I've restarted the autobot builds; windows doesn't find any sdl anymore (??). Travis builds are fine in that regard (use old as instructed), but it shows two errors:
- the test write_engine_options_should_write_a_pretty_json_file needs to be adapted for the renamed variable
- an older SDL is not handled gracefully or forced to use the local copy:
 ```
src/sgp/Video.cc:175:3: error: use of undeclared identifier 'SDL_RenderSetIntegerScale'
                SDL_RenderSetIntegerScale(GameRenderer, SDL_TRUE);
```

simonmarcel:
The SDL_RenderSetIntegerScale has a minimum version requirement of 2.0.5 (https://wiki.libsdl.org/SDL_RenderSetIntegerScale)

I updated the inlcluded version for MSVC to 2.0.8 because that was what I was working with. Mac and mingw would also need to be updated accordingly. SDL for the Linux build is taken from the local system i suppose?

I think I could also implement the scaling variant without the SDL_RenderSetIntegerScale function. But would need some additional time.

lynxlynxlynx:
It's enough to ifdef it with a sdl version check and let the option be a noop for noncompliant systems. Eventually they'll catch up, it's not a major release to warrant a lot of waiting.

simonmarcel:
Changed the input field type for the scaling selection as it makes more sense to have a non editable input field. Also allows a more fitting description to be put in place. Feel free to suggest changes to how the scaling modes should be named.

How do I make an #ifdef based on the SDL version as you suggested? I have no idea how to do that.

lynxlynxlynx:
`#if SDL_VERSION_ATLEAST(2,0,8)`

simonmarcel:
msvc and mingw builds had missing libs because the files are on gitignore and were not commited. I have no idea why Mac does not build. I did not saw anything missing in the SDL folder and I have no Mac to test this on.

lynxlynxlynx:
@selaux any thoughts? It fails already during FindSDL2.cmake.

selaux:
Looking at the differences between this branch and master, it seems like the structure of the new `SDL2.framework` is different than the old one. More specifically, what was previously a symlink (e.g. `Headers` directory) is now somehow a text file. Not sure if this is an extraction issue or it is just like that in the distributed version.

selaux:
Files instead of symlinks is actually what I get as well when I extract the `.dmg` file using 7zip. Not sure how it behaves on macOS though. Easy solution would be to just replace the files with symlinks manually...

simonmarcel:
Borrowed a macbook and it was indeed an extraction issue. Opening the dmg on mac and commiting the files from there resolved the issue.

lynxlynxlynx:
All green, many thanks!

selaux:
Nice work, thx for the PR! :tada:

BardBun:
Hello, as this option is now available, I would like to request another needed function for it:

When choosing the pixel perfect option, please add another dropdown menu to adjust the initial starting window size: 1x, 2x, 3x, 4x, 5x... with 2x (1280x960) being the default pre-selected option, and 1x and the other options being selectable.

As currently, you have to always manually resize the window for it to jump to 1280x960.


This is of course not a problem in fullscreen mode, but when you play in windowed mode it can be really annoying, especially when it leaves a few pixels of black bars around the game.

lynxlynxlynx:
Why not just use the maximize window button?

PR #720, opened 2018-07-24
fix possible division by zero / SIGFPE

aellwein:
A SIGFPE can happen if cycling through items in Alt+w cheat mode. With this change the exception is gone.

### DIFF
--- a/src/game/Tactical/Interface_Utils.cc
+++ b/src/game/Tactical/Interface_Utils.cc
@@ -214,7 +214,7 @@ void DrawItemUIBarEx(OBJECTTYPE const& o, const UINT8 ubStatus, const INT16 x, c
 	}
 	else if (item->isAmmo())
 	{
-		value = 100 * o.ubShotsLeft[ubStatus] / item->asAmmo()->capacity;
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Thanks. Please also address the other two places, since you're at it:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e4faa6ae4a3567dc91a708c456cd4c210eabf96b/src/game/Tactical/ShopKeeper_Interface.cc#L2588
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e4faa6ae4a3567dc91a708c456cd4c210eabf96b/src/game/Tactical/World_Items.cc#L327

lynxlynxlynx:
Thanks!

That one test failure on one platform is unrelated:

> failures:
>     tests::build_engine_options_from_env_and_args_should_return_an_error_if_datadir_is_not_set
> test result: FAILED. 38 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

aellwein:
Thanks!

PR #729, opened 2018-10-07
Build MinGW on travis

selaux:


selaux:
This is finished from my side. I didn't test the resulting package yet though.

I also fixed that when the upload fails the build fails. Now it should succeed either way. Not sure what is wrong there, might be an IP blacklist or something, because it is happening reproducibly only for the linux builds.

.travis.yml
```diff
@@ -58,20 +75,28 @@ script:
...
-  - if [ "$PUBLISH_BINARY" == "true" ] && [ "$TRAVIS_OS_NAME" == "linux" ]; then
+  - if [ "$TRAVIS_MINGW" != "true" ]; then
+      make rust-unit-tests;
+    fi
+  - if [ "$TRAVIS_MINGW" != "true" ]; then
+      ./ja2 -unittests;
+    fi
+  - if [ "$TRAVIS_MINGW" != "true" ]; then
+      ./ja2-launcher -help;
+    fi
```
lynxlynxlynx: These three could be in the same block, but it's not that important. The diff looks fine, so merge away, unless you want to wait for testers/testing.

selaux: Merged them...

lynxlynxlynx: Are you sure they don't need to be and-chained (eg. make failing, launcher passing -> build passing)?

selaux: True. Fixed it on master...

PR #730, opened 2018-10-12
Build VS2017 on appveyor

selaux:
Updates gtest to most recent version.

Makes CI builds quite long unfortunately. Maybe we should decide for one version.


lynxlynxlynx:
Looks nice, but I agree, there's no need for so many windows builds. I vote for the latest toolchain on 64-bit.

lynxlynxlynx:
I think it's enough to do 32-bit builds only at release time.

lynxlynxlynx:
One last thing: if we still allow for local gtest, why remove it? Did you forget to readd the updated version?

selaux:
It is now dowloaded during the build, just like FLTK (see `dependencies/lib-gtest/CMakeLists.txt)`

selaux:
This way we hopefully end up here at some point: https://github.com/ja2-stracciatella/ja2-stracciatella/pull/711. Just in smaller incements.

lynxlynxlynx:
Ah, great!

PR #731, opened 2018-10-13
Use more readable labels in launcher

selaux:
Uses `Russian` instead of `RUSSIAN`.


lynxlynxlynx:
So the commandline versions are still uppercase, right?

Issue #119, opened 2014-02-15
Incorrect body positions on high resolution


lynxlynxlynx:
Originally reported by: **Gennady Trafimenkov (Bitbucket: [gennady](http://bitbucket.org/gennady), GitHub: [gennady](http://github.com/gennady))**

---

On high resolution positions of mercs and soldiers might not be correct.
See two attached images.  Resolution 1600x900, before_scroll.png is taken right after loading the game, after_scroll.png after scrolling a little bit up.
Position of the merc and soldiers changed for one tile.  After scrolling down they return into the original position.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/119

selaux:
I can still reproduce this in master. Actually it seems to me that not the merc is changing position, but the map. When scrolling the map suddenly a different viewport is rendering while the merc stays at the same screen position.

selaux:
I've looked into this a bit. Can anyone decipher what [CalcRenderParameters](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/Build/TileEngine/RenderWorld.cc#L5598) is doing? There is a lot of magic going on in that function and it seems to me like it could influence the rendering viewport.

lynxlynxlynx:
I've came to the same conclusion and have a feeling that the whole system could be much simpler. It shows that the initial design wasn't influenced much by OO design. Several conversions could be dropped by just keeping more state.

selaux:
I bisected the issue and it looks to me that this is a problem that is there since high-res rendering was introduced, I bisected it to this commit 3e55aa4fce91ff6eac4623b8c84b4daaf9c4f43a.

maxrd2:
This should be fixed with commit 599301d (#733). Mercs, soldiers and items are now drawn on correct positions.

Issue #225, opened 2016-02-20
[master bug] viewport/placement related bugs


lynxlynxlynx:
To limit confusion between the similar reports, I opened a new meta bug.

Here is the gist:
- [x] #171 we show one grid line (diagonal) at the bottom too many, one too much at the top
  - this potentially affects #73 too
- [x] #195 npcs at the top row can't be targetted for headshots
- [x] #125  enemies (all npcs?) can get placed on inaccessible squares
- [x] #119 glitches while scrolling

derSteFfi:
now that you put it out that way, sounds like an off-by-one in the area to draw to me?

lynxlynxlynx:
I'm investigating. Is just the viewport wrongly positioned, the accessibility mask (let's pretend  that exists) or both? I think all this came to be after a refactoring in 2013.

JAsmine-ja2:
Great to see this issue(s) getting investigated. This is imo one of the two most annoying issues with Stracciatella atm.

derSteFfi:
right now, i see two parts here:
1. We most certainly have an off-by-one somewhere (as mentioned earlier)
2. I think we have a general problem with the way we handle scaling. In Vanilla, due to the fixed resolution, the ratios of the tactical map, the squad overview, radar screen, etc. where fixed. We sclae the tactical view up, making x:y-ratio variable. This is worked around in many places, but most of the time (at least from what I saw so far) by introducing scaling factors and rounding results back to full pixels. This works for most places, but e.g. #73 and #119 could be a symptom of that. My only solution right now is to scale the tactical only with fixed x:y-ratio. We get black sections of the screen then, but we also have them where we can't scale things up (main menu, right of the radar screen, etc.)

Maybe the best way is to sort out the off-by-one first and investigate all the other issues again afterwards.

selaux:
:+1: Maybe there is some integer division happening somewhere. Can you point to the place in code where the viewport is scaled?

derSteFfi:
i saw it in RadarScreen.cc when fixing #73. During the process i grepped for 'scale' (aifair) to find some decleration and around 50 lines popped up where a vairable was named [g,f,ub,s]Scale[x,y]...

lynxlynxlynx:
There is integer division happening, for example in the part I'm looking at for the viewport bit, but it's only for rounding to a multiple of cell size. No use of parentheses, so I hope it doesn't break in release builds (the pattern is just x / y * y).

[...]

PR #733, opened 2018-10-22
Fixed tactical world rendering when scrolling to top left

maxrd2:
This patch solves #119 
Please do not merge it yet.. I need to test it properly, and check it against other possibly related issues first.

### DIFF
--- a/src/game/TileEngine/Isometric_Utils.cc
+++ b/src/game/TileEngine/Isometric_Utils.cc
@@ -125,14 +125,16 @@ UINT8 const gPurpendicularDirection[NUM_WORLD_DIRECTIONS][NUM_WORLD_DIRECTIONS] 
 
 void FromCellToScreenCoordinates( INT16 sCellX, INT16 sCellY, INT16 *psScreenX, INT16 *psScreenY )
 {
-	*psScreenX = ( 2 * sCellX ) - ( 2 * sCellY );
-	*psScreenY = sCellX + sCellY;
+	// cartesian to isometric
+	*psScreenX = (sCellX - sCellY) * HALF_TILE_WIDTH;
+	*psScreenY = (sCellX + sCellY) * HALF_TILE_HEIGHT;
 }
 
 void FromScreenToCellCoordinates( INT16 sScreenX, INT16 sScreenY, INT16 *psCellX, INT16 *psCellY )
 {
-	*psCellX = ( ( sScreenX + ( 2 * sScreenY ) ) / 4 );
-	*psCellY = ( ( 2 * sScreenY ) - sScreenX ) / 4;
+	// isometric to cartesian
+	*psCellX = floor((FLOAT(sScreenX) / HALF_TILE_WIDTH + FLOAT(sScreenY) / HALF_TILE_HEIGHT) / 2.0f);
+	*psCellY = floor((FLOAT(sScreenY) / HALF_TILE_HEIGHT - FLOAT(sScreenX) / HALF_TILE_WIDTH) / 2.0f);
 }
 
 // These two functions take into account that our world is projected and attached
--- a/src/game/TileEngine/RenderWorld.cc
+++ b/src/game/TileEngine/RenderWorld.cc
@@ -165,6 +165,9 @@ static INT16 gsStartPointY_M;
 static INT16 gsEndXS;
 static INT16 gsEndYS;
 // LARGER OFFSET VERSION FOR GIVEN LAYERS
+// NOTE: Larger viewport offset values are used for static world surface rendering. That surface is blitted
+//       during scrolling to speed up rendering and make scrolling smoother.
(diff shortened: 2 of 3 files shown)


maxrd2:
Think it's all good for merging. Should i make separate PR for each commit or is it ok like this?
It should fix all frustrating issues referenced in #225

selaux:
Thanks for the extensive PR :+1:. I think we can keep all this stuff in a single PR. I would like to do some testing of these changes myself, but I don't know when I will get around to do this though.

selaux:
Allright I got around to a little testing and noticed some issues (tested the starting sector `A9` in `1280x720` resolution)

- The green radar rectangle changes size by probably one pixel when scrolling to the top of the map as well as when scrolling to the bottom
- When scrolling to the top-left of the map and scrolling up and down, the rendered tiles jump by one tile at some specific scroll position

I didn't test the individual issues yet, but I think those two issues should be fixed before more in-depth testing...

maxrd2:
Scrolling the top left issue is what this patch intended to fix, and it fixed it for 1600x900 and other resolutions I've tried. It apparently broke it on 1280x720, will update it soon.

Green radar rectangle changing size is consequence of adding MIN() to rendering code so it doesn't cut the rect off. There are more tiles on tactical map shown than there are on the radar and rectangle is supposed to show exactly location on map. Will try to update the map drawing code to show pixels for all tiles that are shown on map if that's possible so this will get fixed.

maxrd2:
Have re-fixed the issue #119 on some resolutions there was a problem with integer division/rounding.
Have tested it on 640x480, 1280x720, 1600x900, 1920x1080. Am pretty sure it will be fine on other resolutions too.

maxrd2:
I can make separate pull request for the other issues. Top-left rendering problem is not actually related to radar map display and other 2 issues. That way should be easier to test it. Should I do that?

selaux:
Yeah, sounds like a good idea.

maxrd2:
Have managed to find another case (moving map way outside of the usable area to the north) where jumping was occurring and it was all due to bad integer division rounding. Am pretty sure #119 is solved for good with these.
Have retested it on 640x480, 1280x720, 1600x900, 1920x1080.

maxrd2:
Have rebased to current master, and cleaned up coordinate conversion code in hope it will be more obvious.
Here's an attempt at explaining what was done here, and some extra info for future reference. Hope it will be easier to review and merge this now.

JA2 uses 3 types of coordinates: screen, world and map.
Screen coordinates are isometric projection of Map and World (cartesian) coordinates. World units (sometimes referred as cell units) are CELL_*_SIZE multiples of Map units.

file `src/game/TileEngine/WorldDef.h` contains defines CELL_*_SIZE and WORLD_TILE_*
```c++
#define CELL_X_SIZE             10 // worldX = 10 * mapX
#define CELL_Y_SIZE             10 // worldY = 10 * mapY
#define WORLD_TILE_X            40 // screenX = 40 * (map isometric)
#define WORLD_TILE_Y            20 // screenY = 20 * (map isometric)

#define HALF_TILE_WIDTH		((WORLD_TILE_X / CELL_X_SIZE) / 2)	// equals to 2
#define HALF_TILE_HEIGHT	((WORLD_TILE_Y / CELL_Y_SIZE) / 2)	// equals to 1
```

```c++
// world to screen conversion
screenX = (worldX - worldY) * HALF_TILE_WIDTH;
screenY = (worldX + worldY) * HALF_TILE_HEIGHT;

// screen to world conversion
worldX = (screenX / HALF_TILE_WIDTH + screenY / HALF_TILE_HEIGHT) / 2;
worldY = (screenY / HALF_TILE_HEIGHT - screenX / HALF_TILE_WIDTH) / 2;

// map to world(cell) conversion
worldX = mapX * CELL_X_SIZE;
worldY = mapY * CELL_Y_SIZE;
```

Sometimes world/screen coordinates are negative (in top left).
Old code used integer division so world of -25 when converted to map would be calculated as `-25 / 10 = -2.5`, integer casting would "round" that to `-2` instead of `-3` (which `floor(-2.5)` would do).
Conversion from isometric to cartesian involves formula of cartesian (x, y) to produce other isometric (x) and vice versa.
There were a bunch of (bugged) ifs() to detect these cases that would alter cart/iso(x, y) iso/cart x/y were negative, have changed everything to use floating point division. There are no more rounding errors which need to be worked around with code. There are no more unreadable/non-obvious if() conditions.

lynxlynxlynx:
LGTM and thanks for the extra elaboration!

maxrd2:
You're welcome. It took me awhile to wrap my head around all those coordinates and integer constants used in calculations, hope that will save time for anyone looking at it in the future.

selaux:
:top: Tested a little and seems to work nicely. Thanks for investing some time here...

PR #734, opened 2018-10-23
Added brightness/gamma setting to ja2.json. closes #709

### LINKED ISSUE #709: Too low brightness or gamma values
First of all: great job! Thank you very much for this project. I can play the game with a decent resolution on my notebook without issues (or sort of). 

### What is wrong
My main problem: the in-game brightness (or gamma) values are pretty low, so that I have to push my screen brightness to 100%, which in turn is unusable with all other Windows applications. So I have to run the game only in full screen mode and every time switching back to Windows dim brightness down to (acceptable) 40%.

I see that you have managed to extract many hard coded values to json files. Is there a way to allow to configure brightness or gamma via json? If not, would this be possible? I'm not a game developer, but quick googling revealed for example this: https://msdn.microsoft.com/en-us/library/windows/desktop/bb173460(v=vs.85).aspx and this: https://stackoverflow.com/questions/1338398/adjust-brightness-contrast-gamma-of-scene-in-directx. So may be it is possible somehow to add this to the game on startup?

Many thanks again for this great project!

maxrd2:
As the title says. It allows to set brightness value so it corrects the gamma in game.

### DIFF
--- a/assets/distr-files-linux/ja2_manpage
+++ b/assets/distr-files-linux/ja2_manpage
@@ -20,6 +20,11 @@ Path to vanilla data
 \fB\-res WxH\fR
 Screen resolution, e.g. 800x600. Default value is 640x480
 .TP
+\fB\-brightness GAMMA\fR
+Screen brightness (gamma multiplier) value to set where 0.0 is completely dark and 1.0 is normal brightness. Default value is 1.0
+.br
+NOTE: some video drivers do not support setting the gamma value and this option may not have any effect in those cases
+.TP
 \fB\-resversion VERSION\fR
 Version of the game resources.
 Possible values are: DUTCH, ENGLISH, FRENCH, GERMAN, ITALIAN, POLISH, RUSSIAN, RUSSIAN_GOLD.
--- a/rust/src/stracciatella.rs
+++ b/rust/src/stracciatella.rs
@@ -164,6 +164,8 @@ pub struct EngineOptions {
     mods: Vec<String>,
     #[serde(rename ="res", serialize_with = "serialize_resolution", deserialize_with = "deserialize_resolution")]
     resolution: (u16, u16),
+    #[serde(default)]
+    brightness: (f32),
     #[serde(rename = "resversion")]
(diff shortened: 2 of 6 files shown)


selaux:
Hi. First of all thanks for the PR. I just tested this out and it had no effect at all on my system. Might be specific to my configuration though. What systems / OSs did you test this on?

maxrd2:
I'm on linux. Apparently game should be started through ja2-launcher in order for ja2.json to get updated.
Then you can edit the value in file there. ir just "brightness": x.xx can be added to file manually

lynxlynxlynx:
Lgtm. We also have a parameter list in the man file, so please add it also there.

selaux:
Allright did some testing and it looks like this only works on my windows machine for me (tried 2 linux installations). Even on windows it only works within a range of `0.25` to `1.75` for the brightness parameter. @lynxlynxlynx Did you do some testing? Did it work for you?

lynxlynxlynx:
No, the review was done on my phone — just came back from travelling. But even sdl warns that gamma changes aren't supported on all hw, so it may be just that.

maxrd2:
I can confirm it works on linux with official nvidia binary blob. Have tried values 0.0 (completely black) and 2.0 for pretty bright.
Regarding the man page i have found assets/distr-files-linux/ja2_manpage, but that one contains parameters for executable, where is the man file for json file that i should modify?

selaux:
`assets/distr-files-linux/ja2_manpage` is the correct file to modify. The parameter for the exectuable was added here: https://github.com/ja2-stracciatella/ja2-stracciatella/pull/734/files#diff-28a02437d5ebb8167f6cf627caee584dR321. I think we can merge regardless of those hardware incompatibilities and think about whether we can fix them somehow after...

maxrd2:
Thanks for explanation have amended the commit with man page changes.
I was doing 'ja2 --help' on system installed binary instead of locally built version and it didn't show brightness so it got me confused.

lynxlynxlynx:
Thanks!

PR #736, opened 2018-10-24
Remove redundant screen coordinate variables

maxrd2:
Removed the redundant and duplicate screen coordinate variables that were only causing issues.
Map corner coordinates (top-left, top-right, BL, BR) get converted to screen coordinates in code for rendering, and those are stored as 4 (x,y) points representing 4 corners.

In theory top-left-x should be same as bottom-left-x, in practice due to how map is projected on screen they will be a little (half tile) different - if top row is odd and bottom row is even.
Those vars are used only for calculations for rendering on screen eg radar map projection and those discrepancies in values that are supposed to be the same cause only bad calculations.

So this pull request drops redundant vars and renames them simply to top, right, bottom, left.


lynxlynxlynx:
Is this ready?

maxrd2:
Yes. It simplifies whole coordinate situation a little.
I will at some point revisit radar minimap and make that pixel perfect too. But will make another PR for that when it's ready.

Issue #725, opened 2018-09-29
Visual Studio Build/Compilation Errors


arnodick:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

Hi Stracciatella folks,

I'm trying to get Stracciatella compiling on Windows 7 with Visual Studio 2017 but having difficulty.

So far I've used:

`cmake -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain-msvc.cmake -G "Visual Studio 15 2017" ../`

Which successfully builds the VS project when run from a _bin folder.

Then I had to set _SILENCE_TR1_NAMESPACE_DEPRECATION_WARNING in the preprocessor definitions for the gtest and gtest_main projects. This had something to do with std::tr1 being deprecated in Visual Studio, and I'm not sure if it's a good long term solution. It might just be that gtest moves away from std:tr1 and releases a version that will work in VS 15.

Now I'm trying to get the ja2 project to compile, but getting this error:

`ja2-stracciatella-master\_bin\dependencies\lib-fltk\fltk-internal-prefix\src\fltk-internal\src\flstring.h(28): fatal error C1083: Cannot open include file: 'strings.h': No such file or directory [...\ja2-stracciatella-master\_bin\dependencies\lib-fltk\fltk-internal-prefix\src\fltk-internal-build\src\fltk.vcxproj]`

As far as I can tell there is a mistaken build directive called HAVE_STRINGS_H that is being set when it shouldn't be. It should be using string.h instead, I think. I've tried setting HAVE_STRINGS_H to 0 in CMakeCache for fltk, but it didn't have any affect.

When building INSTALL I'm also getting a variety of errors like this:

`SGP.obj : error LNK2019: unresolved external symbol _create_engine_options referenced in function _SDL_main`

Which I'm not sure how to resolve. There are many lines of this, each mentioning a different symbol.

(I'm also not entirely sure which project I should be building in the fist place. ja2, ja2-launcher, INSTALL?)

Any help would be much appreciated! Would love to help getting build/compilation working consistently.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
[...]

PR #739, opened 2018-11-04
Correctly set toolchain and local gtest in toolchain files

selaux:


lynxlynxlynx:
LGTM, but I haven't tested. Just merge if it solves something. :+1:

selaux:
Should close #725, will wait for feedback before merging.

PR #740, opened 2018-11-05
Fixed game starting completely dark.

maxrd2:
When starting game directly without ja2-launcher and no brightness setting defined in json or command line it would default to 0.0, and whole screen would turn completely dark.

I'd appreciate if someone with more experience in rust and serde_json would check this to confirm this is correct solution. Apparently serde(default) would set this value to 0.0 instead of using what was defined here: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/src/stracciatella.rs#L196

I thought serde(default) is just supposed to add line to json if not present using the default value from that line... apparently not.

Edit: this was introduced by changes from my pull request #734

### DIFF
--- a/rust/src/stracciatella.rs
+++ b/rust/src/stracciatella.rs
@@ -164,8 +164,7 @@ pub struct EngineOptions {
     mods: Vec<String>,
     #[serde(rename ="res", serialize_with = "serialize_resolution", deserialize_with = "deserialize_resolution")]
     resolution: (u16, u16),
-    #[serde(default)]
-    brightness: (f32),
+    brightness: f32,
     #[serde(rename = "resversion")]
     resource_version: ResourceVersion,
     #[serde(skip)]


selaux:
I think the correct one would be `#[serde(default = 1.0)]`

selaux:
See: https://serde.rs/attr-default.html

maxrd2:
Have already looked there after trying ``#[serde(default = 1.0)]`` which doesn't work because it expects string-with-the-callback-function-name instead of default value.
The default value is already defined here: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/src/stracciatella.rs#L196, but it doesn't get picked up for some reason when ``#[serde(default)]`` is there. 

The example at https://serde.rs/attr-default.html is using the function ``default_resource()``.

Updating the patch to use the function, which also fixes it, but have hoped there is something that looks cleaner, or that there is a way it could pick default brightness value directly from EngineOptions::Default::default()

selaux:
Allright. I looked once more at the documentation but the following fact is more or less undocumented unfortunately (extracted it from this PR: https://github.com/serde-rs/serde/pull/780).

The `#[serde(default)]` macro that the struct is annotated with already creates the behavior we want. Annotating the field again with `#[serde(default)]` will use the default value for the type in that field. So simply removing the `#[serde(default)]` annotation from the `brightness` field gives us what we want.

selaux:
This is a basic example derived from the serde docs: https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=32cf9cdb1d074b04b27d2e36f4366c77

maxrd2:
ok.. so how i had it fixed originally by removing ``#[serde(default)]`` was ok.
Will push that one again :)
Have added the ``#[serde(default)]`` originally because otherwise it didn't write brightness entry to json file when it's created for the first time.

selaux:
Th for recognizing and the fix BTW. I created a PR for serde docs to clarify this: https://github.com/serde-rs/serde-rs.github.io/pull/84

Issue #171, opened 2015-09-27
Inaccessible area in tactical view (SAME PROBLEM AS ISSUE #125)


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

This bug has a very negative impact on gameplay.
To understand it, look at the pictures.
The first one is from vanilla JA2 Gold:
![SCREEN2.jpg](https://bitbucket.org/repo/6zkzG5/images/2364441947-SCREEN2.jpg)

And here the same situation (same map file) in JA2 Stracciatella
![SCREEN1.jpg](https://bitbucket.org/repo/6zkzG5/images/2181310021-SCREEN1.jpg)

As you can see, there is way more area visible in the south of the map. On both screenshots, I scrolled down south as much as possible.

And now comes the bad thing about this: the "new" area in the Stracciatella map is inaccessible for the player. But enemies can (and will) end up there if invading the map from south.

If an enemy ends up in this area, he is stuck and invisible (SCREEN062 and SCREEN063) and cannot be hurt without cheat codes.

Apart from the problem with the enemies, most maps are not designed to be looked at in this area and as a result, they look a bit weird (roads ending before map border etc.)

This happens on ALL maps.
(SAME PROBLEM AS ISSUE #125)

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/171

lynxlynxlynx:
This San Mona Brothel screenshot is a good test area:
http://pds15.egloos.com/pds/200912/19/94/f0092894_4b2cb1afca61f.jpg

My testing shows that we only show one extra line, independent of resolution.

JAsmine-ja2:
I created a test map which shows quite well, what exactly is going wrong in Stracciatella vs Vanilla JA2.

Note that I scrolled to each corner as much as possible.

Showing Vanilla JA2 1.12 Gold vs. Stracciatella 0.16.0 (JA2 data from the same as vanilla) on the exact same map.




![vscreen000](https://user-images.githubusercontent.com/17213795/34689588-545e00d6-f4b6-11e7-8c30-aae3f6abca6f.jpg)

![screen000](https://user-images.githubusercontent.com/17213795/34689513-0fee618e-f4b6-11e7-8614-b1335da6521d.jpg)


[...]

Issue #73, opened 2013-08-31
Issue with radarmaps


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

The green square in the radarmaps gets hidden by the lower and the right borders of the radarmap-window. The upper and left borders behave normal.
![radarmap issue](http://img.xrmb2.net/images/932468.jpeg)
left side: normal behavior on the upper-left corner.
right side: wrong behavior on the lower-right corner.

v0.12.189, English JA2 Gold, ja2-ENGLISH-1024x768.bat

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/73

derSteFfi:
the code that draws this is:
```C
			// Find the diustance from render center to true world center
			INT16 const sDistToCenterX = gsRenderCenterX - gCenterWorldX;
			INT16 const sDistToCenterY = gsRenderCenterY - gCenterWorldY;

			// From render center in world coords, convert to render center in "screen" coords
			INT16 sScreenCenterX;
			INT16 sScreenCenterY;
			FromCellToScreenCoordinates(sDistToCenterX, sDistToCenterY, &sScreenCenterX, &sScreenCenterY);

			// Subtract screen center
			sScreenCenterX += gsCX;
			sScreenCenterY += gsCY;

			// Get corners in screen coords
			// TOP LEFT
			INT16 const sX_S = g_ui.m_tacticalMapCenterX;
			INT16 const sY_S = g_ui.m_tacticalMapCenterY;

			INT16 const sTopLeftWorldX     = sScreenCenterX - sX_S;
			INT16 const sTopLeftWorldY     = sScreenCenterY - sY_S;
			INT16 const sBottomRightWorldX = sScreenCenterX + sX_S;
			INT16 const sBottomRightWorldY = sScreenCenterY + sY_S;

			// Determine radar coordinates
			INT16 const sRadarCX = gsCX * gdScaleX;
			INT16 const sRadarCY = gsCY * gdScaleY;

			INT16 const sWidth  = RADAR_WINDOW_WIDTH;
			INT16 const sHeight = RADAR_WINDOW_HEIGHT;
			INT16 const sX      = RADAR_WINDOW_X;

[...]

PR #741, opened 2018-11-05
Fix radar map issues

maxrd2:
Have separated this from pr #733, as they were not really related.
It should solve #171, #73, #195, #125 once properly tested which are all listed in #225.

### DIFF
--- a/src/game/Tactical/Interface.h
+++ b/src/game/Tactical/Interface.h
@@ -46,6 +46,11 @@
 
 #define ROOF_LEVEL_HEIGHT		50
 
+#define SCROLL_LEFT_PADDING -30
+#define SCROLL_RIGHT_PADDING 10
+#define SCROLL_TOP_PADDING -50
+#define SCROLL_BOTTOM_PADDING -20
+
 
 // Interface level enums
 enum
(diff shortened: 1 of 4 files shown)


maxrd2:
@selaux https://github.com/ja2-stracciatella/ja2-stracciatella/pull/733#issuecomment-435701845
> Allright I got around to a little testing and noticed some issues (tested the starting sector A9 in 1280x720 resolution)
>    The green radar rectangle changes size by probably one pixel when scrolling to the top of the map as well as when scrolling to the bottom

Have improved a green rectangle drawing a little bit so it resizes less. It is now mostly happening on south part of the map, because there were two unusable lines shown south and radar thumbnail is displaying only usable tiles.
Since radar thumbnail is loaded from image resource and is not generated I can't do much about it.

Only thing I can do about it right now is to make sure that rectangle is always drawn at same size, but that might affect it's accuracy by few tiles. Right now borders should overlap pixel on minimap that represents exact map tile (it can be tested by moving soldiers around on small enough resolution).

There is also some inaccuracy and voodoo around making ``gs<Corner>World<X/Y>`` variables work since there is also integer division used during conversion from screen to map coordinates. Will try to address that asap.

selaux:
LGTM :+1:.

Issue #708, opened 2018-02-27
Tile corruption when scrolling around


moll:
I'm not sure why or what triggers it yet, but seeing tile corruption when navigating around. I mostly do so through the arrow keys, but I think scrolling with the mouse has the same per-tile movement.

See the building left of Blood. Parts of airplane poster on the top-left room also corrupt the grass tile seen through the window.

![screenshot from 2018-02-27 01-45-21](https://user-images.githubusercontent.com/156837/36706031-f67d5842-1b5f-11e8-8df7-3b56018ca937.png)

![screenshot from 2018-02-27 01-45-16](https://user-images.githubusercontent.com/156837/36706035-fb14ee10-1b5f-11e8-9b19-47f183ce66ec.png)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  00cb9fef9 | GOG v1.12 (17794)  |    Probably not due to wide resolution   |         |

This is on Arch Linux on GeForce 960 GTX with the resolution set to 2520x1080px.

Any other intel I could provide to help debug this?

moll:
I have a feeling it's related to the somewhat non-standard resolution of 2520x1080px. Restarting with 2880x1080px doesn't reproduce this, but switching back to 2520x1080px does.

lynxlynxlynx:
Is it visible also without the mercs moving? We have an older report for that here:  #551

moll:
Everyone was stationary in my tests above. I didn't really notice the corruption until I got to Drassen, so I'd say no heavy corruption during movement. I don't remember what my "Merc Lights during Movement" setting was if that's relevant, but I can test if necessary given that the corruption was reproduceable with that non-standard resolution.

maxrd2:
@moll Could you please confirm this still happens on current master?

moll:
Yay. Thx. I'll report back next time I get around to compiling and playing JA2!

flaviojs:
I remember seeing corruption like this (inside Drassen mines) when I used version 0.16.1.
I don't remember seeing it in current master de0701d089dcd947fd49ff918ae016a621f2dfcf.

lynxlynxlynx:
Thanks, closing then.

PR #743, opened 2018-11-14
Various memory fixes.

maxrd2:


maxrd2:
Here are the fixes for issues I've noticed while doing other things.
Issue #708 might be caused by corrupted zbuffer, that 2520x1080px resolution might have a pitch that is multiple of 16 or something. Commit b830736 should fix it if that's the case.

src/game/Utils/Cinematics.cc
```diff
@@ -172,7 +172,8 @@ catch (...) { return 0; }
 
 void SmkCloseFlic(SMKFLIC* const sf)
 {
-	SmackClose(sf->SmackerObject);
+	if(sf->SmackerObject)
+		SmackClose(sf->SmackerObject);
```
lynxlynxlynx: I think this would better fit into SmackClose to mimic free/delete, but then again, this is its only user, so it's not that important.

lynxlynxlynx:
LGTM!

src/game/Utils/Cinematics.cc
maxrd2: right... will change it

maxrd2: rebased and pushed this change

lynxlynxlynx:
Thanks!

PR #745, opened 2018-11-17
Fix deprecation of env::home_dir

### LINKED ISSUE #744: new rust test failure + std::env::home_dir deprecated
Mac is failing now, but there's also a hint:
> warning: use of deprecated item 'std::env::home_dir': This function's behavior is unexpected and probably not what you want. Consider using the home_dir function from https://crates.io/crates/dirs instead.
>    --> src/stracciatella.rs:419:11
>     |
> 419 |     match env::home_dir() {
>     |           ^^^^^^^^^^^^^
>     |
>     = note: #[warn(deprecated)] on by default
>     Finished dev [unoptimized + debuginfo] target(s) in 58.78s
>      Running `/Users/travis/build/ja2-stracciatella/ja2-stracciatella/ci-build/rust/debug/deps/stracciatella-adc025079a4479a4 --nocapture`
> running 39 tests
> thread 'tests::build_engine_options_from_env_and_args_should_return_an_error_if_datadir_is_not_set' panicked at 'assertion failed: `(left == right)`
>   left: `Ok(EngineOptions { stracciatella_home: "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/ja2-test.FMWz8OE3ZIWT/.ja2", vanilla_data_dir: "/some/place/where/the/data/is", mods: [], resolution: (1100, 480), brightness: 1.0, resource_version: ENGLISH, show_help: false, run_unittests: false, run_editor: false, start_in_fullscreen: true, start_in_window: false, scaling_quality: PERFECT, start_in_debug_mode: false, start_without_sound: false })`,
>  right: `Err("Vanilla data directory has to be set either in config file or per command line switch")`', src/stracciatella.rs:1086:9

selaux:
Should close #744. Looking into the test failures next.


selaux:
I also needed to update the base image to xenial in order to make the mingw build succeed, as the mingw version on trusty was too old to contain the `SHGetKnownFolderPath` symbol.

selaux:
Done.

PR #754, opened 2019-01-13
Fixed bad z-buf write during render dirty. refs #753

### LINKED ISSUE #753: Segmentation fault on soldier death
If the death animation of a soldier finishes while he is on screen, the game crashes with a segmentation fault. Moving the camera so the soldier disappears before the animation is finished doesn't 

I get differing frame stacks, which is probably due to faulty memory allocation.
Sometimes it crashes during a libc call in the radeonsi driver, last time it crashed in Blt8BPPDataTo16BPPBufferTransZTransShadowIncObscureClip() at /src/game/TileEngine/RenderWorld.cc:3306

I have bisected the issue and landed at commit 25503e9ab721d2f67276889015d448e8eca1a7e3

Reverting this commit fixed the issue for me.

|  current head | JA2 Gold English  |    yes    |    yes     |

Savegame: If you shoot the soldier with Len he should die immediately:
[CrashDuringDeath.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/2752428/CrashDuringDeath.zip)

How to reproduce: Kill the soldier
OS: Gentoo linux

maxrd2:
This fixes #753. Commit 25503e9 was considering sizeof pointer instead of data it pointed to.

### DIFF
--- a/src/game/TileEngine/Render_Dirty.cc
+++ b/src/game/TileEngine/Render_Dirty.cc
@@ -207,7 +207,7 @@ void RestoreBackgroundRects(void)
 			}
 			else if (b->pZSaveArea != NULL)
 			{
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

PR #757, opened 2019-01-16
Fixed IMP attitude generation

### LINKED ISSUE #756: TODO0008: Attitude generation is broken
### Attitude generation results are different that quiz answers suggest

Already reported as #405 
Closed as invalid but clearly doesn't work

### How to reproduce
Select quiz answers:
5
3
1
1
3
4
4
4
2
1
3
5
1
4
4
1
This selection has exactly 1 attitude selected by 2 answers (aggressive). All other attitudes have at most 1 answer.
This should generate an aggressive merc regardless of what number RNG picks.
But it doesn't.

There are 2 bugs next to each other that cause this. First is TODO0008.
Second, the final assignment when picking a stat should be
iAttitude = i;
instead of
iAttitude = iCounter2;

After fixing those I can finally generate an aggressive merc.

|  0.16.1 | eg. JA2 Gold En  |    yes    |    no     |

akdjka:
Fixes #756

### DIFF
--- a/docs/archeology/TODO
+++ b/docs/archeology/TODO
@@ -24,7 +24,6 @@ TODO0021
 - TODO0005 Make it endian agnostic
 - TODO0006 The IMP personality quiz evaluation seems to have some bugs
 - TODO0007 Deduct AIM_PENALTY_SMG when firing a SMG even when second hand is free?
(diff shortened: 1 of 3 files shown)


akdjka:
Closing due to branching cleanup, will open a new one with the same code

PR #767, opened 2019-01-20
TODO0006: fixed a fallthrough

akdjka:
Of TODO0006 comments quite a few felt arbitrary, maybe the authors wanted it that way. Well, actually most did.
One didn't and I fixed it today.

### DIFF
--- a/docs/archeology/TODO
+++ b/docs/archeology/TODO
@@ -22,7 +22,6 @@ TODO0021
 - TODO0001 implement blitters and related functions
 - TODO0003 Implement sound streaming
 - TODO0005 Make it endian agnostic
-- TODO0006 The IMP personality quiz evaluation seems to have some bugs
 - TODO0007 Deduct AIM_PENALTY_SMG when firing a SMG even when second hand is free?
 - TODO0008 The random call here seems to be off-by-one. If it returns its maximum value then no assignment to iAttitude happens, because the next loop tries to find the n+1th peak of n peaks.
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
If the other comments don't make sense, they should be removed. Or at least the todo itself, since gcc doesn't like implicit fallthroughs anymore.

akdjka:
I didn't say they don't make sense.
I just don't know what to do about them. F.e. 
```
		// XXX TODO0006 the effects seems odd. answer 3 is even more aggressive than answer 2
```
Yes it is. But I have a feeling the authors wanted it like that.
```
		case 6: break; // asshole // XXX TODO0006 code and commend disagree
```
Yes. But there would be 3 answers that give ASSHOLE. At quick glance all attitutes have exactly 2 answers that grant them. For sure the only one which has more is NORMAL. I have a hunch that not adding another ASSHOLE was intentional.
```
		// XXX TODO0006 cases 2 and 3 probably interchanged
		case 3: break; // none
		case 2: AddAnAttitudeToAttitudeList(ATT_ARROGANT); break;
```
Yes. But answers seem to be consistent with the code. So it seeems like a formatting issue. Not sure though.

lynxlynxlynx:
I agree with your interpretation. Either way, a good rule of thumb is to trust the designers/developers in case of doubt. So we can just remove the markings and forget about it.

akdjka:
OK, I removed the other TODO0006 comments

PR #768, opened 2019-01-20
Removed silly camo kit addition

akdjka:
There's a silly piece of code in IMP character generation.
If the character is CAMOUFLAGED they get a camo kit.
First, CAMOUFLAGED characters don't need camo kits, this is entirely useless for them.
Second, one can't create a CAMOUFLAGED IMP so this is dead code.


lynxlynxlynx:
The first one makes thematic sense, but the second argument wins. :+1:

PR #771, opened 2019-01-22
#769: Fixed cleaning knives from blood

### LINKED ISSUE #769: Cleaning blood from knives
### One can clean blood from a knife
There are (at least 2 ways to clean up a knife).
One is surely a bug:  throw a bloody knife through a window which still has glass. Glass stops the knife and cleans it up.
The other is arguably a feature: throw it at a ground.

Fixing the latter is trivial and if there's interest I can do it.

Fixing the first is not trivial. At the point where the game decides about item placement there's no information about the item itself; just a bullet with a flag "is knife".

Best fix? My first idea is to extend the bullet class with a pointer to the item it represents which is null for all bullets that are not items? As well as remove redundant fields (like status).
Though simply adding a "is bloody knife" flag would work as well.

I may try fixing the first as well.

Attached a save that makes testing both cases easy.

|  ce14a7344a8 | eg. JA2 Gold Dutch  |    yes/no    |    yes/no     |

[s.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/2776907/s.zip)

akdjka:
OK, this should fix #769.

However, the code base of JA2 is realy mix of styles. Including many I would call bad. I searched discussion on what style to use on future commits and failed to find any. I tried to be consistent with what's around, to the point of adding tabs in the middle of a line which I personally despise.

### DIFF
--- a/src/game/Tactical/Bullets.h
+++ b/src/game/Tactical/Bullets.h
@@ -15,6 +15,7 @@
 #define BULLET_FLAG_TANK_CANNON	0x0020
 #define BULLET_FLAG_BUCKSHOT		0x0040
 #define BULLET_FLAG_FLAME		0x0080
+#define BULLET_FLAG_BLOODY		0x0100 // bloody knives have both bloody and knife flags
 
 struct BULLET
 {
--- a/src/game/Tactical/Bullets.h
+++ b/src/game/Tactical/Bullets.h
@@ -7,15 +7,15 @@
 #include "Tile_Animation.h"
 
 
-#define BULLET_FLAG_CREATURE_SPIT	0x0001
-#define BULLET_FLAG_KNIFE		0x0002
-#define BULLET_FLAG_MISSILE		0x0004
-#define BULLET_FLAG_SMALL_MISSILE	0x0008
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
LGTM!

Yes, I don't think we have a bug open yet for a style guide. The first big task was to harmonize the indentation, which is luckily already behind us. Generally we agree that it's best to keep the current format for existing changes, however for new stuff or harmonization efforts we agreed somewhere that `if (statements)` is the way to go, no double spacing around the parens. And at least personally I prefer explicit braces just to avoid potential bugs in the future (coverity also warns about this).

Feel free to suggest something concrete.

akdjka:
I don't think it matters much what style is used but it's important to have some style. This really impacts readability. If I were to suggest a particular rule only one comes to my mind - no tabs in the middle of the line. ;)
And any indentation style. Mixing, like it is now, is very bad...

I will adjust my commit to add explicit braces and remove extra spaces.

akdjka:
OK, did some style improvements. I touched directly adjacent areas as well....

selaux:
@akdjka Thanks.

Issue #791, opened 2019-03-02
-Wlogical-not-parentheses


z33ky:
### What is wrong
Compiler (clang 7) complains:
```
[202/376] Building CXX object CMakeFiles/ja2.dir/src/game/Tactical/Points.cc.o
../src/game/Tactical/Points.cc:1681:6: warning: logical not is only applied to the left hand side of this bitwise operator [-Wlogical-not-parentheses]
        if (!item->getItemClass() & IC_GRENADE)
            ^                     ~
../src/game/Tactical/Points.cc:1681:6: note: add parentheses after the '!' to evaluate the bitwise operator first
        if (!item->getItemClass() & IC_GRENADE)
            ^
             (                                )
../src/game/Tactical/Points.cc:1681:6: note: add parentheses around left hand side expression to silence this warning
        if (!item->getItemClass() & IC_GRENADE)
            ^
            (                    )
1 warning generated.
[207/376] Building CXX object CMakeFiles/ja2.dir/src/game/Tactical/OppList.cc.o
../src/game/Tactical/OppList.cc:4900:7: warning: logical not is only applied to the left hand side of this bitwise operator [-Wlogical-not-parentheses]
                if (!tgt.uiStatusFlags & SOLDIER_MONSTER) continue;
                    ^                  ~
../src/game/Tactical/OppList.cc:4900:7: note: add parentheses after the '!' to evaluate the bitwise operator first
                if (!tgt.uiStatusFlags & SOLDIER_MONSTER) continue;
                    ^
                     (                                  )
../src/game/Tactical/OppList.cc:4900:7: note: add parentheses around left hand side expression to silence this warning
                if (!tgt.uiStatusFlags & SOLDIER_MONSTER) continue;
                    ^
                    (                 )
1 warning generated.
```

I'm guessing the code is wrong and should read `if (!(thing & FLAG))`.

lynxlynxlynx:
Yes, good job clang and thanks for the report. If you're willing to open a PR fixing this, even better. :)

PR #793, opened 2019-03-02
#791: Fix flag checking

z33ky:


src/game/Tactical/Points.cc
```diff
@@ -1678,7 +1678,7 @@ INT16 MinAPsToThrow(SOLDIERTYPE const& s, GridNo gridno, bool const add_turning_
 	const ItemModel *item = GCM->getItem(in_hand);
 	// Gennady: This is a very strange piece of code.
 	//          Be very careful with it.
-	if (!item->getItemClass() & IC_GRENADE)
+	if (!(item->getItemClass() & IC_GRENADE))
```
lynxlynxlynx: Ahh, the comment made me remember this is not a good fix. The discussion is in #287 and it's more likely this whole block should be removed or redone.

z33ky: Changed to conserve previous behavior and mention the issue.

lynxlynxlynx:
Thanks!

PR #807, opened 2019-03-29
SoundMan

flaviojs:
I heard a flinch-inducing sound one time while using a shotgun. >.<

After inspecting the code I found that the sound mixing code can overflow/underflow, which is the probable cause of that terrible sound, so I went ahead and made a fix (avoid overflow/underflow and add clipping).
~~I'm not sure if VC allows this kind of array (mix in SoundCallback), someone else will need to test it.~~

After that I noticed out-of-date comments about changing the sound frequency in smacker code and a couple of other minor issues, so I reworked the way sound samples are created from buffers.

While adding sound clipping I wanted to change the format to AUDIO_F32SYS, but decided to ask first.
Can I change the format?
It doubles the amount of memory needed by the samples, it avoids the need of a separate mix buffer and it simplifies other ways of handling sound that would be clipped.

### DIFF
--- a/src/sgp/Smack_Stub.cc
+++ b/src/sgp/Smack_Stub.cc
@@ -48,7 +48,6 @@ void SmackPrintFlickInfo(unsigned long width, unsigned long height, UCHAR8 scale
 UINT32 SmackGetAudio(const smk SmkObj, UCHAR8** audiobuffer)
 {
 	UINT32 audiolen = 0;
-	UINT32 i, index;
 	UCHAR8* paudio;
 
 	if (!audiobuffer)
@@ -154,7 +153,7 @@ Smack* SmackOpen(SGPFile* FileHandle, UINT32 Flags, UINT32 ExtraFlag)
 	flickinfo->FramesPerSecond = usf;
 	audiobuffer = NULL;
 	audiolen = SmackGetAudio(flickinfo->Smacker, &audiobuffer);
-	if (audiobuffer != NULL);
+	if (audiobuffer != NULL)
 	{
 		//SmackWriteAudio( audiobuffer, audiolen); // are getting right audio data?
 		// shoot and forget... audiobuffer should be freed by SoundMan
--- a/src/sgp/Smack_Stub.cc
+++ b/src/sgp/Smack_Stub.cc
@@ -2,6 +2,7 @@
 #if !defined(_MSC_VER)
 	#include <strings.h>
 #endif
+#include "Debug.h"
 #include "Smack_Stub.h"
 #include "Sound_Control.h"
 #include "SoundMan.h"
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Thanks! In a brief reading it looks nice, but it's the type of thing that really requires testing.

No opinion on AUDIO_F32SYS, but I don't think we worry too much about low memory device portability.

flaviojs:
I just realized that `SoundCallback` is executed in another thread... so the terrible sound might be from that too.

`SOUNDTAG.State` is marked volatile, which implies that the compiler will never reorder it or optimize it, but it does not guarantee that different threads are seeing the same data.
You need a (sync) barrier to guarantee that they see the same. (like a mutex)

This was a minor problem when computers only had one core since the OS switching threads involves a barrier, but now almost everything has multiple cores so the threads can run at the same time without a barrier.
If you see a volatile in C/C++, be on guard!!!

src/sgp/SoundMan.cc
```diff
@@ -723,11 +723,13 @@ static SOUNDTAG* SoundGetChannelByID(UINT32 uiSoundID)
 
 static void SoundCallback(void* userdata, Uint8* stream, int len)
 {
-	SDL_memset(stream, 0, len);
+	INT32  mix[ gTargetAudioSpec.samples ];
 
-	UINT16* Stream = (UINT16*)stream;
+	Assert(len * 2 == sizeof(mix));
```
selaux: Is there documentation somewhere that this is always true?

src/sgp/Smack_Stub.cc
```diff
@@ -37,38 +38,51 @@ BOOLEAN SmkAudioSwitch(smk SmkObj, BOOLEAN sw)
...
 {
 	printf ("Video -- Frames: %lu Width: %lu Height: %lu Scale: %d \n", framecount, width, height, scale);
-	printf ("Audio -- FPS: %2.2f  Channels: %d Depth: %d Rate %d\n", usf/1000, a_channels, a_depth, a_rate);
+	printf ("Audio -- FPS: %2.2f  Channels: %u Depth: %u Rate %lu\n", usf/1000, a_channels, a_depth, a_rate);
 }
 
-// read all smackaudio and convert it to 22050Hz on the fly (44100 originally)
-UINT32 SmackGetAudio(const smk SmkObj, const INT16* audiobuffer)
+// read all smackaudio
+UINT32 SmackGetAudio(const smk SmkObj, UCHAR8** audiobuffer)
```
selaux: Why does this use a pointer to a pointer?

src/sgp/Smack_Stub.cc
```diff
@@ -130,27 +144,23 @@ Smack* SmackOpen(SGPFile* FileHandle, UINT32 Flags, UINT32 ExtraFlag)
...
-	audiobuffer = (INT16*) malloc(audiosamples);
-	if (!audiobuffer)
+	audiobuffer = NULL;
+	audiolen = SmackGetAudio(flickinfo->Smacker, &audiobuffer);
+	if (audiobuffer != NULL);
 	{
-		free(flickinfo);
-		return NULL;
+		//SmackWriteAudio( audiobuffer, audiolen); // are getting right audio data?
+		// shoot and forget... audiobuffer should be freed by SoundMan
```
selaux: This comment says that audiobuffer should be freed by Soundman but it is freed two lines below. Is this intended?

selaux:
I think my suspicion from above is true. I get the following output on a 64bit Linux System:

```
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
2019/03/08 20:28:01 ERROR   [Asserts         ] Assertion failed in /ja2-stracciatella/src/sgp/SoundMan.cc, line 787
*** stack smashing detected ***: <unknown> terminated
[1]    12786 abort (core dumped)  ./ja2
```

src/sgp/SoundMan.cc
flaviojs: Your result below proves that it's not true.
Looks like SDL can change the number of samples it asks for in the callback.

src/sgp/Smack_Stub.cc
flaviojs: It's an old comment that I didn't notice, thanks.

src/sgp/Smack_Stub.cc
flaviojs: SmackGetAudio is the one creating the buffer now so it needs a way to return it along with the length.

src/sgp/SoundMan.cc
```diff
@@ -191,32 +197,31 @@ UINT32 SoundPlay(const char* pFilename, UINT32 volume, UINT32 pan, UINT32 loop,
...
-	buffertag->n_samples = size;
-	buffertag->pData = pbuffer;
-	buffertag->uiFlags =  SAMPLE_STEREO | SAMPLE_ALLOCATED;
-	buffertag->uiPanMax        = 64;
-	buffertag->uiMaxInstances  = 1;
-	IncreaseSoundMemoryUsedBySample(buffertag);
+	if (pbuffer == NULL) return SOUND_ERROR;
+	if (size == 0) return SOUND_ERROR;
+
+	if (depth == 16) format = AUDIO_S16LSB;
```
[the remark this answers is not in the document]
flaviojs: Ok. I wanted to find out the formats of other depths but I haven't been able to find info of the format.

selaux:
:+1: Works fine now. Regarding changing audio format I would wait until we have something like SDL mixer. There already noticeable lags when loading a new music piece, even on performant systems.

Issue #810, opened 2019-04-04
About supporting chinese localization maybe?


ChinaSakhalin:
Uh, sorry for devs and everyone, but I don't know how to use this. I found the original english text file in source code named "_EnglishText.cc", I should translate it into chinese right? I don't know programming, but translation is good, so after I made it, I should give the file to developers like you and you make them work, right?

lynxlynxlynx:
Translating the file would be the first step, correct. The problem is that the JA2 fonts don't have Chinese glyphs though, so first a solution for that needs to be found, for you to not waste work. Maybe there's already a modded font out there?

ChinaSakhalin:
Thank you, sir! Really appreciated! Didn't think the reply come fast like this. I will translate, thank you again!

JAsmine-ja2:
@lynxlynxlynx 
Yes, indeed. Old official v1.13 release has chinese localization with modded fonts included:

![SCREEN000](https://user-images.githubusercontent.com/17213795/55580958-0bca0300-571c-11e9-8d7e-b19b7520d4b6.jpg)
![SCREEN001](https://user-images.githubusercontent.com/17213795/55580960-0bca0300-571c-11e9-9774-03cd757e0dda.jpg)
![SCREEN002](https://user-images.githubusercontent.com/17213795/55580961-0c629980-571c-11e9-9df5-5b1e7da800d2.jpg)

Download available here:
http://kermi.pp.fi/JA_2/v1.13_Releases/Official/Chinese/

@ChinaSakhalin 
I can't judge if the old localization of v1.13 is complete and of good quality - but it can probably be of help for you.

JAsmine-ja2:
Just for reference: @gtrafimenkov brought up the idea for a "master font.slf" which supports all game versions here: #43

ChinaSakhalin:
Thank you all, gentlemen! I really appreciate your replies. And I have finished the translation, and slightly changed GameRes.cc and GameRes.h file, didn't know whether I did right or not. I just upload here, or somewhere else? if I did something wrong, please, no hesitate to tell me. I just want to give a little contribution to your great MOD, I really love the original JA2, and I love your idea, so please, if the translation file is bad, I will correct it. Thank you!

[ja2-stracciatella-0.16.1.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3050478/ja2-stracciatella-0.16.1.zip)

lynxlynxlynx:
[...]

PR #811, opened 2019-04-06
WIP: added Chinese translation #810

lynxlynxlynx:
@ChinaSakhalin provided a basic translation in #810 and I've made all the fixes and code adjustments necessary for it to run via ` ./ja2 -resversion CHINESE` (or the launcher).

There are a few TODOs sprinkled inside, just as a reminder, since I don't have the data or font to fully test. But yes, this is useless without a font, while modified data (STIs) would also make it better.

But apparently the existing font is useless, 1.13 implemented winfont support instead. [JA2.v1.13-4452.sourcecode.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3051039/JA2.v1.13-4452.sourcecode.zip)


flaviojs:
Hmm, wouldn't this chinese translation be better off as a mod instead of a new resversion?

lynxlynxlynx:
I don't see how that could work; depending on the way it is done, it may need a mod on top either way.

flaviojs:
The `-resversion` option tells stracciatella the language of the game data and what to emulate.
This chinese translation is not a direct modification of the original data, but a translation on top of the original data so stracciatella should still try to emulate the original (whatever version that is).

The main problem is the fonts, we have glyph images but no unicode codepoint<->glyph mappings.
The translation table would need to be externalized or real fonts supported (they include mappings).
Translation tables need to be specific to each font (or glyph count), as you can see in issue #837.

flaviojs:

lynxlynxlynx:
Closing in favour of #1445.

Issue #806, opened 2019-03-26
Extra crows when a game is loaded


flaviojs:
### What is wrong
A crow is stuck in the top left section of the map outside the fence in D13 (Drassen).

He starts moving to a corpse when I cut a hole in the fence near him.
The console also shows a lot of "WARNING [Points          ] Invalid movement mode 125 used in ActionPointCost". (125 is CROW_WALK)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  de0701d089dcd947fd49ff918ae016a621f2dfcf | ENGLISH  |        |         |

[SaveGame02.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3009686/SaveGame02.sav.zip)

I wonder... how did the crow get there in the first place? I though they only appeared near corpses.

This map has problems going to/from the outside and the inside of the fence, often forcing me to move to an intermediate destination. Maybe it's related?

flaviojs:
The crows only appear when you load a game.
They appear in the north spawn point and head to the closest corpse.

I still don't know how they are created, but the more crows I try to kill before saving, the more crows appear on load.

Here is a save with 22 extra crows:
![22_crows](https://user-images.githubusercontent.com/1009600/55444980-fb494980-55b0-11e9-9fa7-fffdbc8a153d.png)
[SaveGame02.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3036368/SaveGame02.sav.zip)

flaviojs:
After adding debug prints to `HandleRottingCorpses` I found that all the extra crows have `bInSector` equal to `FALSE`.

flaviojs:
Finally found the cause of the extra crows. :)

In `AddCrowToCorpse` the function `FindRandomGridNoFromSweetSpot` is called with a radius of 2.
That function will check a 5x5 area for reachability and will try up to 2000 times random points in a 2x2 area (BUG).

When `FindRandomGridNoFromSweetSpot` fails it returns `NOWHERE` and `AddCrowToCorpse` will do nothing else. `TacticalCreateSoldier` has already been called, so an extra crow remains. <- THE CAUSE

flaviojs:
A final note, the extra crows are present in vanilla!

Recently I played with vanilla and punched crows for a while, then saved, the next time I loaded I was presented with a bunch of crows coming from the north spawn point.

PR #812, opened 2019-04-07
Fix extra crows when a game is loaded

flaviojs:
Fixes: #806

NOTE: the AP cost of CROW_WALK was changed


lynxlynxlynx:
LGTM, thanks!

Issue #42, opened 2013-03-31
Autodetect version of the game resource


lynxlynxlynx:
Originally reported by: **Gennady Trafimenkov (Bitbucket: [gennady](http://bitbucket.org/gennady), GitHub: [gennady](http://github.com/gennady))**

---

At the moment the user must select appropriate version of the game resources before starting the game by using -resversion command line key.

Need to implement automatic detection algorithm, so there will be no need for the user to do it manually.  That should improve usability of the program.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/42

JAsmine-ja2:
Since this task is covered in #607, can we close this issue?

lynxlynxlynx:
not sure, this code should also be in core, so it'd work when launching it directly.

Manuel-K:
True. If we had automatic detection in the engine, we could completely remove the settings from the launcher, although some override functionality could be useful.

flaviojs:
Looking at GameRes.cc, some versions have simple distinguishing features:

RUSSIAN -> files inside `russian` dir
POLISH -> files inside `polish` dir
ITALIAN -> files inside `italian` dir
DUTCH -> files inside `dutch` dir
GERMAN -> files inside `german` dir

but what about the other versions ENGLISH, FRENCH and RUSSIAN_GOLD?
How can they be distinguished?

flaviojs:
An easy way to handle this is to gather a list of file-hash pairs for each version and then use it to compare.

lynxlynxlynx:
File presence is a bit stabler approach, when possible, since I think there are multiple versions of the same locale games.

flaviojs:
Another important note, are the russian/polish/italian/dutch/german directories and files inside SLF archives?
I only have the english version so I can't check.

selaux:
I think we can close this with https://github.com/ja2-stracciatella/ja2-stracciatella/pull/820 and https://github.com/ja2-stracciatella/ja2-stracciatella/pull/856. I tested all versions available to me (German (Steam), French (Steam), Polish (Steam), Russian (Steam), Gold English (Steam), English (Steam)). :tada:

PR #820, opened 2019-04-24
Launcher detects version

flaviojs:
Reference: #42

### Check resource paths:
Can guess GERMAN, DUTCH, ITALIAN, POLISH, and RUSSIAN.
It is missing ENGLISH, FRENCH and RUSSIAN_GOLD. (no unique resources?)

IMPLEMENTED

### Use text:
EDT files have text in the target language, so we could test the unicode ranges being used.
According to `LoadEncryptedData@DefaultContentManager.cc` the text is "encrypted" with ROT-1 and can be badly encoded!

REJECTED I don't have files to test this approach

### Compare hashes:
Could identify game language, specific versions of the game, and mods too.
We need to know the hashes in each version/mod. (md5? sha1? other?)
EDT files are obvious targets since they have text in the target language.
Voice files are good targets if they are also translated.

IMPLEMENTED

### Compare file sizes:
The file size of a file can change between versions, so it can be used as a cheap initial comparison.

IMPLEMENTED

-----

What this PR adds:
 * resource pack for the english version of JA2 in GOG: https://www.gog.com/game/jagged_alliance_2
 * resource packs for the german, french, polish, and russian versions of JA2 in Steam: https://store.steampowered.com/app/545210/Jagged_Alliance_2_Classic/ (provided by selaux)
 * rust code to read/write slf files
 * rust code and binary to create resource packs
 * rust code to guess the version from the game resources
  * launcher button to guess the version


lynxlynxlynx:
I asked the 1.13 team about the files (waiting).

rust/src/res.rs
```diff
@@ -0,0 +1,195 @@
...
+    pub path: String,
+
+    // The properties of the resource.
+    //
+    // # Properties
+    //
+    //  * archive_path (string) - the resource path of the archive that contains this resource
+    //  * archive_slf (bool) - true if the resource is a SLF archive
+    //  * archive_slf_num_resources (number) - number of resources inside the SLF archive
+    pub properties: Map<String,Value>
```
selaux: Could we make the string an enum here, so we don't have arbitrary properties?

rust/src/res.rs
```diff
@@ -0,0 +1,195 @@
...
+pub struct ResourcePack {
+    // A friendly name of the resource pack for display purposes.
+    pub name: String,
+
+    // The properties of the resource.
+    //
+    // # Properties
+    //
+    //  * with_archive_slf (bool, default = false) - the contents of SLF archives are included
+    pub properties: Map<String,Value>,
```
selaux: Could we make the string an enum here, so we don't have arbitrary properties?

rust/src/res.rs
```diff
@@ -0,0 +1,195 @@
...
+        write!(f, "ResourceError({:?})", self.desc)
+    }
+}
+
+
+// Gets the extension of the path in uppercase.
+fn uppercase_extension(path: &Path) -> Result<String,Box<Error>>
+{
+    if let Some(extension) = path.extension() {
+        return Ok(extension.to_str().unwrap().to_uppercase());
```
selaux: You could use `to_string_lossy` to get rid of the unwrap...

rust/src/res.rs
selaux: Or maybe even a `Vec<ResourceProperty>` with

```rust
enum ResourceProperty {
    ArchivePath(PathBuf),
    ArchiveSlf(bool),
...
}
```

selaux:
Awesome work, this is a good use case for a first version of a rust resource loader :+1:.

selaux:
Have you considered using something like [nom](https://github.com/Geal/nom) to do the heavy lifting for the slf format parsing?

flaviojs:
This is my first experience with rust and the format information is scarce, so I opted to keep it simple and avoided libraries. I'll happily look into [nom](https://github.com/Geal/nom) after the TODOs in slf.rs are resolved.

flaviojs:
`ResourcePack` is intended to be a representation of a json file that identifies the resources in a particular version/mod. In fact, it can already be serialized to json and deserialized from json with serde_json.

Required data lives directly in the structs, the rest is optional and goes in the properties.

The properties are arbitrary on purpose. It is the only place where whoever manages the file can place additional information like an author, an url, a revision, a license, any valid json really.

I can turn the known property names into an enum (with an Unknown "catch everything else"), but this code might interact with C code later so it might turn out as wasted effort if done right now. I'll wait until most of the PR is done.

rust/src/res.rs
flaviojs: see https://github.com/ja2-stracciatella/ja2-stracciatella/pull/820#issuecomment-486643636

rust/src/res.rs
flaviojs: see https://github.com/ja2-stracciatella/ja2-stracciatella/pull/820#issuecomment-486643636

rust/src/res.rs
flaviojs: This function is only called where the whole filename is known to have valid utf8, so the unwrap is avoiding the allocation of a temporary string. I'll keep the unwrap.

Also, there is no path that produces an error so I can return a String instead of a Result.

selaux:
Sounds good to me...

lynxlynxlynx:
Didn't get any info on the other languages.

flaviojs:
Without files in other languages to test, it's foolish to try using text, I'll go for hashes.

I'm uneasy about the json format of ResourcePack after reading a sample json file.
I can code the json serialize/deserialize in any way, so I'd like to leave the actual json format for you guys to decide on... (shamelessly trying to dump stuff I'm not good at onto someone else)

Things that I think the format should have:
 * a display name for the launcher
 * a list of resources
 * allow arbitrary notes from the maintainer
 * allow resources that are inside different types of archives (for the game it's the inner resources that really matter, not the archive itself)
 * how to get the data of a resource (currently it's path and archive_path)
 * a way to identify the resource (different types of hashes, maybe file size too?)
 * ja2-stracciatella specific information (like which versions are supported?)

This is the current format:
```json
{
  "name": "Friendly name for display purposes",
  "properties" : {
    "note": "everything inside properties is optional, but properties itself is required",
    "note2": "when with_archive_slf is true it expects archive_path, archive_slf and archive_slf_num_entries in resource properties",
    "note3": "arbitrary properties can be added by the file maintainer (notes like an author, an url, a revision, a license, any valid json really)",
    "with_archive_slf": true
  },
  "resources": [
    {
      "path": "foo/bar.txt",
      "properties": {}
    },
    {
      "path": "baz.slf",
      "properties": {
        "note": "everything inside properties is optional, but properties itself is required",
        "note2": "archive_slf is only generated when with_archive_slf is true",
        "archive_slf": true,
        "archive_slf_num_resources": 1
      }
    },
    {
      "path": "inside/slf.txt",
      "properties": {
        "note": "this is a resource that is inside a slf file",
        "note2": "should we allow archives inside archives? archive_path would need to support it (points backwards so we know how to get the data)",
        "note3": "including archive contents as resources means that there can be multiple resources with the same path",
        "archive_path": "baz.slf"
      }
    }
  ]
}
```

lynxlynxlynx:
It looks ok, but I don't see how you can make much progress without the files. What would you hash?

flaviojs:
I can only create a resource pack for my version of the game (English v1.12 from GOG).

The launcher creates a resource pack when he wants to auto-detect the version in the game dir.
Hashing resources and recording that information is what I'll do next.
Comparing known resource packs to guess the version comes after that.

We can add a button to save/generate a resource pack to the launcher or create a standalone tool, then other people will be able to create issues to include their versions.

flaviojs:
I was going to include CRC, but you can choose the CRC32 of a file by modifying 4 bytes so it's not worth it. =~

flaviojs:
Where should the resource packs go? (json files that help identify a version)

lynxlynxlynx:
I'd put them in assets/externalized/.

flaviojs:
Btw, a friend showed me the files from Jagged Alliance (DOS game) and I saw what is probably the first version of SLF. (no extension yet)

I can add reading/writing capabilities in that format, are you interested?
It's not clear how they handled paths and there is no identifying extension so I won't be using it in the resource packs.

lynxlynxlynx:
I'm of the opinion that it would just be a maintenance burden, since there'd be no users.

flaviojs:
The json resource pack for my version (with file sizes, slf contents, and md5 hashes) is 4.2MB by default and 6.6MB with pretty print. [resource_packs.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3146292/resource_packs.zip)

I was thinking of including all the hash algorithms but it's already bigger than I though... I'm uneasy, what do you guys think about big json files?

lynxlynxlynx:
This is just for finding the unique values right? I don't see a need to have all the file hashes in the tree, let alone several copies.

flaviojs:
Guessing the version isn't reliable yet (not enough data), so I want to add a launcher button:
guess_game_version

I also want to make a command line utility to create resource packs (easier to handle).

Any objections?

lynxlynxlynx:
Well no, but I still think it's silly to have this stored in the tree. At least a separate repo if you really want it versioned.

flaviojs:
Is it possible to use staticlib stracciatella from a separate repo?

Hmm, maybe move slf.rs and res.rs to the separate repo and have stracciatella use it instead?
It looks like you already have a repo for tools, [ja2-open-toolset](https://github.com/ja2-stracciatella/ja2-open-toolset), so that will do if it's ok to mix python and rust there.
Then the next question is, what is the crate name?
It will be in your repo so you should choose the name. (ja2-open-toolset? ja2rust? ja2rs? ja2?)
It will produce a library and an executable, maybe more executables in the future.

lynxlynxlynx:
I don't mind the code being here, it's the json monsters that would needlessly increase the repo size.

flaviojs:
Now it's 34.6kB (removed slf contents and made individual hash properties).
That's what I committed and will use. It should work in my case because it already comes patched to v1.12.

The same might not work for the original version + patches, it depends on how the patch process works.

flaviojs:
Hmm... I don't understand this build error.
Do you guys know what is going on?

lynxlynxlynx:
The build system hasn't changed, so it must be a change in the image.

selaux:
Looks like rust is using the wrong linker (it should use mingw one and actually uses `/usr/bin/ld`. Maybe it's this issue: https://github.com/rust-lang/rust/issues/32859, although idk why it is surfacing right now...

selaux:
@flaviojs Can you try cherry-picking 669140636f28244c1870823d11302bc2da4ecb1f and see if that helps?

flaviojs:
Yup, it's fixed now. :)

flaviojs:
Now it just needs to put the `resource-pack` executable next to `ja2` and `ja2-launcher`.
Any tips?

The rest is ready for a final review.
I'm formatting my rust code with `rustfmt`, you might want to customize it's behavior.

The current implementation is "dumb" because there is not enough data.
Mod handling should only be added after we know resource packs of other versions.

rust/src/bin_resource_pack.rs
```diff
@@ -0,0 +1,196 @@
...
+            Arg::with_name("archive")
+                .help("Adds the archive contents as resources")
+                .long("archive")
+                .value_name("TYPE")
+                .possible_values(&["slf"])
+                .takes_value(true)
+                .multiple(true),
+        )
+        .arg(
+            Arg::with_name("datadir")
```
selaux: Is this really the data dir? Or is a directory containing the data dir? We had that mixup in the launcher once and it was not really clear to users.

flaviojs: It is the same thing that is being stored in the options and that ja2 receives as a command line argument.
Personally I find it confusing since it is actually the game dir, where a "Data" directory exists.
But it's what you use right now so I just followed along.

selaux: Yes that still is the issue I explained above. I'm fine with leaving it as it is right now and rename it everywhere at some point.

selaux:
I also just tested with my resources (Ja2 Steam version if I remember correctly). Unfortunately it didn't work. Is it possible that the resource pack json you added was not created with `--archive slf`? Because it only contains the slf files not their content as far as I can see.

selaux:
Regarding packaging you can add it to be installed [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/CMakeLists.txt#L43), which will install it into `${CMAKE_BINARY_DIR}/lib-stracciatella` and then add it [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/CMakeLists.txt#L302) which will install into the final package. I would rename it to `ja2-resource-pack` though, as it will be installed into global binaries for linux systems and it's less ambiguous that way.

rust/src/bin_resource_pack.rs
```diff
@@ -0,0 +1,196 @@
...
+                .value_name("PATH")
+                .takes_value(true)
+                .multiple(true),
+        )
+        .arg(
+            Arg::with_name("hash")
+                .help("Adds the hash of the file data to the resource properties")
+                .long("hash")
+                .value_name("TYPE")
+                .possible_values(&["md5", "sha1", "blake2s", "blake2b"])
```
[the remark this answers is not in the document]
flaviojs: It's to account for the opinions of different people regarding hash algorithms.
It's fine if you guys decide on only 1 algorithm (it's your project), but not me.

From what I know:
(1) some people don't want to touch anything considered broken (md5 and sha1)
(2) some people think "it's good enough for XYZ so it's good enough"
(3) some people think "I mostly want speed, not all this security"
(4) ...dunno about other common profiles...

blake2 is there for (1), sha1 is there for (2), md5 is there for (3).
I also considered CRC, but you can easily change 4 bytes in the file to get any CRC32 that you want, so I rejected it.

I'm in the (3) category, for me file size + md5 is enough.

rust/src/guess.rs
```diff
@@ -0,0 +1,376 @@
...
+            let b_path = b.get_str("archive_path").unwrap_or("");
+            a_path.cmp(b_path)
+        } else {
+            a.path.cmp(&b.path)
+        }
+    });
+    return resources;
+}
+
+struct GuessError {
```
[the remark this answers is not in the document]
flaviojs: Indeed, I forgot to implement the trait for this error. =~

rust/src/res.rs
```diff
@@ -0,0 +1,624 @@
...
+            if entry.state == SlfEntryState::Ok {
+                let mut resource = Resource::default();
+                let path = header.library_path.clone() + &entry.file_path;
+                resource.path = path.replace("\\", "/");
+                resource.set_property("archive_path", &slf.path);
+                // record file size
+                if self.with_file_size {
+                    resource.set_property("file_size", entry.length);
+                }
+                // TODO include archive inside archive?
```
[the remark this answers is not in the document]
flaviojs: Ok

rust/src/res.rs
```diff
@@ -0,0 +1,624 @@
...
+        return match self {
+            ResourceError::Text(desc) => desc,
+            ResourceError::IoError(err) => err.description(),
+        };
+    }
+}
+
+impl fmt::Display for ResourceError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.description())
```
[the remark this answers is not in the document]
flaviojs: Ok

rust/src/res.rs
```diff
@@ -0,0 +1,624 @@
...
+}
+
+/// Iterator that returns the paths of all the filesystem files in breadth-first order, files before dirs.
+#[derive(Debug, Default)]
+struct FSIterator {
+    files: VecDeque<PathBuf>,
+    dirs: VecDeque<PathBuf>,
+}
+
+impl FSIterator {
```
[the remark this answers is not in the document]
flaviojs: Rust already abstracts filesystem stuff with std::path/std::fs, I don't think we need a module for filesystem related stuff.

However, in my eyes this is a self-contained utility struct so I wouldn't mind moving it to a util/utils module.

rust/src/guess.rs
```diff
@@ -0,0 +1,376 @@
...
+#[derive(Default)]
+pub struct LoggedGuess {
+    pub log: String,
+    pub vanilla_version: Option<VanillaVersion>,
+}
+
+type GuessResult<T> = Result<T, GuessError>;
+
+impl LoggedGuess {
+    fn guess_vanilla_version(&mut self, gamedir: &Path) -> GuessResult<()> {
```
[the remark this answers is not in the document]
flaviojs: Not now, since the module will probably need to be redone when we have enough resource packs to know what would be a good design.

rust/src/bin_resource_pack.rs
```diff
@@ -0,0 +1,196 @@
...
+        }
+    }
+
+    let mut pack = graceful_unwrap(
+        "Creating",
+        builder.execute(matches.value_of("name").unwrap()),
+    );
+
+    if let Some(values) = matches.values_of("property") {
+        let mut name: Option<String> = None;
```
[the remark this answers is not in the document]
flaviojs: I named 2 values so `.number_of_values(2)` is implied.
That's a good way to read the values, I'll do that.

rust/src/bin_resource_pack.rs
```diff
@@ -0,0 +1,196 @@
...
+    if let Some(values) = matches.values_of("archive") {
+        for archive in values {
+            builder.with_archive(archive);
+        }
+    }
+
+    if let Some(values) = matches.values_of_os("datadir") {
+        for datadir in values {
+            let mut found = false;
+            for entry in graceful_unwrap("Reading datadir", fs::read_dir(datadir)) {
```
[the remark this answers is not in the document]
flaviojs: Look at this again, I failed to detect multiple data dirs in a case-sensitive filesystem.
Will do that and use `filter_map`.

flaviojs:
I'm not surprised it didn't detect the steam version you have.
The one I committed doesn't include slf to avoid "json monsters". https://github.com/ja2-stracciatella/ja2-stracciatella/pull/820#issuecomment-489483405, https://github.com/ja2-stracciatella/ja2-stracciatella/pull/820#issuecomment-490375053

Can you share a resource pack of the steam version with file size, md5 and archive slf so I can compare against my version and see what is different?

rust/src/bin_resource_pack.rs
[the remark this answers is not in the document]
lynxlynxlynx: I'm of the same opinion. Nobody is going to mount an attack through this and the sample size makes collisions pretty much impossible.

selaux:
Allright I tried again and the issue I discovered is that my resources dont contain `Imp.dat`, which makes the guess algorithm abort early at `return Err(format!("resource {:?} not found", want.path).into());`. If I replace these two lines with a simple `break;` (which I understand as the algorithm ignoring this file), detection works :tada:!. So my suggestion would be to do this.

Regarding the `Data/Temp` directory, it seems like it is written here: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/game/Laptop/Files.cc#L385. So indeed it makes sense to exclude those files.

selaux:
I've now downloaded all non-english versions of JA2 from steam and created resource packs: [steam_jagged_alliance_2_classic.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3179439/steam_jagged_alliance_2_classic.zip).

One more thing I noticed is that detection does not use the selected directory, but the one from `ja2.json`. I will add a comment how to fix this.

src/launcher/Launcher.cc
```diff
@@ -244,3 +245,28 @@ void Launcher::startEditor(Fl_Widget* btn, void* userdata) {
...
 }
+
+void Launcher::guessVersion(Fl_Widget* btn, void* userdata) {
+	Launcher* window = static_cast< Launcher* >( userdata );
+    auto choice = fl_choice("Comparing resources packs can take a long time.\nAre you sure you want to continue?", "Stop", "Continue", 0);
+    if (choice != 1)
+        return;
+
+    char* log = NULL;
+    if (guess_resource_version(window->engine_options, &log)) {
```
selaux: Add `window->writeJsonFile();` here in order for detection to use the currently selected directory.

selaux:
After that I think we could merge this and work incrementally in other PRs. @lynxlynxlynx do you agree?

lynxlynxlynx:
I guess it can't hurt. What I'd appreciate is an update to the intro post, so it'll be clear what was done in this PR once we have to write the changelog.

rust/src/bin_resource_pack.rs
flaviojs: I'll change the names in my code to gamedir.

rust/src/bin_resource_pack.rs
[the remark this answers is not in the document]
flaviojs: Ok, I'll keep md5 and remove the rest.

src/launcher/Launcher.cc
flaviojs: I see, the problem is that engine_options is only updated when you launch the game or the editor.

Since changing the input fields doesn't update the json, then guessing the version shouldn't update it either. I'll change the rust-C api instead.

flaviojs:
What should I do to the log text?

I think it should be displayed or kept somewhere since the current strategy is far from perfect.
Right now it's printing to the console to avoid being truncated by slog/fltk.

selaux:
I'm not too sure yet how to handle the logs. Ideally we could use the rust `log` crate and write some adapter for ourselves that then logs via SLOG, so we use the same logging setup everywhere. But I think this is out of scope for this PR.

selaux:
Thanks for all the effort that you put into the PR. I will merge now, so further rework can be done in other PRs.

PR #826, opened 2019-05-04
fixed clang Warnings

lynxlynxlynx:
also has a small vanilla ai bugfix included.

-Wall generates more-than-my-screen-buffer-full of extra warnings and I haven't tried -Wextra. I think we should clean it up and use -Werror for debug builds.

The spammer was ENUM_BITSET, which looks useless and random sampling of the users show no uses of the functions. Can probably be outright killed... It turns out that's tricky. While there are some useless users, converting to an old school templated class enum (the cxx11 keyword doesn't help — we need the methods) only almost worked. I couldn't get the bool operator to work, while the rest seemed fine. The simplest way out is to not use the enums as types in the first place.

Then there's the:
- ubLoop = ubLoop FIXME in LOS.cc:3472 (https://github.com/figment/JA2-1.13/commit/3b3ecaf2997d080a1f4442516abdb3eafe7f53d3 just removed it)
- TrashMapTile(), MapOptimize() FIXME (introduced by ja2s)

### DIFF
--- a/src/game/GameInitOptionsScreen.cc
+++ b/src/game/GameInitOptionsScreen.cc
@@ -177,7 +177,7 @@ static void GetGIOScreenUserInput(void);
 static void RestoreGIOButtonBackGrounds(void);
 static void DoneFadeOutForExitGameInitOptionScreen(void);
 static void DisplayMessageToUserAboutGameDifficulty(void);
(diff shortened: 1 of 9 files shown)


selaux:
:+1:

PR #828, opened 2019-05-05
fix a few more warnings and enable most of -Wall and -Wextra

lynxlynxlynx:


lynxlynxlynx:
This now builds everywhere, but it looks like msvc can't handle well even the flags it supports. -Wall caused a 4x increase in build time, so I suggest we just skip it there.

selaux:
Sounds good to me...

PR #830, opened 2019-05-18
Add some rust doc comments to exported things from rust/src/config

selaux:
I really liked the adding of doc comments in #820, so I added comments to all exported things in `rust/src/config`. This should give a better overview over what exists in rust code and the responsibilities of the individual modules.

To see the current docs you can run `cargo doc && firefox target/doc/stracciatella/index.html` in the `rust` subdirectory (adapt for your favorite other browser :wink:)

### DIFF
--- a/rust/src/config/engine_options.rs
+++ b/rust/src/config/engine_options.rs
@@ -4,21 +4,36 @@ use crate::Resolution;
 use crate::ScalingQuality;
 use crate::VanillaVersion;
 
+/// Struct that is used to store the engines configuration parameters
 #[derive(Debug, PartialEq)]
 pub struct EngineOptions {
+    /// Path to configuration directory in the users home directory
     pub stracciatella_home: PathBuf,
+    /// Path to the vanilla game directory
     pub vanilla_data_dir: PathBuf,
+    /// List of enabled mods
     pub mods: Vec<String>,
+    /// Resolution the game will start in
     pub resolution: Resolution,
+    /// Gamma correction parameter
     pub brightness: f32,
+    /// Vanilla game version that the user is in posession of
     pub resource_version: VanillaVersion,
+    /// Whether to show help on startup and exit
     pub show_help: bool,
+    /// Whether to run unittests on startup and exit
     pub run_unittests: bool,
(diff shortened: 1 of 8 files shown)


rust/src/config/ja2_json.rs
```diff
@@ -36,17 +37,19 @@ fn build_json_config_location(stracciatella_home: &PathBuf) -> PathBuf {
...
 
 impl Ja2Json {
+    /// Construct a Ja2Json instance from the stracciatella home directory
     pub fn from_stracciatella_home(stracciatella_home: &PathBuf) -> Self {
         let path = build_json_config_location(stracciatella_home);
         Ja2Json { path }
     }
 
-    pub fn get_content(&self) -> Result<Ja2JsonContent, String> {
+    fn get_content(&self) -> Result<Ja2JsonContent, String> {
```
lynxlynxlynx: was this change intentional?

rust/src/config/engine_options.rs
```diff
@@ -4,21 +4,36 @@ use crate::Resolution;
 use crate::ScalingQuality;
 use crate::VanillaVersion;
 
+/// Struct that is used to store the engines configuration parameters
 #[derive(Debug, PartialEq)]
 pub struct EngineOptions {
+    /// Path to configuration directory in the users home directory
```
lynxlynxlynx:
```suggestion
    /// Path to configuration directory in the user's home directory
```

rust/src/config/stracciatella_home.rs
```diff
@@ -1,6 +1,7 @@
 use dirs;
 use std::path::PathBuf;
 
+/// Find ja2 stracciatella configuration directory inside the users home directory
```
lynxlynxlynx:
```suggestion
/// Find ja2 stracciatella configuration directory inside the user's home directory
```

rust/src/config/ja2_json.rs
selaux: Yes, the function is not publicly used, so I decided to make it not public. Should have added that to a separate commit.

Issue #638, opened 2017-12-16
Launcher - ideas & minor enhancements


JAsmine-ja2:
Testing the win 64 release of 0.16.0

1) The user is able to edit the predefined resolutions. This shouldn't be possible.

1a) The user is also able to edit the language selection field. Shouldn't be possible, too.

2) The warning box when using an invalid custom resolution has no "heading" (text in the upper frame). It could have a heading like "Warning".

3) After the warning for invalid resolution pops up, the launcher could reset the custom resolution to 640x480 for added convenience. Right now, this has to be done manually.

4) When selecting an invalid data directory, the launcher shows a warning when launching the game. However, the warning is very unclear:
![launcher](https://user-images.githubusercontent.com/17213795/34071783-508c2ec0-e27c-11e7-89d9-fb2eaf133751.jpg)

4a) We also need an appropriate warning for wrong language selected. ATM, it is the same warning as 4)

5) Maybe, the launcher could already check for correct data files when selecting the data directory (and not wait until you try to launch the game).

5.1) The same is true for the editor, which needs specific files (editor.slf)

~~6) The launcher executable itself could maybe use the logo.~~ (duplicate of #607)

![logo](https://user-images.githubusercontent.com/17213795/34072393-f13d34c2-e286-11e7-8540-847ab6aca883.jpg)

7) Should the launcher be renamed to

"stracciatella launcher" ? Beacuse that is, what it does. It can't launch regular JA2.

JAsmine-ja2:
The launcher executable itself could maybe use the logo.

![logo](https://user-images.githubusercontent.com/17213795/34072393-f13d34c2-e286-11e7-8540-847ab6aca883.jpg)

JAsmine-ja2:
7) Should the launcher be renamed to

"stracciatella launcher" ?

Manuel-K:
Some of those things are duplicates of #607.

1. True, but I don't know if it is possible to prevent that.
2. I know, but I don't think FLTK supports changing the title of predefined warning boxes. If we want to create a custom form, we could choose the title.
3. IMO a bad idea. I hate it when programs do that. Imagine a small typo that you do not notice, the warning pops up and you have to start again because everything you've entered is gone.
4. Yes, the  error message is horrible and not useful for the end user, but this is not a part of the launcher. The error message is caused by ja2.exe dying – we should definitely change that.
[...]

Issue #692, opened 2018-01-14
Improve the error message when data_dir is not valid.


Manuel-K:
The current error message is cryptic at best.

Note: This is a reminder to myself.

EDIT by lynx:

> terminate called after throwing an instance of 'boost::filesystem::filesystem_error'
>   what():  boost::filesystem::directory_iterator::construct: No such file or directory: "/some/place/where/the/data/is/data"
> Aborted (core dumped)
it should at least display the path to ja2.json.

lynxlynxlynx:
This needs a recheck, since boost was ripped out and it's rust that's doing the FS stuff.

ja2-derek:
If we are only talking about the error message, this is how it looks in the new launcher:

![image](https://user-images.githubusercontent.com/63151803/78739106-23b71100-7986-11ea-903b-66f75f183573.png)

lynxlynxlynx:
No, this was the cmd. But it's fine now, closing:
> $ ./ja2
> Vanilla data directory has to be set either in config file or per command line switch

Issue #290, opened 2016-04-12
Support multiple data paths


selaux:
To ensure maximum moddability, I would change the way we load game resources (images, sounds, etc.). Currently all game resources are loaded from the data directory. For us to be able to provide patching functionality I would suggest changing the logic in the following way:

When a game resource is loaded (lets say from `Loadscreens` for example), locating it should
- First look for a `Loadscreens` library/dir in the current mod directory and look if it includes the resource then use that
- Second look for a `Loadscreens` library/dir in the `externalized/data` folder and look if it includes the resource then use that
- Finally look for the `Loadscreens` library/dir in the `data` folder and use the resource from there

This would allow the following things:
- Ship e.g. a free version of `editor.slf` mentioned in #30 with the application
- Allow modders to overwrite single resources of the game without the need to provide vanilla data, which is not allowed to be distributed, in their mods
- Allow us to provide resources that we might want to use somwhere (e.g. our shiny new logo)

BTW: This is somehow linked to https://github.com/ja2-stracciatella/ja2-open-toolset. My first goal for the open toolset would be to compose a free `Editor.slf` for Stracciatella

JAsmine-ja2:
If I understand it properly, we could with this method also make sure that the official patches are applied to the game. So, the user wouldn't have to worry about the version of the game he has.

selaux:
True, but this would require us to ship the patches with the application, which is illegal :smile:. But it would for example allow us to provide our own patches and you as a modder to e.g. change the face of a single merc (or is this already possible and I dont get it?)

lynxlynxlynx:
We use multiple search paths and separation for unhardcoded and overriden files to great effect in GemRB. But I'd never call it patching. I think it will just confuse people, as you're talking about swapping assets, not modifying them.

derSteFfi:
I like the idea, not only for modding but also for the editor for example. If documented clearly and properly, this is really good.

ja2-derek:
I think this is all done.

PR #833, opened 2019-05-20
Rework launcher.

### LINKED ISSUE #630: Need two rust functions to implement display of game.json location in the launcher
`find_ja2_executable_directory` that returns the full path of the executable directory.
`find_ja2_executable_name` that returns the name of the executable.

I'd drop `find_ja2_executable` instead. It seems to be rather hacky. Unless I'm mistaken, it depends on the length of the launcher executable's name.

flaviojs:
GUI reorganized and uses tabs.
Show location of game.json. (respects EXTRA_DATA_DIR)
Change predefined resolutions to a menu button.
The window is resizable.

Reference: starting from https://github.com/ja2-stracciatella/ja2-stracciatella/issues/607#issuecomment-493200823

Fixes #630 

Show the location of ja2.json at the bottom (can be copied)
Add visual reminder when the resolution is invalid (inspired by 3. in #638)
Add title to the invalid resolution alert (inspired by 2. in #638)
Add visual indicator when the engine settings changed. (`*` prefix in ja2.json path) 
Add reload and save buttons.
Add confirmation dialog when the data dir or editor.slf are not found.


flaviojs:
Now that I have a good grasp of the launcher, I kinda wish the lib and executable roles were reversed.
I'm imagining the C++ launcher exposing only a GUI as a lib, and the actual logic being in rust.
But that's out of scope for this PR and too advanced for my current rust skills. xD

flaviojs:
Screenshots:
tab_play
tab_data
tab_settings

lynxlynxlynx:
Looks ok to me. Not a user though.

flaviojs:
Now the location of ja2.json appears at the bottom:
ja2_json_path_at_bottom

src/launcher/StracciatellaLauncher.cc
```diff
@@ -118,7 +118,12 @@ StracciatellaLauncher::StracciatellaLauncher() {
       o->end();
       Fl_Group::current()->resizable(o);
     } // Fl_Tabs* o
-    stracciatellaLauncher->size_range(465, 300);
+    { ja2JsonPathOutput = new Fl_Output(0, 300, 465, 25);
+      ja2JsonPathOutput->tooltip("ja2.json contains the stracciatella engine options");
+      ja2JsonPathOutput->color(FL_BACKGROUND_COLOR);
+      ja2JsonPathOutput->align(Fl_Align(FL_ALIGN_LEFT|FL_ALIGN_INSIDE));
+    } // Fl_Output* ja2JsonPathOutput
+    stracciatellaLauncher->size_range(465, 325);
```
lynxlynxlynx: What good does displaying the path to ja2.json do? The launcher exists in great part to avoid the need to know about this file, plus it has only two keys, both of which are configurable through the launcher.

flaviojs: For power users, `ja2.json` is essentially a file representation of the launcher options. (or the launcher a GUI for the file)

It just makes sense to me, since there is no help menu in the launcher.

lynxlynxlynx: Must be the launcher then, since we don't save command line options. But using it that way you lose the help — the default has a comment inside, the launcher has tooltips, so I don't understand the argument.

lynxlynxlynx: btw, there was some talk about extra search paths here: #290

flaviojs: I never saw the default comment because I used the launcher to launch the game (that's when the file is saved) way before learning of the file.

I don't see an appropriate place where this path could go as a tooltip and I would lose the ability to copy the path. And if this was a tooltip, shouldn't the game settings path also be a tooltip?

lynxlynxlynx: My argument is that people using the launcher don't need to know about the file at all. #692 is a reminder there's some work to be done for the users that skip the launcher (but can wait if we're swapping the VFS layer).

Also, what would happen if the path was longer than the textbox size?

@selaux @JAsmine-ja2 we need more opinions. :abcd:

flaviojs:
Now there is a visual queue when the resolution is invalid:
invalid_resolution_label

src/launcher/StracciatellaLauncher.cc
flaviojs: When the path is longer you can scroll left and right, like a normal input/output field, or just select everything by double clicking or ctrl+a and copy. (that's an output field, like the game settings)

flaviojs: Uhuu, now I noticed... when I edit that file in a text editor it usually shows the path of the file one way or another, in my current one it's "ja2.json (/home/hp/.ja2)".

I view the launcher as a specialized editor for that file, so that explains why I think it fits.
Now it makes me want to add a save button, a reload button, and show '*' when the settings are modified but not saved, but I'll wait for your feedback first. xD

selaux:
One thing I would add is an area on the play screen where it displays a warning when the data dir is not set correctly (set to the default value).

flaviojs:
Visual indicator when the engine settings changed.
Reload and save buttons.
Visual warning when the data dir or editor.slf not found.
(buttons still work since it only checks when a setting changed)
warnings_reload_save

src/launcher/StracciatellaLauncher.cc
flaviojs: Implemented it since selaux didn't comment on this.

rust/src/stracciatella.rs
```diff
@@ -288,6 +289,89 @@ pub unsafe extern "C" fn guess_resource_version(
...
+        }
+    } else {
+        return ptr::null_mut(); // no home
+    }
+}
+
+/// Returns true if it was able to find path relative to base.
+/// Makes caseless searches one component at a time.
+#[no_mangle]
+pub extern "C" fn check_if_relative_path_exists(
```
selaux: I think this needs to be case insensitive. For me the warnings always show.

src/launcher/Launcher.cc
```diff
@@ -233,6 +226,43 @@ void Launcher::startExecutable(bool asEditor) {
...
 
+bool Launcher::resolutionIsInvalid() {
+	return resolutionXInput->value() < 640 || resolutionYInput->value() < 480;
+}
+
+void Launcher::update(bool changed, Fl_Widget *widget) {
+	// not found warnings
+	if (check_if_relative_path_exists(dataDirectoryInput->value(), "Data", true)) {
+		dataDirNotFound->hide();
+		if (check_if_relative_path_exists(dataDirectoryInput->value(), "Data/Editor.slf", true)) {
```
selaux: I would only show that when the editor button clicks. For most of the people who don't use the editor is is irrelevant.

flaviojs: I like consistency, if the editor button only shows a warning when clicked, then the play button should do the same.

rust/src/stracciatella.rs
flaviojs: I'll look into it.

flaviojs: Fixed.

flaviojs:
Visual warnings replaced with confirmation on click (when not found):
data_dir_not_found
editor_not_found

src/launcher/Launcher.cc
flaviojs: Done.

lynxlynxlynx:
Is this done, are you happy with it?

flaviojs:
I'm happy with it.

selaux:
Tested it out. Very happy with the result as well. Thanks @flaviojs

lynxlynxlynx:
Thanks indeed! :fireworks:

PR #834, opened 2019-06-21
Fix tactical interface rendering on strategic interface.

maxrd2:
This should fix cause of those 3 bug reports.


flaviojs:
Now this looks like a proper fix. :)
I confirm that it fixes those issues.

lynxlynxlynx:
Very nice, thanks!

Issue #719, opened 2018-06-28
"Unable to create directory" message in the launcher


rybakolbasa:
### What is wrong
Hi everybody!

I downloaded the latest JA2S for Win, installed it, and ran the launcher to set the data directory and resolution. Then I ran the game from the launcher and got this:

`2018/05/28 17:56:39 ERROR   [File Man        ] Unable to create tmp directory 'C:\%userprofile%\Documents\JA2\tmp'
`
where %userprofile% is my Windows user account in Users.

Apparently, the game does not run.

There is no tmp folder at the above path. Creating it manually does not help. There is, however, the json file in the /JA2 with the settings I entered in the launcher. The data_dir points to the folder where JA2 is installed (not to the Data folder).

```
{
  "data_dir": "D:\\games\\Jagged Alliance 2",
  "mods": [],
  "res": "1280x720",
  "resversion": "ENGLISH",
  "fullscreen": false,
  "debug": false,
  "nosound": false
}
```

The %userprofile% contains some Russian symbols. I tried to change the temp/tmp environment vars systemwide to something ASCII-only, but to no avail. The launcher still looks for the old path.

The original version of the game starts ok.

Also, for some reason the date stamp in the error log displays May, not June.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.16.1 | JA2 GOG English |    no    |    yes/no     |

OS=Windows 8/64

Any thoughts?

selaux:
Hi, sorry for the late response. Can you try to reproduce this with version `0.15.0`? This will give us a hint where to look for the issue.

selaux:
NB: You need to follow the "Without the optional launcher" instructions from here, as the old version does not have the launcher yet: https://ja2-stracciatella.github.io/how-to-run/

rybakolbasa:
@selaux No problem, I've waited like 15 years already, so a couple of weeks is nothing :)

Here is what I get with 0.15.1:

```
---------------------------
Jagged Alliance 2
---------------------------
ERROR: caught unhandled exception:
boost::filesystem::directory_iterator::construct: Системе не удается найти указанный путь: "\data"
---------------------------
ОК   
---------------------------
```

This means "The system cannot find the specified path".

Here are the contents of the JA2.ini:
`[Ja2 Settings]
[...]

PR #835, opened 2019-06-24
Rustify LibraryDB

flaviojs:
The main differences are:
 * it is thread safe
 * shadowing respects library order and uses the full file path,
   the original shadowed whole libraries based on the longest base path

Before this I actually rustified FileMan and only the free disk space check wasn't possible. I was trying to match the C/C++ interface and the result looked really messy.

Learning from that mess, this time I made a module for rust usage and only then figured out an interface for C. The result looks good. The librarydb module must be thread safe to ensure that calls from C respect the borrow rules of rust.


flaviojs:
Tests are missing for now, first I want to know what you think of the layout.
I'm isolating code that interacts with C in the `c` module and aiming to minimize the logic it contains.
I think the C functions in `stracciatella.rs` should be moved to appropriate submodules of `c`.

rust/src/librarydb.rs
```diff
@@ -0,0 +1,371 @@
...
+}
+
+/// Library file.
+/// Provides access to the data of a library entry through Read and Seek.
+#[derive(Debug)]
+pub struct LibraryFile {
+    /// The index of the entry of this file.
+    index: usize,
+    /// The thread safe library of this file.
+    arc_library: Arc<RwLock<Library>>,
```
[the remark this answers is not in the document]
flaviojs: It would also need a mutex to ensure thread safety from C calls.
Not being linked to the library means not being able to implement a writable library file (not implemented).
Is it desirable to make it self contained? I think it's better like this, I can even check that there are no open library files when the library is dropped (artificial restriction).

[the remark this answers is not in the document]
flaviojs: And it would also need the file handle cloned or reopened for itself.

flaviojs:
I think this is done.

lynxlynxlynx:
Rust is beyond me, but I have a different proposal: would you like to be part of our ragtag team? It doesn't change much, since everyone's changes still have to go through review, but you deserve it.

flaviojs:
Sure, if you don't mind having someone that might disappear to chase something else and become unresponsive.

Right now I'm learning rust and trying to contribute useful code while doing so.
I really like rust, the fights with the borrow checker are teaching me good stuff.
It feels like my C/C++/other programming language skills improve just from learning rust. ^^

lynxlynxlynx:
That description fits pretty much everyone here. :) But I'll wait for either Stefan to chime in first.

selaux:
Sounds good to me :+1:. I hope I find some time to review this PR soon as well...

rust/src/librarydb.rs
```diff
@@ -0,0 +1,504 @@
...
+    }
+
+    /// Opens a library file for reading.
+    /// The file must be dropped before the library database is dropped.
+    pub fn open_file(&self, path: &str) -> io::Result<LibraryFile> {
+        self.locked().open_file(path)
+    }
+}
+
+impl LibraryDBInner {
```
selaux: Do we need this `Inner` construct?  As long as we only have `arc_libraries` we should be good without it right?

rust/src/librarydb.rs
```diff
@@ -0,0 +1,504 @@
...
+                });
+            }
+        }
+        Err(io::ErrorKind::NotFound.into())
+    }
+}
+
+impl Library {
+    /// Opens a library for reading.
+    fn open_library(data_dir: &Path, library: &Path) -> io::Result<Self> {
```
selaux: I would just call this `open` as `library` is already the struct name.

rust/src/librarydb.rs
```diff
@@ -0,0 +1,504 @@
...
+            library_path,
+            library_file,
+            base_path,
+            entries,
+        })
+    }
+
+    /// Finds the file entry that matches the case-insensitive path.
+    /// Returns the entry index or None.
+    fn find_with_case_insensitive(&self, path: &str) -> Option<usize> {
```
selaux: I would just call this `find_case_insensitive`

rust/src/librarydb.rs
```diff
@@ -0,0 +1,504 @@
...
+                            Ok(file_type) => Some((x.file_name(), file_type)),
+                            _ => None,
+                        }) // (OsString,FileType)
+                        .collect();
+                    if !found.is_empty() {
+                        if found.iter().any(|x| &x.0 == os_str) {
+                            path.push(os_str); // perfect match
+                            continue;
+                        }
+                        let want_ascii_uppercase = want.to_ascii_uppercase();
```
selaux: Same as above, but paths are definitely not only ascii, so `to_uppercase` should be used here...

src/sgp/LibraryDataBase.cc
```diff
@@ -1,3 +1,4 @@
+#if 0 // rustified
```
selaux: Should be removed then, right?

src/sgp/LibraryDataBase.h
```diff
@@ -1,6 +1,9 @@
 #ifndef _LIBRARY_DATABASE_H
 #define _LIBRARY_DATABASE_H
 
+#include "stracciatella.h"
+
+#if 0 // rustified
```
selaux: Should be removed then, right?

src/sgp/SGPFile.h
```diff
@@ -16,6 +16,9 @@ enum SGPFileFlags
 	SGPFILE_REAL = 1U << 0
 };
 
+struct LibraryFile;
+
+#if 0 // rustified
```
selaux: Should be removed then, right?

rust/src/librarydb.rs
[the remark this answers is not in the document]
selaux: Fine for me to keep it that way...

rust/src/c/librarydb.rs
```diff
@@ -0,0 +1,231 @@
...
+#[no_mangle]
+pub extern "C" fn LibraryFile_Close(file: *mut LibraryFile) {
+    let _drop_me = from_ptr(file);
+}
+
+/// Seeks a library database file.
+/// `from` expects FileSeekMode values (see src/sgp/SGPFile.h)
+/// Sets the rust error.
+#[no_mangle]
+pub extern "C" fn LibraryFile_Seek(file: *mut LibraryFile, distance: i64, from: c_int) -> bool {
```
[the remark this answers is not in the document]
flaviojs: Because it's expecting FileSeekMode values (not defined in rust) and C/C++ treats it as an int (not a class enum). It will probably be changed to an enum later, when more filesystem stuff gets rustified.

rust/src/c/mod.rs
```diff
@@ -0,0 +1,241 @@
+//! This module and it's submodules contains code for C.
```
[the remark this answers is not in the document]
flaviojs: I'll do that after this is merged. :)

rust/src/librarydb.rs
```diff
@@ -0,0 +1,504 @@
...
+    } else if b < 0 {
+        a.checked_sub(-b as u64)
+    } else {
+        Some(a)
+    }
+}
+
+/// Gets the case insensitive representation of a path in the library database.
+fn case_insensitive_path(path: &str) -> String {
+    path.to_ascii_uppercase().replace('\\', "/")
```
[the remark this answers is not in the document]
flaviojs: Because unicode is [complicated](https://www.w3.org/International/wiki/Case_folding)... a true case insensitive comparison is not just converting to uppercase/lowercase and comparing.

I'm not confident that my implementation would be correct in all cases, so it's better to stick to ascii until rust can do case insensitive comparisons.

rust/src/c/librarydb.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    #[test]
+    fn reading() {
+        let dir = TempDir::new("c_librarydb").unwrap();
+        let data_dir = dir.path();
+        make_data_slf(&data_dir);
+        make_foo_slf(&data_dir);
+        make_foobar_slf(&data_dir);
+
+        // relative path of library is case insensitive, read works, seek works
```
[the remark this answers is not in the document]
flaviojs: These tests are for the C side so I'll eventually move them to C++ unit tests.
I'll leave these as they are for now and split the ones in module `librarydb`.

rust/src/librarydb.rs
flaviojs: It splits simple access (`LibraryDB`) and guarded access (provides access to `LibraryDBInner`).
I'm not happy with the name but it is needed because we have custom functions (I can't do `impl Vec<Arc<RwLock<Library>>>`).

rust/src/librarydb.rs
flaviojs: Can you give me an example of an existing non-ascii path?
I only found ascii in the original game data.

lynxlynxlynx: Paths can be absolute here, so things like ~gyerk/trúgȧmês/ja2/...

selaux: Yes. E.g. https://github.com/ja2-stracciatella/ja2-stracciatella/issues/719 is a bug that happens because we dont handle non-ascii paths correctly...

rust/src/c/librarydb.rs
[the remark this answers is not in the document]
selaux: Sounds good...

rust/src/c/librarydb.rs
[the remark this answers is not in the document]
selaux: Fine for me.

rust/src/librarydb.rs
flaviojs: The `data_dir` part is not treated as case insensitive so it's not a problem.
Only the `file_name` part is treated as case insensitive.

Hmm, currently it is possible to exit `data_dir` while traversing `file_name` and that is not good.
Will make it panic in that case.

flaviojs:
Developer guidelines should be documented somewhere. I was under the impression that old code should be commented out because of some comments in the source code (forgot where I saw it).

Regarding your guidelines so far I've learnt that:
1) I should wait for at least one ACK before merging
2) I should remove the old code when I rustify stuff

rust/src/c/librarydb.rs
```diff
@@ -0,0 +1,231 @@
...
+#[no_mangle]
+pub extern "C" fn LibraryDB_AddLibrary(
+    ldb: *mut LibraryDB,
+    data_dir: *const c_char,
+    library: *const c_char,
+) -> bool {
+    let ldb = unsafe_mut(ldb);
+    let data_dir = path_from_c_str_or_panic(unsafe_c_str(data_dir));
+    let library = path_from_c_str_or_panic(unsafe_c_str(library));
+    match ldb.add_library(data_dir, library) {
```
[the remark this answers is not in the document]
flaviojs: Hmm, I have a feeling there is a better way to handle this in rust.
I'll note down the idea and get back to it after moving the other C stuff into the `c` module.

rust/src/librarydb.rs
flaviojs: Done.

rust/src/librarydb.rs
flaviojs: Renamed it to `find`and clarified what it expects.

flaviojs:
Only the ascii case sensitivity comments are left. Are they resolved?

selaux:
True that there are only ascii names in data directories, but thinking of mods we should support case insensitive unicode. I think we dont need to implement this ourselves, but could use, e.g. https://docs.rs/unicase/2.4.0/unicase/

selaux:
And yes, we should document contribution practices somewhere, I agree...

flaviojs:
Ok. There is other case insensitive code that needs updating so I'll merge this as is and make a new PR.

PR #838, opened 2019-07-12
Support json with comments in rust.

flaviojs:
Ignores comments when reading and can write comments.
RapidJSON can also read json with comments.
Note that json with comments is invalid json and valid json5/hjson. (except shell style comments)

The help field in DEFAULT_JSON_CONTENT (ja2.json) can be written as the comment of data_dir. Should I do that?


selaux:
Yes, otherwise it isn't used anywhere :wink:.

flaviojs:
Personally I would just switch to a format that supports comments, like toml ([v0.5.0](https://github.com/toml-lang/toml/blob/master/versions/en/toml-v0.5.0.md) is considered stable and v1 should be backwards compatible with it).

But regardless of the format, it should be possible to write config comments from the code.
That's what drove me to create this PR.

lynxlynxlynx:
Looks fine to me. I don't see a point in changing formats yet again though — more than just configs are in json.

flaviojs:
So... can I merge this? I have the ability to merge now but dunno what's your policy regarding this kind of thing.

lynxlynxlynx:
It just needs another ack, there's no problem with self-merging. They were a bit implicit here, sorry. And thanks. :)

PR #839, opened 2019-07-19
Change all missnamed data dirs to game dirs

flaviojs:
To preserve backwards compatibility, ja2.json treats data_dir as an alias to game_dir.

The serde field attribute 'alias' requires serde_derive >= 1.0.85.
cbindgen was blocking serde_derive updates, so it was updated to 0.9.
The lock file also had to be updated.

### DIFF
--- a/rust/src/config/cli.rs
+++ b/rust/src/config/cli.rs
@@ -33,6 +33,12 @@ impl Cli {
         opts.long_only(true);
         opts.optmulti(
             "",
+            "datadir", // TODO remove this deprecated option in version >= 0.18
+            "Set path for the vanilla game directory. DEPRECATED use -gamedir instead.",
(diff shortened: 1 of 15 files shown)


rust/src/config/cli.rs
```diff
@@ -33,9 +33,9 @@ impl Cli {
         opts.long_only(true);
         opts.optmulti(
             "",
-            "datadir",
-            "Set path for data directory",
-            DATA_DIR_OPTION_EXAMPLE
+            "gamedir",
```
selaux: Same as with the JSON property, I would support both for at least one version...

selaux:
Except for the one comment LGTM.

rust/src/config/cli.rs
flaviojs: Done.

Issue #321, opened 2016-05-24
Multiple mod support


lynxlynxlynx:
Currently only one mod can be used at a time, unless you manually merge their files. We should support several, at least when their overrides don't conflict (=they change different files).

For cmdline, maybe -mod should accept a csv list, but the planned tlk launcher #270 could have a nice checkbox.

However, you can probably easily break savegames by using complex mods and then loading without them or others. Do we need to store the list of mods per save?

For an easier discussion: IE games just store the list per installed game, so it's a simple log in the game dir. All good modding is done through weidu, which is basically a package manager for mods and allows multiple modifications to the same file. I don't think the ja2 community has anything similar, but it was 10 years since I last checked.

JAsmine-ja2:
Something like weidu does not exist for JA2. Even some simple kind of mod manager with a basic GUI that gives some kind of overview over your installed mods doesn't exist.

To be honest, I can't (yet) imagine how multiple modifications to the same file can turn out to anything useful in the context of JA2. But that's probably because my imagination is... limited here :) :)

Gennady had a different idea concerning this issue - you've probably read it already: [(https://github.com/ja2-stracciatella/ja2-stracciatella/wiki/ModPack)]

What I don't like about his idea is the thing of merging mods like NightOps and Wildfire. It is imo impossible. You can't take only the weapons from NO and the maps from WF because the maps from WF have their own pre-placed items from WF. The item indices from WF and NO are incompatible and you'll end up with weird stuff lying around.

A more simplistic approach (my proposal) to multiple mod support could be to have several mods in a mod-folder that are either shipped together with JA2 Stracciatella or it's available as a optional thing.
The mods are put into the folder by the Stracciatella team after they were tested for compatibility to each other.
The launcher #270 checks if you have the "official" Stracciatella mod folder present and the launcher also "knows" which mods are compatible to each other. You can check with boxes which mods you want to use and the launcher greys out incompatible stuff to help you. Maybe you can even override it to your own risk.

This is just a basic idea.

selaux:
We also have to consider savegame compatibility. Currently if you use a mod, you get a different savegame folder, where the saves for this mod are stored. Not sure how to do that with multiple mods (maybe track the mods inside the savegame). Any ideas?

JAsmine-ja2:
Sounds good...
- Track the mods active in the game.
- Have a tooltip showing used mods when hovering over a savestate in "load game".
- Warn about changed mods when trying to load a savegame.

[...]

PR #840, opened 2019-07-21
Multiple mods + launcher

flaviojs:
The launcher can edit the mods in `ja2.json`.
The game can use multiple mods. (order matters)

Launcher screenshot:
mods
Those checkboxes are for selecting which mods you want to move up/move down/remove.

Addresses most of #321

### DIFF
--- a/src/sgp/SGP.cc
+++ b/src/sgp/SGP.cc
@@ -380,12 +380,21 @@ int main(int argc, char* argv[])
 
 	if(get_number_of_mods(params) > 0)
 	{
-		char* rustModName = get_mod(params, 0);
-		std::string modName = std::string(rustModName);
-		free_rust_string(rustModName);
-		std::string modResFolder = FileMan::joinPaths(FileMan::joinPaths(FileMan::joinPaths(extraDataDir, "mods"), modName), "data");
+		std::vector<std::string> modNames;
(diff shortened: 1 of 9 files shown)


src/sgp/SGP.cc
```diff
@@ -380,12 +380,21 @@ int main(int argc, char* argv[])
 
 	if(get_number_of_mods(params) > 0)
 	{
-		char* rustModName = get_mod(params, 0);
-		std::string modName = std::string(rustModName);
-		free_rust_string(rustModName);
-		std::string modResFolder = FileMan::joinPaths(FileMan::joinPaths(FileMan::joinPaths(extraDataDir, "mods"), modName), "data");
+		std::vector<std::string> modNames;
+		std::vector<std::string> modResFolders;
+		auto n = get_number_of_mods(params);
```
lynxlynxlynx: This can be moved up, so it's looked up only once.

lynxlynxlynx:
LGTM, but I'm useless at rust. I've referenced the feature request this fulfills and basically only some warnings are missing. Using mods changes the savegame names, so it might be confusing to users why some saves are or aren't available.

src/sgp/SGP.cc
flaviojs: Done.

PR #843, opened 2019-07-28
Nfc tests

flaviojs:
In #842 I didn't add tests because I wasn't sure what to test.

Here I'm doing the same tests as the `caseless` crate but with the `unicode::Nfc` and also the examples in http://www.unicode.org/reports/tr15.

~There is one TODO left and~ I still want to add some `String/str` related tests if I can think of some.


flaviojs:
I looked at the String and str tests but didn't find useful stuff, so I'll give up on additional tests.

Issue #777, opened 2019-01-26
Attacking: Different result in Stracciatella vs. Vanilla


JAsmine-ja2:
### Attacking leads to a different outcome, depending on game version: Stracciatella vs. Vanilla
Load this savegame (from  #189) in JA2S and attack the enemy with Johny. Johny will hit the enemie's torso, regardless of the aiming level.
Load the same savegame in Vanilla (I tried 1.12 Gold) and perform the same attack. Johny will MISS the enemy.
This should not happen.
[QuickSave.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/2798958/QuickSave.zip)

JAsmine-ja2:
We have a similar situation here: #755
(one seed leading to different results in JA2S vs Vanilla)

chmallon:
The (pseudo-)random number generator works differently than in vanilla.
(actually two: Random() and PreRandom())
Originally it did '$RANDOM * range / RAND_MAX % range'.
(this also has several problems, which are not relevant here)
On Windows RAND_MAX is 32767, so this (somewhat) works.
On platforms with larger RAND_MAX, typically about 2**31, this breaks when doing the calculation in 32 bit.
So due to the altered calculation, you get different results.

For PreRandom() this could be restored to vanilla behavior, because this cycles over the same 256 numbers over and over anyway.
(See vanilla commented out comment in PreRandom().)
The values in guiPreRandomNums[] could be reduced to 32767 and the calculation restored -- while replacing RAND_MAX by 32767.
This should restore consistent behavior for PreRandom() with vanilla.

Consistent behavior for Random() is not even given for vanilla, because it uses a time-based seed.

flaviojs:
This is just my opinion.
For me having a non-uniform random distribution is a bug and it's great that it has been fixed.
Therefore different results from vanilla are expected and this is not a bug and neither is.

On the flipside.
It is understandable that some players would want the exact same result as vanilla, so for them this is considered a bug.

A `game.json` option could be added to get the original behavior, but I think that is something that will eventually break.
All it takes is a single bugfix that leads to calling one more or one less `PreRandom` to start getting different results.
An example of such a fix is #812, which doesn't affect combat because there are no crows during combat.

PR #846, opened 2019-07-31
Rework Random

flaviojs:
1. Use C++ facilities to generate uniform pseudo-random integer numbers.
2. Replace leftover uses of `rand()` with `Random`.
3. Add game policy to allow not recycling prerandom numbers.
    Vanilla had this commented out in the name of optimization, but today computers are much more powerful and this is not a hot path so I don't see value in this optimization.

This PR is based on the opposite view of #777.
For me non-uniform pseudo-random numbers is a bug (`PreRandom` was not uniform) and issue #777 is not a bug.

### DIFF
--- a/src/sgp/Random.cc
+++ b/src/sgp/Random.cc
@@ -1,69 +1,89 @@
 #include "Random.h"
-#include <stdlib.h>
-#include <time.h>
 
+#include <chrono>
+#include <random>
+
+/// Pseudo-random number engine.
+static std::mt19937 gRandomEngine;
+
+/// Uniform distribution in the range [0,UINT32_MAX].
+static std::uniform_int_distribution<UINT32> guiDistribution(0, UINT32_MAX);
+
+// Pregenerated pseudo-random numbers.
 UINT32 guiPreRandomIndex = 0;
 UINT32 guiPreRandomNums[ MAX_PREGENERATED_NUMS ];
 
+/// Pre-generated pseudo-random number engine.
+struct PreRandomEngine {
+	typedef UINT32 result_type;
+	static constexpr result_type min() { return 0; }
+	static constexpr result_type max() { return UINT32_MAX; }
+	result_type operator()()
+	{
+		// Extract the current pregenerated number
+		UINT32 uiNum = guiPreRandomNums[ guiPreRandomIndex ];
+
+		// Replace the current pregenerated number with a new one.
+		// NOTE the original code has this commented out in the name of optimization.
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
LGTM, but I'd go one step further. Since the original prerandom range is nowadays much larger, I see no point in the extra setting. I'd also remove prerandom and prechance completely, since the speed difference should be negligible.

flaviojs:
I wouldn't remove `PreRandom/PreChance`. It provides some predictability with save-scumming during combat, which is a valid strategy.

As for the new extra setting, I'll gladly remove it if we decide to not recycle prerandom numbers.
@selaux your thoughs?

lynxlynxlynx:
The setting would then better suit this switch: whether to use prerandom at all or redirect to random. :)

selaux:
Had a look at the PR and same as @lynxlynxlynx, I would remove the switch, never recycling numbers sounds good to me. I would not even add a switch between random and prerandom, as I don't really see a good usecase for that (except for saveloading yourself through the game :wink:)

flaviojs:
Removed the game policy and reworked the commits for clarity.

@selaux I didn't understand, do you also want direct random values to be used instead of pregenerated values in PreRandom?
I'm not into that but I'll implement it if both of you agree it should be like that.

Btw, when this is merged #777 should lose the `bug` label and be closed with  a `wontfix` label.
It's an "either-or" situation.

src/sgp/Random.cc
```diff
@@ -1,69 +1,89 @@
...
+/// Uniform distribution in the range [0,UINT32_MAX].
+static std::uniform_int_distribution<UINT32> guiDistribution(0, UINT32_MAX);
+
+// Pregenerated pseudo-random numbers.
 UINT32 guiPreRandomIndex = 0;
 UINT32 guiPreRandomNums[ MAX_PREGENERATED_NUMS ];
 
+/// Pre-generated pseudo-random number engine.
+struct PreRandomEngine {
+	typedef UINT32 result_type;
```
lynxlynxlynx: I'd remove this typedef, since it doesn't add much (pretty obvious the types should match) and max() uses a fixed constant anyway.

flaviojs: Can't be removed, `std::uniform_int_distribution` requires that typedef and the other member functions.

selaux:
@flaviojs I just wanted to emphasize that there is a reason for having random and prerandom. Other than that I think it is fine now...

PR #849, opened 2019-08-09
Fix out of bounds access in DecideActionRed.

### LINKED ISSUE #378: Out of bounds access in DecideAction.cc:1757:41
It can happen that BestThrow.bWeaponIn == NO_SLOT and NO_SLOT == -1.
I'm not sure, it may be that BestThrow.bWeaponIn is always -1 in this place.

flaviojs:
BestThrow.bWeaponIn can be NO_SLOT (-1) which leads to a silent out of
bounds access and an usItem value that I can't predict.

Closes #378

### DIFF
--- a/src/game/TacticalAI/DecideAction.cc
+++ b/src/game/TacticalAI/DecideAction.cc
@@ -1749,13 +1749,14 @@ INT8 DecideActionRed(SOLDIERTYPE *pSoldier, UINT8 ubUnconsciousOK)
 
 			return(AI_ACTION_TOSS_PROJECTILE);
 		}
-		else if (BestThrow.bWeaponIn != NO_SLOT)		// toss/throw/launch not possible
+		else		// toss/throw/launch not possible
 		{
 			// if this dude has a longe-range weapon on him (longer than normal
 			// sight range), and there's at least one other team-mate around, and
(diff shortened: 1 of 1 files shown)


src/game/TacticalAI/DecideAction.cc
```diff
@@ -1749,7 +1749,7 @@ INT8 DecideActionRed(SOLDIERTYPE *pSoldier, UINT8 ubUnconsciousOK)
 
 			return(AI_ACTION_TOSS_PROJECTILE);
 		}
-		else		// toss/throw/launch not possible
+		else if (BestThrow.bWeaponIn != NO_SLOT)		// toss/throw/launch not possible
```
lynxlynxlynx: 1. What about the case of !BestThrow.ubPossible && BestThrow.bWeaponIn == NO_SLOT? It's not handled now, but went into this block before.
2. The comment is now confusing, since taken alone it looks wrong.

flaviojs: It will continue to the code after the block, like it should (no item = no range).
I'll move the check to the inner `if`, that should take care of the confusing comment(s).

lynxlynxlynx: Yeah, much better, thanks.

Issue #726, opened 2018-09-29
Cross-build for Windows on Linux Debian using MinGW errors


arnodick:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

Hi Stracciatella folks,

I'm trying to get Stracc compiling for Windows on a Debian Linux machine and running into issues.

I've got the build portion of it working (I think?) and am at the point where I run make.

There seems to be a bug in the build process. The make process reports:
`target 'rust/CMakeFiles/rust-build' failed`

This is referring to the file JA2PROJECT/_bin/rust/CMakeFiles/rust-build.dir/build.make. If I edit this file by commenting out the line:
'rust/CMakeFiles/rust-build'
then the make process continues, but fails later saying:
`/usr/lib/gcc/x86_64-w64-mingw32/4.9-win32/include/c++/bits/basic_string.h:3017:46: error: 'WINvswprintf' is not a member of 'std'`

This seems to be an issue similar to the Windows Visual Studio build, which gave errors about the wrong strings.h/string.h being used.

I'm not sure if there is actually supposed to be a rust/CMakeFiles/rust-build file that is not being created and should be there, or if it's a bug and the cmake process is writing a line that just shouldn't exist.

Some other issues I ran into, that might help others who are trying to do this cross build:
You need mingw-w64, not just normal mingw! Use:
`sudo apt-get mingw-w64`
Rust doesn't include the standard library for Windows cross-compilation for reasons that are beyond me, and you have to install it yourself.
`rustup target add x86_64-pc-windows-gnu`

A side question: is the cross build meant to make the actual JA2 Stracc launcher, or a Visual Studio project? I only ask because if it is the latter then I'll move on to trying other build methods, because I already have a VS project building (but not compiling).

Thanks for any help! Looking forward to getting Stracc building consistently, and am happy to help in any way I can.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.16 | JA2 Gold English  |    no    |    no?     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[...]

PR #854, opened 2019-08-16
Add note about rustup and cross-compilation

juzza1:
Got some errors about a missing "core" crate for rust when compiling on Debian for Windows, this fix solved the issue. Originally mentioned in https://github.com/ja2-stracciatella/ja2-stracciatella/issues/726

### DIFF
--- a/COMPILATION.md
+++ b/COMPILATION.md
@@ -45,6 +45,12 @@ cmake -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain-mingw.cmake path/to/source
 make
 ```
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks, LGTM. But I'll let our resident _rustees_ take another look.

flaviojs:
`x86_64-pc-windows-gnu` is for 64-bit, and `i686-pc-windows-gnu` is for 32-bit.

These days 64-bit is the norm, but ja2-stracciatella be also be compiled for 32-bit so both toolchains should be mentioned.

lynxlynxlynx:
Thanks everyone!

PR #856, opened 2019-08-18
Rewrite guessing vanilla version

selaux:
This rewrites guessing vanilla version to some degree. Motivation for the change is mostly that I think it never worked as intended on my machine (because of not matching paths case insensitively). I found the logic hard to follow as well. 

### Major Changes
- When building resource packs, resources are now read in parallel using rayon
- Iterators and HashMap are used instead of for loops
- Case insensitivity has been introduced
- When comparing resource packs, we first check for language specific resources (e.g. `german` subdirectory), this is just faster (at least on my machine) and less error prone. It will return a game version if ONLY files for this version are found
- We compare resource packs second, in parallel. We calculate percentages for each resource pack, e.g. how many files are missing in data dir that are in the resource pack. We then use the minimal mismatching game version

### Todo

- [x] Use `Nfc` for resource path

### DIFF
--- a/rust/src/guess.rs
+++ b/rust/src/guess.rs
@@ -31,20 +31,23 @@ pub fn guess_vanilla_version(gamedir: &str) -> Guess {
     logged
 }
 
+/// A difference that was detected in resource packs
 #[derive(Debug)]
 enum Difference {
-    OnlyExistsInDataDir(String),
-    OnlyExistsInPack(String),
+    OnlyExistsInDataDir(Option<String>, String),
+    OnlyExistsInPack(Option<String>, String),
     FileSizeMismatch(String, Option<i64>, Option<i64>),
     HashMismatch(String, String, Option<String>, Option<String>),
 }
 
+/// The result of matching a resource pack
 #[derive(Debug)]
 struct MatchResourcesResult {
     number_of_resources: usize,
     differences: Vec<Difference>,
 }
 
+/// Percentages of differences in resource packs
 struct Percentages {
     only_exists_in_data_dir: f64,
     only_exists_in_pack: f64,
@@ -54,6 +57,7 @@ struct Percentages {
 }
 
 impl Percentages {
+    /// Log percentages with log level info
     fn log(&self) {
         info!(
             "Percentage of resources only in game dir: {}",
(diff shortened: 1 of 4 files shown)


rust/src/guess.rs
```diff
@@ -105,64 +154,90 @@ impl Guess {
...
-                best_num = num_polish;
-                best_version = Some(VanillaVersion::POLISH);
-            }
+
+        let version = match (num_dutch, num_german, num_italian, num_polish, num_russian) {
+            (n, 0, 0, 0, 0) if n > 0 => Some(VanillaVersion::DUTCH),
+            (0, n, 0, 0, 0) if n > 0 => Some(VanillaVersion::GERMAN),
+            (0, 0, n, 0, 0) if n > 0 => Some(VanillaVersion::ITALIAN),
+            (0, 0, 0, n, 0) if n > 0 => Some(VanillaVersion::POLISH),
+            (0, 0, 0, 0, n) if n > 0 => Some(VanillaVersion::RUSSIAN),
```
lynxlynxlynx: RUSSIAN_GOLD is not relevant for this? I noticed it's not in the old code either, but better ask.

selaux: Tbh, I just reused the code that was there before. I'm not even sure what the `RUSSIAN_GOLD` version exactly refers to. It seems to work with the russian version from steam (I didn't do exhaustive testing wether it crashes somewhere in the game though)

flaviojs: That version guessing code is based on the paths returned by this function:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c3e9284a195ef38d774ecb4b3c0cf601576cb996/src/game/GameRes.cc#L156

rust/src/guess.rs
```diff
@@ -105,64 +161,95 @@ impl Guess {
...
+    }
+
+    /// Guess version by building resource pack and comparing it to other resource
+    /// packs in externalized directory
+    fn guess_by_resource_matching(&self, datadir: &Path) -> GuessResult<Option<VanillaVersion>> {
+        let results = self
+            .get_pack_paths()?
+            .par_iter()
+            .map(|path| self.compare_pack(datadir, path))
+            .collect::<Result<Vec<_>, _>>()?;
```
flaviojs: This `collect::<Result<Vec<_>, _>>()?` was confusing me.

I would add a comment saying that it stops early and returns an error (non-deterministic) if any of the parallel calls return an error, or just let everything finish and return a combined error message in case of multiple errors (see [`partition_map`](https://docs.rs/rayon/1.1.0/rayon/iter/trait.ParallelIterator.html#method.partition_map)).

rust/src/guess.rs
```diff
@@ -105,64 +161,95 @@ impl Guess {
...
+            (0, 0, 0, 0, n) if n > 0 => Some(VanillaVersion::RUSSIAN),
+            _ => None,
+        };
+
+        match version {
+            Some(version) => info!(
+                "Detected exclusively language specific files for language {:?}",
+                version
+            ),
+            None => info!("Detected no language specific files or multiple different languages"),
```
flaviojs: I would list of languages it detected.

selaux: Added a log line for that.

rust/src/guess.rs
```diff
@@ -36,63 +130,25 @@ pub struct Guess {
...
-        info!("Inspecting resource paths");
+        info!("Inspecting resource paths for language specific files");
         let mut num_dutch = 0;
         let mut num_german = 0;
         let mut num_italian = 0;
         let mut num_polish = 0;
         let mut num_russian = 0;
         for resource in resources {
-            let p = resource.path.to_lowercase().replace("\\", "/");
+            let p = Nfc::caseless(&resource.path);
```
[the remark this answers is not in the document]
selaux: I added tests for case insensitivity.

rust/src/guess.rs
selaux: I added a comment to that. We don't have a way to return multiple errors from that function yet so `partition_map` seems like it would require quite some overhead (collecting all errors and somehow joining them into a single string).

rust/src/guess.rs
```diff
@@ -1,6 +1,9 @@
 //! This module contains code to guess Vanillaversion with resource packs.
 
 use log::{error, info};
+use rayon::iter::IntoParallelRefIterator;
+use rayon::iter::ParallelIterator;
```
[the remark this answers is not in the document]
selaux: Done.

selaux:
I addressed the comments...

Issue #857, opened 2019-08-23
Compliance with "Core Infrastructure Initiative" best practices


lynxlynxlynx:
This is pretty cool:
https://bestpractices.coreinfrastructure.org/en

It really makes you think and is a good TODO of sorts. I've filled it out for ja2s and we're doing quite well:
https://bestpractices.coreinfrastructure.org/en/projects/3116
I didn't bother with silver, but it has some good items too. Too much focus on security for us though.

I don't think we can ever achieve 100 %, mainly due to the sorry original license, but we can strive to implement the rest. It makes for a stronger project that is easier to contribute to.

Any thoughts?

flaviojs:
This looks like a good TODO of sorts.

I would first try to handle warnings.
That includes resolving warnings that were just silenced in old commits.
We don't need to turn on more compiler warnings, default warnings are fine.
But we should include a static analysis tool in one of the build environments (travis/appveyor) to get other warnings.

lynxlynxlynx:
More warnings are always better. If they're not seen, they're not taken care of. Some are of course irrelevant or even false positives, so it does not make sense to enable all (especially not clang's -Weverything). We actually currently don't enable or silence any warnings, but #828 is changing that.

We have a Coverity account, but nobody ran a check since 2016 nor was anyone interested in dealing with the backlog (<300). They recently started supporting recent compilers for free again and also enabled more builds per project, so if we ever want to commit, it would be possible to run it per-push.

flaviojs:
The majority of the warnings are done.
It seems I won't be able to remove all the warnings so I'll focus on something else.

PR #858, opened 2019-08-28
Fix warning C4099

flaviojs:
Reference #857

The AppVeyor console had this warning:
```
warning C4099: 'LibraryDB': type name first seen using 'struct' now seen using 'class'
```


lynxlynxlynx:
LGTM!

PR #855, opened 2019-08-17
Reorganize rust

flaviojs:
Tasks:
-  [x] C code is contained inside module `c`
 - [x] figure out and fix dependencies of module `config`
 - [x] everything is formatted with rustfmt
 - [x] check formatting with travis/appveyor
 - [x] fix [clippy](https://github.com/rust-lang/rust-clippy) lints
 - [x] check clippy lints with travis/appveyor

Postponed:
 - split rust code into subcrates (lib, bin, C static lib)

A different PR:
- rename C functions

PR #859, opened 2019-08-30
Latest free editor.slf in the build system

### LINKED ISSUE #29: Tracona map in the map editor

---

Map editor shows the Tracona map instead of Arulco.
The map is taken from editor.slf - it is OMERTA.STI

Need to think about how to solve the problem.
Probably we can make Arulco map, include it into the distribution and show it instead of OMERTA.STI

Reported by https://bitbucket.org/alex_weil

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/29

flaviojs:
Reference: https://github.com/ja2-stracciatella/free-ja2-resources/issues/5

Tasks:
 - [x] find the latest release of the free editor.slf during config
 - [x] download editor.slf during build
 - [x] don't download by default (delayed until the rest is done)
 - [x] download editor.slf on release builds in travis/appveyor
 - [x] install next to the dirs externalized and mods
 - [x] always try to load the free editor.slf after the existing editor.slf
 - [x] update the editor.slf test in the launcher

The minimum cmake version increased to 3.1 for the json parser.

Closes #29 (see https://github.com/ja2-stracciatella/free-ja2-resources/pull/4)


lynxlynxlynx:
I think it should be off by default.

It's done now, but in case we didn't want to bump our cmake requirement or deal with json, I wrote a microservice just for this sort of thing. It's for a different modding community, but it works for all repos: https://github.com/Gibberlings3/modhub

flaviojs:
I have a linux environment but not the other build environments.
For now it needs to be on for travis/appveyor to test it.
It will be turned off after the rest is done.

flaviojs:
403 Forbidden in the mac build environment, no idea why...

lynxlynxlynx:
Restart it after a while, I'm having random problems in another project too. Different error though.

flaviojs:
I asked GitHub Developer Support and Ivan told me:
> Thanks for reaching out. Sure thing -- the reason why you observed a 403 is that you hit the API rate limits for unauthenticated requests:
>
> https://developer.github.com/v3/#rate-limiting
>
> For unauthenticated requests, the API rate limit is 60 reqs/hour per IP address. The documentation explains how you can get a higher and per-user limit (TL;DR: you need to authenticate), and you should check the Travis CI documentation on how to safely use authentication credentials from jobs that run on Travis.

This means the same is happening to all the build jobs that use that particular machine and don't have an access token configured, not only ours.

@selaux 
You're the repo owner so you should be the token owner too.
Can you create an access token for travis and add it to the repo?

https://gist.github.com/topheman/25241e48a1b4f91ec6d4
https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line

selaux:
I created a personal access token for my account and added it as the `GITHUB_ACCESS_TOKEN` environment variable to the master branch. Unfortunately I don't think there are access tokens for groups.

flaviojs:
Take 2.
Rebased onto master to get the token and implemented by calling curl directly.
It also uses ETag headers to avoid downloading again.

flaviojs:
I don't understand why the mingw build fails to find curl...

selaux:
I think `find_program(CURL_EXECUTABLE "curl${CMAKE_EXECUTABLE_SUFFIX}")` will look for `curl.exe` for mingw, but the build runs on linux. Maybe just removing the `${CMAKE_EXECUTABLE_SUFFIX}` will work?

flaviojs:
That was it, thanks!

flaviojs:
Rebased onto master to catch #855

flaviojs:
Done.

PR #864, opened 2019-09-12
Update rust toolchain to 1.36.0

flaviojs:
Originally I was updating to 1.34.0 for `SlfEntry::to_system_time` but it came with new warnings.
There is no reason to aim for a low version (not a library).
Since new warnings are unavoidable, I updated to the latest stable version.


lynxlynxlynx:
1.34 is the latest on ubuntu 18.10, so I don't think we should go higher yet — just making it harder for people to build.

flaviojs:
Version 18.10 is no longer supported by ubuntu, maybe that's why the rust version is older than the supported releases.

flaviojs:
All the releases supported by ubuntu have version 1.36 or 1.37 and the clippy in version 1.34 has issues.

@lynxlynxlynx Is 1.36 good enough?

lynxlynxlynx:
Yep, that's present in ubuntu.

PR #866, opened 2019-09-18
Use /MD runtime with MSVC

### LINKED ISSUE #641: Releases link against debug and release runtime on MSVC
Basically we need to redistribute debug and release dlls with MSVC releases. I am not sure why exactly.

I dont consider this a critical as it does not increase release size that much, it just clutters the directory.

flaviojs:
Fixes #641


flaviojs:
Note that the debug dlls are still included in the zip/installer.
I haven't removed `CMAKE_INSTALL_DEBUG_LIBRARIES` because it also affects mingw and I don't know if it's needed.

selaux:
I would remove it. According to [Documentation](https://cmake.org/cmake/help/v3.4/module/InstallRequiredSystemLibraries.html) is is only used for MSVC.

flaviojs:
Nice, done.

flaviojs:
So, can this be merged?

lynxlynxlynx:
Thanks!

PR #868, opened 2019-09-19
Set the source and execution charset to utf-8 in MSVC

flaviojs:
Continuation of #867

@lynxlynxlynx better let the tests finish before merging from now on >.<
Even if it works for me, it might not work in the builders, like what happened here.


lynxlynxlynx:
Wow, a single line **comment** change broke the build. I haven't seen such bullcrap since python2.

flaviojs:
The BOM was removed, it's an "invisible" change..

PR #870, opened 2019-09-20
Set locale at the start

### LINKED ISSUE #485: Mac OS Sierra Russian font bug
Hello. I use the compiled version ja2-v0.15.1-macos from your site and watch just such a problem (see screenshots in the attachment) with fonts in some places.

OS: macOS Sierra 10.12.2

### LINKED ISSUE #487: "ERROR saving the game" after Ivan has arrived
When Ivan arrives, the game log shows an entry "???":
ivan

At the same time, the console starts showing invalid character messages:
```
2017/00/22 22:59:11 ERROR   [Font            ] Invalid character given U+7FB5
2017/00/22 22:59:11 ERROR   [Font            ] Invalid character given U+0010
2017/00/22 22:59:11 ERROR   [Font            ] Invalid character given U+F4050170
```

And from this point on, saving the game fails with the message "ERROR saving the game", and the savegame slot where I attempted to save gets deleted.

How to reproduce: load the 
savegame:
[QuickSave.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/722304/QuickSave.sav.zip)

Fast forward time for a couple of hours, try to save.

flaviojs:
This is an attempt to fix #485 and fix #487.

According to  https://stackoverflow.com/a/15440423, the wchar_t representation  is  platform-dependent  and  locale-dependent.

In particular, the LC_CTYPE category of the locale at compile time and at run time should be the same.

~This PR will attempt to make them match.~

This PR sets a unicode locale at the start, which hopefully means it can handle the characters of all game languages. (our linux and mac builders have a utf-8 locales)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,7 @@ option(WITH_FIXMES "Build with fixme messages" OFF)
 option(WITH_MAEMO "Build with right click mapped to F4 (menu button)" OFF)
 option(BUILD_LAUNCHER "Build the ja2 launcher application" ON)
 option(WITH_EDITOR_SLF "Include the latest free editor.slf" OFF)
+set(WITH_CUSTOM_LOCALE "" CACHE STRING "Set a custom locale at the start, leave empty to disable")
 
 set(EXTERNAL_PROJECT_CMAKE_ARGS
     -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}
@@ -61,12 +62,16 @@ set(EXTERNAL_PROJECT_CMAKE_ARGS
     -DCMAKE_FIND_ROOT_PATH_MODE_INCLUDE=${CMAKE_FIND_ROOT_PATH_MODE_INCLUDE}
 )
 
+## Build
+
 if(MSVC)
 	# set the source and execution charset
 	add_compile_options("/utf-8")
 endif()
 
-## Build
+if(NOT "${WITH_CUSTOM_LOCALE}" STREQUAL "")
+    add_definitions("-DWITH_CUSTOM_LOCALE=\"${WITH_CUSTOM_LOCALE}\"")
+endif()
 
 set(BUILD_SHARED_LIBS OFF CACHE BOOL "")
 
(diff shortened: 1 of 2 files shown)


flaviojs:
From the old code, freebsd also has locale issues.

flaviojs:
It seems like macosx has BSD code mixed in, mostly in the kernel and drivers.
It is core stuff, so I wouldn't be surprised if the locale implementation is also like FreeBSD.
I'll assume mac matches FreeBSD if I can reproduce wide string problems in FreeBSD.

flaviojs:
The current freebsd (12.0) fails the call to set LC_CTYPE to "UTF-8" so maybe it's related to an old version of freebsd?

The more I read about locale stuff in C/C++ the more troublesome it seems. What a headache-inducing topic... there is basically no truly portable way to handle this.

lynxlynxlynx:
Is that even a valid locale? Usually it's something like en_US.UTF-8.

flaviojs:
That's the usual convention. Dunno, maybe that's how they named the non-ascii "C" locale before the current convention was adopted.

You can make and use custom locales, so in practice the locale name is a black box.

lynxlynxlynx:
Sure, but does it fail because it's missing or something else? I doubt you meant just setting the env var.

flaviojs:
The env vars are used to set the locale and there is always the classic locale as fallback. It is unspecified when the env vars are used.

Probably there is only consistency with ASCII, which to my knowledge is available in all encodings...
For stuff that can't be represented, it might ignore and let it pass, it might replace with ? or �, it might return an error in sprintf/swprintf calls, or have another kind of resolution.

I can't test, but I'm assuming a locale that can't represent russian characters was used in those issues and that their standard behavior is to error in swprintf when that happens. Our code assumes it always succeeds.

We don't rely on the other aspects of the locale, so it should work as long as the locale can represent all the characters we use.

flaviojs:
Hmm, I can't think of anything else that might be causing VS2005 to hang...

flaviojs:
Done.
Unfortunately the C++ locale can't be set until the hang with VS2015 is figured out.

lynxlynxlynx:
LGTM, but haven't tested.

PR #871, opened 2019-09-22
Panic triggers only happen for soldiers

### LINKED ISSUE #851: Crash from panic trigger when there are no soldiers.
I first got this crash when testing the fix for #688.

Example 1: [crash1.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3496535/crash1.zip)
 * shoot the torso
 * skip your turn
 * kill everyone with Alt+O (with cheats enabled)

Example 2: [crash2.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/3496579/crash2.zip)
 * shoot the legs

-----

I'm using revision 298096f574ae53599783d59ff1523806a7a9607b and the crash is in `ClosestPanicTrigger`:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/298096f574ae53599783d59ff1523806a7a9607b/src/game/TacticalAI/PanicButtons.cc#L368

There are no army guys here, only hicks (civilians), so this is performing a division by 0.
I wouldn't be surprised if there is other code like this, since fighting the army is your "main job".

I see 2 ways of fixing this:
 * exit the function if `pSoldier` is not part of the army (easy)
 * extend the logic to non-army enemies (hard)

flaviojs:
Fixes #851

### DIFF
--- a/src/game/Tactical/Soldier_Ani.cc
+++ b/src/game/Tactical/Soldier_Ani.cc
@@ -2321,7 +2321,10 @@ no_cry:
 					INT8 bPanicTrigger;
 
 					bPanicTrigger = ClosestPanicTrigger( pSoldier );
-					SetOffPanicBombs(pSoldier, bPanicTrigger);
+					if (bPanicTrigger != -1)
+					{
+						SetOffPanicBombs(pSoldier, bPanicTrigger);
+					}
 					// any AI guy has been specially given keys for this, now take them
 					// away
 					pSoldier->bHasKeys = pSoldier->bHasKeys >> 1;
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
LGTM, though maybe the 8118 block should remain in the failure path too.

flaviojs:
Amended the commit. That code block is part of failure path now.
I'm too unfamiliar with the AI and animation spaghetti so I can't judge if it's appropriate.

lynxlynxlynx:
I don't think anyone is familiar, but this way the change is smaller, so fewer chances of breaking some exotic corner case.

PR #874, opened 2019-10-02
Move the C API of rust code to a separate crate.

flaviojs:
The C API was already moved to the `c` module.
This PR brings the separation a step further by moving it to a separate crate.

The `stracciatella` crate is a static rust library (rlib).
The `stracciatella_c_api` crate is a static C library.

The rust directory was turned into a cargo workspace.
This means both crates use the same dependencies and you can run cargo as usual on that directory.

I also noticed it was including multiple versions of some crates.
This was due to dev dependencies so I just updated them while I was at it.


selaux:
Thanks. Rust stuff looks like it is getting better and better :slightly_smiling_face:

PR #875, opened 2019-10-06
Add RustPointer

flaviojs:
Manually freeing rust memory is an easy recipe for resource/memory leaks, specially when exceptions are involved.

This PR introduces the use of `std::unique_ptr` with a custom deleter, aliased as `RustPointer`.
`SGPFile` is excluded because it feels unsafe to put `RustPointer` there.

### DIFF
--- a/rust/stracciatella_c_api/cbindgen.toml
+++ b/rust/stracciatella_c_api/cbindgen.toml
@@ -1,7 +1,10 @@
 autogen_warning = "/* DO NOT MODIFY MANUALLY! Changes will be discarded during build. */"
-include_guard = "stracciatella_h"
 include_version = true
 language = "C++"
+
+# Top of stracciatella.h
+header = '''
+#pragma once'''
 
 [parse]
 parse_deps = true
--- a/rust/stracciatella_c_api/cbindgen.toml
+++ b/rust/stracciatella_c_api/cbindgen.toml
@@ -1,14 +1,38 @@
 autogen_warning = "/* DO NOT MODIFY MANUALLY! Changes will be discarded during build. */"
 include_version = true
 language = "C++"
+sys_includes = [
+    # std::unique_ptr
+    "memory",
+]
 
 # Top of stracciatella.h
 header = '''
 #pragma once'''
 
+# Bottom of stracciatella.h
+trailer = '''
+/// Callable that deletes rust memory.
+struct RustDelete
+{
+	void operator()(char* ptr) const { CString_destroy(ptr); }
+	void operator()(EngineOptions* ptr) const { EngineOptions_destroy(ptr); }
+	void operator()(LibraryDB* ptr) const { LibraryDB_destroy(ptr); }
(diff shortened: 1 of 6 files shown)


src/externalized/DefaultContentManager.cc
```diff
@@ -426,13 +420,13 @@ SGPFile* DefaultContentManager::openGameResForReading(const char* filename) cons
...
 
-				LibraryFile* libFile = LibraryFile_open(m_libraryDB, filename);
+				RustPointer<LibraryFile> libFile(LibraryFile_open(m_libraryDB.get(), filename));
 				if (libFile)
 				{
 					SLOGD("Opened file (from library ): %s", filename);
 					SGPFile *file = MALLOCZ(SGPFile);
 					file->flags = SGPFILE_NONE;
-					file->u.lib = libFile;
+					file->u.lib = libFile.release();
```
selaux: Wouldn't it make sense to store the `RustPointer<_>` in `file->u.lib` and reset when `SGPFile` is destroyed?

src/externalized/DefaultContentManager.cc
```diff
@@ -482,13 +476,8 @@ bool DefaultContentManager::doesGameResExists(char const* filename) const
...
 			{
-				LibraryFile* libFile = LibraryFile_open(m_libraryDB, filename);
-				if (libFile)
-				{
-					LibraryFile_close(libFile);
-					return true;
-				}
-				return false;
+				RustPointer<LibraryFile> libFile(LibraryFile_open(m_libraryDB.get(), filename));
+				return static_cast<bool>(libFile);
```
selaux: Should it be `libFile.get()` here?

src/externalized/DefaultContentManager.cc
flaviojs: I don't feel that SGPFile is safe to use like that. I can't pinpoint the reason why, but when the feeling is this strong it usually means I got "burned" by something similiar in the past.

src/externalized/DefaultContentManager.cc
flaviojs: `std::unique_ptr` has a bool operator, I'm using it by casting to bool.

selaux: :+1:

rust/stracciatella_c_api/cbindgen.toml
```diff
@@ -1,12 +1,39 @@
...
 include_version = true
 language = "C++"
+sys_includes = [
+    # std::unique_ptr
+    "memory",
+]
+
+# Top of stracciatella.h
+header = '''
+#pragma once'''
```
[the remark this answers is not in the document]
flaviojs: Currently `cbindgen` has no way to place extra code inside the guard.

I didn't opt for an extra guard wraping the new code (one of the alternatives) because we already rely on `#pragma once` in the C++ code.

If we want to avoid it, we should also remove it from the C++ code.
We target C++11 and always use it in the same filesystem so I think it's fine to rely on it.

PR #876, opened 2019-10-12
Retry binary uploads

selaux:
I retriggered the latest nightly build and the upload that previously failed succeeded now. Lets try if retying the upload will work...


lynxlynxlynx:
LGTM, but let's wait for the skipping test to be over first.

flaviojs:
There was a merge today, so no need to wait. ^^

PR #877, opened 2019-10-12
Trigger daily coverity builds as well

selaux:
According to https://scan.coverity.com/faq#frequency we have minimum 1 build a day, so we can trigger a coverity build similarily to nightlies every day. Now we can think about how to make the results of the scan more visible to developers (us).

~Ref: https://scan.coverity.com/projects/ja2-stracciatella~
Ref: https://scan.coverity.com/projects/ja2-stracciatella-ja2-stracciatella

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -1,28 +1,28 @@
-name: Nightly trigger
+name: Coverity trigger
 
 on:
   schedule:
-    - cron:  '0 2 * * *'
+    - cron:  '0 3 * * *'
 
 jobs:
   build:
 
     runs-on: ubuntu-latest
-    
+
     steps:
-    - name: Checkout nightly
+    - name: Checkout coverity
       uses: actions/checkout@v1
       with:
-        ref: nightly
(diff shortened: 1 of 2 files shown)


selaux:
BTW: I cant access the Coverity scans. @lynxlynxlynx Do you remember who set the account up? I requested access, but somebody needs to approve...

lynxlynxlynx:
You're already an admin on there with github@ and there are no pending requests. But it's true, the link you posted also displays to me, like I have to add myself to the project. From my dashboard everything is available, defects, settings ...

lynxlynxlynx:
ah, the other one is the old one:
https://scan.coverity.com/projects/ja2-stracciatella-ja2-stracciatella

flaviojs:
LGTM, I'm curious to see what it thinks about our usage of rust. (no rust last time it was triggered)

lynxlynxlynx:
Not much, as I specifically excluded the rust dir today (it still had the old _build hierarchy), but maybe it will have something to say on the boundary.

PR #879, opened 2019-10-14
Convert canonicalized UNC paths to normal paths when possible

flaviojs:
On windows the launcher couldn't find the path to `game.json` because the canonicalized base path was in UNC form and the relative part had a `/` path separator.

With a normal base path it works as expected, so crate [dunce](https://crates.io/crates/dunce) is being used to turn UNC paths into normal paths when possible.

Referencce: https://github.com/rust-lang/rust/issues/65397

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -250,6 +250,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "dunce"
+version = "1.0.0"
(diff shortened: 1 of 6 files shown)


flaviojs:
Now path separators are being normalized too.

Out rust functions that interact with the filesystem are a bit all over the place (in terms of strategy).
I'll refactor them when I rustify the C++ filesystem stuff.

Issue #886, opened 2019-10-15
Ubuntu 19.10 installation issues


anlag:
I stumbled on this project today and decided to give it a go. Armed with a freshly bought discounted copy of JA2 from GOG.com, I attempted to install it on my Ubuntu 19.10 desktop. I had a few issues along the way which I was able to get around, but wanted to note them down here. I realize strictly speaking I should open a separate issue for each one, but I'm a bit reluctant to spam the issue tracker in the first instance when it may just have been me missing something.

Nevertheless, here are the issues I encountered and how I resolved them.

## libpng12 issue
After installing the latest stable version (0.16.1) and trying to execute `ja2-launcher`, I got a library error:

```
ja2-launcher: error while loading shared libraries: libpng12.so.0: cannot open shared object file: No such file or directory
```
Searching issues I found this to be [fixed](https://github.com/ja2-stracciatella/ja2-stracciatella/issues/747) in nightlies, so I installed version 0.16.x-20191012 (subsequent deb builds are 0 bytes) which resolved this problem.

## libfltk_images issue
Having got past the libpng issue, I ran into another missing library.
```
ja2-launcher: error while loading shared libraries: libfltk_images.so.1.3: cannot open shared object file: No such file or directory
```
Unlike the first issue, I was unable to find this referenced in the issue tracker. It was easy enough to resolve however, by simply installing `libfltk-images1.3` from the distribution repositories (`sudo apt install libfltk-images1.3` for completeness.) With this done, I was able to run the launcher.

## Data dir not found
With the launcher running, I needed to specify the game directory, which gave me some pause since all I had was a Windows executable for the installer. The only way I could think of proceeding was to install Wine (`sudo apt install wine`) and then install it through there (`wine setup_jagged_alliance_2_1.12_\(17794\).exe`). The installation crashed with some errors but only after the game directory had been created so I didn't try to resolve them.

Upon specifying the (default) game directory (`~/.wine/drive_c/GOG Games/Jagged Alliance 2`) in the launcher, I received a "Data dir not found" error. I didn't find this specifically addressed in the issue tracker but another issue made me wonder if the spaces in the path might be the cause, so I tried copying the directory to `~/Apps/JA2`. When entering that into the launcher, the game started up.

## Resolution
I can change the resolution in the launcher and the game window does change accordingly, however the actual game content remains at the default 640x480 size, meaning in a larger window the game is surrounded by a lot of black.

[...]

PR #888, opened 2019-10-17
Add function to resolve caseless path to caseful variant

selaux:
This should extract a common functionality used to find a directory case insensitively. Basically now all user editable paths should be canonicalized and made case insensitive with this commit. It also seems to fix one of the issues in #886 where the launcher mishandles spaces.

### DIFF
--- a/rust/stracciatella_c_api/src/c/misc.rs
+++ b/rust/stracciatella_c_api/src/c/misc.rs
@@ -3,14 +3,13 @@
 //! The code in this module is not associated with any module in particular.
 
 use std::ffi::CString;
-use std::path::{Component, PathBuf};
+use std::path::PathBuf;
 use std::ptr;
 
 use stracciatella::config::find_stracciatella_home;
-use stracciatella::fs::canonicalize;
+use stracciatella::fs::{canonicalize, resolve_existing_components};
 use stracciatella::get_assets_dir;
 use stracciatella::guess::guess_vanilla_version;
-use stracciatella::unicode::Nfc;
 
 use crate::c::common::*;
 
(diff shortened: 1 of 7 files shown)


flaviojs:
You must take extra care with `caseless=false`tests because the case sensitivity of the filesystem can't be predicted. With symbolic links, even just part of the path can have a different case sensitivity.

Example: `t!(temp_dir.path(), "FOO", false, <expected result>);`
Here the expected result can be either true or false, it is impossible to predict.

As for the strategy of always comparing with Nfc, I'm unsure... do all filesystems treat names with different umlaut ö variations as the same name?
They probably don't, and that would make `resolve_existing_components` wrong because it can resolve to the wrong path.

As for #886, you should note that 0.16.1 is being used and the launcher changed a lot since then.
The space issue might already be fixed in master.

selaux:
True, I missed the case-sensitivity of the OSs in the tests :confused:.

Regarding always comparing with NFC, I would argue that it should always work the same, regardless of case-sensitivity. So either we always normalize to NFC (in both caseful and caseless matching) or never. Matching with NFC only in the caseless case is confusing in my oppinion. Additionally a user having two existing paths with different unicode variants seems a very unlikely edge case to me. But I'm open for suggestions.

flaviojs:
It only seems weird if you view paths as true unicode strings.
I view them as raw data that can even include invalid unicode and the `\0` character.
I'm not sure how much we support when coming from the C/C++ side, that is something for when filesystem stuff is rustified.

If we ask for a case sensitive path we are saying we want an **exact** match to the argument, so we should make raw comparisons.

If we ask for a case insensitive path we are saying we want a match according to some case folding rules. How we actually apply case folding (in NFC form) is an implementation detail.
With multiple matches, the best possible match is a raw match.

Ideally we would ask a case insensitive filesystem for the correct path components, but we can't.
Currently symbolic link aren't being considered. I don't know if we should follow them here.

flaviojs:
Hmm, I think imposing NFC is fine if we're taking about a virtual filesystem where we control the rules (like the content manager).

But this function is in the filesystem module (fs) so it should respect the filesystem as much as possible.

selaux:
Okay I will change the PR then to the following:

- Dont use NFC when not matching caseless paths
- Follow links when resolving paths

flaviojs:
Not following links might be ok.
Even though I remember reading several times about following links, I don't remember any specific reason why it's needed. :/

flaviojs:

flaviojs:
I'll try to continue this PR.

selaux:
Thanks. Sorry, currently I have only little time for this project.

flaviojs:
With the current implementation the name `resolve_existing_components` looks misleading to me.
Maybe rename it to `join_and_resolve_caseless`?

Note that it does not canonicalize, it only normalizes path separators, everything else is maintained.

rust/stracciatella_c_api/src/c/misc.rs
```diff
@@ -232,28 +219,59 @@ mod tests {
...
-        t!("FOO", true, true);
+
+        t!(temp_dir.path(), "baz", false, false);
+        t!(temp_dir.path(), "baz", true, false);
+
+        t!(temp_dir.path(), "foo", false, true);
+        t!(temp_dir.path(), "foo", true, true);
+        t!(temp_dir.path(), "FOO", true, true);
+
+        t!(temp_dir.path(), "foo/bar", false, true);
```
selaux: Does it make sense to add a test where the base dir is `temp_dir.path().join("Foo")` to test whether the base path is resolved case insensitively as well?

selaux:
I cant approve since I openened the PR, but other than the one comment this looks good to me.

rust/stracciatella_c_api/src/c/misc.rs
flaviojs: The result will depend on the case sensitivity of the filesystem, so it can't be tested.

PR #889, opened 2019-10-18
Build only when needed

### LINKED ISSUE #887: External project files are included in the deb
Found by @lynxlynxlynx  in #873.

The deb contains the install step files of the external projects.

The problem only happens when `BUILD_ALWAYS` is present.

Without that option the projects are not rebuilt if the sources are modified.
This means it can be removed from`gtest-internal`, but not `libstracciatella`.

Must find an alternate way to trigger a rebuild when the sources change.

flaviojs:
Don't rebuild gtest-internal. (we don't expect changes)

Trigger a rebuild of stracciatela when the sources change.

The md5 checksums and paths of the source files are computed every time cmake runs.

When the computed data is different from the data in the md5 file, the md5 file is updated and the update file is touched if it exists.

The update file is the stamp for the update step of the external project.

Fixes #887


flaviojs:
Hmm, not sure it this works when a rust is added and built, then modified.

flaviojs:
I like the new strategy.
As long as the source globs are correct, it will always rebuild when it needs.

PR #893, opened 2019-10-20
Fix checks for validity of the grid number.

flaviojs:
Coverity detected an access out of bounds in at least one of these.

NOWHERE must not change to keep compatibility with old saves, so instead all checks with >= NOWHERE and < NOWHERE are changed.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/b430600fac6648445be19d524551a43d5472b9be/src/game/TileEngine/Isometric_Utils.h#L13


lynxlynxlynx:
LGTM

PR #894, opened 2019-10-20
Fix uninitialized data in GrenadeTest1 and GrenadeTest2

flaviojs:
Coverity detected uninitialized data in Object.

DeleteObj is already being used to initialize, so the same is done here.


lynxlynxlynx:
This looks ok and that function really deserves a rename.

flaviojs:
It is used to initialize and to delete an object, maybe `ResetObj`?

lynxlynxlynx:
Don't have much preference, eg. ZeroObj or ClearObj would work too.

flaviojs:
Then I'll just leave it as is. I'll leave refactoring to someone that wants to do it.

PR #899, opened 2019-10-23
Fix some coverity issues with format strings

selaux:
Unfortunately gcc does not seem to pick up `SLOG` as format strings. Will have to check how to make it work.


flaviojs:
There are lots of coverity formatting issues...

This is probably not the right place to to ask, but how about adopting https://github.com/zrax/string_theory?

It could replace the UTF8String class and it brings other goodies.
The biggest wins in my eyes is the UTF-8 by default design (matches rust), and type safe formatting (goodbye coverity formatting issues).

flaviojs:

lynxlynxlynx:
Much of this should be obsolete now. :)

PR #903, opened 2019-10-25
Update shields

flaviojs:
Add coverity shield.
Add labels to the shields. 
Use the nightly branch in the travis and appveyor shields.

---

I saw that the other shields are for the master branch while coverity is for nightly.

@selaux @lynxlynxlynx Changing the other shields to nightly seems good to me, any objections?


lynxlynxlynx:
What's the rationale? To me, as a viewer, it seems obvious to expect that they show the current state.

flaviojs:
If we want the absolute latest state, then we should not use shields.io, which shows a cached state.

For me the state of the nightly branch is also the current state, but with a resolution of 24h instead of the shields.io cache time.

In out case it has another use, it tells me if the current nightly is available or not. ^^

lynxlynxlynx:
Ok, then, useful side effects.

flaviojs:
I'm done.

PR #910, opened 2019-10-28
Do not require libdl for OpenBSD

agernler:
On OpenBSD, there is no libdl.  The desired functionality is contained
in libc.  Thus, create a separate cmake case for OpenBSD that does not
add libdl.


selaux:
:+1: Thanks.

PR #912, opened 2019-10-29
Fix unintended integer division in SmackDoFrame.

flaviojs:
Coverity detected an unintended integer division in SmackDoFrame.

"FramesPerSecond" is actually the number of microseconds per frame.
Therefore "framerate" is the number of milliseconds per frame.

"framerate" is double, so it should not be truncated to integer values.


flaviojs:
The Cinematics/Smack_Stub code should probably be rewritten.

The current implementation is guaranteed to desync, hopefully not by much.
At a minimum it should have a "start" time and sync against it.

PR #913, opened 2019-10-30
Reimplement cinematics.

flaviojs:
It uses libsmacker directly.

It supports playing 4 videos at the same time.
Before it had 4 slots but could only play correctly 1 at a time.

It supports playing all sound channels.
Before it only played the first channel.

It does not desynchronize over time.
It synchronizes using an initial tick.
The start time of the sound and the video is still a problem.


flaviojs:
For me, the video/surface pipeline seems to be delaying the images by around 120ms compared to the sound. You can test by decreasing `start_tick` by 120 in `SmkPlayFlic`.

Hardcoding this value is not a solution. The right value probably depends on the system and maybe the settings too.

flaviojs:
NOTE a function to calculate when the next poll should happen can be made, but we already control the frame rate so it is not needed

misanthropos:
I have tried it and approve.
Smack still needs to be loaded fully into memory - but now you could start to stream it from the file. Are you interested in doing that?

flaviojs:
Unfortunately libsmacker only supports a filename for streaming.

Our case is a file inside an archive (slf) so I would have to copy the data to a temporary file and stream it, which I don't think is good.

misanthropos:
Agreed - creating a file just for streaming would only make sense on systems short on RAM.
An alternative could be adding support reading SGP-Files with libsmacker. Could be some work though.... However it should remove the short outages between cutscenes.

flaviojs:
I don't want to customize libsmacker. Making something in rust based on it might be interesting, but filesystem stuff comes first.

Dunno about outages between curscenes, I'm more concerned about the apparent desync between video and audio. I'm not even sure if it's already like that in the smk file or if it's our code.

Do you know of a smk video that looks good to calibrate video and sound?

flaviojs:
See if the commit fixed the outage between cutscenes.

misanthropos:
It feels a bit better - but I think the main reason for the short pause is the size of the flic to be loaded to memory first - especially the part in the bar.
The pause has been there before too - but in the end nothing that one should spend too much time in.
However - the scene, where you see tank after tank passing is not entirely smooth.. like a loop with a short pause after that - never figured out why that is.

flaviojs:
Loading and decoding are done in the main thread.
Hmm, the video delay I experience might be because of this.

Moving that to a separate thread that also prepares the next frame(s) would probably improve things. It's something to try when rust can read smk files. ^^

I'll stay away from this in C/C++.

flaviojs:
Media Player Classic seems to play smk files smoothly and in synch for me, I will use it as reference.

The video with the bar (prague.smk) has a few spot that are good to check if the video and audio match.

The video with the tanks (omerta.smk) is not smooth in the scene where they pass by.
That probably means the video is originally like that.

flaviojs:
@misanthropos I only noticed that you're part of the team now.
This means an ok from you is good enough for me to merge.
Can you do a final check of the changes in this PR?

lynxlynxlynx:
Not my area of knowledge, so just merge it as far as I'm concerned. It clearly improves several things.

src/game/Utils/Cinematics.cc
```diff
@@ -7,200 +7,313 @@
...
+static void SmkSkipFrames(SMKFLIC* sf);
+static void SmkBlitVideoFrame(SMKFLIC* const sf, SGPVSurface* surface);
 
-		if (SmackWait(smkobj)) continue;
 
-		//if (SmackSkipFrames(smkobj)) continue;
+BOOLEAN SmkPollFlics(void)
+{
+	BOOLEAN is_playing = FALSE;
+	FOR_EACH(SMKFLIC, sf, gSmkList)
```
selaux: I think we should not use boost in new code anymore. As when we come to the point where we replace the FileMan stuff with rust, we only use `FOR_EACH` from boost, which to me does not make it worth anymore to keep it as a dependency.

(Please ignore if this is not boost foreach)

src/game/Utils/Cinematics.cc
```diff
@@ -7,200 +7,313 @@
...
 }
 
 
 void SmkInitialize(void)
 {
-	// Wipe the flic list clean
-	memset(SmkList, 0, sizeof(SmkList));
-
-	// Use MMX acceleration, if available
-	SmackUseMMX(1);
```
selaux: This was removed and not re-added anywhere, so we dont use mmx accelleration anymore?

src/game/Utils/Cinematics.cc
```diff
@@ -7,200 +7,313 @@
...
 
 	// Set the blitting position on the screen
-	sf->uiLeft = left;
-	sf->uiTop  = top;
+	sf->left = left;
+	sf->top  = top;
+
+	// Get all the audio data
+	unsigned char audio_tracks;
+	unsigned char audio_channels[7];
```
selaux: The 7 occurs an awful lot in this code. Does it make sense to extract it into a meaningful constant?

src/game/Utils/Cinematics.cc
```diff
@@ -7,200 +7,313 @@
...
+	FOR_EACH(SMKFLIC, sf, gSmkList)
+	{
+		if (!(sf->flags & SMK_FLIC_OPEN)) return sf;
+	}
+	return nullptr;
 }
 
 
-static SMKFLIC* SmkGetFreeFlic(void)
+static void SmkSkipFrames(SMKFLIC* sf)
```
selaux: Maybe a better name: `SmkSkipToCurrentFrame` or something alike...

src/game/Utils/Cinematics.cc
```diff
@@ -7,200 +7,313 @@
...
-static void SmkSetupVideo(void)
+static void SmkBlitVideoFrame(SMKFLIC* const sf, SGPVSurface* surface)
 {
-	UINT32 red;
-	UINT32 green;
-	UINT32 blue;
-	GetPrimaryRGBDistributionMasks(&red, &green, &blue);
-	if (red == 0xf800 && green == 0x07e0 && blue == 0x001f)
+	// get frame (source)
+	// TODO handle flags SMK_FLAG_Y_* (I need a sample of each case)
```
selaux: Is this TODO still relevant?

src/game/Utils/Cinematics.cc
flaviojs: It was function from the radtools dll, it was never used with smacker.

src/game/Utils/Cinematics.cc
flaviojs: I would say yes, classic ja2 does not use the other modes, but mods might.
It's not critical, the video will just have an unexpected height.

src/game/Utils/Cinematics.cc
flaviojs: If smacker had a constant I would use it.
I decided to match the library.

src/game/Utils/Cinematics.cc
flaviojs: It is not from boost.

PR #915, opened 2019-11-02
Use library string_theory.

flaviojs:
Coverity has many complaints about string stuff.

Here I will introduce https://github.com/zrax/string_theory for string stuff.
The biggest selling points of this library are (for me):
* utf8 by default (matches rust)
* can convert to and from utf16/utf32/wchars
* type safe formatting
* no additional dependencies

NOTE version 3 is under development and is header-only, the current version has source files

# Tasks:
- [x] add string_theory to the build system
- [x] convert all code that uses UTF8String
- [x] remove UTF8String and utf8cpp

# Rejected:
- convert format code?
- convert wchar code?

While converting code that uses UTF8String I decided to not fix any bugs here.
Converting formatting and wchar code would fix bugs so they were rejected for this PR.


flaviojs:
Ok, this is ready for review.

The library is built at configure time.
This makes sense because it is essentially a replacement for find_package.

The library is built in release mode without debug symbols.
This is a personal preference, I would also accept RelWithDebInfo but Debug doesn't make sense for me.

The library is used as a library interface.
Modern cmake recommends not using variables for the include dir and link libraries.
This is one way of doing things, better ways are available after bumping the cmake version.

flaviojs:
@lynxlynxlynx can you review this and merge if it's fine?

lynxlynxlynx:
Sure, looks fine.

PR #919, opened 2019-11-11
Fix resource leak in ExplosiveDamageGridNo

flaviojs:
Coverity detected a resource leak in ExplosiveDamageGridNo when ppTile goes out of scope.

I don't really understand this code, but turning ppTile into a vector is enough to fix it.
The vector will be freed when it goes out of scope.


lynxlynxlynx:
Looks odd indeed, but we've also seen worse.

PR #922, opened 2019-11-15
Fix unused values in ChooseHairColor 

flaviojs:
~Coverity detected an unused value of hair in ChooseHairColor~

~This last line is probably leftover code.~

----

Coverity detected unused values of hair in ChooseHairColor

This last line is overriding them and is probably leftover code.


lynxlynxlynx:
I don't get it, how is it unused and not just overriding all the previous logic?

flaviojs:
It's the previous ones that are unused, precisely because this line was overriding them.
Should have made it clearer. =~

flaviojs:
Remade the commit to make it clear.

lynxlynxlynx:
Ah, thanks. All head variants are still used, so yeah, looks fine.

Issue #873, opened 2019-09-30
Nightlies are not built anymore


selaux:
### What is wrong
Nightlies are not built anymore. This is because I had a job running on a personal CI server that actually triggered the nightly build. This CI server does not exist anymore (actually for a long time now :wink: ).

It would be nice to evaluate whether GitHub Actions can be used for that, so everybody in the team has access.

lynxlynxlynx:
Yep, it supports cron, basically:
https://help.github.com/en/articles/events-that-trigger-workflows#scheduled-events-schedule

selaux:
For reference this was the previous script in jenkins:

```groovy
node() {
    def today = new Date().format('yyyyMMdd');
    def buildId = today
    def branch = "master"
    def message = "Nightly build $buildId triggered via API request"
    
    stage "Git Poll"
    git poll: true, url: 'https://github.com/ja2-stracciatella/ja2-stracciatella'
    
    stage "Trigger travis nightly build"
    withCredentials([[$class: 'StringBinding', credentialsId: 'travis-ci-trigger-token', variable: 'TOKEN']]) {
        def headers = [
            [name: 'Travis-API-Version', value: '3'],
            [name: 'Authorization', value: "token ${env.TOKEN}" ]
        ]
        def postBody = """{
            "request": {
                "message": "$message",
                "branch": "$branch",
                "config": {
                    "env": {
                        "global": [
[...]

PR #924, opened 2019-11-16
GitHub CI

flaviojs:
Attempt to replace travis builds.
Travis has random ftp upload failures that will never be fixed.

Reference #873

### DIFF
--- a/.github/scripts/ci-build.sh
+++ b/.github/scripts/ci-build.sh
@@ -112,14 +112,18 @@ for file in ja2-stracciatella_*; do
   elif [[ "$file" == *".dmg" ]]; then
     # based on https://ss64.com/osx/hdiutil.html
     hdiutil verify "$file"
+    device=""
     while IFS=$'\n' read -r line; do
       echo "$line"
       IFS=$'\t' read -ra arr <<< "$line"
+      if [[ "$device" == "" ]]; then
+        device="${arr[0]}"
+      fi
       if [[ "${arr[2]}" != "" ]]; then
         ls -laR "${arr[2]}"
       fi
-    done <<< "$(hdiutil attach \"$file\")"
-    hdiutil detach "$file"
+    done <<< "$(hdiutil attach \"$(pwd)/$file\")"
+    hdiutil detach "$device"
   else
     echo "TODO list contents"
   fi
--- a/.github/scripts/ci-build.sh
+++ b/.github/scripts/ci-build.sh
@@ -1,22 +1,34 @@
 #!/usr/bin/env bash
+#
+# Build script for CI environments, assumes a posix-like host (linux or mac).
+#
+# Requires the following environment variables:
+#   CI_REF - full reference of the current branch, tag, or pull request
+#   CI_TARGET - target we are building for: linux/mingw/mac
+# Unused or optional environment variables:
+#   CI_NAME - name of the job
+#   CI_OS - host runner image: https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners
+#   SFTP_USER - ftp username for uploads (secret)
+#   SFTP_PASSWORD - ftp password for uploads (secret)
 
-set -x
 set -e
+set -x
 
 echo "## prepare environment ##"
-if [[ "${GITHUB_REF}" == "refs/heads/nightly" ]]; then
+if [[ "${CI_REF}" == "refs/heads/nightly" ]]; then
   echo "-- NIGHTLY --"
   export PUBLISH_BINARY="true"
   export PUBLISH_DIR="nightlies"
   export BUILD_TYPE="ReleaseWithDebInfo"
   export VERSION_TAG="$(date +%Y%m%d)"
-elif [[ "${GITHUB_REF}" == "refs/tags/"* ]]; then
+elif [[ "${CI_REF}" == "refs/tags/"* ]]; then
   echo "-- RELEASE --"
   export PUBLISH_BINARY="true"
   export PUBLISH_DIR="releases"
   export BUILD_TYPE="ReleaseWithDebInfo"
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
We get 2,000 minutes per month of this for free or 66 builds if we consider this PR's runtime. We may hit that at some point, with our iterative approach. Does it just stop working afterwards?

flaviojs:
It will probably fail until the next month begins.
It's not uploading right now (can't set secrets), it will take a bit longer when it does.

flaviojs:
AppVeyor has mingw installed, so the build can be converted and moved there.

flaviojs:
Another alternative is to restrict this to nightly and tags (releases), which build as RelWithDebInfo.
Those uploads should never fail, the PR uploads are less critical.
The 2000 minutes per month limit is sufficient for them.

janisozaur:
Where did you get that 2000 minutes number?

lynxlynxlynx:
In the email to all the beta participants, but it's clearly visible on the site too.

janisozaur:
I can't seem to find it, can you provide a link? I've found only https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#usage-limits and there's no such information

selaux:
I think @lynxlynxlynx refers to this https://github.com/features/actions#pricing-details. Although from what I understand from this is that private repositories are limited to 2000 minutes.

janisozaur:
Yes, public repositories are all free without any specified limits (but I imagine there are some, I was told that each job has a timeout of 6 hours). I was very confused when I saw you discussing the 2000 minute limit. I hope that clears up confusion and perhaps will affect your decision.

.github/scripts/ci-build.sh
```diff
@@ -0,0 +1,105 @@
...
+mkdir ci-build
+cd ci-build
+$CONFIGURE_CMD ..
+cat ./CMakeCache.txt
+$BUILD_CMD
+$BUILD_CMD --target package
+
+echo "## test ##"
+if [[ "$CI_TARGET" != "mingw" ]]; then
+  sudo $BUILD_CMD --target install
```
selaux: Isn't that only required for "linux" builds?

.github/workflows/github-ci.yml
```diff
@@ -0,0 +1,48 @@
+name: GitHub CI
+
+on:
+  push:
+    branches: [ master, nightly ]
```
selaux: I think we need to add the coverity stuff to this as well. Not sure how hard it is to integrate manually...

.github/scripts/ci-build.sh
flaviojs: All of them should be tested if possible, mingw can't be tested because it's linux cross compiling to windows. I could be tested if it was built in a windows host.

.github/workflows/github-ci.yml
flaviojs: I can try adding coverity after this is merged. It needs to be done directly in master, so that it's triggered by a push to nightly.

flaviojs:
If public repositories don't have build limits then everything is fine. :)
Even AppVeyor builds could be replaced, but I think that's out of scope for this PR.
Coverity will have to wait since the final test (auto push to nightly) needs to be done directly in master.

@selaux uploads still need to be tested, can you set up the SFTP_* secrets?

selaux:
@flaviojs I added the necessary secrets.

flaviojs:
I don't see a way to rebuild, that's bad... I'll ask support about it.

flaviojs:
Maybe I can make it trigger with a comment like "/ci rebuild", I'll try it out.

janisozaur:
Once a build is finished, you can click on a button to rerun in the webui.

flaviojs:
I see a button in the failed nightly workflow (convenient failure), but I don't see it on successful ones.
In this case I want a rebuild after the checks are successful, I'll try the comment approach.

janisozaur:
https://github.community/t5/How-to-use-Git-and-GitHub/Is-it-possible-to-manually-force-an-action-workflow-to-be-re-run/m-p/25475 suggests there is no such feature yet.

Other solutions suggest adding an empty commit, although that's not strictly the same outcome.

flaviojs:
The comment approach failed.

I did find this: https://github.community/t5/GitHub-API-Development-and/Can-the-Checks-API-be-used-to-trigger-two-runs-when-two-pushed/m-p/18008#M719
Seems like it's possible with the checks API, which is currently in public preview.
If we wait, someone will probably make a bot/app.

In the meantime, uploads can be required instead of optional. Making it fail if any detail goes wrong should be good enough for now. If it passes with something incomplete it can be considered a bug.

flaviojs:
Hmm, do PR builds have access to the repo secrets?

janisozaur:
I would be surprised if they did. PRs can be submitted by third parties at will and you cannot prevent them from pushing code to leak the secrets, which is why they should be unavailable by design.

flaviojs:
Not sure what to do then, uploads are the reason this change was attempted.
Without that this PR makes little sense.

janisozaur:
> Secrets are environment variables that are encrypted and only exposed to selected actions. Secrets are not passed to workflows that are triggered by a pull request from a fork of this repository.

selaux:
This was the same before. PRs from forks never had the secrets exposed afaik.

flaviojs:
Too bad, PRs are impossible but nightly and releases are fine.

I'll use https://github.com/actions/upload-artifact in PRs, at least the files will become available.
There might be a way to upload them somewhere else when the checks API is done.

flaviojs:
I wish there was a github actions yaml validator...

flaviojs:
That's all from me.

PR files are not uploaded but are available as artifacts in the check suite.
If this is acceptable, then merge.

PR #930, opened 2019-11-21
Fix uninitialized members in WeaponModel constructor

flaviojs:
Coverity detected uninitialized members in the WeaponModel constructor.


selaux:
:clap:

Issue #590, opened 2017-08-26
failure when building on msys2


askondro:
make
[  0%] Built target gtest
[  0%] Built target slog
warning: unused manifest key: package.release-debuginfo
       Fresh winapi-build v0.1.1
       Fresh itoa v0.1.1
       Fresh winapi v0.2.8
       Fresh num-traits v0.1.40
       Fresh dtoa v0.2.2
       Fresh libc v0.2.29
       Fresh serde v0.8.23
       Fresh getopts v0.2.14 (https://github.com/rust-lang-nursery/getopts?rev=b8beaf63#b8beaf63)
       Fresh serde_json v0.8.6
       Fresh shell32-sys v0.1.1
       Fresh user32-sys v0.2.0
       Fresh stracciatella v0.1.0 (file:///C:/msys64/home/askon/ja2-stracciatella/rust)
    Finished release [optimized] target(s) in 0.0 secs
[  0%] Built target rust-build
[  2%] Built target smacker
make[2]: *** No rule to make target 'rust/release/libstracciatella.dll.dll.a', needed by 'ja2.exe'.  Stop.
make[1]: *** [CMakeFiles/Makefile2:135: CMakeFiles/ja2.dir/all] Error 2
make: *** [Makefile:152: all] Error 2

[full_log.txt](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1253723/full_log.txt)

Manuel-K:
That's weird: `.dll.dll.a`

Can you please check what extension the libstracciatella that was built has?

askondro:
I has the extension .dll

rust/release/stracciatella.dll

Manuel-K:
@selaux Any ideas?

selaux:
My guess is the issue is somewhere aroun here: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/CMakeLists.txt#L34

I don't know how to build through msys though, so it is really hard to reproduce for me.

Manuel-K:
@selaux Are you sure that it is a good idea to concatenate `${CMAKE_SHARED_LIBRARY_SUFFIX}` and `${CMAKE_IMPORT_LIBRARY_SUFFIX}`?

Manuel-K:
Btw. I can reproduce the issue. 

Basic steps: 
[...]

PR #935, opened 2019-11-30
Install mingw dlls with fixup_bundle

### LINKED ISSUE #934: Mingw exes and dlls missmatch
Discovered in #933.

The current nightly (2019-11-26) mingw exes are 64-bit but the included dlls are 32-bit.
When you try to run `ja2.exe` or `ja2-launcher.exe` you get an error message with code 0xc000007b.

flaviojs:
The included dlls were 32bit and might not work if the mingw-w64 gcc version is different.

Now they are copied from the system at install time.

Fixes #934.


flaviojs:
Note that `install-dlls-mingw.cmake.in` assumes linux cross-compiling to windows (our current build).
It probably has problems with paths under msys2.

Looking at #590 msys2 is close to possible, I'll try it out after this is merged.

PR #937, opened 2019-12-03
Convert wchar code

flaviojs:
wchar_t is compiler dependent and locale dependent.
When a character can't be represented by the locale things start to go wrong.

Unicode supports all characters and the [string_theory](https://github.com/zrax/string_theory) library is utf8 by default.
The aim of this PR is to convert all wchar_t code to unicode using that library.

Lots of code will be affected so merge conflicts are expected as other PRs are merged.
To facilitate conflict resolution, each commit will be centered around a single ~function or structure~ file+header.


flaviojs:
Damn, looking at MOUSE_REGION I realized that I can't use `ST::string` in it because of memset. What to do... this might be a blocker since I don't want to refactor in the PR.

lynxlynxlynx:
You mean memset is being used over MOUSE_REGION? I didn't grep any instances. Do a separate PR to give it the usual constructors first and get rid of any plain copy hacks?

flaviojs:
There is a memset in MouseRegion::MouseRegion, another in MSYS_RemoveRegion, maybe more.
The same will happen with other structures.
The only way to be safe is to replace all uses of memset with something else, which I don't really want to do but might have to if I want to continue this. =~

flaviojs:
I think I have a good strategy now.
The last obstacle seems to be paths so I'll rustify them in a separate PR and rebase after it is merged.

flaviojs:
I can't see any more problems while reviewing the changes.
I'll play test it to Drassen and then consider it done.

flaviojs:
Done, hopefully that's all the problems. :~

lynxlynxlynx:
Big work! There's no way I'm reading the whole diff carefully though. In some sampling I didn't see much of interest:
- some function signatures got deconstified for no obvious reason (non-string params)
- st_fmt_printf_to_format looks like it should be upstream, perhaps we can contribute it back

flaviojs:
I want the arguments to reflect the real function signature to avoid missunderstandings.
I don't want to see consts that are not part of the type. (they don't help the code unless it's a BIIIG function, which is already discoraged).
 * `const int x` and `int const x` are `int` in the signature, but you can't change the local var <= 1 const is not needed
 * `char* const x` is `char*` in the signature, but you can't change the local var <= 1 const is not needed
 * `const char* const x`/`char const* const x` is `const char*` (or `char const*`) in the signature, but you can't change the local var <= 1 const is not needed
 * `int x` is `int` in the signature
 * `const char* x`/`char const* x` is `const char*` (or `char const*`)  in the signature
 * `const ST::string& x`/`ST::string const& x` is `const ST::string&` (or `ST::string const&`) in the signature

----

I wouldn't put `st_fmt_printf_to_format` upstream, since they probably want to discorage the use of printf-style formatting and I think there are vendor-specific formats that clash.

lynxlynxlynx:
So everything builds elsewhere too and works for you. If you're done done, go ahead and merge.

PR #938, opened 2019-12-07
Replace memset code

flaviojs:
In #937 I found that I couldn't use `ST::string` inside `MOUSE_REGION` because of `memset`.

This PR aims to replace all `memset` code with something else.

Not convinced memset is bad?
Here is an article with more problems: https://www.viva64.com/en/b/0360/

---

Fixes too much mix memory, possible uninitialized byte and warnings in `SoundMan.cc`.

Fixes initialization of `gusMapPathingData` in `FindStratPath`.

### DIFF
--- a/src/game/Laptop/BobbyR.cc
+++ b/src/game/Laptop/BobbyR.cc
@@ -446,7 +446,7 @@ static void InitBobbyRayNewInventory(void)
 	UINT16	usBobbyrIndex = 0;
 
 
-	std::fill_n(LaptopSaveInfo.BobbyRayInventory, MAXITEMS, STORE_INVENTORY{});
+	std::fill_n(LaptopSaveInfo.BobbyRayInventory, static_cast<size_t>(MAXITEMS), STORE_INVENTORY{});
 
 	// add all the NEW items he can ever sell into his possible inventory list, for now in order by item #
 	for( i = 0; i < MAXITEMS; i++ )
--- a/src/game/Laptop/BobbyR.cc
+++ b/src/game/Laptop/BobbyR.cc
@@ -480,7 +480,7 @@ static void InitBobbyRayUsedInventory(void)
 	UINT16	usBobbyrIndex = 0;
 
 
-	std::fill_n(LaptopSaveInfo.BobbyRayUsedInventory, MAXITEMS, STORE_INVENTORY{});
+	std::fill_n(LaptopSaveInfo.BobbyRayUsedInventory, static_cast<size_t>(MAXITEMS), STORE_INVENTORY{});
 
 	// add all the NEW items he can ever sell into his possible inventory list, for now in order by item #
 	for( i = 0; i < MAXITEMS; i++ )
--- a/src/game/Tactical/PathAI.cc
+++ b/src/game/Tactical/PathAI.cc
@@ -53,6 +53,8 @@ static BOOLEAN gfDrawPathPoints = FALSE;
 
 #include "Logger.h"
 
+#include <algorithm>
+
 BOOLEAN gfPlotPathToExitGrid = FALSE;
 BOOLEAN gfRecalculatingExistingPathCost = FALSE;
 UINT8 gubGlobalPathFlags = 0;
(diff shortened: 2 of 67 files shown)


flaviojs:
Reworked commits so that the C/C++ includes come last.

flaviojs:
Hmm, this is gonna take a lot of commits grouping by function... maybe I should group them by filename instead.

flaviojs:
Done, it's ready for review.

src/game/Laptop/BobbyR.cc
```diff
@@ -444,7 +446,7 @@ static void InitBobbyRayNewInventory(void)
 	UINT16	usBobbyrIndex = 0;
 
 
-	memset( LaptopSaveInfo.BobbyRayInventory, 0, sizeof(STORE_INVENTORY) * MAXITEMS);
+	std::fill_n(LaptopSaveInfo.BobbyRayInventory, static_cast<size_t>(MAXITEMS), STORE_INVENTORY{});
```
selaux: Wont that result in the array filled with references to the same `STORE_INVENTORY{}` struct?

src/game/TacticalAI/FindLocations.cc
```diff
@@ -552,7 +554,7 @@ INT16 FindBestNearbyCover(SOLDIERTYPE *pSoldier, INT32 morale, INT32 *piPercentB
 #if defined( _DEBUG ) && !defined( PATHAI_VISIBLE_DEBUG )
 	if (gfDisplayCoverValues)
 	{
-		memset( gsCoverValue, 0x7F, sizeof( INT16 ) * WORLD_MAX );
+		std::fill_n(gsCoverValue, WORLD_MAX, 0x7F7F);
```
selaux: What does `0x7F` actually stand for in this case?

src/game/TacticalAI/AIMain.cc
```diff
@@ -69,7 +71,7 @@ void InitAI(void)
 #ifdef _DEBUG
 	if (gfDisplayCoverValues)
 	{
-		memset( gsCoverValue, 0x7F, sizeof( INT16 ) * WORLD_MAX );
+		std::fill_n(gsCoverValue, WORLD_MAX, 0x7F7F);
```
selaux: What does `0x7F` actually stand for in this case?

src/game/Tactical/PathAI.cc
```diff
@@ -719,13 +721,13 @@ INT32 FindBestPath(SOLDIERTYPE* s, INT16 sDestination, INT8 ubLevel, INT16 usMov
...
-	memset( pathQ, 0, iMaxPathQ * sizeof( path_t ) );
-	memset( trailTree, 0, iMaxTrailTree * sizeof( trail_t ) );
+	std::fill_n(pathQ, iMaxPathQ, path_t{});
+	std::fill_n(trailTree, iMaxTrailTree, trail_t{});
 
 #if defined( PATHAI_VISIBLE_DEBUG )
 	if (gfDisplayCoverValues && gfDrawPathPoints)
 	{
-		memset( gsCoverValue, 0x7F, sizeof( INT16 ) * WORLD_MAX );
+		std::fill_n(gsCoverValue, WORLD_MAX, 0x7F7F);
```
selaux: What does `0x7F` actually stand for in this case?

src/game/Laptop/BobbyR.cc
flaviojs: Nope, `LaptopSaveInfo.BobbyRayInventory` is an array of concrete structures, not references.
`STORE_INVENTORY{}` is a temporary default/zero-filled structure that is assigned (copied) to each element of the array.

src/game/TacticalAI/FindLocations.cc
flaviojs: It's a magic debug number.

src/game/TacticalAI/AIMain.cc
flaviojs: It's a magic debug number.

src/game/Tactical/PathAI.cc
flaviojs: It's a magic debug number.

flaviojs:
This PR touches too much code so it's easy to make mistakes.
To facilitate review and avoid mistakes, I changed the code to the alternative closest to the original memset.

src/game/Laptop/BobbyR.cc
selaux: Thanks for clarification, I just wanted to make sure...

PR #943, opened 2020-01-05
Fix remaining calls to CalibreModel::getName()

flaviojs:
The return value changed when UTF8String was replaced with ST::string.
Probably forgotten because the file didn't reference UTF8String directly.

Bug introduced in #915
The item calibre name would show up as garbage.


lynxlynxlynx:
LGTM. The github failure looks unrelated.

PR #944, opened 2020-01-07
Fix bad format specifier

flaviojs:
Bug introduced in #941


flaviojs:
Detected by coverity, I'm glad it's working again. :)

PR #947, opened 2020-01-07
Fix warning C4805

flaviojs:
Was complaining about comparing bool with BOOLEAN in SetEditingStatus.
Now it's comparing bool with bool.

Reference #857


lynxlynxlynx:
We could get rid of all BOOLEAN, but it would likely require bumping the save game version too.

flaviojs:
I don't see a compelling reason to change all of them.
If we used a style that recommended bool over BOOLEAN, maybe.

PR #949, opened 2020-01-08
Fix expression result unused warning

flaviojs:
Reference #857


lynxlynxlynx:
Thanks for doing all this stellar work!

PR #975, opened 2020-01-12
Fix result not used warning in Launcher

flaviojs:
Reference #857

### DIFF
--- a/src/launcher/Launcher.cc
+++ b/src/launcher/Launcher.cc
@@ -237,7 +237,11 @@ void Launcher::startExecutable(bool asEditor) {
 		cmd += std::string(" -editor");
 	}
 
-	system(cmd.c_str());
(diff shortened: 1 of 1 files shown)


src/launcher/Launcher.cc
```diff
@@ -237,7 +237,11 @@ void Launcher::startExecutable(bool asEditor) {
 		cmd += std::string(" -editor");
 	}
 
-	system(cmd.c_str());
+	int ret = system(cmd.c_str());
+	if (ret != 0)
+	{
+		SLOGW("there was an error while running '%s' (%d)", cmd.c_str(), ret);
```
lynxlynxlynx:
```suggestion
		SLOGW("There was an error while running '%s' (%d)", cmd.c_str(), ret);
```

PR #976, opened 2020-01-12
Fix result not used warning in BuildListFile

flaviojs:
fgets returns nullptr when eof is reached and nothing is read.

Reference #857

### DIFF
--- a/src/game/AniViewScreen.cc
+++ b/src/game/AniViewScreen.cc
@@ -281,11 +281,8 @@ static void BuildListFile(void)
 		return;
 	}
 	//count STIs inside header and verify each one's existance.
-	while( !feof( infoFile ) )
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Doesn't this introduce an off-by-1 error, for the last blocks?

flaviojs:
According to the fgets docs it's only null when nothing is read (was at eof).

According to the feof docs it only checks if the flag is set (does not set it).
An off-by-1 is indeed possible when fgets reads nothing.
By doing so it avoids processing garbage (previous line again?).
What a pleasant surprise, I fixed a bug that I didn't know about. ^^

But now that I look closer at that newline removal, it only works properly on windows and when the last line is empty. I'll fix it.

PR #980, opened 2020-01-13
Fix wrong npc id used in RefreshNPCScriptRecord

flaviojs:
`ubNPC` is 200 here, and this code is clearly meant to process `i`.


lynxlynxlynx:
Nice, could be the reason for some npc quest problems.

PR #984, opened 2020-01-14
Fix sign-compare warnings in tests

flaviojs:
Reference #857


lynxlynxlynx:
It looks fine, though the unrelated rust failure is not helpful.

flaviojs:
Remade commit with a few forgotten changes.

Hmm, the sporadic failure downloading a crate probably wouldn't happen if they were cached.

PR #985, opened 2020-01-15
Fix sign-compare warnings in Finances.cc

flaviojs:
For negative iCurrentPage values the next page and last page buttons will show up now.

Reference #857


flaviojs:
Page 0 is the summary so a negative value is guaranteed to have a next page.
I don't see how it goes negative though...

This file is hard to grasp (needs a rewrite), I wouldn't be surprised if there was a bug somewhere.

PR #988, opened 2020-01-16
Unbreak strategic pathing

oldlaptop:
Strategic pathing (easily visible when trying to use the helicopter; the behavior isn't perfectly consistent, but generally it won't take any paths longer than two sectors) has apparently been broken since 0e013cb67be05e598a3de1fbb245b6e5cff7b49a, which switched from byte assignment with memset to item assignment with std::fill, but keeping the same value (so each element is 0x000000FF instead of 0xFFFFFFFF). This should restore the correct behavior.


flaviojs:
Good catch, thank you. :)

PR #992, opened 2020-01-17
Make sound id variables UINT32

flaviojs:
The sound manager returns UINT32, so the variables for that should be UINT32 too.

Not to be confused with enum SoundID.
The enum and it's variables should be renamed to avoid confusion...

Fixes some sign-compare warnings.
For an unknown reason there are equivalent statements that don't warn... gcc bug?

Reference #857


lynxlynxlynx:
For the equivalents, you mean eg. iPositionSndID? I guess it's not warning, since it's not used in any comparison.

flaviojs:
No, it was some of the statements that I changed here. It's nothing serious.

PR #994, opened 2020-01-17
Fix sign-compare warnings of random values

flaviojs:
Reference #857


flaviojs:
I don't think the other side of the comparisons can be negative.
If they can then this fixes the wrong result (negative values would become big positive values).

PR #996, opened 2020-01-18
Fix sign-compare warning in AddCreaturesToBattle

flaviojs:
I don't know why, but the warning only happened in the double sum...
Since I don't understand, I'm converting the code instead of casting.

Reference #857


lynxlynxlynx:
Much cleaner this way, yes. :+1:

PR #998, opened 2020-01-19
Fix error message in DefaultContentManager::getNewString

flaviojs:
%d is not suitable for size_t

Introduced in #995
Found by coverity


lynxlynxlynx:
Merged, but github is acting up.

PR #999, opened 2020-01-20
Remove boost

flaviojs:
This PR aims to remove all the boost code, one function at a time.


flaviojs:
Good, I went smoother than expected. :)
I'll cleanup and remake the commits before asking for a review.

flaviojs:
Ok, this is ready for review.

src/sgp/FileMan.cc
```diff
@@ -672,70 +634,72 @@ FILE* FileMan::openForReadingCaseInsensitive(const std::string &folderPath, cons
...
-	std::vector<std::string> paths;
-	boost::filesystem::path path(dirPath);
-	boost::filesystem::directory_iterator end;
-	for(boost::filesystem::directory_iterator it(path); it != end; it++)
+	std::vector<std::string> ret;
+	std::vector<std::string> paths = FindAllFilesInDir(dirPath, sortResults);
+	for (std::string& path : paths)
 	{
-		if(boost::filesystem::is_regular_file(it->status()))
+		RustPointer<char> path_ext(Path_extension(path.c_str()));
```
selaux: Ideally we should do most of this logic in rust already, but for the time being it think it is ok...

src/game/Editor/Sector_Summary.cc
```diff
@@ -1981,7 +1981,12 @@ static void SummarySaveMapCallback(GUI_BUTTON* btn, INT32 reason)
 		{
 			if( gubOverrideStatus == READONLY )
 			{
-				FileClearAttributes(GCM->getMapPath(gszDisplayName));
+				std::string path = GCM->getMapPath(gszDisplayName);
+				if (!Fs_setReadOnly(path.c_str(), false))
```
selaux: Instincitvely I would say that `FileClearAttributes` does not equal to `Fs_setReadOnly`, but rather the opposite (removing read only flags). But that is just derived from the names of the functions.

rust/stracciatella_c_api/src/c/path.rs
```diff
@@ -0,0 +1,53 @@
...
+//! [`std::path`]: https://doc.rust-lang.org/std/path/index.html
+
+use std::ptr;
+
+use crate::c::common::*;
+
+/// Gets the extension of the path.
+/// Returns null if there is no extension.
+#[no_mangle]
+pub extern "C" fn Path_extension(path: *const c_char) -> *mut c_char {
```
selaux: We could also keep `Path` as a type for C++ and convert to `c_char` whenever needed in C++ code. This way we would represent paths as their original OS-dependent representation. I know its not done in other parts of the code, but might be a thought for the future...

rust/stracciatella_c_api/src/c/librarydb.rs
```diff
@@ -32,19 +32,14 @@ pub extern "C" fn LibraryDB_push(
...
-        Err(err) => {
-            remember_rust_error(format!("{:?}", err));
-            false
-        }
-        Ok(_) => {
-            forget_rust_error();
-            true
-        }
+    if let Err(err) = ldb.add_library(data_dir, library) {
+        remember_rust_error(format!("{:?}", err));
```
selaux: Might be good to include data_dir and library in the error message.

rust/stracciatella_c_api/src/c/librarydb.rs
```diff
@@ -79,23 +72,18 @@ pub extern "C" fn LibraryFile_close(file: *mut LibraryFile) {
...
-        Err(err) => {
-            remember_rust_error(format!("{:?}", err));
-            false
-        }
-        Ok(_) => {
-            forget_rust_error();
-            true
-        }
+    if let Err(err) = seek_result {
+        remember_rust_error(format!("{:?}", err));
```
selaux: Might be good to include library file name in the error message.

rust/stracciatella_c_api/src/c/librarydb.rs
```diff
@@ -106,18 +94,13 @@ pub extern "C" fn LibraryFile_read(
...
-        Err(err) => {
-            remember_rust_error(format!("{:?}", err));
-            false
-        }
-        Ok(_) => {
-            forget_rust_error();
-            true
-        }
+    if let Err(err) = file.read_exact(buffer) {
+        remember_rust_error(format!("{:?}", err));
```
selaux: Might be good to include library file name in the error message.

rust/stracciatella_c_api/src/c/fs.rs
```diff
@@ -0,0 +1,209 @@
...
+pub extern "C" fn Fs_setReadOnly(path: *const c_char, readonly: bool) -> bool {
+    forget_rust_error();
+    let path = path_from_c_str_or_panic(unsafe_c_str(path));
+    let result = fs::metadata(&path).and_then(|x| {
+        let mut permissions = x.permissions();
+        permissions.set_readonly(readonly);
+        fs::set_permissions(&path, permissions)
+    });
+    if let Err(err) = result {
+        remember_rust_error(format!("Fs_setReadOnly {:?} {}: {}", path, readonly, err));
```
selaux: I like this format for rust_errors :+1:.

rust/stracciatella_c_api/src/c/fs.rs
```diff
@@ -0,0 +1,209 @@
...
+        }
+        Ok(tempdir) => into_ptr(TempDir { inner: tempdir }),
+    }
+}
+
+/// Checks if the path exists.
+#[no_mangle]
+pub extern "C" fn Fs_exists(path: *const c_char) -> bool {
+    let path = path_from_c_str_or_panic(unsafe_c_str(path));
+    fs::metadata(&path).is_ok()
```
selaux: You could just use `path.exists()` here.

rust/stracciatella_c_api/src/c/fs.rs
```diff
@@ -0,0 +1,209 @@
...
+use std::time;
+
+use stracciatella::fs;
+
+use crate::c::common::*;
+use crate::c::misc::VecCString;
+
+/// A directory in the filesystem that is automatically deleted.
+/// The contents of the directory are deleted before the directory is deleted.
+pub struct TempDir {
```
selaux: Should we extract TempDir into its own file and make it independent from `Fs`?  As `Fs_createTempDir` is also not really dependent on `Fs`.

On second thought, is it really necessary do define our own `TempDir`? Can't we just return the one from the crate and define the `TempDir_` functions on that one?

rust/stracciatella/src/fs.rs
```diff
@@ -71,3 +103,73 @@ pub fn resolve_existing_components(path: &Path, base: Option<&Path>, caseless: b
...
+        // use statvfs in unix family
+        // TODO determine which of statvfs64/statfs64/statvfs/statfs are available
+        use std::os::unix::ffi::OsStrExt;
+
+        let mut bytes = path.as_os_str().as_bytes().to_vec();
+        bytes.push(0);
+        let mut data: libc::statvfs = unsafe { std::mem::zeroed() };
+        let result = unsafe { libc::statvfs(bytes.as_ptr() as *const libc::c_char, &mut data) };
+        return match result {
+            0 => Ok(data.f_bfree as u64 * data.f_bsize as u64),
```
selaux: You could use `u64::from()` in order to get rid of the `#[allow(clippy::cast_lossless)]`, right?

selaux:
I saw one warning, when loading a game. Not sure if it is related:

```
game/Tactical/Tactical_Save.cc: SynchronizeItemTempFile() Reported 0, should be 9
```

src/sgp/FileMan.cc
flaviojs: Yeah, I plan to rethink paths when the non-boost filesystem stuff is moved to rust.

src/game/Editor/Sector_Summary.cc
flaviojs: It is "removing the the read only flag" by setting the flag to false.
Maybe it should have a more neutral name? How about `Fs_updateReadOnly` or `Fs_changeReadOnly`?

rust/stracciatella_c_api/src/c/fs.rs
flaviojs: True, but I want to make it clear that it's touching the filesystem.

rust/stracciatella_c_api/src/c/fs.rs
flaviojs: It relies on the filesystem functions to create a special type of directory, so I think it qualifies for `Fs`.

Unfortunately `cbindgen` won't emit an opaque type for the original `TempDir`, so the wrapper is needed.

rust/stracciatella_c_api/src/c/path.rs
flaviojs: I plan to rethink paths when the non-boost filesystem stuff is moved to rust.

rust/stracciatella/src/fs.rs
flaviojs: With `u64::from()` you'll get a different warning when the value is already `u64`.
It depends on the target system, in the CI the linux and mac runners had different types.

flaviojs:
That warning is not related, it was around before I started playing with ja2s.
At the time I looked at the source and didn't understand what the warning really meant.
Everything seemed to work fine so I just ignored it.

flaviojs:
I wonder why I didn't implement `fmt::Display` for `LibraryFile` before...

rust/stracciatella_c_api/src/c/librarydb.rs
flaviojs: Done

rust/stracciatella_c_api/src/c/librarydb.rs
flaviojs: Done

rust/stracciatella_c_api/src/c/librarydb.rs
flaviojs: Done

selaux:
Thank you very much :heart:

PR #1001, opened 2020-01-29
Rework unicode locale initialization

flaviojs:
Windows does not support the utf8 locale, not sure how to proceeed here.

*nix, macosx and *bsd are set according to the info from python.

Problems are logged after the logger is initialized.


flaviojs:
Note: this fixes the warning that was being printed in windows when `ja2` was executed.

lynxlynxlynx:
If it works, just go ahead with the merge.

Issue #665, opened 2018-01-04
Externalizing game data


JAsmine-ja2:
Externalizing game data is definitely one of the most important parts of the "modding platform" aspect of the Stracciatella project.

### System
- [x] new hotkeys
- [x] stat progress bar in GUI
- [x] item shadows
- [x] game speed aka milliseconds per game cycle
### General gameplay settings
- [x] starting cash
- [x] drop all
- [x] always reveal all drops, regardless of distance
- [x] CtH settings
- [x] stealing/losing deliveries at Drassen airport
- [x] manually start turn-based mode
- [x] extra attachments on armor
- [x] elite soldier EXP level
- [x] min status % of items in enemy hands
- [x] minor AI improvements
- [x] enable multiple interrupts

### System
- [x] **music** - this feature (#387) is not clearly documented. Issues remain (#664).
### IMP creation
- [x] **IMP starting inventory** - extra items for skills and traits aren't externalized
- [x] **IMP creation attribute points** - works, but has issues (#644)
### Items
- [x] **ammo types** - can't be edited
- [x] **weapons** - number of indices can't be increased
- [x] **magazines** - number of indices can't be increased

### Army
- [ ] garrison groups (permanent presence in certain sectors)
- [ ] patrol groups (mobile squads moving around)
- [ ] army composition (details on the groups like number and probability of being replaced)

### Mapping:
- [ ] sector names
- [ ] sector loadscreens
- [ ] cities (location, size, name ...)
- [ ] movement costs (moving from one sector to another)
- [ ] accessibility of setors ("borderland", ocean, "bad sectors"...)
- [ ] SAM sites (sector, location of the computer)
- [ ] air control table (the area that is defended by each SAM site)
- [ ] BR shipping destination
- [ ] Bloodcats (sectors, number, chance to encounter)
- [ ] Coolness (per sector)
- [ ] underground sectors (increase max. no.)

### Items:
- [ ] armor
- [ ] other items
- [ ] attachments (increase max. no.)
- [ ] items, that are always available regardless of their coolness vs player's progress (#173)

### Misc. stuff:
- [ ] delivery methods of BR (cost, time)
- [ ] chopper refuelling stations / bases
- [ ] Possibility to increase stats with lvl 0 (zero)
- [ ] No. of mercs (AIM, MERC) that can die while "on assignment" per difficulty
- [ ] Progress %'s when Madlab, Mike and Iggy are made available
- [ ] Dialogues
[...]

Issue #652, opened 2018-01-01
hotkey wishlist #2


lynxlynxlynx:
Extracting the remaining stuff from #82:

> How about adding some hotkeys that reduce tedious micromanagement? 

- [ ] Ctrl-Shift-R to reload all weapons (our [alt-r](https://ja2-stracciatella.github.io/how-to-run/#extra-hotkeys) currently does not work for the whole squad)
- [ ] Ctrl-Shift-F/Ctrl-Shift-U (_or better_) to remove all attachments/magazines from all items on the ground in the current sector.
- [ ] a hotkey for closing websites (esc closes the whole laptop)

wvx:
The patch from 2016/2017 Inventory Management Extras (IME) does the 1.13 remove all ammo/attachments and stacks the accessible sector inventory at the active character's grid, the hotkey is shift-f but this could be changed if there is another preference - hotkeys should one day be configurable.

Reload all weapons is really easy (tm). We could just use your code from CTRL-N and do a reload instead of face autoswap.

wvx:
From #82, IME alt-shift click on map screen inventory will sell for 0. This effectively deletes the item but the idea is to have a configuration to sell for a small % of the original value as in 1.13

PR #1002, opened 2020-01-29
Update to unmerged mods for comment

wvx:
The "hack bypass" patches are a stop gap just to allow the game to be played but do not fix the underlying cause. These crashes are also unverified by other players.

Most changes are opt-in via game.json settings.


src/game/Tactical/Soldier_Control.cc
```diff
@@ -9085,6 +9215,751 @@ static void SetSoldierPersonalLightLevel(SOLDIERTYPE* const s)
...
+			DeleteObj(&pSoldier->inv[i]);
+		}
+	}
+	else //WORKAROUND for enemy soldiers use sector equipment
+	{
+		for(int i=0; i<NUM_INV_SLOTS; ++i)
+		{
+			switch(pSoldier->inv[i].usItem)
+			{
+				case KEY_1 ... KEY_32:
```
lynxlynxlynx: This is what breaks the build ... since it appears to be a non-standard gcc extension.

wvx: Thank you looking into the error, I have pushed an update (squashed into IME) in an attempt to solve this problem.

lynxlynxlynx:
Do you have a longer list already? Let me know once you're done.
In general some things will be out of scope for us, but on the other hand anything that externalizes settings/values is always welcome. Then there's the gray area in between.

wvx:
I do not intend to add significantly to these mods. IME contains many functions that could be useful for source-modders to deal with sector inventory and equipping soldiers. 

Issue #652 Ctrl-Shift-F is included in IME as Shift-F.

I like that there is an option for the militia to not get default equipment and so the player must scrounge to improve the sectors defense from the spoils of winning battles. But that is just my preference. The militia are auto-equipped so there is no necessity for micro-management. These features are disabled by default.

Militia control was intended for micro-managing militia placement in preparation of a sector invasion. But it can also be used to guide a battle with one auto-recruit in a sector which otherwise had no player merc. Or just enter a sector and do stuff as if radioed in. The militia can also be moved to another sector and un-recruited to serve as a guard, see #665 'Army'.

wvx:
Added attempt for: GCC: -Wmaybe-uninitialized,  -Wsign-compare, -Wstringop-trunction,  -Wformat-overflow=

flaviojs:
I don't quite get this PR... is it done?

This kind of looks like a bunch of mod PRs bundled together.
I would prefer one PR per mod, like this it's difficult to accept.
Mods that preserve default vanilla behavior are easily accepted (don't forget to document the vanilla values in new json settings).
Mods that change the default behavior need more care.

lynxlynxlynx:
We had a similar exercise before, where we just cherry-picked what was in scope or valuable (since it wasn't just about features previously either). Wasn't done for this one yet, but it's clear not all should be accepted. I can look into it starting with May, but other opinions on what to keep are of course welcome.

wvx:
Thank you for taking a look at the PR. I appreciate that it would be nicer to have separate PR per change. The patches are in order of dependency and perhaps eligibility. There should be no problem to include the first patches and skip everything else but I expect that it will be difficult to skip too many and include later ones only.

I will work on updating the PR to account for recent commits to master, maybe in the next weeks.

lynxlynxlynx:
I've opened a cleaned up PR for the compilation fixes.
Hacks are dubious and should mostly be asserts.

c02c3e11 - this is only for the end/delete display, so not affecting AI, but can't it be made to use the actual LOS range of the (first) selected soldier? What was the reason for the change?

d08fcd2 - needed an update and full of whitespace changes, but otherwise good. Queued for separate PR.

6be35c15 - fine, ripped out and queued.
abaadfe - looks fine, but depends on a truckload of code from IME a875539, so skipped.
a875539 - not sure if finished, but much of it seems out of scope to me, skipped.

4642ee85 - looks good, but tactical_realtime_save_allowed should not be part of this and please only make it affect the fully rested case. The others are very rare, so the player should get notified if they happen. :poland: 

"mods":
- 94ca0f31: don't play ironman then!?
- 838e119: out of scope
- 718bb8d: looks unfinished, but also out of scope
- b4fd08f7: unfinished
- ae178ac8: I like it, it's an often requested feature, but I'd skip the NPC part and if you look at game.json, we already have an imp object where the new settings should go (renamed)  :poland: 
- 8d1ad686: looks fine, though I'd be surprised if there wasn't a function already to get that SECTORINFO and there's no need for so many parentheses  :poland: 

I'll look at the three externalization commits next.

Stuff that could still be included is marked with the _Polish_ flag. :grin:

ja2-derek:
>     * [8d1ad68](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/8d1ad686ce07b6ec108bda45bf796d2cec1ae8f6): looks fine, though I'd be surprised if there wasn't a function already to get that SECTORINFO and there's no need for so many parentheses  🇵🇱

Should be doing it like `SectorInfo[ SECTOR( gWorldSectorX, gWorldSectorY ) ]`

lynxlynxlynx:
702dd9f2: acceptable, but it will clearly have merge conflicts.

cb7fbe035: nice, but please use a lambda instead of the final define, assert that PROGRESS_PORTION_TOTAL is 100 and then you don't need to do the divisions, ubKillsPerPoint has a bad fallback (should be 7). If ubKillsPerPoint is 0, shouldn't usKillsProgress also be 0? 

5d1119da: vanilla externalizations are fine, the rest is out of scope

wvx:
There have been some recent changes to master that conflict with some commits and which I have not looked at yet.

abaadfe731179249443a307013008a04203a2730 IME is essentially a collection of helper functions that could save future mod writers from reinventing the wheel and focus the extra time on their mod. SoldierAutoSwitchGoggles was easy to implement because of the IME functions to place, attach and detach. The IME functions have not been added as a separate commit because there did not seem to be a justification on their own, without a mod shell. Perhaps SoldierAutoSwitchGoggles could be that justification for separating the other mods if IME functions can be included as a dependency.

a87553991a82cab2ce4618c1e0598e14319b9ea9 IME works as advertised. The vehicle inventory graphics I hope will be done by someone else but I use it as-is. 

718bb8d237380ee5cdd75c30383dd5b59d5d128a This mod makes the elite encounters feel special and the battles more difficult. It combos well with the earlier AI mod that let AI choose shots based on chance to hit and your multiple-interrupts mod. Elites at level 10 will often get the interrupt.

cb7fbe03564ab16bf9f5387da9369e7ee9110487 The reason for the division is to allow the user some flexibility in their choice of values, as implemented they can choose any values and the division will rescale to 100. There is another mistake besides the default value that has been corrected but I did not push some weeks ago because of new conflicts with master. The kills_per_point_0 should be changed to kills_per_point_1 etc. If ubKillsPerPoint is 0 I decided was a nonsense value but as suggested it could be used as an ignore value. Feel free to make any changes.

b4fd08f7466ea1761fee7a91533609f696b8f3c0 does still have the roof glitch, it is easy to reproduce and will some day be looked at. This mod is useful to abstract the aiming clicks as used by 5d1119da2e77ce448962414e42a75e3bb17c2d3f

ae178ac85f170770a1d74f63bcd9450c39dffc6f The NPC addition can be seen as an externalization of the NPC inventory and stats, all that is missing is a JSON loader. It is offered as an option to use in other mods or to play.

94ca0f312e5b8af88c3526afde78ec715ab60668 I prefer pure ironman but Straciatella is unstable in my testing and has been for years hence the "hack" commits which have saved me no small amount of frustration when an ironman game crashes. The mod still disables turn-based saves. But a reasonable question has been asked - why not just play with saves enabled and choose not to reload? And I offer the same reason that games such as JA2 and many others have ironman modes.

5d1119da2e77ce448962414e42a75e3bb17c2d3f Perhaps other players, like myself will enjoy experimenting with these parameters. They are optional.

I appreciate that some of these commits contain large mods and that they significantly change the gameplay. They are inspired by 1.13 features that I enjoyed and offered so that others may have the opportunity to enjoy and customize their game beyond vanilla. I have no expectation that any will be included. Feel free to choose the parts that you like.

lynxlynxlynx:
IME: 10 different things definitely don't belong to the same commit, especially when the message mentions WIP several times. As far as goggles go, we have our putDayHeadGear already, which the new code doesn't use, so it looks like a lot of duplication.

cb7fbe0: Are you sure that rounding errors can't produce a value < 100? It's cleaner and easier to understand if we just force it to a total of 100. If the user can't add up numbers, we don't have to jump through hoops for it, since the error message will make it clear.

ae178ac Still, it's doing two things at once.

94ca0f3 Sure, but it can stay as a local change.

All the game.json values need a fallback if the key isn't present. Now that we'll be making it easier to use an outside file, it could easily get outdated.

wvx:
I will try to make an effort to separate IME into dependency functions as one commit and then as many commits that are independent if the team would prefer that. Maybe IME base, SoldierAutoSwitchGoggles, then others.

SoldierAutoSwitchGoggles will choose to place face items as the situation changes such as mask in gas, sun glasses or UV goggles above ground or remove if nothing is suitable. Using the hotkey twice will not toggle unless the situation changes. The function also accounts for an earlier mod to allow face items as helmet attachments. The AI has also been given this function and they benefit from acquired or placed items. If they do not already look for dropped gear and a future mod is made that does so then the logic is already in place to correctly switch.

ae178ac85f170770a1d74f63bcd9450c39dffc6f can be separated into IMP / NPC parts.

cb7fbe03564ab16bf9f5387da9369e7ee9110487 I like that the user is free to choose numbers instinctively and we can add some post-processing that ensure that 100 progress is reached for the sake of game events and future mods but I agree that this may be an unnecessary complication for a minor benefit.

src/game/Tactical/Weapons.cc
```diff
@@ -797,7 +797,7 @@ static BOOLEAN UseGun(SOLDIERTYPE* pSoldier, INT16 sTargetGridNo)
 	}
 	else
 	{
-		uiHitChance = CalcChanceToHitGun( pSoldier, sTargetGridNo, pSoldier->bAimTime, pSoldier->bAimShotLocation );
+		uiHitChance = CalcChanceToHitGun( pSoldier, sTargetGridNo, pSoldier->bAimTime, pSoldier->bAimShotLocation, true );
```
lynxlynxlynx, line 800: These changes fix the erroneus degradation of attachments and should be a separate commit. We could include it sooner, since the circle stuff still needs the roof fix.

src/game/Tactical/UI_Cursors.cc
```diff
@@ -162,6 +162,34 @@ UICursorID GetProperItemCursor(SOLDIERTYPE* const s, GridNo const map_pos, BOOLE
...
+{
+	bool const is_throwing_knife = GCM->getItem(pSoldier->inv[HANDPOS].usItem)->getItemClass() == IC_THROWING_KNIFE;
+	SOLDIERTYPE const* const target = gUIFullTarget;
+	UINT8 chance_to_hit, chance_to_clear, chance_on_target;
+
+	if (!target)
+	{
+		chance_to_hit = SoldierToLocationChanceToGetThrough(pSoldier, map_pos, gsInterfaceLevel, pSoldier->bTargetCubeLevel, 0);
+	}
+	else
```
lynxlynxlynx, line 175: does this also work for melee?

wvx: I will look into adding melee case to the AAC mod. There is CalcChanceHTH for melee and maybe we would have to consider special cases like if the attacker is unseen.

src/game/Tactical/Weapons.cc
wvx: accurate aim circle has now been fixed but there is another issue. Shots aimed at the head that should hit will always miss depending on cursor position or something. SoldierToSoldierBodyPartChanceToGetThrough and CalcChanceToHitGun will report 100 chance-to-hit and therefore so will the mod. The actual bullet looks like it is hitting the level below. The result is identical with the mod disabled. This could be separated in the next update.

lynxlynxlynx: If you can easily reproduce it, definitely report it as a separate issue.

lynxlynxlynx:
Ah, didn't notice you force pushed again, will check some of the splits.

lynxlynxlynx:
Done in #1133. There were some other candidates, but most of the commits weren't independent, so in the end I ignored them if they didn't build due to needed missing code.

PR #1003, opened 2020-02-06
Use c++ memory

flaviojs:
The aim of this PR is to ensure constructors/destructors are always called.

It is needed for #937, which will make some POD structures become non-POD. The constructor/destructor needs to be called to avoid nasty surprises.

Functions like `MemAlloc` and `MemFree` will not call the constructor/destructor. I have no specific strategy to ensure the constructor/destructor is called, it's gonna be handled on a case-by-case basis.

----

FIXMEs added


flaviojs:
Hmm, replacing C++ new/delete seems like the way to go.
I'll restart and replace all mallocs/frees.

flaviojs:
Hmm, should I place FIXMEs in the bug locations instead fixing them here?

lynxlynxlynx:
Does it work without the fixes? If not it'd be just a very slight improvement to have a separate PR.

flaviojs:
This PR will work without the fixes.
I'll place FIXMEs and fix them after this PR is merged.

flaviojs:
Done.

With the notable exception of stuff that is directly read from file or directly written to file, everything else should have the constructor/destructor called.

lynxlynxlynx:
This is very tedious to review. I suggest the lazy way out, merging and letting coverity complain if anything is off that a compiler and your test runs didn't catch.

flaviojs:
Parts of the code where coverity detected problems changed, so I expect it to complain regardless of the result of the review.

The lazy approach is fine (just merge) since I'll be here to fix the problems. :)

PR #1005, opened 2020-02-25
Fix sound in SDL 2.0.6

flaviojs:
SDL 2.0.6 crashes in SDL_ResampleAudio and does not request enough memory for SDL_Convert_U8_to_F32_SSE2 (corrupts memory).

To avoid problems we use a custom resampler (linear interpolation) and allocate more buffer memory.

The name of the smacker audio streams was useless. Now it has the channel index and the filename.

---

Reference #608 #615

Tested with a manual compile of SDL 2.0.6, no more crashes. ^^

### DIFF
--- a/src/game/MainMenuScreen.cc
+++ b/src/game/MainMenuScreen.cc
@@ -76,7 +76,6 @@ static void HandleMainMenuInput(void);
 static void HandleMainMenuScreen(void);
 static void RenderMainMenu(void);
 static void RenderGameVersion(void);
-static void RenderSDLVersionWarningIfNeccessary(void);
 static void RenderCopyright(void);
 static void RestoreButtonBackGrounds(void);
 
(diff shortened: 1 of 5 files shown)


flaviojs:
Looks like it make a mistake somewhere, resampled sounds seem too fast.

flaviojs:
Thanks.

While test playing my fixed code I noticed memory corruption...
It appears that SDL 2.0.6 doesn't request enough buffer memory.
I'm inclined to move audio conversion to rust if I find a good crate.

lynxlynxlynx:
Considering sdl has already been at 2.0.10 for a while, is all this even worth the bother? Time heals all.

flaviojs:
SDL 2.0.7+ has [audio streams](https://wiki.libsdl.org/Tutorials/AudioStream), which simplifies things (music should be streamed).
I wanted to use them but that means not supporting ubuntu xenial (sdl2 2.0.4).
The only I'm trying to make SDL 2.0.6 work is because it's in the supported range.

flaviojs:
Found the solution for the memory corruption, everything is good now.

While trying rust crates for audio I got more familiar with interpolation techniques and decided to implement linear interpolation, which is cheap and produces little audible noise.

lynxlynxlynx:
If you're not looking for further review, just merge it.

PR #1008, opened 2020-03-20
Refine the build system

flaviojs:
I am not sure how far I will take this PR, there is no specific objective.

Are you dissatisfied with something in the build system?
If yes, tell me about it and I might do something about it here.


wvx:
Hello, could you make it so that `make` does not need to download anything. So that we can build off-line with just the git source.

Some time ago I got a compile on Android working under XSDL or something like that. So any change that allows for an easy build to get the game playable on an Android device would be great.

Thank you.

flaviojs:
1) cargo(rust) gets the dependencies from the web
2) sdl2 is required, it can use the system lib or use one of the 3 lib variants in git (urg... bloat)
3) rapidjson is required, it can use the system lib or used the included code
4) fltk is optional (needed for the launcher), it can use the system lib or download and build it
5) gtest is optional, it can use the system lib or download and build it
6) editor.slf is optional, it can download it

If your system provides the libraries then you just need to figure out how to make cargo work offline.

If your system does not provide the libraries, you can probably compile them yourself and use an environment variable to point to your library. (look at the corresponding FindXXX file)

flaviojs:
Oh, I forgot the new string_theory library.
Right now it always downloads, I can make it search for a system lib.

wvx:
Thanks for taking a look at the online dependencies. If you feel like it, try and get a native compile on an Android (ARM) device. I got it to work once after much tweaking and not again. The game played well except for the double cursor but some devices can disable the OS cursor.

flaviojs:
The critical issue for any work on ARM is setting up an automated build+test.

We use github, travis and appveyor right now, if you find (or make) a good ARM example for one of them then I can look into it. Note that build times are long so using emulation (like qemu) will probably hit the time limit.

flaviojs:
Done.

I'll stop here since my `find_package` knowledge is too low to do the rest with confidence.

lynxlynxlynx:
Looks fine to me, besides jumping the gun on the version update.

flaviojs:
I didn't jump the gun, I changed the versioning strategy to semantic versioning:
 * 1.16.x is an invalid version
 * 1.16.1-dev < 1.16.1 < 1.16.2-dev < 1.16.2 < 1.17.0-dev < 1.17.0

Since the next version will be 0.17.0, I'm using 1.17.0-dev instead of 1.16.2.

flaviojs:
Hmm, this is only marginally related to the build system so I'll split it into it's own PR.

lynxlynxlynx:
This tagging is imposed by the standard? To me it looks backwards — I'd expect 1.16.1-dev to come after 1.16.1, since we don't know the next release number ahead of time.

flaviojs:
It uses the same logic as release candidates. (1.16.1-rc < 1.16.1)

The patch number is reserved for bug fixes only, so 1.17.1 would be the next version in the 1.17.x branch, and 1.18.0 would be the next version in the master branch.

The major version is reserved for breaking changes. If it's still 0 then the minor version can have breaking changes.

Err, I just noticed that I wrote 1.XXX in these posts and should have written 0.XXX.

lynxlynxlynx:
Yeah, standard stuff, but we don't know ahead of time what sort of promises we'll be breaking, which fields will need to be upped. Perhaps having a version for HEAD is just silly, since it's the build "number" that matters.

flaviojs:
We are at major version 0 so there are no promises. We can break as much as we want and HEAD is always "(next minor)-dev" until we decide to release a 1.0.0.

When releasing 1.0.0 we must decide what is a breaking change. HEAD is "(next minor)-dev" before a breaking change is added and becomes "(next major)-dev" with the breaking change.

Not having released 1.0.0 makes things easy for us. :)

PR #1009, opened 2020-03-29
Rustify paths

flaviojs:
This PR aims to add support for all paths by rustifying them, needed for #937.

The strategy is to use utf8 strings without embedded nuls in the C++ code.
`AnyPath` converts to/from `Path` in a way that is compatible with `String` and `CString`.

### DIFF
--- a/src/game/Utils/STIConvert.cc
+++ b/src/game/Utils/STIConvert.cc
@@ -53,9 +53,6 @@ static BOOLEAN ConvertToETRLE(UINT8** ppDest, UINT32* puiDestLen, std::vector<ST
 
 void WriteSTIFile(UINT8* const pData, SGPPaletteEntry* const pPalette, const INT16 sWidth, const INT16 sHeight, const char* const cOutputName, const UINT32 fFlags, const UINT32 uiAppDataSize)
 {
-
-	FILE *							pOutput;
-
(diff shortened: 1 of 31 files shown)


flaviojs:
Done.

I test played it up to Drassen.

src/game/Utils/STIConvert.cc
```diff
@@ -127,28 +130,43 @@ void WriteSTIFile(UINT8* const pData, SGPPaletteEntry* const pPalette, const INT
...
 
 	// write app-specific data (blanked to 0)
+	const uint8_t zero = 0;
 	for (uiLoop = 0; uiLoop < Header.uiAppDataSize; uiLoop++)
 	{
-		fputc( 0, pOutput );
+		if (!File_writeAll(file.get(), &zero, sizeof(zero)))
+		{
+			RustPointer<char> err(getRustError());
+			SLOGW("WriteSTIFile: %s", err.get());
```
lynxlynxlynx, line 166: Opening a random commit ... it seems like we could use another helper function or three for all this logging.

flaviojs: Maybe after all the `wchar_t` code is converted (#937).

While coding, the `printf` format strings were an annoyance so I will probably target them.
Logging functions are one of those targets.

PR #1013, opened 2020-04-04
Add fltk to build dependencies

lucasmr:
Tried to compile without fltk, the ja2-launcher was not built.


lynxlynxlynx:
Thanks, but no, it's an optional dependency, not required, so it doesn't fit on that list. The real question worth looking into is why it didn't fallback to our internal copy.

lucasmr:
Alright, no problem.

PR #1015, opened 2020-04-06
Quick fix for FPE in CalcBestThrow when MinAPsToThrow returns 0

### LINKED ISSUE #1006: Occasional FPEs (div by zero) when AI tries to throw a grenade
I've noticed in the last few months (verified on the indicated hash) that the game occasionally bombs with a floating-point exception; upon further investigation it's a division by zero at https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/game/TacticalAI/Attacks.cc#L944. Digging to the bottom of it is unfortunately more time than I have to devote to this (not nearly familiar enough with the codebase). I'm afraid I don't have a reliable repro, but the attached save "often" crashes: move Carlos one square to the "west", so as to get LoS (once the smoke clears) on the row of squares between the burned out tank and the sandbags, and the AI will *usually* shoot him (once the smoke clears) and then trigger the crash. Github identifies #701 as possibly related, but I'm not able to evaluate that possibility.

I don't recall ever seeing a crash in this sort of situation in vanilla or in older (before the past 6-8 months or so) Stracciatella, but of course I can't be certain. It does seem to occur often enough that I think I'd have noticed earlier if it was present.

Tested on voidlinux, gcc 9.2.0; other environment attributes available on request. 

| 38f5f8f3ddaf21d29c224c33a7f9c856150e3f6f  | JA2 from GOG  |    unknown, suspected no    |    unknown, suspected yes     |

[fpe.zip (happens to be a combat quicksave)](https://github.com/ja2-stracciatella/ja2-stracciatella/files/4259526/fpe.zip)

flaviojs:
The FPE is a divide by zero while calculating iHitRate.

MinAPsToThrow warned and returned 0 on unexpected input, now it just warns.

It is unclear how a launcher should be trully handled:
1) GLAUNCHER has item class IC_LAUNCHER.
2) MinAPsToAttack calls MinAPsToShootOrStab for item class IC_LAUNCHER.
3) CalcBestThrow calls MinAPsToThrow and has the comment "NB grenade launcher is NOT a direct fire weapon!"

Fixes  #1006

### DIFF
--- a/src/game/Tactical/Points.cc
+++ b/src/game/Tactical/Points.cc
@@ -1685,12 +1685,10 @@ INT16 MinAPsToThrow(SOLDIERTYPE const& s, GridNo gridno, bool const add_turning_
 	if (!item)
 	{
 		SLOGW("MinAPsToThrow - in-hand item is missing");
-		return 0;
 	}
 	else if (!(item->getItemClass() & (IC_GRENADE | IC_THROWN))) // match MinAPsToAttack
 	{
 		SLOGW("MinAPsToThrow - in-hand item '%s' has unexpected item class 0x%x", item->getInternalName().c_str(), item->getItemClass());
-		return 0;
 	}
 
 	if (gridno != NOWHERE)


lynxlynxlynx:
LGTM!

PR #1016, opened 2020-04-08
Allowing more Town Sectors

ja2-derek:
This is not an issue with vanilla, but for future map mods.

Before this change,  `g_town_sectors` in Strategic_Town_Loyalty [limits to 40](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/36708ba01af1564cb6fd4e13cbc58533cd8d0e0d/src/game/Strategic/Strategic_Town_Loyalty.cc#L78). If the world has more than 39 town sectors, `BuildListOfTownSectors` will [iterate past the end of array](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/36708ba01af1564cb6fd4e13cbc58533cd8d0e0d/src/game/Strategic/Strategic_Town_Loyalty.cc#L692), causing a "stack-based buffer overrun"  exception.

This change should bring no side-effects other than wasting a few more CPU cycles [iterating on empty elements](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/36708ba01af1564cb6fd4e13cbc58533cd8d0e0d/src/game/Strategic/Strategic_Town_Loyalty.h#L91).

There are 36 town sectors in Vanilla, 40 in Wildfire maps. The new `MAX_NUM_OF_TOWNS_SECTORS`  value of 50 is chosen rather arbitrarily, so that is enough for Wildfire and not unnecessarily large.

### DIFF
--- a/src/game/Strategic/Strategic_Town_Loyalty.cc
+++ b/src/game/Strategic/Strategic_Town_Loyalty.cc
@@ -73,9 +73,10 @@
 // town loyalty table
 TOWN_LOYALTY gTownLoyalty[ NUM_TOWNS ];
 
+#define MAX_NUM_OF_TOWNS_SECTORS 50
 
 // Town names and locations
-TownSectorInfo g_town_sectors[40];
+TownSectorInfo g_town_sectors[MAX_NUM_OF_TOWNS_SECTORS];
 
 
 #define BASIC_COST_FOR_CIV_MURDER	(10 * GAIN_PTS_PER_LOYALTY_PT)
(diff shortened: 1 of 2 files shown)


flaviojs:
I would prefer if `g_town_sectors` was turned into a `std::vector<TownSectorInfo>`. (no upper limit)
Want to try it? (we compile the sources as C++11)

derek9gag:
> I would prefer if `g_town_sectors` was turned into a `std::vector<TownSectorInfo>`. (no upper limit)
> Want to try it? (we compile the sources as C++11)

Sure! That sounds great. I will update.

ja2-derek:
Done! Feels much cleaner and modern now.

selaux:
Would be cool if all town related stuff could be extracted to externalized data, so eventually mods can define their own towns and dont have to use the vanilla stuff. But this is probably out of scope of this PR.

selaux:
I just saw your proposal for that in the Wildfire maps bug ticket 👍

PR #1018, opened 2020-04-09
Externalizing BloodCat placements

ja2-derek:
Externalizing strategic map data from code to json. Bloodcat placements are relatively simple and independent so I am starting here.

See #665 and #1011. 

## Summary

- There are 2 types of placements. First is the "regular" placement which are slots for bloodcat ambush. Second are bloodcats that are always present in the Lair and Arena sectors, which I call "spawns"
- Added models classes and deserialization for Bloodcat placements
- Updated GameInit to use externalized data
- Replaced a few places that uses `SEC_I16` and `SEC_N5` with the externalized fact `isLair` and `isArena`  

## LoadSave

There are some BloodCat placement patching logic in [LoadStrategicAI](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/9f76684de4cbab33ad8e31d756d2e6f5de44ede5/src/game/Strategic/Strategic_AI.cc#L2633), for `( ubSAIVersion < 18 )` and `f( ubSAIVersion < 22 )`.

I suppose these are for only patching saves from very old JA2 versions. I am not touching those.

Issue #1011, opened 2020-04-04
Support for Wildfire maps


ja2-derek:
I was bored with the vanilla maps, so I tried playing on Wildfire maps with Straciatella, and I got a working proof-of-concept. I probably is not the first one to try, but I wonder how much interest there is. I may be able to help make this into a JA2S mod.

I am able to get in-game with WF maps, though I haven't completed a play-through nor tested extensively. I am reporting my progress so far, hoping some of you would also be interested:

## Using Wildfire maps

These are what I did so far:

- Export assets from Wildfire data directory, I am using the Steam version. Extract the below files and place in a mod directory
  - Data/BinaryData/JA2set.dat
  - Data/Interface/b_map.sti, convert to 8-bit pcx as b_map.pcx (See #712)
  - Data/Maps/**
  - Data/Radarmaps/**
  - Data/TileSets/**
- The maps should be loading fine at this point once the mod is enabled
- There are more town sectors in Wildfire maps. To enable those, edit:
  - InitializeStrategicMapSectorTownNames(void), add A1,B1,C14,F2 as town sectors
  - GARRISON_GROUP, add garrison groups to those sectors
  - g_traverse_we, g_traverse_ns and g_traverse_through, change those to town sectors
  
## Known issues

- Some sectors crashes the Map Editor. For example I6 (Estoni). I haven't been able to figure out why.
- Wildfire has different items, so you will be getting wrong pre-placed items(https://github.com/ja2-stracciatella/ja2-stracciatella/issues/321#issuecomment-221561534)
-  Some rooms and  key locations moved, which means:
   - Some NPC AI movement scripts fail, e.g. Kyle in San Mona. The NPC scripts, and any hard-coded gridNo needs to be updated with WF ones.
- Some civilians outside of San Mona are in Kingpin's faction. For example some civilians may attack you in Drassen Airport if Kingpin is hostile.

## What can be done

If there is interest, we can make it into a modpack and provide basic (and experimental) support of Vanilla-like experience on WF maps.

Regarding items differences, I don't know if we can ignore or somehow replace those WF items. I don't know how JA2 1.13 handle those issues.

For the strategic map differences, we can externalize Sector Info and Garrison Groups and make them mod-overridable.

## Screenshots

![screen1](https://user-images.githubusercontent.com/63151803/78450930-d6505080-76b4-11ea-8fa2-880b8b341da0.png)

[...]

Issue #254, opened 2016-03-10
Does Unfinished Business work with Stracciatella?


lauri:
Sorry about (ab)using the issue tracker to ask this question but I couldn't find a definitive answer anywhere.

Is it possible to play JA2: Unfinished Business with Stracciatella? Simply pointing ja2.ini to the UB data files does not seem to work.

JAsmine-ja2:
It would be super-cool if Stracciatella could support both UB and the Demo #238. But that would probably be waay too much work and our (few) resources are probably spent better elsewhere..?

selaux:
Is the source code of Unfinished Business ist not publicly available, ihmo there is actually no way to make this work. We could try to make it possible to at least load the assets from Unfinished Business, but not sure if that will work either.

lynxlynxlynx:
It would become a reverse-engineering project, which is definitely hard, but not impossible. I have no idea about the differences between the two games.

selaux:
Actually I just found out that the source code is indeed avaiable [here](http://kermi.pp.fi/JA_UB/Source/). But finding the differences would be really hard. For now we might want to check first why the assets cannot be loaded.

gdm85:
Code-wise, I would diff the UB sources versus the JA2 sources and try to figure out what are the logic/parameters changes. Haven't had time to give this approach a try but I guess some people here would have better luck at it.

Manuel-K:
@gdm85 Yes, that is the obvious approach to the problem. 

I've located an unchanged version of the JA2 source code and converted the UB source code to unix line endings. The location of the SGP folder is different between the two releases, but that is a trivial fix. Unfortunately the JA2 source code has some language strings inlined. And there are some whitespace changes between the two source trees. This makes things more unclear than they'd have to be.

It's still a lot of work.

Manuel-K:
I'm in the process of cleaning up both source trees and moving things around so that both trees have the same folder structures and (if possible) the files at the same location.

@selaux @lynxlynxlynx Is it okay if I create a new repo at ja2-stracciatella/ja2-ub-comparison?

selaux:
Fine for me.

lynxlynxlynx:
Sure. Don't forget about the various space-ignorant options when comparing.

Manuel-K:
[...]

PR #1020, opened 2020-04-11
Externalizing Town and Town Sectors data

ja2-derek:
Externalization of Town and Town Sector data, in a similar manner as #1018. 

See #665 and #1011.

## Summary

- Externalizing basic town info. See assets/externalized/strategic-map-towns.json for the schema
- Adding TownModel class, which can be retrieved by `townId`
- Replacing the hard-coded tables of `InitializeStrategicMapSectorTownNames`, `sBaseSectorList`, `pTownPoints` and `MilitiaTrainingAllowedInTown`

##  What's not included

* Town names - they are defined in translation string files
* Is Town hidden - codebase has very specific logic for Tixa and Orta. It will require a lot of code change to generalize the logic
* Base sector - it is now calculated at init time

This PR conflicts with #1018. I will re-base when it is merged to master.

### DIFF
--- a/src/externalized/strategic/TownModel.cc
+++ b/src/externalized/strategic/TownModel.cc
@@ -22,7 +22,7 @@ TownModel* TownModel::deserialize(const rapidjson::Value& obj)
 	std::vector<uint8_t> sectorIDs;
 	for ( auto& s : obj["sectors"].GetArray() ) 
 	{
-		sectorIDs.push_back( SECTOR_FROM_COORDINATES( s.GetString() ) );
+		sectorIDs.push_back( SECTOR_FROM_SECTOR_SHORT_STRING( s.GetString() ) );
 	}
 
 	Assert( obj.HasMember("townPoint") && obj["townPoint"].IsObject() );
--- a/src/externalized/strategic/TownModel.cc
+++ b/src/externalized/strategic/TownModel.cc
@@ -1,0 +1,41 @@
+#include "TownModel.h"
+
+#include "JsonObject.h"
+
+TownModel::TownModel(int8_t townId_, std::vector<uint8_t> sectorIDs_, SGPPoint townPoint_, bool isMilitiaTrainingAllowed_)
+		: townId(townId_), sectorIDs(sectorIDs_), townPoint(townPoint_), isMilitiaTrainingAllowed(isMilitiaTrainingAllowed_) {}
+
+const uint8_t TownModel::getBaseSector() const
+{
+	int8_t minX = 99, minY = 99;
+	for ( auto sectorID : sectorIDs ) {
+		int8_t x = SECTORX(sectorID), y = SECTORY(sectorID);
+		if ( x < minX ) minX = x;
+		if ( y < minY ) minY = y;
+	}
+	return SECTOR( minX, minY );
+}
+
+TownModel* TownModel::deserialize(const rapidjson::Value& obj)
+{
+	Assert( obj.HasMember("sectors") && obj["sectors"].IsArray() );
+	std::vector<uint8_t> sectorIDs;
+	for ( auto& s : obj["sectors"].GetArray() ) 
+	{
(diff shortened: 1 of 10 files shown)


src/externalized/strategic/TownModel.cc
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	Assert( obj.HasMember("townPoint") && obj["townPoint"].IsObject() );
+	auto tp = obj["townPoint"].GetObject();
+	SGPPoint townPoint = SGPPoint();
+	townPoint.iX = tp["x"].GetInt();
+	townPoint.iY = tp["y"].GetInt();
+
+	JsonObjectReader reader(obj);
+;	return new TownModel(
```
lynxlynxlynx, line 35: There's an extraneous semicolon here.

lynxlynxlynx:
`TownModel.cc:25:63: error: ‘SECTOR_FROM_COORDINATES’ was not declared in this scope`

lynxlynxlynx:
LGTM.

Should also the names be externalized? I don't know if WF left them the same, but now there's just the choice of the existing ones.

ja2-derek:
I am not sure. I can put the English towns names in the JSON, but we will lose the localizations like this:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/8e2396f0021f97b36cd89ba2d894725ecb796e38/src/game/Utils/_RussianText.cc#L217-L231

Or should we externalize the whole translation file instead? What do you think?

ja2-derek:
Town names in Wildfire.

image

lynxlynxlynx:
How to make shit sound Spanish 101. ;) So for WF, either some extra code is needed, with San Mona as the outlier, or the names replaced in all languages.

In the longterm the whole translations as non-code sounds better, but let's not get distracted. I suggest we add an array to the cities json to hold the original and all the translated versions. Then your WF mod can still achieve its thing by simply overriding it with its own file. 

Is the id used also for something else? I guess since there's a 1:1 map, it doesn't really matter.

ja2-derek:
I personally prefer vanilla town names lol

lynxlynxlynx:
The point is that any new data projects can then have any names they like, it's not just for WF. But we can leave it be for now.

flaviojs:
It's better to implement only what is needed since using json implies coding a lot logic.

I would wait for a true scripting language (lua? js?) to be integrated before thinking of what future mods need.

lynxlynxlynx:
It's not just about the future though, but also eg. Unfinished Business. We haven't reviewed and merged all the code changes yet #254 , but the more is externalized, the fewer conflicts need to end up as ifdefs etc.

ja2-derek:
While we are at town names... do you know Russian or Polish? What does "locative" mean and how important is it?

It only seems make a difference in RUS and PL.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/8e2396f0021f97b36cd89ba2d894725ecb796e38/src/game/Utils/_RussianText.cc#L234-L249

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/8e2396f0021f97b36cd89ba2d894725ecb796e38/src/game/Utils/_PolishText.cc#L234-L249

lynxlynxlynx:
It's from declination. Without it, nouns can look completely wrong in a sentence. I think it's present in all Slavic languages, no Romanic and don't know about the rest.

ja2-derek:
I think I will not put those inside the TownModel, but rather put that under externalized/strings, and handle with `ContentManager::loadStringRes`, just like ammo calibres.

lynxlynxlynx:
That works too, but is harder to use. Considering the low activity of the community, I don't think it matters that much.

ja2-derek:
Yes, it would be less straight-forward to mod creators, but more convenient too translators.

lynxlynxlynx:
This one is now complete, right — I can merge?

ja2-derek:
Yes please!

lynxlynxlynx:
Thanks!

PR #1021, opened 2020-04-11
Externalizing Movement Costs

ja2-derek:
Externalizing Movement Costs on the strategic map (Strategic_Movement_Costs.cc). 

Again, see #665 and #1011.

This PR conflicts with #1018 and #1020. This needs to be re-based and re-tested before merging to master.

## Summary

- JSON data is a direct translation from the original C++ code
- Extracted tables replaced the arrays: `g_traverse_we`, `g_traverse_ns`, `g_traverse_through` and `g_travel_rating`
   - Noted that `ubTravelRating` is not used for anything meaningful
- Removed a unit test which tests on the table data; Added equivalent Assert statements in model class.

## Testing

Manually tested like this:

1. Start a new game, with 0.16.1
2. Hire somebody and proceed to land in A9; GABBI-killed all enemies in A9
3. Try to move across strategic map. Record the ETA to different destinations.
4. Repeat 1~3 after applying this PR to master. Verify ETAs are exactly the same
5. Apply the data file override for Wildfire Maps mod, which modifies some traversibility values. Repeat 1~3 and verify getting different ETAs.

Expected results:
| From | To | Travel ETA  (0.16.1) |
|---|---|---|
| A9 | A10 | 5m |
| A9 | A8 | 1h 44m |
| A9 | A1| 15h 40m |
| A9 | A15 | 10h 13m |
| A9 | I6 | 12h 52m|
| A9 | H1| 15h 30m |
| A9 | I16 | 15h 45m |
| A9 | P1 | 22h 6m |
| A9 | K4 | 22h 32m |
| A9 | M13 | 20h 31m |


assets/externalized/strategic-map-movement-costs.json
```diff
@@ -0,0 +1,115 @@
+/**
+ * Movement costs of moving across the strategic map
+ */
+{
+	// West-east traversibility between adjacent sectors
+	"traverseWE": [
```
selaux, line 6: Really stupid question: Shouldn't this be calculatable by assigning a short name to each sector and then executing a calculation like:

```
fromKind = GetSectorTraversabilityMovingFrom()
toKind = GetSectorTraversabilityMovingTo()
swich (fromKind, toKind) {
    ("T", "SR") => MOVEMENT_COST_TOWN_TO_SPARSE_ROAD,
     ...
}
```

This way we would only need one mapping table of pairs of traversability rating short names to movement costs, where we can extract the `MOVEMENT_COST_XXX` from.

assets/externalized/strategic-map-movement-costs.json
```diff
@@ -0,0 +1,115 @@
...
+	/* I */ [ "G", "G",  "CR", "D",  "S",  "X",  "S",  "X",  "A",  "A",  "P",  "P",  "X",  "X",  "S",  "S"  ],
+	/* J */ [ "G", "CR", "CR", "W",  "W",  "SR", "S",  "A",  "X",  "A",  "A",  "S",  "S",  "FR", "S",  "E"  ],
+	/* K */ [ "G", "CR", "W",  "X",  "W",  "PR", "AR", "AR", "AR", "PR", "PR", "PR", "SR", "SR", "D",  "E"  ],
+	/* L */ [ "C", "CR", "W",  "W",  "D",  "PR", "P",  "A",  "P",  "SR", "X",  "X",  "C",  "W",  "D",  "E"  ],
+	/* M */ [ "G", "CR", "WR", "F",  "DR", "DR", "P",  "P",  "P",  "TR", "G",  "G",  "S",  "W",  "E",  "E"  ],
+	/* N */ [ "G", "G",  "X",  "X",  "X",  "TR", "CR", "CR", "TR", "TR", "G",  "G",  "E",  "E",  "E",  "E"  ],
+	/* O */ [ "G", "G",  "X",  "X",  "G",  "G",  "G",  "T",  "T",  "G",  "G",  "G",  "G",  "E",  "E",  "E"  ],
+	/* P */ [ "G", "G",  "X",  "G",  "G",  "G",  "G",  "G",  "G",  "G",  "G",  "G",  "G",  "E",  "E",  "E"  ]
+	],
+	// It is supposed to mean how travelled a sector is. The higher the travel rating, the more people go near it.
```
selaux, line 67: IDK if I would even include this if it is not used.

assets/externalized/strategic-map-movement-costs.json
ja2-derek: We need the traversibility ratings (not just the cost) for at least traversibility-through. This is used for things like loading screen:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/a5c4da215c7721561e0516f546ca540032f35cac/src/game/Loading_Screen.cc#L69-L76

Also, `SectorInfo.ubTraversability` (array of traversibility rating) is part of the save game. If we really skip the whole rating thing, it will break the save game compatibility.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/a5c4da215c7721561e0516f546ca540032f35cac/src/game/Strategic/LoadSaveSectorInfo.cc#L32

assets/externalized/strategic-map-movement-costs.json
ja2-derek: I don't know either. 

Save games still expects the Travel Rating value.  We can either leave it here in JSON and never touch it again (and in case somehow we want to reference or use it in the future), or we can hard-code the values all to zero.

lynxlynxlynx: Let's keep it.

flaviojs:
You are using `Assert` to validate json data, but asserts can be removed at build time.
They should only be used in situations where the developer has full control (source code).
Json data can be modified by the user so the developer has no control over what it gets.

ja2-derek:
That's a good point. I didn't think thoroughly on this. Our `Assert`  seems only does  SLOGA, that's not  even the C assert.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/sgp/Debug.h#L10

Maybe I should just throw runtime_error with proper message?

flaviojs:
SLOGA aborts if asserts are enabled:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/b22697bf9c53239ddf6d1c35e9ed4b195c0db6d7/src/sgp/Logger.cc#L16-L21
Use SLOGE instead of SLOGA? Or only throw and leave the logging decision to the caller.

Other than that I don't see anything wrong.
I don't have an opinion regarding what gets externalized.

lynxlynxlynx:
Thanks!
(both errors were just due to temporary network timeouts)

PR #1022, opened 2020-04-12
ja2_json.rs: fixed replacement var error message suggestion

lynxlynxlynx:
fixes migration scenario when you still have data_dir.

edit: forgot to add [ci skip], so killed them manually.


flaviojs:
LGTM

PR #1024, opened 2020-04-13
Externalizing Town Names

ja2-derek:
Started out in [#1020](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1020#issuecomment-612481515_). This PR externalizes town names, so future mods (Unfinished Business maybe?) can override

Tried to override town names in a mod. <details><summary>Result:</summary></details>

## Summary

- Externalizing localization strings from `Utils/_*Text.cc` to `externalized/strings/strategic-map-town*.json`
- Replacing all occurences `pTownNames` and `g_towns_locative`
- New string functionalities use `string_theory`. Existing codebase expects `wchar_t*` and uses `swprintf`. For now just converting the `ST::string` to `const wchar*`
    - Adding a macro syntactic sugar for repeated code of `->to_wchar().c_str()`
    - Replacing `swprintf` is left to #937.

## Conflicts

Conflicts with #1020 and #937.


flaviojs:
```cpp
// Macro to convert string_theory string to const wchar_t*
#define WCHAR_BUF(st_string)	( st_string->to_wchar().c_str() )
```
This is a big no-no!
`to_wchar()` returns a temporary object that will be freed when `c_str()` returns.
That means that the pointer returned by `c_str()` will be freed memory.

Personally, I would wait for #937 to be merged and remake this PR on top of that.

ja2-derek:
Ah. I didn't know about that. Thank you. Let's wait for #937 to be merged then.

flaviojs:
`getTownName` and `getTownLocative` will access invalid memory when `townId` is too big, I would return `ST::null` on those cases.

flaviojs:
`getTownName`/`getTownLocative` return new objects (in the stack).
When you chain with `.c_str()` they are temporary objects that are freed when `c_str` returns and the `c_str` pointer will be freed memory.

You can either:
1) put the object in a temporary variable that must live as long as the `c_str` pointer is needed
2) or return a const reference in `getTownName`/`getTownLocative` (the `c_str` pointer is valid as long as the original `ST::string` exists)

ja2-derek:
Is this  https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1024/commits/bed6873bda8485c940261f9ef5d2f42637a09f24 better?

flaviojs:
Yup, avoiding `c_str` also works. :)
Since you have full control over the format string, it would be better to use [`ST::format`](https://github.com/zrax/string_theory/wiki/Format-String-Reference) instead of `st_format_printf` (converts the format string at runtime and calls `ST::format`).

I don't see anything wrong with the current code.

flaviojs:
Thanks.

PR #1025, opened 2020-04-15
Externalizing Army Group definitions

ja2-derek:
Externalizing ARMY_COMPOSITION and GARRISON_GROUP.

See #1011 and #665. Conflicts with my other externalization PRs.

EDIT: also PATROL_GROUP, as it is very related.

## Summary

- Externalizing ARMY_COMPOSITION and GARRISON_GROUP structs into json. 
- Replacing `gArmyComp` array with std::vectors

### Data Validation

1. Start a new game
1. Take a dump of the original ArmyComposition and GarrisonGroup with VC++  debugger
1. Take a dump again after the patch
1. Compare line-by-line with a diff tool


flaviojs:
I don't see `SAVED_ARMY_COMPOSITIONS` being enforced.
What happens when the json has more army compositions and you save+load a new game?

ja2-derek:
The save will not have all the Army Comps, if JSON defined too many. Game will very likely to crash, when a Garrison Group references a missing army comp.

Similar case if Army Comp is removed from JSON after a game is saved. The save can reference a non-existence Army Comp and crash.

I will add a check to enforce `SAVED_ARMY_COMPOSITIONS` and `SAVED_GARRISON_GROUPS` when JSON is loaded, and also in the JSON warn modders that previous saves will crash if data is removed.

flaviojs:
Will it really crash when they change? If they are saved then I would expect the saved ones to be used and new changes to only apply to new games.

To make sure the rest of the code is correct, I would add an `Assert` for the size just before they are saved.

If you want to use composition names then you must enforce the correct mapping to the internal enum values. Currently the user can change the value of the json mappings, but the C++ values remain the same.

Also, in general, bad user data should give useful feedback instead of crashing or being converted to unexpected data.

ja2-derek:
The possible crash is here:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/aa64f038c6c23f41571c8ba23a6d64b9d066b438/src/game/Strategic/Strategic_AI.cc#L2355-L2358

When loading, we only read up to `gArmyComp.size()` items. If the JSON has fewer comps than the save, it will not read all the data from save. Or unless we change the logic here to keep reading data until we see the first empty ARMY_COMPOSITION struct - with "empty" meaning all zero values.

The names are mapped at initialization time, so any missing mapping will error out at initialization. After a game is started, only the numeric IDs are saved.

Agreed on more defensive handling of user data, and better feedback messages.

flaviojs:
I would just load all `SAVED_ARMY_COMPOSITIONS` to `gArmyComp`.
Then, if you want, pop army compositions from the back if they are empty and never referenced.

Imagine I swap the id field of the "QUEEN_DEFENCE" and "ROADBLOCK" entries in the json.
The C++ code that uses QUEEN_DEFENSE will use the "ROADBLOCK" json entry and the ROADBLOCK C++ code will use the "QUEEN_DEFENCE" json entry, which is counter intuitive...
If you want to use names then prepopulate the mappings with the known C++ names and add new names after that.

ja2-derek:
Just found a crash that can occur when Patrol Group and Garrison Group sectors overlap. I did not realize there is an assumption that these 2 types do not overlap. 

I will go back to review the the points raise above, and maybe together with the Patrol Group externalization.

ja2-derek:
Expanding this PR to cover also Patrol Groups, as it is closely related to Garrison and Composition,  and externalized in a similar manner.

I don't have a good answer to renaming of default groups can cause confusion.  I still prefer to use name rather than ID, for better usability and readability. Rather I have added more data validation.

##  Differences from previous commit

- Rebased to HEAD of master
- Added data validation and description in JSON
    - Unexpected ID-Name pair will result in warning log (e.g. swapping QUEEN_DEFENCE and ROADBLOCK in json)
    - Arrays too large or too small will result in runtime error
    - Unmappable names will result in error
- Load game take whatever saved in the file. Warn if it is different from JSON


Save/Load game are touched. Most changes are in Strategic_AI.cc. I will be testing it further in-game .

flaviojs:
I took a closer look at the code before the PR:
1) the code uses the first `NUM_ARMY_COMPOSITIONS` indexes so they must always exist, even if the data is "empty" => you must enforce they exist to avoid crashes/corruption
2) the code assumes they are placed in the correct index and the badly named field `iReadability` is actually the id/index => you must enforce the correct placement to preserve savefile compatibility with vanilla

I also see you copying/moving vectors around.
You should do that only when transforming or passing ownership of the data.
Try to use `const std::vector<T>&` or `std::vector<T>&` when possible.

ja2-derek:
> 1. the code uses the first NUM_ARMY_COMPOSITIONS indexes so they must always exist, even if the data is "empty" => you must enforce they exist to avoid crashes/corruption

Yes, it is enforced in ArmyCompositionModel::validateData: 

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/b11867a75b9c7aef411483815c1186ac726fb298/src/externalized/army/ArmyCompositionModel.cc#L50-L55

A save will always write 60 Army Comp entries no matter what. Now we enforce to have at least `n` Army Comps (n = `NUM_ARMY_COMPOSITIONS`). A mod can have more Army Comps, but vanilla should be able to load the first `n` army compositions and disregard anything after. I have not tested compatibility - I don't think a save from a modded game will work well in vanilla even if it doesn't crash.

ja2-derek:
> 2. the code assumes they are placed in the correct index and the badly named field iReadability is actually the id/index => you must enforce the correct placement to preserve savefile compatibility with vanilla

Searched for `iReadability` in the curent master HEAD and found no reference to that field anywhere in code. So if the field holds a wrong value, it will not affect anything. But indeed it will be confusing, that the game may not behave as the modder intended.

I added a check to WARN against mismatch of array-index and `iReadability` (renamed to `compositionId`)
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/b11867a75b9c7aef411483815c1186ac726fb298/src/externalized/army/ArmyCompositionModel.cc#L60-L63

assets/externalized/army-compositions.json
```diff
@@ -0,0 +1,329 @@
...
+ * Limitations:
+ *  1. You cannot have more than 60 army compositions. This is limited by the
+ *     value of SAVED_ARMY_COMPOSITIONS.
+ *  2. DO NOT remove, rename or re-order the default Army Compositions, or you
+ *     risk crashing the game.
+ *  3. Adding or removing Army Compositions may cause problems with previous
+ *     saves.
+ * 
+ * The below default compositions are referenced in the code, and their ordering
+ * index not be altered:
```
lynxlynxlynx, line 28:
```suggestion
 * index should not be altered:
```

assets/externalized/army-garrison-groups.json
```diff
@@ -0,0 +1,66 @@
+/**
+ * Defining garrison groups (stationary enemy troops) that should present at game start.
```
lynxlynxlynx, line 2:
```suggestion
 * Defining garrison groups (stationary enemy troops) that should be present at game start.
```
Can't believe Canadians make so many typos.

assets/externalized/army-patrol-groups.json
```diff
@@ -0,0 +1,50 @@
+/**
+ * This defines the Army Patrol Groups at the start of game. There can only
```
lynxlynxlynx, line 2:
```suggestion
 * This defines the Army Patrol Groups at the start of the game. There can only
```

src/externalized/army/ArmyCompositionModel.cc
```diff
@@ -0,0 +1,88 @@
...
+{
+	return {
+		compositionId,
+		priority,
+		elitePercentage,
+		troopPercentage,
+		adminPercentage,
+		desiredPopulation,
+		startPopulation,
+		{ 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 }
```
lynxlynxlynx, line 21: These zeroes are just relics for save game compatibility? We should document it, since it looks odd in all these models.

lynxlynxlynx:
@flaviojs do you have any further comments?

src/externalized/army/ArmyCompositionModel.cc
flaviojs: I avoid list initialization like this because it's hard to search and hides simple mistakes like using the wrong order or missing a field (can result in uninitialized memory).
It's also a pain to change when the structure changes.

This test shows list initialization forms that are known to initialize all the memory in our CI compilers (I recommend zero-initialize forms):
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/ed9e1c24622747aecd6767b01c70a818406948de/src/sgp/SGP.cc#L601

I prefer a more search friendly pattern:
```cpp
ARMY_COMPOSITION tmp = ARMY_COMPOSITION{};// or ARMY_COMPOSITION tmp{};
tmp.iReadability = compositionId;
tmp.bPriority = priority;
tmp.bElitePercentage = elitePercentage;
tmp.bTroopPercentage = troopPercentage;
tmp.bAdminPercentage = adminPercentage;
tmp.bDesiredPopulation = desiredPopulation;
tmp.bStartPopulation = startPopulation;
return tmp;
```

flaviojs:
The army group loading is still an issue for me because positions can be shifted.
I would read everything first and only then limit the size:
```cpp
gArmyComp.assign(SAVED_ARMY_COMPOSITIONS, ARMY_COMPOSITION{});
FileRead(hFile, gArmyComp.data(), SAVED_ARMY_COMPOSITIONS * sizeof(ARMY_COMPOSITION));
// read other stuff...
// limit the size of gArmyComp (after gGarrisonGroup)
size_t numArmyCompositions = NUM_ARMY_COMPOSITIONS;
for (const GARRISON_GROUP& gg : gGarrisonGroup)
{
  numArmyCompositions = std::max<size_t>(numArmyCompositions, gg.ubComposition + 1);
}
gArmyComp.resize(numArmyCompositions);
// warn/error when an id does not match the position...
```

src/externalized/army/ArmyCompositionModel.cc
ja2-derek: I totally understand your point ([#1017](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1017/files) 🤦‍♂️).

But can still use `const` class member with zero-initialize constructor? The class data should be read-only.

flaviojs: `ARMY_COMPOSITION ArmyCompositionModel::toArmyComposition() const`

This const means you are will not change the contents of ArmyCompositionModel.
You are only reading the members to fill a different structure, so there is no problem.

ja2-derek:
> 
> 
> The army group loading is still an issue for me because positions can be shifted.
> I would read everything first and only then limit the size:
> 
> ```c++
> gArmyComp.assign(SAVED_ARMY_COMPOSITIONS, ARMY_COMPOSITION{});
> FileRead(hFile, gArmyComp.data(), SAVED_ARMY_COMPOSITIONS * sizeof(ARMY_COMPOSITION));
> // read other stuff...
> // limit the size of gArmyComp (after gGarrisonGroup)
> size_t numArmyCompositions = NUM_ARMY_COMPOSITIONS;
> for (const GARRISON_GROUP& gg : gGarrisonGroup)
> {
>   numArmyCompositions = std::max<size_t>(numArmyCompositions, gg.ubComposition + 1);
> }
> gArmyComp.resize(numArmyCompositions);
> // warn/error when an id does not match the position...
> ```

I see that you are trying to validate the consistency between GarrisonGroups and ArmyCompositions. 

We enforce consistency in JSON data at game launch. The data from ContentManager is only used once at Campaign Init. We guarantee data consistency at Campaign Init, so the data within a save must be consistent. Save/load game do not reference JSON data. 

I will add data validations after loading to ensure the save has the correct composition IDs, and defines all the hard-coded references.

flaviojs:
There are programs to manipulate save files. Their data should not be considered safe.

lynxlynxlynx:
At some point the user becomes responsible though and we shouldn't bother. I'm not drawing any particular lines though.

ja2-derek:
Re-based commits to resolve conflicts.  Also addressed PR comments:

- Uses zero-initialize form for structs
- Added validation on Army Comp data from saved  game
- Ensuring compositions referred by Garrison Groups exist

src/externalized/army/ArmyCompositionModel.cc
```diff
@@ -86,3 +86,20 @@ void ArmyCompositionModel::validateData(std::vector<const ArmyCompositionModel*>
...
+	{
+		SLOGW(ST::format("There are too few loaded Army Compositions. Expected: {}; Actual: {}", NUM_ARMY_COMPOSITIONS, armyCompositions.size()));
+	}
+	for (size_t i = 0; i < armyCompositions.size(); i++)
+	{
+		// Warn if default compositions are altered. It might work, but no guarantee.
+		auto comp = armyCompositions[i];
+		if (comp.iReadability != i)
+		{
+			SLOGW(ST::format("Army Composition has incorrect ID. The save might be co	rrupted. Expected: {}; Actual: {}", i, comp.iReadability));
```
lynxlynxlynx, line 102:
```suggestion
			SLOGW(ST::format("Army Composition has incorrect ID. The save might be corrupted. Expected: {}; Actual: {}", i, comp.iReadability));
```

src/game/Strategic/Strategic_AI.cc
```diff
@@ -2282,41 +2282,43 @@ void LoadStrategicAI(HWFILE const hFile)
...
-	auto buff3 = new GARRISON_GROUP[SAVED_GARRISON_GROUPS]{};
-	FileRead(hFile, buff3, SAVED_GARRISON_GROUPS * sizeof(GARRISON_GROUP));
-	gGarrisonGroup = std::vector<GARRISON_GROUP>(buff3, buff3 + iGarrisonArraySize);
-	delete[] buff3;
+	auto buffGG = new GARRISON_GROUP[SAVED_GARRISON_GROUPS]{};
+	FileRead(hFile, buffGG, SAVED_GARRISON_GROUPS * sizeof(GARRISON_GROUP));
+	gGarrisonGroup = std::vector<GARRISON_GROUP>(buffGG, buffGG + iGarrisonArraySize);
+	delete[] buffGG;
+
+	// resize gArmyComp accoording to what are actually referenced in GarrisonGroups and PatrolGrouops
```
lynxlynxlynx, line 2310: It doesn't appear to care about patrol groups and something is wrong with the sentence. Maybe:
"resize gArmyComp according to count of actual references in GarrisonGroups and PatrolGrouops"?

ja2-derek: Updated comment to clarify. Sorry foor the keyeboard chattering...

lynxlynxlynx:
Thanks!

PR #1026, opened 2020-04-16
Use semantic versioning

flaviojs:
Change the current version to "0.17.0-dev".

When possible, use the commit hash as the build metadata.

Only GameVersion.cc receives GAME_VERSION to avoid big rebuilds when the commit changes.

----

This was split from #1008


flaviojs:
Adopting this means the version must be set to:
 * `0.(next minor).0-dev` in the master branch
 * `0.(minor).(patch)` in release tags
 * `0.(minor).(next patch)-dev` in the 0.(minor).x branches

When releasing `1.0.0` we must decide what is a breaking change.
The master branch version will be `(major).(next minor).0-dev` before a breaking change and `(next major).0.0-dev` with a breaking change.

Having a tag means the version is a prerelease, so `0.17.0-dev` comes before `0.17.0`.
We are at major version 0 so there are no promises, we can break stuff as much as we want.

src/game/GameVersion.cc
```diff
@@ -1,10 +1,12 @@
 #include "GameVersion.h"
 
 #ifndef GAME_VERSION
-	#define GAME_VERSION "XXXXX"
+	#define GAME_VERSION XXXXX
```
lynxlynxlynx, line 4: Can this still be reached?

flaviojs: Same as before, it not reached if our build system is used.
Should I remove?

lynxlynxlynx: I'd remove it, yes, since it's quite sane to expect people to use our build system. :)

flaviojs: Done

PR #1027, opened 2020-04-16
Fix negative array index read in EnableQDSButtons

flaviojs:
Detected by coverity.

### DIFF
--- a/src/game/Strategic/Quest_Debug_System.cc
+++ b/src/game/Strategic/Quest_Debug_System.cc
@@ -2734,7 +2734,7 @@ static void EnableQDSButtons(void)
 
 	if( gfUseLocalNPCs )
 	{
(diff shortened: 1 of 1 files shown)


src/game/Strategic/Quest_Debug_System.cc
```diff
@@ -2734,7 +2734,7 @@ static void EnableQDSButtons(void)
 
 	if( gfUseLocalNPCs )
 	{
-		bool const enable = IsMercInTheSector(gubCurrentNpcInSector[gNpcListBox.sCurSelectedItem]) != -1;
```
lynxlynxlynx, line 2737: You could also just remove the extraneous extra scope at the start of the function and use the result here for brevity.

flaviojs: I don't want to. =~
I avoid restructuring in coverity PRs to make sure the original problem is easy to see.

PR #1029, opened 2020-04-16
Externalizing random NPC placements

ja2-derek:
Quick PR to get NPC placements to work with map mod, but also a question to ask. I will further test and polish, if we agree on the approach.

See also #665 and #1011.

## Summary

- Some NPCs (e.g. DEVIN, CARMEN, HAMOUS (and his truck)) are randomly placed in several possible sectors
- Externalizing the random placements to JSON

## Question

The Strategic Event handler is very specific to JA2 storyline and campaign. There is not a lot of flexibility even by externalizing the sector list. Not saying we have to follow, but 1.13 allows overriding by Lua script.

Should we also introduce a scripting engine here?

### DIFF
--- a/assets/externalized/strategic-map-npc-placements.json
+++ b/assets/externalized/strategic-map-npc-placements.json
@@ -5,7 +5,7 @@
  *
  * Fields:
  *  - profileId:    The numeric ID of the NPC merc profile.
- *  - sectors:    Sectors in which the NPC can appear. If empty, the NPC will never appear. All of these sectors should to have NPC merc placed, or else the NPC may get "assigned" to sector but not appearing.
+ *  - sectors:    Sectors in which the NPC can appear. If empty, the NPC will never appear. All of these sectors should have the NPC merc preplaced, or else the NPC may get "assigned" to a sector but not appear.
  */
 [
 	{
--- a/assets/externalized/strategic-map-npc-placements.json
+++ b/assets/externalized/strategic-map-npc-placements.json
@@ -1,0 +1,21 @@
+/**
+ * Lists the sectors where randomly placed NPCs can appear. Each NPC here has his own corresponding code sections.
+ *
+ * Fields:
+ *  - profileId:    The numeric ID of the NPC merc profile.
+ *  - sectors:    Sectors in which the NPC can appear. All of these sectors should to have NPC merc+-+ placed, or else the NPC may get "assigned" to sector but not appearing.
+ */
+[
+	{
+		"profileId": 63,    // HAMOUS, together with his Ice-cream truck (PROF_ICECREAM, 162)
+		"sectors": [ "G6", "F12", "D7", "D3", "D9" ]
+	},
+	{
+		"profileId": 61,    // DEVIN
+		"sectors": [ "G9", "D13", "C5", "H2", "C6" ]
+	},
+	{
+		"profileId": 78,    // CARMEN
+		"sectors": [ "C5", "C13", "G9" ]
+	}
+]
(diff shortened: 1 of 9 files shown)


src/game/Strategic/Strategic_Event_Handler.cc
```diff
@@ -772,29 +773,16 @@ void HandleEarlyMorningEvents( void )
 			{
 				// ok, Devin's sector not loaded, so time to move!
 				// might be same sector as before, if so, oh well!
-				switch( Random( 5 ) )
+				uint8_t randomPlacement = Random(5);
```
lynxlynxlynx, line 776: Make it `placement->sectorIds.size()` instead of 5 and there's both more freedom and the exception can be removed. Same with the others.

lynxlynxlynx:
I'd try to keep focused and first get WF compatibility out of the way. Just avoiding feature creep.

derek9gag:
Sure.

I was thinking full-blown scripting is better in the long term, and making changes now may complicate the effort later. But we can still go back in VCS history.

ja2-derek:
Sure.

I was thinking full-blown scripting is better in the long term (I am already thinking of UB support...), and making changes now may complicate the effort later. But we can still go back in VCS history.

flaviojs:
I think that right now there is too much spaghetti code for full-blown scripting.
And spaghetti code is very hard to get rid of, you're welcome to try if you want a challenge.

As for this PR:
1) deserializing assumes the sector names are valid
2) the DEVIN/HAMOUS/CARMEN code assumes there is at least 1 placement
3) it should be documented that adding other npcs does nothing (it needs scripting to be generalized)

assets/externalized/strategic-map-npc-placements.json
```diff
@@ -1,9 +1,11 @@
 /**
- * Lists the sectors where randomly placed NPCs can appear. Each NPC here has his own corresponding code sections.
+ * Lists the sectors where randomly placed NPCs can appear. 
+ * 
+ * Adding other NPCs here will not have any effect, as each NPC here requires his own corresponding sections in codebase.
  *
  * Fields:
  *  - profileId:    The numeric ID of the NPC merc profile.
- *  - sectors:    Sectors in which the NPC can appear. All of these sectors should to have NPC merc+-+ placed, or else the NPC may get "assigned" to sector but not appearing.
+ *  - sectors:    Sectors in which the NPC can appear. Must not be empty. All of these sectors should to have NPC merc placed, or else the NPC may get "assigned" to sector but not appearing.
```
lynxlynxlynx, line 8: Is there some deeper engine need for this? Because now you can't for example disable Hamous ever appearing just by editing the sectors out.

ja2-derek: No, I don't think so. 

Indeed we can support disabling NPCs by adding empty check and allowing empty sector lists instead.

assets/externalized/strategic-map-npc-placements.json
```diff
@@ -1,9 +1,11 @@
 /**
- * Lists the sectors where randomly placed NPCs can appear. Each NPC here has his own corresponding code sections.
+ * Lists the sectors where randomly placed NPCs can appear. 
+ * 
+ * Adding other NPCs here will not have any effect, as each NPC here requires his own corresponding sections in codebase.
  *
  * Fields:
  *  - profileId:    The numeric ID of the NPC merc profile.
- *  - sectors:    Sectors in which the NPC can appear. All of these sectors should to have NPC merc+-+ placed, or else the NPC may get "assigned" to sector but not appearing.
+ *  - sectors:    Sectors in which the NPC can appear. If empty, the NPC will never appear. All of these sectors should to have NPC merc placed, or else the NPC may get "assigned" to sector but not appearing.
```
lynxlynxlynx, line 8:
```suggestion
 *  - sectors:    Sectors in which the NPC can appear. If empty, the NPC will never appear. All of these sectors should have the NPC merc preplaced, or else the NPC may get "assigned" to a sector but not appear.
```

lynxlynxlynx: final comment, I swear. :)

lynxlynxlynx:
Thanks!

PR #1030, opened 2020-04-17
Use DataWriter and DataReader

flaviojs:
I was going to fix a coverity issue but INJ_STR was in the way.
Coding in macros (like INJ_STR) is something to avoid.
So instead I'm changing the save/load stuff to use DataWriter/DataReader.


flaviojs:
This touches save file data so it still needs a play-through to Drassen.

flaviojs:
Done.

PR #1031, opened 2020-04-18
Skip Invalid Items on Maps

ja2-derek:
This is intended to improve map mod support (https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1011#issuecomment-609579529). 

Some items in Wildfire re-use item IDs, that maps to internal items in Vanilla. (e.g. LAW rocket is item#160 which is "Creature Gas" in Vanilla). To better support WF maps, we skip those items, as such items just do not exists in Vanilla.

An assumption here is: if an item is marked `ITEM_NOT_EDITOR`, it is not supposed to placed in the map file as a world item. 

# Summary

- Adding the `ITEM_NOT_EDITOR` flag to the "creature gas" internal items. Before this PR they do show up in the Map Editor, but they should not.
- Add a check, so when loading from map file, `ITEM_NOT_EDITOR` items are not placed. A warning is logged, because it means the map is not used as it intended.

---

The WF map version of Drassen SAM site has a "LAW Rocket". It will be skipped:
image


flaviojs:
This is a smart solution, but it is basically just avoiding a problem.

A better solution would be to have a json that maps ja2 wildfire item ids to ja2 item ids in the maps you specify.

flaviojs:
There is nothing with this code, merging.
Thanks.

PR #1032, opened 2020-04-18
Fixing Key Name display in Map Editor

ja2-derek:
See the below screenshots. One on top is 0.16.1, the other is HEAD.

image

This is because LOCK.ubEditorName has type of UINT8[]. Previously it could be interpreted as if it is `char[]`, but `ST::format` prints it as `true`.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/b22697bf9c53239ddf6d1c35e9ed4b195c0db6d7/src/game/Editor/EditorItems.cc#L322


selaux:
Thanks.

PR #1041, opened 2020-04-20
Fix divide by zero in GetAvailableWorkForceForMineForPlayer

flaviojs:
Detected by coverity.

~A zero should never happen but `GetAvailableWorkForceForMineForEnemy` has this check, so we match it.~


flaviojs:
Wait... the same a divide by zero coverity issue exists for `GetAvailableWorkForceForMineForEnemy` so this is not the right solution.

flaviojs:
I guess coverity considers that the data can change between calls.

If coverity still complains after this, then the issue is with Assert.

flaviojs:
Merging.
I want to know what coverity says.

PR #1042, opened 2020-04-20
Allowing mods to replace World Items on maps

ja2-derek:
This should be a better approach than #1031. See also #1011.

# Summary

- Allow mods to have a mapping, to replace or ignore pre-placed items on tactical maps. 
- The replacement mapping should be empty by default. It is intended to be a quick compatibility fix.
- Replacements are global. Replacement logic is kept simple for now. We can expand later if there are specific needs. 

## Example

The replacement mapping for Wildfire maps would be something like:
```json
[
	{ "from": 66, "to": "M24" },             // VSS        => M24
	{ "from": 67, "to": "DRAGUNOV" },        // V-94       => DRAGUNOV
	{ "from": 68, "to": "G11" },             // F2000      => G11

	{ "from": 265, "to": "SILENCER" },       // Silencer .45  => SILENCER
	{ "from": 269, "to": "SILENCER" },       // Silencer P90  => SILENCER

	{ "from": 160, "to": "ROCKET_LAUNCHER" },        // LAW Rocket => LAW
	{ "from": "DISCARDED_LAW", "to": "ROCKET_LAUNCHER" } // Empty LAW  => LAW
]
```


flaviojs:
Right now this is only useful for full map mods, which is fine.

There should be an error when the item name is not found.

The mappings should be from item id to item id since nothing else can be changed and nothing guarantees that the item model won't be freed while you store a copy.

When mapping to item id NONE the item should be skipped in `LoadWorldItemsFromMap`.

ja2-derek:
> There should be an error when the item name is not found.

It uses `ItemSystem::getItemByName` to look up by name, which throws a `std::runtime_error` if item is not found.

Agreed on the point about using item index.

ja2-derek:
Resolved conflicts and pushed. (74efe50..d817b8084)

## What's new in the new commit

- JSON does not allow NULL. We use #0 ("NADA") to skip item
- Replacement mapping is now a mapping from item index to item index
- Reverted some changes in World_Items.cc from previous PR #1031. This PR made it redundant

flaviojs:
I would keep the invalid item check from #1031 unless there is a way for them to appear in the map.

I just had a thought... does this apply when you load a map? If it does, then it will be changing an already changed map and can produce unexpected consequences. Imagine I set X->Y and Y->X, the items will flip on every save-load.

ja2-derek:
> I would keep the invalid item check from #1031 unless there is a way for them to appear in the map.

Pushed again. Now keeping the invalid check, in case the mod has not mapped every invalid item.

> I just had a thought... does this apply when you load a saved game? If it does, then it will be changing an already changed map and can produce unexpected consequences. Imagine I set X->Y and Y->X, the items will flip on every save-load.
I have just tried this in game. Tried moving between loaded sectors, loading a game, and moving sectors after loading a game. In all cases, the function `LoadWorldItemsFromMap` only see items from the original map .dat file.

src/externalized/ItemSystem.h
```diff
@@ -8,4 +8,5 @@ class ItemSystem
 {
 public:
 	virtual const ItemModel* getItemByName(const ST::string &internalName) const = 0;
+	virtual const ItemModel* getItem(uint16_t itemIndex) const = 0;
```
lynxlynxlynx, line 11: This looks redundant. Did you mean to add a non-static GetItemFromValue instead?

lynxlynxlynx:
I've restarted the mac build, but it timeouts at the same place (building rust deps). Hopefully it will resolve itself later.

src/externalized/ItemSystem.h
derek9gag: This is to expose an existing method in DefaultContentManager via ItemSystem interface. 

It probably would touch many more files if we rename it

ja2-derek: This meant to expose an existing method in `DefaultContentManager` via the `ItemSystem` interface.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/2ae26c4e46a1c268bd93af7e8d5573090b2ae797/src/externalized/DefaultContentManager.cc#L1005

It would probably touch many more files if we are to rename it.

lynxlynxlynx: Ah, missed the one use later down, never mind.

lynxlynxlynx:
Thanks!

PR #1046, opened 2020-04-24
Fix divide by zero in GetAvailableWorkForceForMineForEnemy

flaviojs:
Detected by coverity.


lynxlynxlynx:
Am I blind, how can that be a div-by-0?

ja2-derek:
I think it believes the results from calling twice `GetTownSectorSize( bTownId )` can be different.

flaviojs:
GetTownSectorSize is a function that uses external data, which can change at any time in C/C++.

If coverity was a tool that analyzed the generated executable it could probably conclude that a divide-by-zero will not happen here, but it deals with the incomplete picture (compile units) so it can't conclude that.

To handle this I could dismissing it manually in coverity (and risk it showing up again) or pull the data to a local variable before use.

PR #1049, opened 2020-04-24
Replace PaletteRepID with ST::string

flaviojs:
This fixes the buffer overflow false-positive in SetMercSkinAndHairColors.

This PR touches save files so it still needs a play to Drassen.


flaviojs:
My `B13.dat` has uninitialized `SOLDIERCREATE_STRUCT` palette ids, which are invalid utf8.
What to do with bad data in original files...

lynxlynxlynx:
Uninitialized as in runtime random or are they constant?

flaviojs:
It is data that was uninitialized when the map file was created.
For us it is invalid data being read from the map file (constant but not predictable).

lynxlynxlynx:
Can't we detect it though and then use an empty string? Or is just part of it invalid and we need to use the chars that come before that?

flaviojs:
Uninitialized data is unpredictable, it can even be valid utf8 data.

Right now I'm considering two possibilities:
1) fallback to a specific string if any of the unicode is invalid (the string depends on the context)
2) use the unicode replacement character in the invalid parts

lynxlynxlynx:
What matters is if it holds game relevant data, not so much whether it's proper unicode. Does the original rely on these funky values in the airport, do they get ignored, something else?

flaviojs:
In this particular case the data only matters if the `fVisible` field is is set.

But the solution should be appropriate for other invalid string data being read, since I want to enforce utf8 everywhere.

lynxlynxlynx:
I suggest you write some temporary code to see if any other sector has this problem, since the scale may affect our approach.

flaviojs:
Will do that.

flaviojs:
I had a walk around the surface and found there maps: B13,C13,D13, B6,C6,C5,D5,D6, E9, F8, I13, J9, J12
Not a complete check, but enough to show that it's common.

lynxlynxlynx:
Sounds like it's worth trying with #2 then.

flaviojs:
Done.

src/game/Tactical/Overhead_Types.h
```diff
@@ -290,11 +290,11 @@ enum BoxingStates
...
 // PALETTE SUBSITUTION TYPES
 
-typedef CHAR8 PaletteRepID[ 30 ];
+#define PaletteRepID_LENGTH 30
 
 
 // MACROS
 // This will set an animation ID
-#define SET_PALETTEREP_ID( a, b )			( strcpy( a, b ) )
+#define SET_PALETTEREP_ID( a, b )			( (a) = (b) )
```
lynxlynxlynx, line 298: Something to remove at one point — the macro adds nothing.

flaviojs: Yup.

lynxlynxlynx:
So, any visible changes in game? The build problem is the same upstream one as before.

flaviojs:
During the game everything appeared to be the same as before.

lynxlynxlynx:
Good job then. :) If you're done, you have my ack.

PR #1053, opened 2020-04-26
Fix uninitialized member in GUI_BUTTON::GUI_BUTTON

flaviojs:
Detected by coverity.


flaviojs:
Note that this is actually an indirect false positive.
SpecifyButtonSoundScheme sets the value of ubSoundSchemeID indirectly.

PR #1055, opened 2020-04-29
Cleanup filesystem stuff

flaviojs:


flaviojs:
I saw several uses of `.c_str()` on temporaries, which I believe results in dangling pointers (unless C++ extends the temporary lifetime).
I'll be cleaning up or converting string stuff next.

lynxlynxlynx:
I don't know if you're done, but looking at this diagonally, it seems fine.

PR #1058, opened 2020-05-05
Fix Mine Text display

ja2-derek:
Mine income rates going beyond 100% in strategic map. The numerator is a daily rate, but the denominator is a periodic rate. 

Both should have been daily rate. Ref: https://github.com/ja2-stracciatella/ja2-stracciatella/commit/a3df1905455a29f21dbbca2647a0688a40c23e3d 

@flaviojs


lynxlynxlynx:
LGTM, thanks!

PR #1060, opened 2020-05-06
Remove meaningless Assertions on unsigned int

ja2-derek:
New defects introduced in #1021.  Detected by Coverity.

This PR removes the meaningless assertions that compares unsigned int to 0.


lynxlynxlynx:
Thanks!

PR #1059, opened 2020-05-06
Externalizing SAM site locations

ja2-derek:
This PR externalizes SAM site locations. The goal is to allow map mods to override SAM site locations. Though the SAM site list is now a `std::vector`, adding or removing SAM sites are not allowed.

See also #1011 and #665.

### Variables converted

- `pSamList[]`:  Now dynamically defined by JSON
- `pSamGridNoAList[]`:  This is the "anchor" tile of the SAM computer terminal. This is always the lower tile (the one with higher gridNo)
- `pSamGridNoBList[]`:  The other tile of the SAM computer
- `gbSAMGraphicList[]`:  It is now derived from SAM gridNos

### Variables not changed
- `fSamSiteFound[]`  Not changing for save compatibility

### Restrictions

Restrictions enforced in `SamSiteModel.cc`.

- Must define exactly 4 SAM sites
    - Breaks Saves, if more than 4 SAM sites
    - Breaks hard-coded references from Skyrider and Meanwhile dialogues, if fewer than 4 SAMs
- SAM gridNos must be 2 adjacent tiles

### What's not covered

- Airspace control table
- Hard-coded references to SAM sectors, such as `HandleSkyRiderMonologueAboutDrassenSAMSite()`, `ChooseSuitableGarrisonToProvideReinforcements()`
- Land type strings. SAM sectors (D2, I8, etc) are hard-coded-mapped to respective land type strings. It is another big chunk of work to externalize sector land type mappings. The land type display will be wrong if modder moved the SAMs sites to other sectors.
- A section of code in Assignments.cc commented out, saying: `/* No point in allowing SAM site repair any more.  Jan/13/99.  ARM` - _**Should all these be deleted?**_

### Testing

Manually tested in game.

1. Recruited Skyrider and he told us about Drassen SAM site
1. Verified other SAM sectors are not named as SAM sites
1. Took over Drassen SAM site.
1. Skyrider told us about other SAM sites. Those sectors are now named SAM sites.
1. Damaged SAM computer. SAM computer looks damaged. Confirm `bSAMCondition` reduced via logs.
1. Queen's Army re-took Drassen SAM site
1. Re-took Drassen SAM site. Confirmed the SAM computer in good condition.

PR #1061, opened 2020-05-07
Externalizing parameters to NPC Actions and quests Facts

ja2-derek:
This PR covers 2 externalizations, for the evaluation logic of `NPC_ACTION_*` and `FACT_*`. The two are implemented in the same way. Logic is still in code, so the externalization here is limited. Modders can only modify the numeric values, not until we somehow externalize/script the code logic.

See also #1011 and #665. ContentManager changes likely conflicts with #1059.

## Summary

- Externalized "parameters", keyed by NPC_ACTION and FACT enum IDs
- Only numeric values are parameterized - only gridNo and money amount
- All parameters are mandatory, as all are hard-coded references
- The model classes provide fallback mechanism. It allows defaulting to Vanilla values, if the parameters is not defined
    - This is to avoid game crashing on bad or missing data file, for example from version upgrade
    - The game is allowed to proceed with fallback value, but a WARN log will be printed to remind the player to fix it.

## Testing

Each NPC action and Fact will have to be tested in game individually.


lynxlynxlynx:
Thanks!

PR #1064, opened 2020-05-10
Fixing an Out-of-Bounds check in PathAI

ja2-derek:
## Motivation

Both editor and game crashes when trying to load the Wildfire version of Estoni I6.dat. The exception was array out-of-bounds when the algorithm attempts to access gridNo beyond `WORLD_MAX`(25600)

## Analysis 

The error is when the game tries to add a soldier to the map. The soldier is placed very close to bottom-left corner. In this case the soldier is at 24719, which is 5 tiles away from edge.

The algorithm then tries to find a "sweet spot" using FindBestPath without a destination. I believe the algorithm is doing a search a few tiles around the soldier placement tile.

There is a out-of-bounds check in code, but it never took effect. The check happens after the gridNo is used. Though I haven't tried, I suspect it to be reproducible by placing a soldier at the far bottom-left. 

This is an independent issue from #1063.

### DIFF
--- a/src/game/Tactical/PathAI.cc
+++ b/src/game/Tactical/PathAI.cc
@@ -980,6 +980,11 @@ INT32 FindBestPath(SOLDIERTYPE* s, INT16 sDestination, INT8 ubLevel, INT16 usMov
 
 			newLoc = curLoc + DirIncrementer[ubCnt];
 
+			if ( newLoc < 0 || newLoc >= GRIDSIZE )
+			{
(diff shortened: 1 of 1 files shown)


ja2-derek:
Not really relevant, but apparently ja2-1.13 has a similar fix

https://github.com/aimnas/ja2-1.13-source-mirror/blob/master/Tactical/PATHAI.cpp#L3104-L3109

PR #1066, opened 2020-05-10
Fixing an error with Town data

ja2-derek:
My bad. The town data I committed in #1020 has an error. Now the town sectors are exactly the same as 0.16.1.

smap_0161
smap_master


lynxlynxlynx:
What about the Omerta label? Looks like it needs a y=10.

Issue #707, opened 2018-02-27
Game.json in user directory (possibly ~/.ja2)


moll:
Hey!

Thanks for the great JA2 icecream! Looking forward to getting back in the game. I'd like to tweak some settings in game.json, but when installed through a Linux distro package manager (Arch Linux through https://aur.archlinux.org/packages/ja2-stracciatella-git/, for example) game.json is located in /usr/share and isn't user-editable and is shared by everyone on the machine. Have you considered supporting game.json from ~/.ja2? That'd fit nicely with other user settings.

Thanks!

lynxlynxlynx:
We have, yes, as the plan is to have the launcher mention the file #607.

lynxlynxlynx:
Quoting myself from a closed PR:
> I don't like the (default) paths personally, since EXTRA_DATA_DIR will be undefined for most, so we'll default to the binary location. Which is not handy for installed or built-only versions: first has permission implications, while both suffer from losing the settings on a reinstall/rebuild.
> 
> So, I think we should first check if the file is in the home settings dir, alongside the saves and ja2.json.

#833 introduces a nice get_game_json_path where the change can be made.

flaviojs:
It's fine to load a game.json from a yet-to-specified location in the home dir but that means there will be 2 game.json (EXTRA_DATA_DIR points to where the original is installed).
But first we need to define what happens to the original.
Is the original game.json ignored? Is it loaded as default values? Maybe both are possible and it depends on something inside the json file?

What about the other json files? They live next to game.json so maybe they should be treated the same?

What about mods? The mods directory is installed next to the externalized directory (where game.json lives) so maybe they should be treated the same?

The acceptable home dir path for game.json depends on the answers to the questions above.

PS- as a final note, I think the name EXTRA_DATA_DIR is confusing given the current usage (tell the code where the externalized/mods asset dirs are installed to), maybe it should be renamed to ASSETS_DIR

lynxlynxlynx:
It's stored in stracciatella_home, nothing yet-to-be-specified. We could treat them as overrides, but all the values have defaults anyway, so I don't see a benefit in loading more than one file.

Externalized/ is for power users, so I wouldn't worry about that right now.

Mods are static, so any changes will be rare there.

flaviojs:
If all values have defaults when they are not in the json, then it's fine to just shadow the original.

[...]

PR #1067, opened 2020-05-10
Replace LibraryDB with Vfs

flaviojs:
~v1: Rework LibraryDB~
v2: Replace LibraryDB with Vfs

Supports filesystem directories and slf archives.
Can be expanded to support other archive types.

Vfs and VfsFile are independent but not thread-safe.
Make sure each object is only used by one thread at a time.


flaviojs:
I'm considering expanding LibraryDB to also handle data dirs. But that means it's a virtual file system.
If this is a virtual filesystem, maybe it should also handle writable paths like the current directory, TMPDIR and DEVINFO_DIR.

I can't decide if I should go with this... give me your thoughts.

ja2-derek:
I wish that .SLF can be used in mod directories.

It seems opening a large number of small files (at least on Windows) is pretty slow. Having a mod of thousands of resources (mostly tileset and tilecaches) makes the game slow. File caching kicks in eventually, but the first load can be quite slow. I am talking about ~10s freeze when loading a map. This is very slow, for a 20 years old game running on a modern PC.

lynxlynxlynx:
I don't know how things are currently organized or what the impact would be, so ¯\_(ツ)_/¯.

flaviojs:
Allowing slfs in mod dirs should be easy. ;)

As for the current structure, #707 contains my old findings about how things are organized (excluding TMPDIR and DEVINFO_DIR).

Readable dirs are easy, the logic is similar to the current handling of slfs (filenames are case insensitive and shadowed).

Writable dirs are tricky, shadowing might not be appropriate.
What should happen if I delete a file in a writable dir and it was shadowing another file? I can't decide.

I'll try implementing readable data dirs and skip writable dirs for now...

flaviojs:
I ended up reimplementing LibraryDB as a virtual file system Vfs.

The ordering is based on my old proposal (https://github.com/ja2-stracciatella/ja2-stracciatella/issues/707#issuecomment-509862638) with a few changes:
 * the current dir is not handled by Vfs, it gave me problems that I didn't understand
 * the free editor.slf at the end (fallback)

For now it is read-only.

flaviojs:
Hmm, going for writable dirs seems like a bad idea for now.

As a final note, this PR changed the order of the "current dir".
Before it came after mods and externalized.
Now it comes first.

lynxlynxlynx:
Is "current dir" still our tmp dir, not whatever CWD ja2 was ran from?

flaviojs:
Yup, the current dir is set to out tmp dir in `FileMan::switchTmpFolder`.

flaviojs:
Hmm, I'm tempted to just go ahead and merge so I can continue my other PR...

lynxlynxlynx:
Yeah, let's go ahead. Things are finally congealing for a nicely cut 0.17 release.

selaux:
Yeah, sorry currently I am not able to review stuff, quickly...

Issue #1072, opened 2020-05-16
Clean up ContentManager


ja2-derek:
Follow up of https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1059#discussion_r421571872

These are memory leaks in ContentManager, but are insignificant, as the init is only done once.

- [ ] Organize the code in loadStrategicLayerData and loadGameData
- [ ] Clean up JSON document objects - `delete` or use smart pointers
- [ ] Delete object pointers in vectors, in `~DefaultContentManager()`

PR #1074, opened 2020-05-18
Externalizing Underground Sector Info

ja2-derek:
Visiting an underground sector that has a map .DAT but not corresponding `UNDERGROUND_SECTORINFO`  crashes the game.

See also #1011 and #665.

### Summary

- Extracted data from Campaign_Init..cc:BuildUndergroundSectorInfoList to JSON
- Generalized the per-DIFLEVEL numbers of enemies as a "base num + random" formula
- Only changed Campaign Init - in-memory linked-list and Save/Load are not touched
- Validation to protect from crash due to missing data

### DIFF
--- a/assets/externalized/strategic-map-underground-sectors.json
+++ b/assets/externalized/strategic-map-underground-sectors.json
@@ -1,0 +1,274 @@
+/*
+ * Defines the underground sectors that can be occupied by enemies, creatures,
+ * etc. It also contains the network of cave connections critical for strategic
+ * creature spreading, as we can't know how the levels connect without loading
+ * the maps. Any changes to the maps, require changes accordingly.
+ *
+ * Number of enemies and variances are specified as array of integer, in the
+ * form of [ n_EASY, n_MEDIUM, n_HARD]. defaults to [0,0,0] if not given.
+ *
+ * Variances can be specified to add per-game randomness. It is specified as
+ * arrays of integers: [ var_EASY,  var_MEDIUM, var_HARD]. also defaults to
+ * [0,0,0] if not given.
+ *
+ * The formula to calculate the number of enemies is: 
+ *    ( n_DIFLEVEL + Random(var_DIFLEVEL) )
+ * where Random is a function returning a random number between 0 and 
+ * (var_DIFLEVEL - 1) inclusively.
+ *
+ *
+ * Field definitions:
+ *  - sector:    Sector X,Y, such as A1 and P16
+ *  - sectorLevel:    Sector Z, an integer between 1 and 3
+ *  - adjacentSectors:    Directions to the adjacent sectors, as an array of directions, e.g. ["N","E","S","W"]
+ *  - numTroops:    The numbers of enemy troops in sector, as an array of 3 integers, e.g. [1,2,3] meaning: 1 enemy on DIF_LEVEL_EASY; 2 on DIF_LEVEL_MEDIUM; 3 on DIF_LEVEL_HARD
+ *  - numElites:    The numbers of enemy elites in sector, by game difficulty
+ *  - numCreatures:    The numbers of Crepitus in sector, by game difficulty. (SciFi mode only)
+ *  - numTroopsVariance:    The variance added to numTroops, by game difficulty.
+ *  - numElitesVariance:    The variance added to numElites.
+ *  - numCreaturesVariance:    The variance added to numCreatures.
+ */
+[
+    {
+        //Miguel's basement.  Nothing here.
+        "sector": "A10",
+        "sectorLevel": 1
+    },
+    {
+        //Chitzena mine.  Nothing here.
+        "sector": "B2",
+        "sectorLevel": 1
+    },
+    {
+        //San Mona mine.  Nothing here.
+        "sector": "D4",
+        "sectorLevel": 1
+    },
+    {
+        "sector": "D5",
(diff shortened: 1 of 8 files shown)


assets/externalized/strategic-map-underground-sectors.json
```diff
@@ -0,0 +1,274 @@
...
+    {
+        "sector": "J9",
+        "sectorLevel": 1,
+        "numTroops": [8, 11, 15]
+    },
+    //J9 feeding zone
+    {
+        "sector": "J9",
+        "sectorLevel": 2,
+        "numCreatures": [4, 6, 8],        // ubNumCreatures = (2 + gGameOptions.ubDifficultyLevel*2 + Random( 2 ))
```
lynxlynxlynx, line 61: Your math is fine; these comments are redundant.

src/externalized/strategic/UndergroundSectorModel.cc
```diff
@@ -0,0 +1,126 @@
...
+	std::array<uint8_t, NUM_DIF_LEVELS> vals = {};
+	if (!obj.HasMember(fieldName))
+	{
+		return vals;
+	}
+	auto arr = obj[fieldName].GetArray();
+	if (arr.Size() != NUM_DIF_LEVELS)
+	{
+		SLOGE(ST::format("The number of values in {} is not same as NUM_DIF_LEVELS({})", fieldName, NUM_DIF_LEVELS));
+		throw std::runtime_error("The number of elements does not match the number of difficulty levels");
```
lynxlynxlynx, line 39: Do we catch these exceptions somewhere or are they fatal? It seems two logging calls are redundant.

src/externalized/strategic/UndergroundSectorModel.cc
```diff
@@ -0,0 +1,126 @@
...
+
+void UndergroundSectorModel::validateData(const std::vector<const UndergroundSectorModel*> ugSectors)
+{
+	// check for existence of hard-coded references
+	// the list below is based on the occurrences of FindUnderGroundSector in codebase
+	std::array<uint8_t, 3> basementSectors[] = {
+		{ 13, 5,  1 },  // Draassen mine
+		{ 9,  8,  1 },  // Cambria mine
+		{ 14, 10, 1 },  // Alma mine
+		{ 4,  8,  2 }   // Grumm mine
```
lynxlynxlynx, line 113: So basically modders can only add underground sectors, not move or remove them. I guess this is not a problem for Wildfire, since the maps are so similar?

src/externalized/strategic/UndergroundSectorModel.cc
ja2-derek: They are unhandled exceptions. On Windows it will be a message box saying "exited with status code -107374091". Cannot see the error message without attaching a debugger. Also, I have not figuredd  out how to properly do `ST::String` and subtitutions with std::runtime_error.

Need a bit guidance and suggestions here. I can add this to #1072 to clean up all occurences.

src/externalized/strategic/UndergroundSectorModel.cc
ja2-derek: Yes for now, and only these 4 sectors, because  of creature spreading:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/2ad95c3a7b26ca6c8436e3389173b3be8c9de80a/src/game/Strategic/Creature_Spreading.cc#L332-L333

Alternatively, we can add null-check in Creature_Spreading.cc, and we only WARN on missing.

And later when we externalize creature spreading or make it optional,  we can remove this check.

lynxlynxlynx: It's confusing, since in the json all the mine sectors are in the required part, not just these entrances. But yeah, should also be externalized eventually.

src/externalized/strategic/UndergroundSectorModel.cc
lynxlynxlynx: It appears unification would only be possible for debug builds (SLOGA). Are you saying the native message box will not contain our message or were you referring to the title?

src/externalized/strategic/UndergroundSectorModel.cc
ja2-derek: Oh right. The line off comment "Do all of the mandatory underground mine sectors" was copied from the original code.

Those other mine sectors can be moved and removed, as long as they are consistent with Data/Maps. Only these 4 creature lair entrances are referenced by code.

src/externalized/strategic/UndergroundSectorModel.cc
ja2-derek: Yes I am saying the message box will not contain our message, so the user cannot tell what went wrong.

lynxlynxlynx: So then isn't it simplest to just throw with no message, since it's getting logged beforehand anyway?

ja2-derek: Updated as said.

src/externalized/strategic/UndergroundSectorModel.cc
ja2-derek: Removed that line of confusing comment.

lynxlynxlynx:
Thanks!

PR #1085, opened 2020-05-26
Vvx feats

lynxlynxlynx:
cherry-picked goodies, but apparently I forgot to switch the branch, so the vvx-fixes one should come first — only the last two commits are new. #1002


src/game/Tactical/Turn_Based_Input.cc
```diff
@@ -2729,6 +2747,25 @@ static void SetBurstMode(void)
...
 }
 
+static void GroupAutoReload()
+{
+	// modified from SwitchHeadGear
+
+	SOLDIERTYPE* const selectedSoldier = GetSelectedMan();
+	if (selectedSoldier == NULL) return;
+
+	FOR_EACH_IN_SQUAD(k, selectedSoldier->bAssignment)
```
flaviojs, line 2757: `FOR_EACH_IN_SQUAD` assumes `selectedSoldier->bAssignment` is lower than `NUMBER_OF_SQUADS`.
I would add an assert to ensure the assumption is true.

This code is probably based on `SwitchHeadGear`, which could also use an assert.

lynxlynxlynx: Did something better, just grabbing the current squad, which is also globally tracked. The tactical map check also ensures it's one of the 20 normal slots.

lynxlynxlynx:
Restarted the autocancelled gh jobs, so now all tests pass.

PR #1091, opened 2020-05-28
Fixing a typo in InitializeMines

ja2-derek:
This has to be a typo, as this only initializes the value of the first element, and leaves the rest uninitialized


lynxlynxlynx:
Thanks

PR #1096, opened 2020-06-03
Adjusting log levels

### LINKED ISSUE #1057: Log level adjustments
Playing through the game now, and found some places where I think logging level is inappropriate. I will add to this list as I find more.

 - [ ] `**** Bobby Rays Ordering Error ****` (BobbyRGuns.cc) - This is not a program error, just the player trying to order more than quantity in stock. This should be removed, or at DEBUG level.
 - [x] `opening mod's resource: %s` (ModPackContentManager.cc) - It becomes spammy when you have a full map mod, and the game loads hundreds of tileset files each time entering a sector. Should be DEBUG. (Removed in #1067)
- [ ] `Enemies on team 1 prompted to go on RED ALERT!` (AIMain.cc) - Should be DEBUG log. 
- [ ] `Map version is greater than the current version (old ja2.exe?)` (Map_Information.cc) - The game will continue to load the map regardless. This should be WARN

ja2-derek:
Closes #1057. 

Logging changes only, as described in the issue.

### DIFF
--- a/src/game/Laptop/BobbyRGuns.cc
+++ b/src/game/Laptop/BobbyRGuns.cc
@@ -1119,7 +1119,6 @@ static void SelectBigImageRegionCallBack(MOUSE_REGION* pRegion, INT32 iReason)
 
 
 static UINT8 GetNextPurchaseNumber(void);
-static void ReportBobbyROrderError(UINT16 usItemNumber, UINT8 ubPurchaseNum, UINT8 ubQtyOnHand, UINT8 ubNumPurchasing);
 
 static void PurchaseBobbyRayItem(UINT16 usItemNumber)
 {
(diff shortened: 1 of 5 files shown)


src/game/Laptop/BobbyRGuns.cc
```diff
@@ -1373,13 +1371,3 @@ static UINT8 CheckPlayersInventoryForGunMatchingGivenAmmoID(ItemModel const* con
 	}
 	return n_items;
 }
-
-static void ReportBobbyROrderError(UINT16 usItemNumber, UINT8 ubPurchaseNum, UINT8 ubQtyOnHand, UINT8 ubNumPurchasing)
```
lynxlynxlynx, line 1377: this also has a declaration above.

ja2-derek: Updated

src/game/Tactical/OppList.cc
```diff
@@ -1375,16 +1375,14 @@ static INT16 ManLooksForMan(SOLDIERTYPE* pSoldier, SOLDIERTYPE* pOpponent, UINT8
...
-		SLOGE("ManLooksForMan - TARGET is inactive etc ID %d(%s)to ID %d",
-			pSoldier->ubID, pSoldier->name.c_str(), pOpponent->ubID);
 		return(FALSE);
 	}
 
 
 	// if he's looking for a guy who is on the same team
 	if (pSoldier->bTeam == pOpponent->bTeam)
 	{
-		SLOGE("ManLooksForMan - SAME TEAM ID %d(%s)to ID %d",
```
lynxlynxlynx, line 1387: Like I said in the issue, I can hit this also with the same soldier in both roles, so since this then returns false, we can add an independent `pSoldier == pOpponent` check right at the start.

lynxlynxlynx:
Thanks!

PR #1098, opened 2020-06-05
Externalizing mines

ja2-derek:
For #665, so we can customize mines, incomes, etc. Head miners are not covered in this PR.

## Externalized data

- Mine sectors - surface and underground (`gMineLocation`, GetIdOfMineForSector)
- Mine income   (`guiMinimumMineProduction`)
- Town ID (`gMineLocation`)
- Other misc attributes such as mineral type (`gubMineTypes`), and some special attributes mentioned below

## Generalizations

1. `mineId == MINE_SAN_MONA` => `mine.isAbandoned`
2. `mineId == MINE_ALMA` => `mine.noDepletion` / `mine.headMinerAssigned`
    Alma mine always has Matt as head miner, and is never selected for depletion, due to quests
2. `ubDepletedMineIndex == MINE_DRASSEN`: `mine.delayDepletion`
   If Drassen mine is chosen for depletion, we will delay the depletion. I believe this is for early-game game balance

## Save compatibility

The save file does not carry the number of mines defined. So any change to the number of mines breaks saves. This is checked in validation and noted in JSON,

PR #1105, opened 2020-06-07
Fixing head miner assignment

ja2-derek:
Bug introduced in #1098. Head miner all assigned to the same (wrong) town ID. 

This affects nightly and master since #1098.


lynxlynxlynx:
Nice

PR #1110, opened 2020-06-08
Tidy up old changelog

ja2-derek:
Was looking through old changes today. 

Did a regex substitution, from `^\(r(\d\d)` to  `\- \(r(\1)`, after line 253, so old logs are formatted nicely in Markdown.

### DIFF
--- a/changes.md
+++ b/changes.md
@@ -250,364 +250,364 @@
 # JA2-Stracciatella History of Changes
 (this is the old changelog of the original project)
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks, looks fine. But please add "[ci skip]" to such work, so we don't needlessly burden the bots. Appveyor is annoyingly busy already.

ja2-derek:
> 
> 
> Thanks, looks fine. But please add "[ci skip]" to such work, so we don't needlessly burden the bots. Appveyor is annoyingly busy already.

Sure!

PR #1115, opened 2020-06-10
Translations: Fix json syntax in shipping destinations

flex-lab:
[ERROR] externalized/DefaultContentManager.cc: Failed to parse 'strings/shipping-destinations-ger.json'


ja2-derek:
Thanks for the fix. I will try to make a unit test to parse all JSON files for syntax errors

PR #1116, opened 2020-06-11
Adding unit test to parse every JSON in externalized dir

ja2-derek:
To prevent JSON syntax errors. Some files are not always processed in loadGameData(), in particular the localization files.

No other JSON problems found other than those already addressed in [#1115](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1115/files?w=1).

### DIFF
--- a/src/externalized/DefaultContentManager_unittests.cc
+++ b/src/externalized/DefaultContentManager_unittests.cc
@@ -7,14 +7,6 @@
 
 #include "gtest/gtest.h"
 
-#ifdef __cpp_lib_filesystem
-#include <filesystem>
-namespace fs = std::filesystem;
-#else
-#define _SILENCE_EXPERIMENTAL_FILESYSTEM_DEPRECATION_WARNING
-#include <experimental/filesystem> // remove when we are on C++ 17
-namespace fs = std::experimental::filesystem;
-#endif
 
 #define TMPDIR "temp"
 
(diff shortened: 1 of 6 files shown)


ja2-derek:
Linux and mingw64 build fine, just need the JSON fix for tests to pass.

Mac build is failing due to missing `stdc++fs` library. I will check if this can be easily fixed, or else may re-implement the FS stuff with primitives, avoiding `<filesystem>` totaly.

src/externalized/DefaultContentManager_unittests.cc
```diff
@@ -127,4 +137,37 @@ TEST(TempFiles, deleteFile)
...
+TEST(ExternalizedData, readAllData)
+{
+	DefaultContentManager* cm = DefaultContentManagerUT::createDefaultCMForTesting();
+	ASSERT_TRUE(cm->loadGameData());
+}
+
+std::vector<ST::string> recursivelyListAllFiles(ST::string dir, std::string suffix)
+{
+	std::vector<ST::string> files;
+	for (auto& p : fs::recursive_directory_iterator(dir.to_std_string()))
```
lynxlynxlynx, line 149: We have nothing in FileMan or our rust fs layer to do this? The whole std::filesystem include block looks fragile. CC @flaviojs

ja2-derek: How about now? Added `ls -R` to FileMan

lynxlynxlynx: Looks better to me, thanks. If this is now done and Flávio doesn't comment in a while, go ahead and merge it.

ja2-derek:
Rebased to master

lynxlynxlynx:
I suspect you missed the last commits, since the test breaks on the same json.

PR #1117, opened 2020-06-11
Misc fixes: gcc warning and Coverity defects

ja2-derek:
Tried doing a build with gcc today, and saw a bunch of new compiler warnings. These should not affect anything.

### DIFF
--- a/src/game/Strategic/Strategic_AI.cc
+++ b/src/game/Strategic/Strategic_AI.cc
@@ -293,7 +293,7 @@ static void SendReinforcementsForGarrison(INT32 iDstGarrisonID, UINT16 usDefence
 
 static void RequestAttackOnSector(UINT8 ubSectorID, UINT16 usDefencePoints)
 {
-	INT32 i;
+	UINT32 i;
 	for( i = 0; i < gGarrisonGroup.size(); i++ )
 	{
 		if( gGarrisonGroup[ i ].ubSectorID == ubSectorID && !gGarrisonGroup[ i ].ubPendingGroupID )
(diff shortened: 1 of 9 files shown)


ja2-derek:
> 
> 
> [463dbdb](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/463dbdbe15c792d9e34d92da34817f361dc007da) Hi ja2-derek, we did the same commit today.

Ah cool!

ja2-derek:
Took away the signed-unsigned stuff. Thanks @wvx

wvx:
> Took away the signed-unsigned stuff. Thanks @wvx

No problem if you prefer your commit. The commit was hidden, so it is understandable that we duplicated the effort and I was asked to separate the PR. Thanks for also looking at the compiler warnings.

lynxlynxlynx:
So can the other commit be cherrypick it into your branch? Seems it does quite some more.

ja2-derek:
> 
> 
> So can the other commit be cherrypick it into your branch? Seems it does quite some more.

Done!

src/game/Strategic/Strategic_AI.cc
```diff
@@ -3139,7 +3140,7 @@ static void RequestHighPriorityGarrisonReinforcements(size_t iGarrisonID, UINT8
 	}
 	ubDstSectorX = (UINT8)SECTORX( gGarrisonGroup[ iGarrisonID ].ubSectorID );
 	ubDstSectorY = (UINT8)SECTORY( gGarrisonGroup[ iGarrisonID ].ubSectorID );
-	if( iBestIndex != -1 )
+	if( iBestIndex != (size_t)-1 )
```
ja2-derek, line 3143: How does it work? Won't the cast change `-1` into `INT_MAX`? Can it still compare to a signed int?

wvx: Yes, as I understand it it should be something like SIZE_MAX which is OK if "size_t iBestIndex = -1;" is equivalent to "size_t iBestIndex = SIZE_MAX;" There are other cases where -1 was inteded to returned as SIZE_MAX (but specified only as -1) to flag a special case such as in "size_t const patrol_id = FindPatrolGroupIndexForGroupID(g.ubGroupID);"

Maybe, as you suggest, we should be using the limits, such as SIZE_MAX in this case, to be clear on the intention.

ja2-derek: Aha. I didn't realize `iBestIndex` is unsigned

lynxlynxlynx:
Thanks guys!

PR #1118, opened 2020-06-12
Attempt to improve GitHub CI build time

ja2-derek:
Attempt to improve on #1101 and #1102. Only covers GitHub CI builds.

## Summary of changes

- Split the `ci-build.sh` script into 3 steps - setup, build and publish
- Add `ccache` to the CI build process (https://ccache.dev)
    - CMake will make use of it if installed
- Cache  C++ objects with `ccache` (`~/.ccache`)
    - Prefers the last build from same Git branch, falls back to last master build
- Cache Rust deps and output
    - Cache `~/.cargo`, `~/.rustup` and `$PROJECT_DIR/dependencies/lib-stracciatella/target`
    - Assuming Rust code not changed often, cache is used only if the entire `rust/` directory is untouched
- Add concurrency in make (`-j 4`)
- `rustup` with `minimal` [profile](https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles); install `rustfmt` and `clippy` separately

## Results

- Linux build with primed cache takes ~3 minutes (using system-provided libfltk)
- MinGW and Mac builds with primed cache take ~6 minutes

PR #1119, opened 2020-06-16
Fix msys2-mingw build on AppVeyor

ja2-derek:
AppVeyor build broken since #1118, because the script `.github/scripts/ci-buidl.sh` was split. I didn't realize it has been used by AppVeyor msys2-mingw build.

### DIFF
--- a/.github/scripts/ci-build.sh
+++ b/.github/scripts/ci-build.sh
@@ -53,7 +53,7 @@ elif [[ "$CI_TARGET" == "msys2-mingw32" ]]; then
   export CONFIGURE_CMD="${CONFIGURE_CMD} -DCPACK_GENERATOR=ZIP"
   export RUSTUP_HOME="$(cygpath -w ~/.rustup)"
   export CARGO_HOME="$(cygpath -w ~/.cargo)"
-  export PATH=$PATH:$HOME/.cargo/bin
+  export PATH="$PATH:$HOME/.cargo/bin"
   export RUN_INSTALL_TEST=false # no sudo
 
 elif [[ "$CI_TARGET" == "mac" ]]; then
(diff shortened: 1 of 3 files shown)


ja2-derek:
Fixed. We should be disabling that msys2 build though.

.github/scripts/ci-build.sh
```diff
@@ -46,15 +46,14 @@ if [[ "$CI_TARGET" == "linux" ]]; then
...
-  export RUSTUP_INIT_ARGS="${RUSTUP_INIT_ARGS} --target=x86ST_64-pc-windows-gnu --profile=minimal"
   export RUN_TESTS=false
 
 elif [[ "$CI_TARGET" == "msys2-mingw32" ]]; then
   export CMAKE_GENERATOR="MSYS Makefiles"
   export CONFIGURE_CMD="${CONFIGURE_CMD} -DCPACK_GENERATOR=ZIP"
   export RUSTUP_HOME="$(cygpath -w ~/.rustup)"
   export CARGO_HOME="$(cygpath -w ~/.cargo)"
-  export RUSTUP_INIT_ARGS="${RUSTUP_INIT_ARGS}-i686-pc-windows-gnu --default-host=i686-pc-windows-gnu"
+  export PATH=$PATH:$HOME/.cargo/bin
```
lynxlynxlynx, line 56:
```suggestion
  export PATH="$PATH:$HOME/.cargo/bin"
```
In case the paths ever get whitespace in them

lynxlynxlynx:
Why the change from the minimal profile? Not good enough or did those two extra components make it identical to the default. Or something third?

ja2-derek:
> The default profile includes all of components in the minimal profile, and adds rust-docs, rustfmt, and clippy.

Using `minimal` we need to install fmt and clippy with a second command. I do not see any significant difference in the time spent between default and minimal + fmt + clippy.

PR #1120, opened 2020-06-19
Clean up SET_PALETTEREP_ID

### LINKED ISSUE #1113: Cannot change merc colors in Map Editor

### What is wrong
Details, details, details.

|  master | JA2 Steam version |    no    |    yes     |

### Symptoms

Unable to change colors of merc detailed placement, in Map Editor. 
- Open any map
- Switch to MERC mode, choosen any person. Enable detailed placement. Switch to colors/palette tab
- Click on the arrows to change colors. Nothing happens

### Analysis

Should be a result of #1049, in which:

- `char*` and `strcpy` replaced with `ST::string` and `=`
  `strcpy` overwrites content at the memory address, and it could be inside some structs; ST::string assignment is probably replacing the variable reference with another string instance
- `SET_PALETTEREP_ID` probably fails in certain cases, for example when temp variables are involved. May need to be a reference or pointer here.

EditorMercs.cc:[ProcessMercEditing](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c13a6e53601fcd41f11d36fadf75c45e67cbf25a/src/game/Editor/EditorMercs.cc#L264)() handles the merc color changes. There might be other places affect as well.

ja2-derek:
Cleans up and removes `SET_PALETTEREP_ID`. Replaces with `=` assignment, or `.set()`.

Fixes #1113 and follows up on https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1049#discussion_r415354629. This should clear the one remaining bug before [0.17.x](https://github.com/ja2-stracciatella/ja2-stracciatella/issues?q=is%3Aopen+is%3Aissue+milestone%3Av0.17)

# Summary

I took the patch from https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1113#issuecomment-641044901. On a second thought, I think there is nothing wrong with doing `ST::string::set`. It is part of the public API.

Then I went through each usage of `SET_PALETTEREP_ID`. All other usages are fine with a simple `=` assignment. So I replaced all of them and it should clear things up.

### DIFF
--- a/src/game/Editor/EditorMercs.cc
+++ b/src/game/Editor/EditorMercs.cc
@@ -267,45 +267,45 @@ void ProcessMercEditing(void)
 
 	// Handle changes to the merc colors
 	SOLDIERTYPE& s = *g_selected_merc;
-	ST::string soldier_pal;
-	ST::string placement_pal;
+	ST::string* soldier_pal;
+	ST::string* placement_pal;
 	UINT8 ubType;
 	switch (iEditWhichStat)
 	{
 		case 0:
 		case 1:
 			ubType        = EDIT_COLOR_HEAD;
-			soldier_pal   = s.HeadPal;
-			placement_pal = gpSelected->pDetailedPlacement->HeadPal;
+			soldier_pal   = &s.HeadPal;
+			placement_pal = &gpSelected->pDetailedPlacement->HeadPal;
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
Thanks!

PR #1123, opened 2020-06-21
Giving names to lock pauses

ja2-derek:
Game_Clock.h has an `enum` of numbered lock pause reasons. This PR is a refactor to give them names with some meaning.

All renames are done by the "Rename" operation (Ctrl-R, Ctrl-R) in Visual Studio, so the rename should be safe and consistent.

`LOCK_PAUSE_09` and `LOCK_PAUSE_10` are not used in codebase, and hence removed.


lynxlynxlynx:
LGTM

PR #1125, opened 2020-06-22
Upload builds to google storage

selaux:
Should fix https://github.com/ja2-stracciatella/ja2-stracciatella/issues/873.

Not finished yet. But it is only testable using a PR...


ja2-derek:
Is this something we can use? https://docs.travis-ci.com/user/deployment-v2/providers/gcs/

selaux:
So I looked into this, but in the end it would be nice to end up with the same (or similar) code for all CI systems, so there is not so much duplication.

lynxlynxlynx:
Is this now done?

selaux:
No, I didn't get around to it yet.

selaux:
It is done now: https://storage.googleapis.com/ja2-builds/index.html#pull-requests/1125/

lynxlynxlynx:
Thanks!

Issue #667, opened 2018-01-05
slf archive and folder confusion


JAsmine-ja2:
![edit](https://user-images.githubusercontent.com/17213795/34604730-2f2c6cc6-f209-11e7-8587-aba984a709f6.jpg)

EDIT:

the editor loads map files from maps.slf. when saving a map, the editor seems to check maps.slf for the existence of the specific map file. This generates the "map exists" warning. When saving, the new map file is saved to the folder, where nothing is overwritten, because the folder is empty.

PR #1126, opened 2020-06-23
Supporting more data directories

ja2-derek:
For #707. Building on top of the VFS introduced in #1067.

# Summary

- Adding VFS_ORDER for user home directory
    - The effective VFS order is like https://github.com/ja2-stracciatella/ja2-stracciatella/issues/707#issuecomment-509862638, except the tmp dir
- Adding mods and assets directory in user home (`~/.ja2` or `%userprofile%\documents\JA2`)
    - Mods can be loaded from one or more directories. VFS ordering determines which files to use
    - User home directory (`~/.ja2/data`) has the higher  priority than Stracciatella assets (`/usr/share/ja2`)
- Update launcher to list mods from both assets and home directories
-------

If this PR is accepted, we may need to update documentations to tell users what they can do. They may need to create `~/.ja2/data/` and copy from `/usr/share/ja2`. We do not support JSON merging, so a customized `game.json` in home directory will replace the default file.

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -157,7 +157,7 @@ static ST::string LoadEncryptedData(ST::string& err_msg, STRING_ENC_TYPE encType
 }
 
 DefaultContentManager::DefaultContentManager(GameVersion gameVersion,
-						const ST::string &configFolder,
+						const ST::string &userHomeDir,
 						const ST::string &gameResRootPath,
 						const ST::string &externalizedDataPath
 	)
@@ -173,7 +173,7 @@ DefaultContentManager::DefaultContentManager(GameVersion gameVersion,
 	 * exist we should use them.  If doesn't exist, then use lowercased names.
 	 */
 
-	m_configFolder = configFolder;
+	m_userHomeDir = userHomeDir;
 	m_gameResRootPath = gameResRootPath;
 	m_externalizedDataPath = externalizedDataPath;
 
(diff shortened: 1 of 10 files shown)


lynxlynxlynx:
Nice, you don't shy away from rust either. The whole thing is a bit over my head though, so I'll leave the review to others.

ja2-derek:
Might  also help https://github.com/ja2-stracciatella/ja2-stracciatella/issues/667.

Map can still be loaded with `ModPackContentManager`, but saving and the overwrite warning can use `~/.ja2/data`.

src/externalized/DefaultContentManager.cc
```diff
@@ -191,34 +191,47 @@ DefaultContentManager::DefaultContentManager(GameVersion gameVersion,
...
+		if (throwOnError)
+		{
+			throw std::runtime_error("Failed to add to VFS");
+		}
 	}
+}
 
+void DefaultContentManager::init()
+{
+	AddVFSLayer(VFS_ORDER::ASSETS_USERHOME, FileMan::joinPaths(m_userHomeDir, "data"));
```
selaux, line 227: Should this be `"data"` or `"externalized"?`

ja2-derek: It is a new directory so it's up to us. The codebase treats externalized and data the same  I chose `data` because this directory can also contain binary assets.

ja2-derek:
I'm going to merge this if there are no objections

PR #1129, opened 2020-06-27
Fix tactical screen scroll freeze

### LINKED ISSUE #1122: Unable to scroll screen when entering turn mode

### What is wrong
Details, details, details.

|  0.16.1 and master | JA2 1.03 Steam|   no    |    not sure     |

When your merc sees an enemy when the mouse is held down, the scroll will not work. It will get back to normal if you do some keyboard shortcut, or switch to strategic map view and switch back.

To reproduce, start a fresh new game:
- After arrival in A9, shoot the gun so enemies will start coming
- Hold the left mouse button down (e.g. drag a selection box and hold)
- When enemy comes, the game switches to turn-based mode
- Release the mouse now. The screen is no longer scrollable. Radar map not working.
- Some actions that can unlock the scroll:
  - Moving the merc,
  - Selecting another merc by clicking in tactical screen (not the bottom menu),
  - Switching to strategic map screen and switch back,
  - Auto bandaging by hotkey, or
  - Quick save and load (Alt+S, Alt+L)

Maybe the mouse state is messed up.

See an example here: https://youtu.be/Pr0I1_EOD5Y?t=819

Cannot reproduce in vanilla (Steam version 1.03).

OS is Win 10.

ja2-derek:
Analysis here: https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1122#issuecomment-650511871

Fixes #1122. Preserves the behavior of #565.

This fix ensures scrolling is un-freezed when we enter combat mode during rubber-banding. Also re-tested #550.

### DIFF
--- a/src/game/Tactical/Interface_Control.cc
+++ b/src/game/Tactical/Interface_Control.cc
@@ -577,8 +577,8 @@ void EraseInterfaceMenus( BOOLEAN fIgnoreUIUnLock )
 	PopDownOpenDoorMenu( );
 	DeleteTalkingMenu( );
 
-	// Stop Rubberbanding every time a menu is erased/opened
-	gRubberBandActive = FALSE;
+	// Cancel Rubberbanding every time a menu is erased/opened
+	EndRubberBanding(true);
 	ResetMultiSelection();
 }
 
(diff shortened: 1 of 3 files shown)


src/game/Tactical/Interface_Control.cc
```diff
@@ -596,8 +596,6 @@ void ResetInterfaceAndUI( )
 
 	EraseRenderArrows( );
 
-	EndRubberBanding( );
```
ja2-derek, line 599: This is already called in EraseInterfaceMenus()

lynxlynxlynx:
Nice!

Issue #1102, opened 2020-06-06
Use caching on build bots / speed up builds


lynxlynxlynx:
All three of our CI platforms support caching between builds (to some extent). We should use it to both speed build times and be less of a burden on the infrastructure.

Low hanging fruit: make sure to install all available dependencies, so the bundled one don't get used (eg. gtest). 
Perhaps string_theory can also be cached.
Crates should be possible too.

ja2-derek:
Not sure how much time is taken to set up dependencies. Perhaps we should add some timing logs, or split them up into smaller sections.

`~/.cargo` should be very easy : https://docs.travis-ci.com/user/caching/#rust-cargo-cache

In my local, gcc compilation is the slowest. How about keeping and re-using the intermediate object files for PR builds (i.e. except nightly and releases)?

lynxlynxlynx:
On the github workflows, timing is included by default.

Interesting idea about *.o — I wonder if/how well it could work. :thinking:

ja2-derek:
Travis has the timing for each command run. 

GitHub CI only shows the overall time for the entire `cibuild.sh` script. We may want to split it.

But do we still need Travis build on PR? I think we only need Travis for nightly.

ja2-derek:
I suggest:

- [ ] Disable Travis on PR, and focus on GitHub CI only
- [ ] Split `.github/scripts/ci-build.sh`for timing in logs
- [ ] Experiment with:
    - [ ] `make -j $NUM_CORES` 
    - [ ] Caching on GitHub CI ([docs](https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows))
        - [ ] `ccache` - https://ccache.dev/
        - [ ] `~/.cargo`

lynxlynxlynx:
I don't think -j would make sense, since it's just heavier on the infra. Some of the bots explicitly discourage it, though I don't remember which.

ccache — great idea, forgot about that. We'll perhaps have to make the cache per-runner though, so eg. the different msvc versions don't bust it for each other.

lynxlynxlynx:
It also looks like we do two mingw builds, which is not really necessary. I suggest we drop either; the crosscompiling one is probably more worthwhile as far as testing the system goes.

ja2-derek:
> It also looks like we do two mingw builds, which is not really necessary. I suggest we drop either; the crosscompiling one is probably more worthwhile as far as testing the system goes.

[...]

PR #1132, opened 2020-06-27
Run rust lint and fmt only once

ja2-derek:
Small improvement for #1102. This should save around 2 minutes on each AppVeyor build. 

`rustfmt` and `clippy` are both static analysis. We only need to run once per commit to rust code, and no point to run on each and every platform.

After this PR rust code checks will only run on the Linux build on GitHub CI, which is currently the fastest build. This should speed up all other builds.


lynxlynxlynx:
Thanks!

Issue #1127, opened 2020-06-25
Small issues on master


ja2-derek:
Tracking small, non-blocker issues I found on current master. Some are related to my recent PRs. I will keep updating this.

- [x] Log says "externalized/DefaultContentManager.cc: Unsupported file type." when it is actually a directory not found. (#1131)
- [x] The title of the IMP Skills Trait screen is a few pixels too high. See this [screenshot](https://user-images.githubusercontent.com/63151803/85189813-bf4cf200-b2e4-11ea-852e-ba2656b6388c.png) (#1130)
- [x] (Editor, 0.16.1) Unnamed placeholder items towards the ends of AMMO and WEAPONS lists in Editor. Those are the `CLIP_NOTHING` in `magazines.json` and `unused` in `weapons.json` (#1191)
- [x] IMP Skill Traits introduced a couple of "high impact" defects on Coverity  (#1190)
- [x] Compiler warning:
    - [x] Missing `()` in function call: https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1174/commits/a1d2919104d4c6c3e72cf62d874d12d828a89888#diff-60d1ebf19ff040db3da437082175c7feR1153   (#1190)
- [x] If JSON has a missing data field or field contains an unexpected data type, logs only say "Assertion failed: false, ... rapidjson/document.h, line 1053", without mentioning which JSON file or which field. (#1133, #1167)

lynxlynxlynx:
Forgot about this, but I'm working on tackling the last item for game.json parsing.

lynxlynxlynx:
For the last item, missing keys are now taken care of, but invalid values still assert.

Issue #32, opened 2013-03-25
Return support of 90210 code


lynxlynxlynx:
Originally reported by: **Gennady Trafimenkov (Bitbucket: [gennady](http://bitbucket.org/gennady), GitHub: [gennady](http://github.com/gennady))**

---

In some versions of the original game (for example, Russian and Russian Gold) it was possible to enter code 90210 instead of xep624 and get previously generated IMP character.

This is not supported in JA2-Stracciatella at the moment.

Return support of it back. It wouldn't hurt anyone.

Reported by https://bitbucket.org/alex_weil

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/32

lynxlynxlynx:
vxx already did something more flexible, you just have to remember the nickname:
https://bitbucket.org/vxx/ja2-stracciatella/commits/ca194b2ca636dca814781c50ea594166a7d9a6d6?at=master

lynxlynxlynx:
now here:
https://github.com/ja2-stracciatella/ja2-stracciatella/commit/a78ca6677528f63aa330b3da783278c8e3a30de0

lynxlynxlynx:
#1133 adds basic support for importing IMPs. What's remaining is simplifying it so you don't enter the nickname, but already trigger it on the main IMP page (nickname instead of the password). And the export format can be streamlined and/or made non-binary.

icelynjennings:
Relevant code appears present, albeit commented out, in the codebase of JA2.5 Unfinished Business:

https://github.com/Manuel-K/ja2-ub-comparison/blob/ee3f53866641bccd4544a45bfdde51bac55c0c2c/src_ub/game/Laptop/IMP%20HomePage.c#L390-L395

lynxlynxlynx:
It would still need adapting, since we can have an arbitrary amount of profiles in a savegame dir, not just one. Not that it would be hard, just take the most recently created one.

PR #1133, opened 2020-06-28
Vvx and more

lynxlynxlynx:
This cherry-picks a few of the updated changes from vvx, finishes them where they were incomplete and then cleans up game.json and its handling a bit.

#32
#665
#1127

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -80,7 +80,14 @@
     //  Enables extra attachments: boosters and break lights on leggings, goggles on helmets.
     "extra_attachments": false,
 
-    //  Enables middle mouse button to trigger a look into cursor direction.
+    //  Load an IMP merc from a previous save game by typing in same nickname on IMP form.
+    //  Make sure to use the same capitalization!
+    //  This setting must be already enabled during saving to enable a later load.
+    //  vanilla setting: false, false
+    "imp_load_saved_merc_by_nickname": false,
+    "imp_load_keep_inventory": false,
+
+    //  Enables middle mouse button to trigger a look into cursor direction
     //  vanilla_setting: false
     "middle_mouse_look": true,
 
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -80,13 +80,6 @@
     //  Enables extra attachments: boosters and break lights on leggings, goggles on helmets.
     "extra_attachments": false,
 
-    //  Load an IMP merc from a previous save game by typing in same nickname on IMP form.
-    //  Make sure to use the same capitalization!
-    //  This setting must be already enabled during saving to enable a later load.
-    //  vanilla setting: false, false
-    "imp_load_saved_merc_by_nickname": false,
-    "imp_load_keep_inventory": false,
-
     //  Enables middle mouse button to trigger a look into cursor direction
     //  vanilla_setting: false
     "middle_mouse_look": true,
@@ -106,14 +99,15 @@
     "tactical_head_damage_multiplier": 1.5,
     "tactical_legs_damage_multiplier": 0.5,
 
-    //  IMP attribute point defaults:
-    //  Warning: Setting min attribute points to a value less than 15 allows
-    //           the creation of a character with less than 15 health. This
-    //           is a dying character!
-    //  Zero attribute point bonus is what you receive for setting a skill to 0.
-    //  Bonus attribute points is your initial pool of points to distribute.
-    //  vanilla_setting: 85, 35, 15, 40
+    //  IMP creation tweaks
     "imp": {
+        //  IMP attribute point defaults:
+        //  Warning: Setting min attribute points to a value less than 15 allows
+        //           the creation of a character with less than 15 health. This
+        //           is a dying character!
+        //  Zero attribute point bonus is what you receive for setting a skill to 0.
+        //  Bonus attribute points is your initial pool of points to distribute.
+        //  vanilla setting: 85, 35, 15, 40
         "max_attribute_points": 85,
         "min_attribute_points": 35,
         "zero_attribute_points_bonus": 15,
@@ -122,7 +116,14 @@
         // Replace the IMP personality quiz with a form to select skills directly
         // NOTE: does not set any personality
         // vanilla setting: false
-        "pick_skills_directly": false
+        "pick_skills_directly": false,
+
+        //  Load an IMP merc from a previous save game by typing in same nickname on IMP form.
+        //  Make sure to use the same capitalization!
+        //  This setting must be already enabled during saving to enable a later load.
+        //  vanilla setting: false, false
+        "load_saved_merc_by_nickname": false,
+        "load_keep_inventory": false
     },
 
     "progress": {
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -3,7 +3,7 @@
     //  The settings in this file have an influence on the game in general.
 
     //  --------------------------------------------------
-    //  New hotkeys
+    //  New hotkeys and input
     //  vanilla setting for all hotkeys: false
     //  --------------------------------------------------
 
@@ -14,6 +14,14 @@
     //  Press 'ctrl+q' to swap hand items
     //  Press 'shift+j' to climb through open window
     "extra_hotkeys": true,
+
+    //  Press d in realtime mode to manually start turn based mode.
+    //  vanilla setting: false
+    "can_enter_turnbased": false,
+
+    //  Enables middle mouse button to trigger a look into cursor direction
+    //  vanilla setting: false
+    "middle_mouse_look": true,
 
     //  --------------------------------------------------
     //  System settings
@@ -79,14 +87,6 @@
 
     //  Enables extra attachments: boosters and break lights on leggings, goggles on helmets.
     "extra_attachments": false,
-
-    //  Enables middle mouse button to trigger a look into cursor direction
-    //  vanilla_setting: false
-    "middle_mouse_look": true,
-
-    //  Press d in realtime mode to manually start turn based mode.
-    //  vanilla_setting: false
-    "can_enter_turnbased": false,
 
     //  Packages at Drassen airport never get stolen, no need to bribe Pablo.
     //  vanilla_setting: false
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -116,5 +116,31 @@
         // NOTE: does not set any personality
         // vanilla setting: false
         "pick_skills_directly": false
-    }
+    },
+
+    "progress": {
+        //  Progress weight of parameter on campaign progress
+        //  vanilla settings: kills 25, control 25, income 50
+        "weight_kills": 25,
+        "weight_control": 25,
+        "weight_income": 50,
+
+        //  Progress point progression
+        //  vanilla settings: 0_easy 7, 1_medium 10, 2_hard 15
+        "kills_per_point_0": 7,
+        "kills_per_point_1": 10,
+        "kills_per_point_2": 15,
+
+        //  Minimum first time progress to trigger event
+        //  vanilla settings: madlab: 35, mike: 50, iggy: 70
+        "event_madlab_min": 35,
+        "event_mike_min": 50,
+        "event_iggy_min": 70
+    },
+
+    //  Allowed unhired merc deaths at given difficulty level
+    //  vanilla settings: 1, 2, 3
+    "unhired_merc_deaths_difficulty_0": 1,
+    "unhired_merc_deaths_difficulty_1": 2,
+    "unhired_merc_deaths_difficulty_2": 3
 }
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -135,9 +135,9 @@
 
         //  Progress point progression
         //  vanilla settings: 0_easy 7, 1_medium 10, 2_hard 15
-        "kills_per_point_0": 7,
-        "kills_per_point_1": 10,
-        "kills_per_point_2": 15,
+        "kills_per_point_easy": 7,
+        "kills_per_point_medium": 10,
+        "kills_per_point_hard": 15,
 
         //  Minimum first time progress to trigger event
         //  vanilla settings: madlab: 35, mike: 50, iggy: 70
@@ -148,7 +148,7 @@
 
     //  Allowed unhired merc deaths at given difficulty level
     //  vanilla settings: 1, 2, 3
-    "unhired_merc_deaths_difficulty_0": 1,
-    "unhired_merc_deaths_difficulty_1": 2,
-    "unhired_merc_deaths_difficulty_2": 3
+    "unhired_merc_deaths_easy": 1,
+    "unhired_merc_deaths_medium": 2,
+    "unhired_merc_deaths_hard": 3
 }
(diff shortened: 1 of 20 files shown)


src/game/SaveLoadGame.cc
```diff
@@ -1132,6 +1138,85 @@ static void SaveMercProfiles(HWFILE const f)
 	}
 }
 
+const char * IMPSavedProfileCreateFilename(const char *nickname_cstring)
```
ja2-derek, line 1141: All `const char*` should be replaced with `ST::string`. We do not need any of the `c_str()`, `malloc` or `free`.

`sprintf` should be replaced by `ST::format`

src/game/SaveLoadGame.cc
```diff
@@ -1132,6 +1138,85 @@ static void SaveMercProfiles(HWFILE const f)
...
+
+	SGPFile *f = IMPSavedProfileOpenFileForRead(nickname);
+	FileSeek(f, sizeof(MERCPROFILESTRUCT), FILE_SEEK_FROM_START);
+	FileRead(f, pSoldier->inv, sizeof(OBJECTTYPE) * NUM_INV_SLOTS);
+	FileClose(f);
+}
+
+void SaveIMPPlayerProfiles()
+{
+	#define LASTIMPPROFILE NPC169 // (PLAYER_GENERATED_CHARACTER_ID+5)
```
ja2-derek, line 1200: `(PLAYER_GENERATED_CHARACTER_ID+5)` is correct, isn't it? Why don't we use it? We don't even need this #define.

src/game/Laptop/IMP_Begin_Screen.cc
```diff
@@ -289,6 +300,20 @@ static void BtnIMPBeginScreenDoneCallback(GUI_BUTTON *btn, INT32 reason)
 			iCurrentImpPage = IMP_MAIN_PAGE;
 			fButtonPendingFlag = TRUE;
 		}
+
+		if(GCM->getGamePolicy()->imp_load_saved_merc_by_nickname)
+		{
+			if(IMPSavedProfileDoesFileExist(pNickNameString.c_str()))
```
ja2-derek, line 306: This is quite a bit different from  `90210` or JA2 1.13, where the nickname is to be typed in place of the XEP624 passcode. 

It would be nicer if we can skip the entire creation process, without even showing the first screen.

src/game/SaveLoadGame.cc
```diff
@@ -1132,6 +1138,85 @@ static void SaveMercProfiles(HWFILE const f)
...
+		if (mercprofile->bLife == 0) continue;
+		SOLDIERTYPE* const pSoldier = FindSoldierByProfileID(i);
+		if (!pSoldier) continue;
+		if (pSoldier->bTeam != OUR_TEAM) continue;
+		if (pSoldier->ubWhatKindOfMercAmI != MERC_TYPE__PLAYER_CHARACTER) continue;
+
+		SGPFile *f = IMPSavedProfileOpenFileForWrite(mercprofile->zNickname.c_str());
+		if (!f) continue;
+
+		mercprofile->ubSuspiciousDeath = i - PLAYER_GENERATED_CHARACTER_ID; // save voice_id, field not used for resuscitated merc
```
ja2-derek, line 1214: Rather than doing this hack, why not just write the Profile ID and Portrait ID into the file along with the structs?

This is also what JA 1.13 does : https://github.com/aimnas/ja2-1.13-source-mirror/blob/b988df876f36e373ca21e0001a91aefe87ecdd50/Laptop/IMP%20Confirm.cpp#L1456

src/game/SaveLoadGame.cc
```diff
@@ -1132,6 +1138,85 @@ static void SaveMercProfiles(HWFILE const f)
...
+{
+	const char *profile_filename = IMPSavedProfileCreateFilename(nickname);
+	bool fexists = Fs_exists(profile_filename);
+	free((void*)profile_filename);
+	return fexists;
+}
+
+SGPFile* const IMPSavedProfileOpenFileForRead(const char *nickname)
+{
+	if (!IMPSavedProfileDoesFileExist(nickname)) return (SGPFile *)-1;
```
ja2-derek, line 1158: We can just throw an exception here.

src/game/SaveLoadGame.cc
ja2-derek: Or if we are planning for future (for UB - there is a new IMP profile ID), this define should be placed in Solder_Profile.h.

src/externalized/policy/DefaultGamePolicy.cc
```diff
@@ -1,50 +1,72 @@
...
+	starting_cash_hard = gp.getOptionalInt("starting_cash_hard", 30000);
+
+	f_drop_everything = gp.getOptionalBool("drop_everything");
+	f_all_dropped_visible = gp.getOptionalBool("all_drops_visible");
+
+	multiple_interrupts = gp.getOptionalBool("multiple_interrupts");
+
+	enemy_weapon_minimal_status = gp.getOptionalInt("enemy_weapon_minimal_status", 0);
+
+	JsonObjectReader ai = JsonObjectReader(gp.GetValue("ai"));
```
ja2-derek, line 27: This is now the last mandatory field in game.json :)

lynxlynxlynx: All three groupings/objects are, but since this is the first time things can go wrong, we don't need to care yet. It will become important if we add another grouping.

src/game/Laptop/IMP_Confirm.cc
```diff
@@ -220,12 +222,45 @@ static void BtnIMPConfirmYes(GUI_BUTTON *btn, INT32 reason)
...
+
+		SOLDIERTYPE* const pSoldier = FindSoldierByProfileID(PLAYER_GENERATED_CHARACTER_ID + LaptopSaveInfo.iVoiceId);
+		if (!pSoldier) return;
+
+		if (fLoadingCharacterForPreviousImpProfile && gamepolicy(imp_load_keep_inventory))
+		{
+			IMPSavedProfileLoadInventory(gMercProfiles[PLAYER_GENERATED_CHARACTER_ID + LaptopSaveInfo.iVoiceId].zNickname.c_str(), pSoldier);
+		}
+
+		if (pSoldier->ubID == 0 && FindObj(pSoldier, LETTER) == NO_SLOT)
```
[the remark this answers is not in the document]
lynxlynxlynx: Or just drop it directly.

src/game/SaveLoadGame.cc
lynxlynxlynx: @wvx do you remember why you tried both constants?

src/game/Laptop/IMP_Begin_Screen.cc
lynxlynxlynx: Good idea, but we need to also increase the char limit for it to work. No biggie.

src/game/SaveLoadGame.cc
lynxlynxlynx: Once we get rid of them all, we should just poison it via `__attribute__ ((deprecated("some message.")))` to avoid reintroduction. Works on gcc/clang, not elsewhere.

src/game/Laptop/IMP_Confirm.cc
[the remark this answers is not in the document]
wvx: The ideal would be to use JSON to export/import to abstract from the compiler-struct memory format. This is not difficult. A text file would make it easier to edit for creating mod scenarios such as the intended "All you need is kill".

The data was placed in an unrelated field because it was an easy hack.

src/game/Laptop/IMP_Begin_Screen.cc
wvx: I would like the flow to be (optionally) automatic so that a player could jump right back in after failing the solo/duo straight-for-Meduna scenario.

src/game/SaveLoadGame.cc
wvx: This looks like I missed some of the split for NPC save/load. The first define would be just for IMP save/load and then it was extended to include NPCs so that the scenario would more closely match the source story.

lynxlynxlynx:
I also now see it didn't handle not importing imps.

wvx:
> I also now see it didn't handle not importing imps.

5dd2417de57297deeeaf2bfcfca8aea3f230e41b fix attempt

src/game/SaveLoadGame.cc
wvx: 2cb2534ed77b72c0a845c9d59b2ba8b865fd689f ST::string attempt

ja2-derek: That commit LGTM 👍 . Thanks!

assets/externalized/game.json
```diff
@@ -115,6 +116,39 @@
...
+        //  Minimum first time progress to trigger event
+        //  vanilla settings: madlab: 35, mike: 50, iggy: 70
+        "event_madlab_min": 35,
+        "event_mike_min": 50,
+        "event_iggy_min": 70
+    },
+
+    //  Allowed unhired merc deaths at given difficulty level
+    //  vanilla settings: 1, 2, 3
+    "unhired_merc_deaths_difficulty_0": 1,
```
selaux, line 151: For starting cash we call them easy medium hard. Here it should be the same:

```
    "starting_cash_easy":   45000,
    "starting_cash_medium": 35000,
    "starting_cash_hard":   30000,
```

assets/externalized/game.json
```diff
@@ -115,6 +116,39 @@
...
+    "progress": {
+        //  Progress weight of parameter on campaign progress
+        //  vanilla settings: kills 25, control 25, income 50
+        "weight_kills": 25,
+        "weight_control": 25,
+        "weight_income": 50,
+
+        //  Progress point progression
+        //  vanilla settings: 0_easy 7, 1_medium 10, 2_hard 15
+        "kills_per_point_0": 7,
```
selaux, line 138: See below comment: easy, medium, hard

src/game/SaveLoadGame.cc
wvx: The ideal would be to use JSON to export/import to abstract from the compiler-struct memory format. This is not difficult. A text file would make it easier to edit for creating mod scenarios such as the intended "All you need is kill".

The data was placed in an unrelated field because it was an easy hack.

src/game/Laptop/IMP_Confirm.cc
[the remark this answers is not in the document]
wvx: The letter placement is not ideal and does not account for the possible case where all slots are filled. If you know how to drop the letter on the map then that would be a possible fix. Can items on the ground be destroyed?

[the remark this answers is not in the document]
lynxlynxlynx: I checked it and it's too early to fall back to dropping it to the ground. Haven't had much time to look at all this again yet — will integrate your fixes then, thanks.

lynxlynxlynx:
Cleaned up taking into account the comments. Only the IMP import commit changed and the last one fixes the json key names wrt difficulty. 
I suggest we defer the export format and full 90210 support for later

src/game/Laptop/IMP_Begin_Screen.cc
```diff
@@ -289,6 +300,16 @@ static void BtnIMPBeginScreenDoneCallback(GUI_BUTTON *btn, INT32 reason)
 			iCurrentImpPage = IMP_MAIN_PAGE;
 			fButtonPendingFlag = TRUE;
 		}
+		else if (GCM->getGamePolicy()->imp_load_saved_merc_by_nickname && IMPSavedProfileDoesFileExist(pNickNameString.c_str()))
+		{
+			fLoadingCharacterForPreviousImpProfile = true;
+			LaptopSaveInfo.iVoiceId = IMPSavedProfileLoadMercProfile(pNickNameString.c_str());
```
ja2-derek, line 306: We don't need the `.c_str()`. It accepts `ST::string`

src/game/Laptop/IMP_Confirm.cc
```diff
@@ -220,12 +222,24 @@ static void BtnIMPConfirmYes(GUI_BUTTON *btn, INT32 reason)
...
-		LaptopSaveInfo.fIMPCompletedFlag = TRUE;
+		LaptopSaveInfo.fIMPCompletedFlag = AddCharacterToPlayersTeam();
+		if (!LaptopSaveInfo.fIMPCompletedFlag) return; // only if merc hiring failed: no charge, give it another go
+
+		SOLDIERTYPE* const pSoldier = FindSoldierByProfileID(PLAYER_GENERATED_CHARACTER_ID + LaptopSaveInfo.iVoiceId);
+		if (!pSoldier) return;
+
+		if (fLoadingCharacterForPreviousImpProfile && gamepolicy(imp_load_keep_inventory))
+		{
+			IMPSavedProfileLoadInventory(gMercProfiles[PLAYER_GENERATED_CHARACTER_ID + LaptopSaveInfo.iVoiceId].zNickname.c_str(), pSoldier);
```
ja2-derek, line 233: This function also takes a `ST::string`, no need for `c_str()`

src/game/Laptop/IMP_Confirm.cc
```diff
@@ -220,12 +222,24 @@ static void BtnIMPConfirmYes(GUI_BUTTON *btn, INT32 reason)
...
+			// readd letter, since it just got wiped and almost certainly is not present in the import
```
ja2-derek, line 234: typo "readd"?

src/game/Tactical/Merc_Hiring.cc
```diff
@@ -56,6 +56,16 @@ extern BOOLEAN gfFirstHeliRun;
 // Saved in general saved game structure
 INT16 g_merc_arrive_sector = START_SECTOR;
 
+void CreateSpecialItem(SOLDIERTYPE* const s, UINT16 item)
+{
+	OBJECTTYPE o;
+	CreateItem(item, 100, &o);
```
ja2-derek, line 62: Should we use the enum ITEMDEFINE?

src/game/Laptop/IMP_Confirm.cc
lynxlynxlynx: Perhaps it should have a dash, re-add. I'll add it to be clearer.

src/game/Tactical/Merc_Hiring.cc
lynxlynxlynx: you mean for the param? Nothing uses it yet.

ja2-derek: I mean 100 can be replaced with the enum `LETTER` of from `enum ITEMDEFINE`

ja2-derek: I mean 100 can be replaced with the enum `LETTER` of from `enum ITEMDEFINE`

src/game/Laptop/IMP_Confirm.cc
ja2-derek: Ah! I see it now

src/game/Tactical/Merc_Hiring.cc
lynxlynxlynx: No, that's the status percent.

ja2-derek: :facepalm: I see

ja2-derek:
LGTM!

PR #1142, opened 2020-07-08
Initialize VFS at a single place from EngineOptions

selaux:
This is a proposal to move VFS initialization to a single place: The rust vfs module. Imho this makes it easier to grasp how the VFS is initialized compared to the VFS initialization at multiple places in C++ code using the order enum.

It also allows for SLF files in the mods directories.

The first two commits are cleanup commits that are kind of unrelated.

### DIFF
--- a/rust/stracciatella/Cargo.toml
+++ b/rust/stracciatella/Cargo.toml
@@ -12,7 +12,7 @@ path = "src/stracciatella.rs"
 
 [[bin]]
 name = "ja2-resource-pack"
-path = "src/bin_resource_pack.rs"
+path = "src/bin/resource_pack.rs"
 required-features = ["bin"]
 
 [dependencies]
--- a/rust/stracciatella/src/bin/resource_pack.rs
+++ b/rust/stracciatella/src/bin/resource_pack.rs
@@ -11,10 +11,6 @@
 //! ```
 //!
 
-mod file_formats;
-mod res;
-mod unicode;
-
 use std::fmt::Debug;
 use std::fs;
 use std::path::Path;
--- a/rust/stracciatella/src/file_formats/mod.rs
+++ b/rust/stracciatella/src/file_formats/mod.rs
@@ -1,3 +1,57 @@
 //! This module contains code to read and write fileformats that are used by Jagged Alliance 2
 
+use std::io::ErrorKind::{InvalidData, InvalidInput};
+use std::io::{Error, Read, Result, Write};
+
 pub mod slf;
+
+/// Trait that adds extra functions to Read.
+pub trait StracciatellaReadExt: Read {
+    /// Reads and discards unused bytes.
+    fn read_unused(&mut self, num_bytes: usize) -> Result<()> {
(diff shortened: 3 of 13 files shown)


ja2-derek:
I think moving it to Rust makes it less accessible to changes. I believe most of us are more comfortable with C/C++ than Rust. Though It is great, if this is not going to change frequently.  Not sure if Flavio has more in mind about VFS (e.g. make VFS configurable like 1.13 maybe?)

Looking at the changes, does it now support SLF files in mod directories? And does it include the `~/.ja2/data` [user home directory](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1126/commits/30ca8737ac0fb277e6d5f7a2a674b4c8d84799c0#diff-a601ac1290ef97e8e5b1a9b583d3a7c7R227) for customized config?

lynxlynxlynx:
I don't mind having it in a single place.

selaux:
I don't mind either way. I just thought this is easier to grasp on first read. Also I wanted to work a little on some Rust code :wink:. I think I forgot about `~/.ja2/data`, it is not included yet.

Yes support for SLF in mods is now there. I think this is important as mods should not be treated differently then the original data folder (e.g. copying some resources from other "versions" like Wildfire should just work, without the need to extract anything. Although i dont think that new mods will use SLF (as it is harder to produce than just having all the individual files in the mods directory.

Let me know if you want me to finish the PR (adding `~/.ja2/data`) or if this should just be closed.

ja2-derek:
> Yes support for SLF in mods is now there. I think this is important as mods should not be treated differently then the original data folder (e.g. copying some resources from other "versions" like Wildfire should just work, without the need to extract anything. Although i dont think that new mods will use SLF (as it is harder to produce than just having all the individual files in the mods directory.

I prefer SLFs for mod releases because of the overhead of opening files on Windows. On Windows, opening only several files rather than hundreds should make quite a difference in load speed.

selaux:
Generally when creating the PR I also wondered about performance impacts of the VFS. Currently we do a lot of disk access when opening a single file. E.g. for a file in a vanilla SLF, we acess ~3 other locations first and check if a file exists there. With mods it can get even worse. I can imagine for systems with a hard disk that is not good at random accesses, this might be quite slow, but for my system with an SSD I didn't notice any impact. Not sure how we can test for these kinds of possible regressions.

lynxlynxlynx:
I doubt it matters, probably other things dominate our performance (eg. unaccelerated drawing).

ja2-derek:
It does have an effect in performance on Windows. I believe it mainly affects game init, game save/load and enter sectors.

You can see in the profiling report below, `vfs::VfsFile_Open` and `vfs::dir::DirFs::open` are some of the more expensive functions.

image

lynxlynxlynx:
So, can we finish this and merge?

selaux:
Rebased and added home directory.

lynxlynxlynx:
There are merge conflicts now.

PR #1145, opened 2020-07-10
Drassen to be consistent with other mines before depletion

### LINKED ISSUE #1144: Drassen mine doubles production before depletion

### What is wrong
Details, details, details.

|   master | Steam version  |    yes?  |    N/A  |

## Background

At the start of every campaign, one mine is chosen for depletion after 10 days of mining. If Drassen is chosen, it should have 20 days of production before depletion.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/f288e5da14768357a5fcf8e4b83ced1140919e33/src/game/Strategic/Strategic_Mines.cc#L135-L142

The head miner warns about depletion at after about 7.5 days (or 15 days for Drassen):

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/f288e5da14768357a5fcf8e4b83ced1140919e33/src/game/Strategic/Strategic_Mines.cc#L147-L148

The mine extraction rate is updated when the mine is close to depletion:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/f288e5da14768357a5fcf8e4b83ced1140919e33/src/game/Strategic/Strategic_Mines.cc#L341-L342

For all other mines, the production rate will go up slightly by 250 per day, and will run out after ~2 days. If it is Drassen, it will double the production rate and run out after ~2 days.

Here is my question - is this a bug? It is awkward to me to double the production rate all in a sudden, and that the mine will actually run out at around day 17 instead of day 20. 

A more correct way, to me, is to set the `uiOreRunningOutPoint` of Drassen to 12.5%, so it start the depletion sequence near day 17.5 instead. Doing so the production rate depletion handling is consistent with other mines, and it will actually last 20 days as promised. But that is a deviation from Vanilla. 

What do you think?

ja2-derek:
Fixes #1144. 

This is about the Drassen mine, if selected for depletion, being inconsistent with other mines. With this PR any depleting mine will always run out in ~2.5 days after head miner's warning (assuming 100% production).

### Before this PR

Drassen head miner warns on depletion at around the 15th day, then it doubles the production rate, and runs out in about 2.5 days.

If any other mines are selected for depletion, the head miner warns after ~7.5 days. Production rate stays similar, and mine runs out in about 2.5 days.

### After this PR

Drassen head miner to warn after ~17.5 days. The production rate will stay about the same, and the mine runs out in another ~2.5 days. Total ore supply is the same. This behaviour is now consistent with other mines.

Depletion behaviours of other mines are unchanged.

### DIFF
--- a/src/game/Strategic/Strategic_Mines.cc
+++ b/src/game/Strategic/Strategic_Mines.cc
@@ -39,6 +39,7 @@
 #define MINE_PRODUCTION_NUMBER_OF_PERIODS 4						// how many times a day mine production is processed
 #define MINE_PRODUCTION_START_TIME				(9 * 60)		// hour of first daily mine production event (in minutes)
 #define MINE_PRODUCTION_PERIOD						(3 * 60)		// time seperating daily mine production events (in minutes)
+#define MINE_PRODUCTION_DAYS_AFTER_HEAD_MINER_WARNING 2.5					// how many more days the mine can still produce after head miner warns of depletion
 
 
 // this table holds mine values that change during the course of the game and must be saved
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

PR #1148, opened 2020-07-11
No traversal from I3 to I2

### LINKED ISSUE #492: Runtime error in sector I2
I'm getting "runtime error" ... load the save, go to map, choose sector I2 and compress the time to reproduce it.
[QSI2.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/758764/QSI2.zip)

ja2-derek:
Fixes #492, as described in https://github.com/ja2-stracciatella/ja2-stracciatella/issues/492#issuecomment-306898896

`G` is ground barrier, which disallows tactical traversal.


lynxlynxlynx:
We can fix the other pair when someone hits it.

ja2-derek:
Do you know where it is? I tested some other sectors around the edge but couldn't find anything wrong.

lynxlynxlynx:
No, it'd require some digging. Likely easiest to look at which minimaps are missing and then check their neighbours.

PR #1156, opened 2020-07-16
set version to 0.17.0

lynxlynxlynx:
my testing went fine as well :)
this also includes the tag


ja2-derek:
Should this not be merged to the `0.17.x` branch?

Master should be `0.18.x-dev`

lynxlynxlynx:
If we think we need it. The version would otherwise be bumped right after packaging.

ja2-derek:
True. We can create the branch if there is a worthy bug in 0.17.

Otherwise we can go straight for 0.18.0 instead of maintaining a "bug-fix" branch.

PR #1157, opened 2020-07-17
Fix upload dirs for releases

selaux:
It seams that CI builds dont upload into the correct directory (they use the directory name as file name).

Second issue is that travis does not seem to run for tagged releases. I am not sure why.


ja2-derek:
https://docs.travis-ci.com/user/customizing-the-build#building-specific-branches

> Note that safelisting also prevents tagged commits from being built. If you consistently tag your builds in the format v1.3 you can safelist them all with regular expressions, for example /^v\d+\.\d+(\.\d+)?(-\S*)?$/.

selaux:
Thanks. I will add it to this PR.

lynxlynxlynx:
LGTM

PR #1106, opened 2020-06-07
Externalizing Creature Lairs and Spreading

ja2-derek:
Fixes #1087.

See also #1095 and #665.

## Summary

- Extracted JSON data from Creature_Spreading.cc (`strategic-map-creature-lairs.json`)
- Removed `isInfestible` from Mine. It is redundant. The existence of a CreatureLairModel means the same.

### Model fields data

_All data come from Creature_Spreading.cc_

- InitLair`${TOWN}` => `"sectors"`
- InitCreatureQuest: `curr->uiFlags |= SF_PENDING_ALTERNATE_MAP;` => `"entranceSector"`
- ChooseTownSectorToAttack => `"attackSectors"`
- MineClearOfMonsters => MineModel::`mineSectors` (#1098)
- GetWarpOutOfMineCodes => `"warpExit"`

## Does not cover

 - Hard-coded Tixa dungeon references (in StrategicMap.cc, etc)

PR #1159, opened 2020-07-19
Fix creature town attack

ja2-derek:
Found bugs when investigating a [bug report](http://thepit.ja-galaxy-forum.com/index.php?t=msg&goto=360622&#msg_360622 ).

 - a67c898e6f54dad93d041acfe2051882f776ed35: `gubSectorIDOfCreatureAttack` is not correctly updated, which should mean all attacks will happen at the mine entrance sector. (Bug since #1106)
- fd2b3bea9431109329d7e779105c6898e6e68be1: `giLairID` never got reset during `ReStartingGame`. This seems to be an old issue (v0.16 or earlier). Loading a game with an active lair, then starting a new campaign game cause the next game to have the wrong `giLairID`, and this wrong ID will get saved.

Needs more testing, and further investigation to see if it is related to the reported bug.

### DIFF
--- a/src/game/Strategic/Creature_Spreading.cc
+++ b/src/game/Strategic/Creature_Spreading.cc
@@ -640,14 +640,6 @@ void DeleteCreatureDirectives()
 	giLairID = 0;
 }
 
-void ClearCreatureQuest()
-{
-	//This will remove all of the underground sector information and reinitialize it.
-	//The only part that doesn't get added are the queen's lair.
-	BuildUndergroundSectorInfoList();
-	DeleteCreatureDirectives();
-}
-
 void EndCreatureQuest()
 {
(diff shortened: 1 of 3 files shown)


ja2-derek:
Tested the commits I made so far, and they should work as intended.

I don't have further findings on the bug report. We need further input from the reporter.

ja2-derek:
I don't have anything to add for this. This PR is ready for review and merge.

I will continue investigation (and probably new PR) once the forum bug report has more updates.

lynxlynxlynx:
Sent him a PM reminder, let's see ...

ja2-derek:
Maybe should we merge the bug-fixes in this PR and go ahead with 0.17.0?

lynxlynxlynx:
Agreed!

PR #1162, opened 2020-07-23
Ignore invalid items when reading files

ja2-derek:
Fixes https://github.com/ja2-stracciatella/mod-nightops-maps/issues/5

Changes `ExtractObject`, which reads `OBJECTTYPE` from saves and maps. 

For example, NightOps maps H13 contains a few items not defined in JA2:S. The principle here is we should be more defensive reading files, as we now have mods. On reading an invalid item index, instead of de-referencing null pointer, the game should issue a warning and continue to load rest of the file.

### DIFF
--- a/src/game/Tactical/LoadSaveObjectType.cc
+++ b/src/game/Tactical/LoadSaveObjectType.cc
@@ -4,6 +4,7 @@
 
 #include "ContentManager.h"
 #include "GameInstance.h"
+#include "ItemModel.h"
 
 
 void ExtractObject(DataReader& d, OBJECTTYPE* const o)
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Is this safe? As in the partial IC_NONE items don't cause problems later?

ja2-derek:
Loaded that map fine, but let me test it further.

ja2-derek:
Changed the implementation - now the item is replaced by a `NONE`, instead of a partial object with invalid ID.

This function is used in used to load saves and maps. Saves by JA2:S itself should never have invalid item. Soldier inventory in maps also uses this - and `NONE` object is a normal case for inventory slots.

lynxlynxlynx:
Thanks!

PR #1170, opened 2020-08-03
docs/Release-checklist.md: converted to github checklist

lynxlynxlynx:
merged with the one for 0.17 and cleaned up


ja2-derek:
LGTM.

BTW - do we maintain/support the Arch Linux [package](https://aur.archlinux.org/packages/ja2-stracciatella-git/) in any way, @selaux?

lynxlynxlynx:
I don't think so, we've just prodded them in the past (AUR is a community repo).

PR #1172, opened 2020-08-04
Allow inconsistency in minor map version

ja2-derek:
This is for https://github.com/ja2-stracciatella/mod-nightops-maps/issues/7. The map crashes without this PR.

There is a check in World Def loading, that `gMapInformation.ubMapVersion` must be equal to `ubMinorMapVersion`. Both values come from the map .DAT file, but somehow they are inconsistent in certain NightOps maps. My guess is that the map author did some manual/hex editing when adapting the maps for vanilla. 

I changed the throw-exception into a `SLOGW`. With this PR, the sectors load fine and I completed the entire game.

### DIFF
--- a/src/game/TileEngine/WorldDef.cc
+++ b/src/game/TileEngine/WorldDef.cc
@@ -2378,7 +2378,7 @@ try
 
 	if (dMajorMapVersion >= 4.00 && gMapInformation.ubMapVersion != ubMinorMapVersion)
 	{
-		throw new std::runtime_error("map version must match minor version");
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Why do we check in the first place? Is there some possible format difference that would make this problematic and we should have a narrower check?

ja2-derek:
I need some time to dig into this...

ja2-derek:
Haven't got the time to finish this. We need to step through the code and might also need to look at the 1.13 map editor.

Or we might try to re-export that Night-Ops map with the editor or a map convertor.

PR #1174, opened 2020-08-04
No strategic movements during bloodcat ambush

### LINKED ISSUE #412: Player can compress time or move the squad while being ambushed by Bloodcats
bloodcat-time
Dimitri was doing his Duty as a mule, bringing Guns to Tony, when he was ambushed by Bloodcats.
I found out, that I was able to compress time while the ambush was still going.

This seems like a bug, because there is clearly some kind of combat going on in the sector and the player usually can't compress time during combat.

Savegame:
[QuickSave.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/513318/QuickSave.zip)

ja2-derek:
Fixes #412.

This PR disallows changing sector, time compression and strategic squad movement during a bloodcat ambush, as if sector is hostile.

Before this PR, the hostile sector check only checks for :
- `INCOMBAT` in gTacticalStatus.uiFlags
- gTacticalStatus.`fEnemyInSector`
- `InAirRaid` _(to be cleaned up)_

Bloodcats in the wild are neutral before they are spotted, similar to hostile civilians. So during an ambush but before seeing a bloodcat, player could still do strategic actions, as combat (turn-based) mode is off and there are no enemies in sector.

Bloodcat ambush is checked by ` gubEnemyEncounterCode == BLOODCAT_AMBUSH_CODE && HostileBloodcatsPresent()`. The check is specific to ambushes. All other kinds of bloodcat encounters are not covered by this check.

### DIFF
--- a/src/game/Strategic/Map_Screen_Interface.cc
+++ b/src/game/Strategic/Map_Screen_Interface.cc
@@ -3880,6 +3880,8 @@ static MoveError CanCharacterMoveInStrategic(SOLDIERTYPE& s)
 			if (gTacticalStatus.uiFlags & INCOMBAT) return ME_COMBAT;
 			// Hostile sector?
 			if (gTacticalStatus.fEnemyInSector) return ME_ENEMY;
+			// Bloodcat ambush?
+			if (gubEnemyEncounterCode == BLOODCAT_AMBUSH_CODE && HostileBloodcatsPresent()) return ME_COMBAT;
 			// Air raid in loaded sector where character is?
 			if (InAirRaid()) return ME_AIR_RAID;
 		}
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
LGTM. HostileBloodcatsPresent doesn't actually check for hostility, but that's an old decision. So I couldn't think of a way to trigger this in bad way. :)

ja2-derek:
Oh right. This PR actually relies on that it doesn't care about hostility. Maybe we should rename the function.

ja2-derek:
This introduced a new compiler warning:
```
ja2-stracciatella\src\game\Strategic\Map_Screen_Interface_Bottom.cc(1153,173): warning C4551: function call missing argument list
```

I will address this in #1127 along with other compiler warnings and coverity defects.

PR #1175, opened 2020-08-05
Improve Pixel-Perfect scaling usability

### LINKED ISSUE #1169: Scaling usability improvements
Agreed. We can think about:

- Disallow / warn against resolution larger than the desktop
- Pixel-perfect scaling to have a minimum scaling of 1, so we display a partial screen rather than a black screen

_Originally posted by @ja2-derek in https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1165#issuecomment-667726647_

ja2-derek:
Resolves #1169.

In Pixel-Perfect mode:
1.  Throws can error if the desktop is too small for the requested game resolution. Pixel-perfect mode cannot scale down. It will result in an empty black screen.
2. Limits the minimize size in Windowed mode. Resizing the window to anything smaller will result in an empty black screen.

### DIFF
--- a/src/sgp/Video.cc
+++ b/src/sgp/Video.cc
@@ -97,6 +97,20 @@ static VideoScaleQuality ScaleQuality = VideoScaleQuality::LINEAR;
 static void RecreateBackBuffer();
 static void DeletePrimaryVideoSurfaces(void);
 
+// returns if desktop resolution larger game resolution
+BOOLEAN IsDesktopLargeEnough()
+{
+	SDL_DisplayMode dm;
+	if (SDL_GetDesktopDisplayMode(0, &dm) == 0)
+	{
+		if (dm.w < SCREEN_WIDTH || dm.h < SCREEN_HEIGHT)
+		{
+			return false;
+		}
+	}
+	return true;
+}
+
 void VideoSetFullScreen(const BOOLEAN enable)
 {
 	if (enable)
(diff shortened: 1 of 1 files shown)


src/sgp/Video.cc
```diff
@@ -97,6 +97,20 @@ static VideoScaleQuality ScaleQuality = VideoScaleQuality::LINEAR;
...
+BOOLEAN IsDesktopLargeEnough()
+{
+	SDL_DisplayMode dm;
+	if (SDL_GetDesktopDisplayMode(0, &dm) == 0)
+	{
+		if (dm.w < SCREEN_WIDTH || dm.h < SCREEN_HEIGHT)
+		{
+			return false;
+		}
+	}
```
lynxlynxlynx, line 111: Shouldn't it also return false in the else case?

derek9gag: `SDL_GetDesktopDisplayMode` returning non-zero means we do not know about the desktop resolution, so it could be anything. 

I opt for letting it try as it is not (yet) a critical error. It may still run into other problems anyway.

ja2-derek: `SDL_GetDesktopDisplayMode` returning non-zero means we do not know about the desktop resolution, so it could be anything.

I opt for letting it try as it is not (yet) a critical error. It may still run into other problems anyway.

PR #1178, opened 2020-08-06
wvx's Accurate Aim Circles

ja2-derek:
I am cherry-picking this mod from @wvx (604a66320afa17ba6b3dae97a467749d41f4493e), because I can no longer properly play the game without it ;)

I understand that this mod is not yet 100% complete. It does not work for melee attacks (https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1002#discussion_r437472430) and may have some edge case bugs (https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1002#discussion_r437472564). From my own experience in game, it is already a great improvement over vanilla.

I think it is good to merge it in, and we can always continue fixing the remaining issues later on. Let me know your thoughts.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -23,6 +23,10 @@
     //  vanilla setting: false
     "middle_mouse_look": true,
 
+    //  Size of aim cricle reflects chance to hit letting the player know
+    //  when enough time units have been spent and which shots are unlikely.
+    "accurate_aim_circle": false,
+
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
We try to be a stable product, so let's not introduce bugs on purpose. Why not fix the problems first?

ja2-derek:
This is going to be an opt-in feature. But I see your point. Let's this re-visit this later.

PR #1179, opened 2020-08-06
Adding the monster spit effect to weapons.json

ja2-derek:
Externalizing the monster-spit to gas-effect mapping, so the code do not have a dependency on the MONSTER_SPIT pseudo-weapons. Some mods (I mean Unfinished Business) re-purpose the monster spits' item ID to be actual weapons.

The new field `usSmokeEffect` is an item index, mandatory for monster "weapons". Values should be one of the CREATURE_GAS items:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/6f2ec3397094cdac5cd3e7efaeb4304831029113/src/game/Tactical/Items.cc#L96-L98

### DIFF
--- a/assets/externalized/weapons.json
+++ b/assets/externalized/weapons.json
@@ -1735,7 +1735,8 @@
         "Sound": "sounds/adult spit.wav",
         "bNotBuyable": true,
         "bNotEditor": true,
-        "bDefaultUndroppable": true
+        "bDefaultUndroppable": true,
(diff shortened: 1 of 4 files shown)


selaux:
Looks good. It might be nice to extract all bullet related stuff into a separate json in the future (model, animation, sounds, smoke, physics related stuff). We might also be able to put it into some existing ones, e.g. `calibres.json`.

PR #1180, opened 2020-08-06
Reduce noise in logs

ja2-derek:
Reducing the log levels of those that does not affect anything, or those that players do not need to know:

1. `Adding directory to VFS` - unless developing mods or changing VFS, I don't think players need to know about this
2. `Skipped adding to VFS` - the caller has already requested to skip errors, so loading this is totally optional
3. `Can't compress time, pause state locked (reason %d). OK unless permanent...` - this happens when you press `+` during a message popup
4. `Schedule callback occurred outside of tactical -- Auto processing!` - this is a normal case, nothing to worry (https://github.com/ja2-stracciatella/ja2-stracciatella/commit/34f35318764671a121239543b9e86554ded499aa#diff-7715cba6d43e93a9ffead03797722f0bL201-L203)
5. `No code support for NPC action` - several NPCs at San Mona work fine despite of this message. This looks like normal case to me.

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -198,12 +198,12 @@ void DefaultContentManager::AddVFSLayer(VFS_ORDER order, const ST::string path, 
 	ST::string error = ST::null;
 	if (Fs_isDir(path.c_str()))
 	{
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Looks reasonable, thanks.

PR #1181, opened 2020-08-07
Externalize SAM air control table

ja2-derek:
Externalizes the SAM site air control table (`ubSAMControlledSectors`). Part of #665.

The JSON file is copy of the original data in code, which uses `SAM site ID + 1` to indicate SAM sites, and `0` to indicate free sectors. Unfortunately the convention of indicating the SAM site location with `01` `02` does not work with JSON.

This was tested with taking over SAM sites in a fresh game, and flying Heli over in red zones.


selaux:
Looks good. Thanks.

lynxlynxlynx:
Hmm, a potential upgrade is to store them as powers of two, not consecutive numbers, so there could be sectors covered by more than one SAM.

PR #1182, opened 2020-08-07
Externalize sector Loading Screens

ja2-derek:
At the high level, there are 2 ways to determine the loading screen for a sector:

1. Specific mapping from sector to loading screen (e.g. P3 => `DAYPALACE`,  A2 => `NIGHTCHITZENA`)
2. If no mapping found, there are rules for land types, SAM sites, underground caves, etc

### What's included

This PR externalizes part 1 of the loading screen logic. There are 2 kinds of data externalized:

1. Loading screen details, which has an index (`LoadingScreenID`), and points to a resource file (.STI) 
2. Mapping from `sector + night/day` to `LoadingScreenID`

With this, mods can freely add new loading screens and then maps specific sectors to them. Though it is not possible to change the rules via JSON, the mods can resort to adding a mapping for every single sector.

Part of #665.

### DIFF
--- a/assets/externalized/loading-screens-mapping.json
+++ b/assets/externalized/loading-screens-mapping.json
@@ -1,0 +1,45 @@
+/**
+ * This defines the loading screen to use for specific sectors. If the sector loading screen is not defined here, the game falls back to rules to determine loading screen.
+ *
+ * Fields:
+ *   - sector:      The short name of the sector, e.g. A9, B10
+ *   - sectorLevel: The Z-level if sector is underground. Defaults to 0, meaning that the sector is on the surface
+ *   - night:       The internal name of the loading screen to use during night time (see loading-screens.json)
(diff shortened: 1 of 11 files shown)


ja2-derek:
I am still looking at the failed check but I don't understand why it happens. It is consistently crashing at the same test even when it is re-run.

All tests work fine on the GitHub CI Linux and Mac builds, and also on my VC++ environment.

ja2-derek:
> 
> 
> I am still looking at the failed check but I don't understand why it happens. It is consistently crashing at the same test even when it is re-run.
> 
> All tests work fine on the GitHub CI Linux and Mac builds, and also on my VC++ environment.

Fixed. It is a unit test only problem. The ContentManager in tests tried to `delete` an un-initialized pointer, i.e. one with garbage value.

selaux:
:+1: . Thanks.

PR #1185, opened 2020-08-16
Externalize (arms) dealer info

ja2-derek:
Externalizes (#665) `ARMS_DEALER_INFO`, which is the info for all bartenders/traders/repairmen in game. Dealer inventory are already externalized, and this PR externalizes the dealers themselves.

## Summary 

The first few commits are method extraction that should not change any functionalities.

The next commits are adding the JSON data, adding model classes, and replacing `ARMS_DEALER_INFO&` with `const DealerModel*`. New dealer flags are added but not yet used in this PR.

## What's next

This PR avoids any non-essential changes, as it already has many lines changed. I intend to make further PRs to generalize some hardcode logic (e.g. repairmen check, electronics repairer (Fredo) check, and so on), and maybe to relax the limit on number of dealers.

I will do a full round of testing after the follow-up PRs.

### DIFF
--- a/src/game/Tactical/ShopKeeper_Interface.cc
+++ b/src/game/Tactical/ShopKeeper_Interface.cc
@@ -1910,11 +1910,8 @@ static void DisplayArmsDealerCurrentInventoryPage(void)
 				//if the dealer repairs
 				else if (DoesDealerDoRepairs(gbSelectedArmsDealerID))
 				{
-					//if the item is damaged
-					if (gpTempDealersInventory[usCnt].uiFlags & ARMS_INV_ITEM_NOT_REPAIRED_YET)
-						fDisplayHatchOnItem = TRUE;
-					else
-						fDisplayHatchOnItem = FALSE;
+					// not displaying the hatch, because it would obscure the repair ETA
+					fDisplayHatchOnItem = FALSE;
 				}
 				else // non-repairman
 				{
--- a/src/game/Tactical/ShopKeeper_Interface.cc
+++ b/src/game/Tactical/ShopKeeper_Interface.cc
@@ -62,6 +62,7 @@
 #include "UILayout.h"
 #include "CalibreModel.h"
 #include "ContentManager.h"
+#include "DealerModel.h"
 #include "GameInstance.h"
 #include "MagazineModel.h"
 #include "WeaponModels.h"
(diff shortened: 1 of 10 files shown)


src/game/Tactical/ShopKeeper_Interface.cc
```diff
@@ -1896,10 +1908,10 @@ static void DisplayArmsDealerCurrentInventoryPage(void)
 					fDisplayHatchOnItem = TRUE;
 				}
 				//if the dealer repairs
-				else if( ArmsDealerInfo[ gbSelectedArmsDealerID ].ubTypeOfArmsDealer == ARMS_DEALER_REPAIRS )
+				else if (DoesDealerDoRepairs(gbSelectedArmsDealerID))
 				{
 					//if the item is damaged
-					if( ArmsDealerInfo[ gbSelectedArmsDealerID ].uiFlags & ARMS_INV_ITEM_NOT_REPAIRED_YET )
+					if (gpTempDealersInventory[usCnt].uiFlags & ARMS_INV_ITEM_NOT_REPAIRED_YET)
```
ja2-derek, line 1914: This has to be a bug. The original code was looking for a `ARMS_INV_ITEM` flag in the `ARMS_DEALER_INFO`.

ja2-derek:
Not sure why the tests is failing only on Mac. I am still looking for clues.

src/game/Tactical/ShopKeeper_Interface.cc
lynxlynxlynx: However the effect was that it was always false. Shouldn't this check be negated? As in, only hatch items that have been repaired already?

lynxlynxlynx:
Other than the comment, LGTM! :+1:

src/game/Tactical/ShopKeeper_Interface.cc
ja2-derek: Indeed... the hatch seems totally useless. 
 1. The repaired item are placed in the bottom section
 2. The repair ETA cannot be read at all with hatch  

See the below screenshot:
image

I think we should just remove the check and leave `fDisplayHatchOnItem` as FALSE.

lynxlynxlynx: Yes, let's remove — the red text is enough.

PR #1186, opened 2020-08-16
Fix warp exit from mine

ja2-derek:
Warp exit is not working in some cases (since 0.17.0). After finishing the quest, but before leaving the lair, save and load the game. It no longer prompts for "Travel to surface"

When the creature quest is finished, `giLairID` becomes -1. On the next game load, `gLairModel` becomes NULL, and we must rely on `giDestroyedLairID` instead.


lynxlynxlynx:
Lgtm!

PR #1188, opened 2020-08-18
Feature: Add chance-to-hit hotkey

dariusk:
I decided to add this for myself and a friend, and thought I would at least offer the work up to the community. The modification is simple:

When the player hovers their cursor over a grid square, pressing ~'q'~ `f` will print the range to hit message and then a message displaying the base chance to hit (before additional aiming), and the chance to hit after cover is accounted for (before additional aiming). If there is no soldier present, it will assume you're shooting in the air above the ground level, in the dead middle of the cursor cube. If there is a soldier present, then the soldier's stance will be taken into account for the cover modifer calculation.

~I chose "q" because it was the only unbound alphanumeric hotkey I could find.~

image

## Potential issues

Some potential issues with this PR, that I am willing to fix up if people can point me in the right direction:

* ~I'm not a C++ programmer. My understanding is that `const_cast` is a dangerous function and frowned upon but I couldn't figure out another way to, well, cast the constant in order to call `CalcChanceToHitGun()`.~ (addressed in the conversation below)
* I'm fairly certain I got the calculations right, at least such as they are consistent with how it's done in-game. Chance to hit is calculated, then multiplied by cover modifier, as seen in the [ubChanceToReallyHit](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c3e9284a195ef38d774ecb4b3c0cf601576cb996/src/game/TacticalAI/Attacks.cc#L255) calculation for the AI. But it's possible I got them wrong! I tested a few different scenarios and they all seemed to make intuitive sense based on the numbers I'm showing, so I think it's okay, but open to suggestions/corrections.


ja2-derek:
First of all thank you for the PR!

> I chose "q" because it was the only unbound alphanumeric hotkey I could find.

Or can we expand the `f` key output? I think they serve very similar purposes. And this should be a great complement to the accurate aim circle. What do you think, @lynxlynxlynx ?

> * I'm not a C++ programmer. My understanding is that `const_cast` is a dangerous function and frowned upon but I couldn't figure out another way to, well, cast the constant in order to call `CalcChanceToHitGun()`.

Maybe just use `SOLDIERTYPE* s` in the function arg list? I know that the variable is not supposed to be altered but this is what all other functions do.

Also, if you must add to `_EnglishText.cc`, you will also need to add to all other languages.

src/game/Tactical/DisplayCover.cc
```diff
@@ -435,6 +435,45 @@ void DisplayRangeToTarget(SOLDIERTYPE const* const s, INT16 const sTargetGridNo)
...
+
+	SOLDIERTYPE* pSoldier = const_cast<SOLDIERTYPE*>(s);
+
+	// Get the range to the target location
+	UINT16 const usRange = GetRangeInCellCoordsFromGridNoDiff(s->sGridNo, sTargetGridNo) / 10;
+
+	if (WeaponInHand(s))
+	{
+		// Get the chance to hit
+		UINT32 const uiHitChance = CalcChanceToHitGun(pSoldier, sTargetGridNo, pSoldier->bAimTime, pSoldier->bAimShotLocation, true );
```
ja2-derek, line 451: `fModify`, the last arg to `CalcChanceToHitGun`, should be `false` unless you are actually firing the gun. Passing `true` means to modify the status (wearing off) the weapon and attachments.

dariusk: Got it, will change.

dariusk:
Great! I'll add the localized text for the languages where applicable (some of the languages currently just use the English text enum but some like German have localized text so I'll add it there as best I can Google Translate).

I think having `f` print both the range and the CTH is a great call, I'll try having it print two log lines. I'll try your solution for the const-casting too.

dariusk:
Okay, I've:

* added Russian and German translations (other languages simply use the English for these actions)
* bound the action to `f` so now (if you're holding a ranged weapon) it shows the range info on one line, then CTH on the next (see screenshot below)
* fixed the issue where checking the CTH caused wear and tear

image

lynxlynxlynx:
I don't have a checkout here, but it looks mostly good. :+1:  The best solution to the const problem is to constify more of the code, not less, though.

dariusk:
Re: consts, unfortunately the new code path goes through a function that could in theory modify the soldier in question, so I think removing the const makes the most sense.

ja2-derek:
Changes LGTM

lynxlynxlynx:
Thanks!

PR #1190, opened 2020-08-19
Fixing compiler and coverity warnings

ja2-derek:
Clearing some Coverity and compiler warnings. There should be no change in functionality. More details in the commit messages.


lynxlynxlynx:
Nice!

PR #1191, opened 2020-08-19
Unused items slots should be IC_NONE

ja2-derek:
The editor displays the "unused", "NOWEAPON", "CLIP_NOTHING" at the end of the item display.  Those are item slots reserved for weapons or magazines, but not used. 

In Vanilla they are of class `IC_NONE` and hence ignored by the editor. In Stracc, it was changed during the externalization and they are currently of class `IC_PUNCH` and `IC_AMMO` (see https://github.com/ja2-stracciatella/ja2-stracciatella/commit/1aa7ef5e6ae2d40ed88e87910555f225fabf9d23).

This PR changes those "items" back to `IC_NONE`, so they will be ignored by the editor.

### DIFF
--- a/src/externalized/MagazineModel.cc
+++ b/src/externalized/MagazineModel.cc
@@ -6,12 +6,13 @@
 
 MagazineModel::MagazineModel(uint16_t itemIndex_,
 				const char* internalName_,
+				uint32_t itemClass_,
 				const CalibreModel *calibre_,
 				uint16_t capacity_,
 				const AmmoTypeModel *ammoType_,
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
LGTM ... this doesn't affect punching or throwing rocks, right?

ja2-derek:
> LGTM ... this doesn't affect punching or throwing rocks, right?

Indeed it affect punching. Punching is done with the `Nothing` item  (also a `NoWeapon` item). I have got the fix for this, but now I got stuck with #1192.

Issue #1198, opened 2020-08-25
Patches from Unfinished Business


ja2-derek:
Here is my proposal to cherry-pick some bug fixes and minor improvements from Unfinished Business source code (#254). My selection criteria are 1) the intention is clear, 2) can be tested and verified, and 3) makes sense in vanilla.

### PATCH152: Notify of "someone else in sector" after battle

Vanilla behaviour: If your squad enters a sector via Strategic Movement with an important NPC (e.g. Skyrider), there is a pop-up saying your squad noticed somebody. But there is no notification at all if there happens to be enemies in the sector.

In UB ([PATCH152](https://github.com/ja2-derek/ja2-ub-comparison/commit/7eb9ea13)), the pop-up appears in Tactical after your win the battle.

### PATCH200: Light up soldier body if fake light is off

When "fake light" is turned off, the soldiers are supposed to have a [personal lighting](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/7bdf5d3f7a7e8787fe3f68a3b5e4e9ae4b04d22f/src/game/Tactical/Soldier_Control.cc#L9071-L9077). But it does not really work in Vanilla, the lighting will be gone once you move the soldier:
![merc_lighting_van](https://user-images.githubusercontent.com/63151803/91233726-670ef200-e764-11ea-9bb3-9af33103603b.png)

Unfinished business fixes the personal lighting ([PATCH200](https://github.com/ja2-derek/ja2-ub-comparison/commit/56c4eb43)):
![merc_lighting_ub](https://user-images.githubusercontent.com/63151803/91233725-65ddc500-e764-11ea-9250-0410d281c466.png)

### PATCH208: Disarming trap when inventory is full

Vanilla behaviour: Disarming a pre-placed bomb trap (e.g. in Cambria SAM site) results in a mine, detonator or something similar. But if the merc's inventory is full, the object creation is cancelled, and there is an "Action Item" placed in the cursor.

This is fixed in UB ([PATCH208](https://github.com/ja2-derek/ja2-ub-comparison/commit/a464eaca)), the created object (mine or detonator) will be placed in the cursor.

### PATCH029: Disallow breaklights on rooftop

Vanilla behaviour: If you throw a breaklight on the rooftop, it either lights up the ground (at the same gridNo) or have no effects at all - and the breaklight is gone.

[...]

PR #1200, opened 2020-08-27
Notify the player of important NPC in sector

ja2-derek:
Ported from Unfinished Business ([PATCH152](https://github.com/ja2-derek/ja2-ub-comparison/commit/7eb9ea13)). See also #1198.

If you travel to or going through a sector with an NPC (e.g. Skyrider), there is a prompt saying something like "Squad 1 noticed someone in sector A1". But if the sector has an enemy presence, then you will be going to battle instead, without any prompt about the NPC. This PR adds a prompt in tactical after winning a battle, that says "There is someone else in sector...".

This feature is more important in UB because there are always enemies in the sector where RPC `MANUEL` is.

## Changes

- Added new message string (English and [German](https://github.com/ja2-derek/ja2-ub-comparison/blob/master/src_common/game/Utils/_Ja25GermanText.c#L155) messages from UB source code)
- Adds an asynchronous check after winning a battle
    - We wait until all other things are done (e.g. win-battle quote, auto-bandaging) 
    - Shares the same check with the one in Strategic Movement (`WildernessSectorWithAllProfiledNPCsNotSpokenWith`)
- Differences from UB PATCH152:
    - UB implementation uses Game Event and persists into saves
    - This implementation uses a callback timer, so we do not need to modify events and saves
    - Notification is lost if player exits tactical immediately after battle


lynxlynxlynx:
Looks fine, thanks!

ja2-derek:
Now the "There is someone else" dialogue pops up after killing all enemies in the first sector (because of Fatima). 

This is probably too big a deviation from Vanilla. I am thinking what to do about this..

lynxlynxlynx:
Another field in the npc data json?

ja2-derek:
My bad. There is actually no problem. I had just messed up the JSON data in my dev env.

The check uses `WildernessSectorWithAllProfiledNPCsNotSpokenWith`, and A9 Omerta is a town sector so the pop up will not be shown.

Issue #939, opened 2019-12-07
Include order in C++


flaviojs:
In rust I'm grouping as std, then other crates, then the crate. Inside each group it's sorted alphabetically.
Main reasons:
1) easier to know where to find docs and source
2) it's intuitive:
2.1) code comes after includes (use) so it also shows the closeness to your code
2.2) alphabetical order is means you don't have to read all the includes in the group

What about in C++?
CONTRIBUTING.md says "Sort includes by type (project, system) and alphabetically."
Is this supposed to be like rust but with the group order reversed?
That is: our includes, dependency includes, core includes

lynxlynxlynx:
Yes. My usual order is: same dir, other dirs, system, so paths in includes aren't scattered all around due to sorting. In some projects the counterpart header has a separate place too, since it makes sense for it to be prominent.

In the existing code it's a mess, with system includes sometimes even in the middle of local ones — nobody cleaned up the style. If you want to unify it, go ahead. I for one don't care if the supraordering goes one way or the other.

flaviojs:
I'll fix the order of the include sections I modify, which right now is C/C++ includes.

PR #1202, opened 2020-08-28
Clean up Map Screen code

ja2-derek:
Noticed there are a lot of dead code in Map Screen (one of the largest files in codebase) when I was about to make a patch. So I decided to make a clean up before making any real changes.

## Summary

- Removed everything related to `fZoomFlag` (which is a cancelled feature)
    - `fZoomFlag` was always FALSE, and we never set its value. So keep everything `if (!fZoomFlag)` and remove all `if (fZoomFlag)`
- Removed commented-out and unused code
- Cleaned up includes (#939) by a combination of:
   - Removing unnecessary `#include`s by trial-and-error
   - '[include-what-you-use](https://github.com/include-what-you-use/include-what-you-use)', which also replaces some includes with forward declarations
   - Sorted includes, using the "[IncludeToolbox](https://marketplace.visualstudio.com/items?itemName=Wumpf.IncludeToolbox)" for VS

## Testing

The changes are limited to the Strategic Map screen, especially the tracing of squad routes and waypoints. There should be no functional differences from before.


lynxlynxlynx:
Beautiful diffstat. :)

PR #1203, opened 2020-08-28
Fixing illegal array access flagged by Coverity

ja2-derek:
These are illegal array access (out-of-bounds access/read), in the Laptop. Mostly because of wrong array bounds and checks.

### DIFF
--- a/src/game/Laptop/Florist_Gallery.cc
+++ b/src/game/Laptop/Florist_Gallery.cc
@@ -334,18 +334,15 @@ static void ChangingFloristGallerySubPage(UINT8 ubSubPageNumber)
 {
 	fLoadPendingFlag = TRUE;
 
-	//there are 3 flowers per page
-	if( ubSubPageNumber == FLOR_GALLERY_NUMBER_FLORAL_IMAGES )
-		ubSubPageNumber = 4;
-	else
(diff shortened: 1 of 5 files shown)


src/game/Laptop/Florist_Gallery.cc
```diff
@@ -334,18 +334,15 @@ static void ChangingFloristGallerySubPage(UINT8 ubSubPageNumber)
 {
 	fLoadPendingFlag = TRUE;
 
-	//there are 3 flowers per page
-	if( ubSubPageNumber == FLOR_GALLERY_NUMBER_FLORAL_IMAGES )
-		ubSubPageNumber = 4;
-	else
-		ubSubPageNumber = ubSubPageNumber / 3;
+	//there are 3 flowers per page, 4 pages in total
+	UINT8 ubPageNumber = std::min(ubSubPageNumber / 3, 3);
```
lynxlynxlynx, line 338: Just asking since it's not only a coverity fix — does the florist still work fine?

ja2-derek: Yes. It still works. I checked all 10 flowers are on the list.

There are 10 available, 3 per page. So there are 4 pages with page numbers 0-3

PR #1204, opened 2020-08-29
Display a message if player choose an inaccessible dropoff sector

ja2-derek:
Porting [PATCH157](https://github.com/ja2-derek/ja2-ub-comparison/commit/f2ecb49d) from Unfinished Business (#1198)

Displays a message saying "New Recruits cannot arrive there" if the player tries to choose an inaccessible sector as the dropoff sector, in the same way as all other dropoff point (bullseye icon) messages.


ja2-derek:
How it looks in game:

image

PR #1207, opened 2020-08-31
Array-bounds checks and clean up

ja2-derek:
Array out-of-bounds access flagged by Coverity.

Also cleaned up the includes of the modified files - removed unnecessary and then sorted.


ja2-derek:
AppVeyor seems to be caused by a new Visual Studio 2019 update - https://developercommunity.visualstudio.com/content/problem/1139955/c-builds-with-windows-sdk-81-fail-after-upgrade-to.html

Issue #1205, opened 2020-08-29
Extending mods with scripting


ja2-derek:
This is a proposal and RFC for adding scripting functionalities for further enabling JA2:S mods. This is all preliminary, so feel free to comment, counter-propose or reject.

# Proof-of-concept

Lua seems to be the choice for embedded scripting in many successful games (JA2 1.13 :-P). I made a simple prototype with Lua and Sol2 (wrapper around plain lua-C bindings API, v3.2). I am able to add items, and modify SectorInfo from Lua script.

[Sol2](https://sol2.readthedocs.io/en/latest/) seems to be one of the more actively maintained (and maybe well promoted) libraries out there. Its API and usages look clean and concise enough to me. But it requires C++ 17.

_Disclaimer: I have no prior experience with Lua or such libraries, other than ~10 lines of scripts on an nginx server_

# Concerns

- Are there any other things we want to tackle before introducing a big chunk of customisation?
- There will be quite a lot of binding code necessary. It could slow down build times.
- Or we could disable all Lua extension functionalities with a compiler "experimental" flag like `#if WITH_LUA` 

# Implementation

It is a two-way binding - that we invoke Lua scripts from certain points in C++ codebase, and also allow Lua to call C++ functions and alter global variables.

A lot from below is taken from or inspired by JA2 1.13 [wiki](http://ja2v113.pbworks.com/w/page/4218358/Modding%20JA2%20with%20LUA%20Scripts).

### Global variables

Some of the global variables that can be read and write from Lua scripts.

| Type | Name | 
|---|---|
| TacticalStatusType | gTacticalStatus | 
| UINT8 | gWorldSectorX |
| UINT8 | gWorldSectorY |
| UINT8 | gbWorldSectorZ |
| UINT8[] | gubFact |
| UINT8[] | gubQuest |
| MERCPROFILESTRUCT[] | gMercProfiles |
| SECTORINFO[] | SectorInfo |
| StrategicMapElement[] | StrategicMap|

### Functions callable

C++ functions to be made available to Lua scripts. These are only examples, there will be a lot more.

| Module | Name | User-defined types |
|---|---|---|
| Overhead | GetMan | SOLDIERTYPE |
| Dialogue_Control | TacticalCharacterDialogue | SOLDIERTYPE | 
| NPC | TriggerNPCRecord | |
| TileEngine/Exit_Grid | AddExitGridToWorld | |
[...]

PR #1209, opened 2020-09-01
Lua scripting engine

ja2-derek:
Basic implementation of #1205.

_Ignore all files under dependencies/lib-lua, and the diff should be more managable._

# Lua

Lua (v5.3.5) and Sol2 (v2.20.6) are bundled and built from source. System packages are not considered at all.  Lua is a small library, and the pre-compiled Lua binaries did not work for me (VC++), probably because of Debug/Release or VS version differences. But the compilation of Lua is quite fast, and always building from source simplifies the CMakeFile quite a bit

**sol2 requires C++ 14**

# Scripting Engine

There are two main module for this so-called "scripting engine". First is the functions available to Lua scripts (FunctionsLibraries.h), and second is the integration points where C code will call lua function if defined (see ScriptingExtensions.h).

Mods using Lua must provide a `data/scripts/main.lua`. This file is the entrypoint (think `int main`), and can include other lua files. The base game will also provide some constants, enums or maybe helper functions in Lua.

The functions exposed to Lua are a slightly simplified versions of the corresponding C functions, and should be restricted to use only by Lua scripts. The functions are all grouped in the `scripting/` module. I intend to generate some form of documentations later on.

# Error handling

All Lua scripts are fully parsed before game starts. If there is an error (e.g. syntax error), an exception is thrown and the game will refuse to start.

If Lua script had a runtime error (within Lua, or within a C function called by Lua), the Lua error will be logged and thrown as `std::runtime_error`. It may or may not be handled by the game depending where it is thrown. The scripting engine will be disabled until the game restarts.

# Example implementation

The `generous-rebel` minimod is updated to use Lua to extend its functionality. It now adds a few items to the initial sectors at the first visit.


ja2-derek:
A (very basic) manual for using scripting in JA2:S mods: [scripting-manual.pdf](https://github.com/ja2-stracciatella/ja2-stracciatella/files/5177403/scripting-manual.pdf)

lynxlynxlynx:
Did you make that manually or is there some doxygen, sphinx or similar behind it?

ja2-derek:
That's generated with Doxygen from annotated comments in code

src/externalized/scripting/ScriptingExtensionsLua.cc
```diff
@@ -128,15 +129,18 @@ static void RegisterGlobals()
...
 	lua.set_function("GetUndergroundSectorInfo", GetUndergroundSectorInfo);
 	
 	lua.set_function("CreateItem", CreateItem);
 	lua.set_function("CreateMoney", CreateMoney);
 	lua.set_function("PlaceItem", PlaceItem);
+
+	lua.set_function("JA2Require", JA2Require);
+	lua.set_function("require",  [](void) { throw std::logic_error("require is not allowed. Use JA2Require instead"); });
+	lua.set_function("dofile",   [](void) { throw std::logic_error("require is not allowed. Use JA2Require instead"); });
+	lua.set_function("loadfile", [](void) { throw std::logic_error("require is not allowed. Use JA2Require instead"); });
```
lynxlynxlynx, line 143: These two have the wrong name in the error.

PR #1210, opened 2020-09-07
Support Unfinished Business ("JA 2.5") tilesets

ja2-derek:
This PR allows us to load both JA2 Vanilla and JA2.5 Unfinished Business maps.

# Background

Vanilla has 50 tilesets. Unfinished Business adds another 20, a total of 70. I believe there are no dispute that tilesets 50~69 are for UB.

# Summary of changes

- `gTilesets` is now 70 elements long
- This is fully compatible with vanilla data; With vanilla data, we only load first the 50 tilesets and the maps only reference those.
- `NUM_TILESETS` is now a variable `ubNumTilesets` read from Tileset DB; it must be either 50 or 70.
- There is a "default tileset" (`GENERIC_1` in vanilla) which has the special tiles such as doors; JA2.5 tilesets (index > 50) uses a different "default"

# Reference

This is ported from UB source code: 

- [PATCH020](https://github.com/ja2-derek/ja2-ub-comparison/commit/c41920c5)
- [PATCH119](https://github.com/ja2-derek/ja2-ub-comparison/commit/a9d2a98a)
- [PATCH120](https://github.com/ja2-derek/ja2-ub-comparison/commit/e3baf3c7)

### DIFF
--- a/src/game/TileEngine/WorldDef.cc
+++ b/src/game/TileEngine/WorldDef.cc
@@ -208,14 +208,17 @@ static void AddTileSurface(const ST::string filename, UINT32 const tileType);
 
 TILE_SURFACE_RESOURCE GetAdjustedTilesetResource(TileSetID tilesetID, UINT32 uiTileType, const ST::string filePrefix)
 {
-	if (tilesetID >= NUM_TILESETS) throw std::logic_error("invavlid tilesetID");
+	if (tilesetID >= gubNumTilesets) throw std::logic_error("invavlid tilesetID");
 
 	ST::string  filename = gTilesets[tilesetID].zTileSurfaceFilenames[uiTileType];
 	if (filename.empty())
 	{
 		// Try loading from default tileset
-		filename = gTilesets[GENERIC_1].zTileSurfaceFilenames[uiTileType];
-		tilesetID = GENERIC_1;
+		tilesetID = (uiTileType >= DEFAULT_JA25_TILESET || (gubNumTilesets == JA25_NUM_TILESETS && uiTileType == SPECIALTILES))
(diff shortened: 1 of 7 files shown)


src/game/TileEngine/WorldDef.cc
```diff
@@ -227,8 +227,8 @@ try
 			// ATE: If here, don't load default surface if already loaded
 			if (gbDefaultSurfaceUsed[i]) continue;
 
-			filename       = gTilesets[GENERIC_1].TileSurfaceFilenames[i];
-			tileset_to_add = GENERIC_1;
+			tileset_to_add = (tileset_id < DEFAULT_JA25_TILESET || i == SPECIALTILES) ? GENERIC_1 : DEFAULT_JA25_TILESET;
```
lynxlynxlynx, line 230: Why is the extra SPECIALTILES check needed? If it is, why just here?

ja2-derek: It is ported from [PATCH119 (R440)](https://github.com/ja2-derek/ja2-ub-comparison/commit/a9d2a98a#diff-f80efc555e07bb1af95205bd9b05cbe8R440). 

I believe the `DEFAULT_JA25_TILESET` do not have `SPECIALTILES`. So we need to fallback one more step to `GENERIC_1`. Maybe it is clearer to negate it?

```c
 tileset_to_add = (tileset_id >= DEFAULT_JA25_TILESET && i != SPECIALTILES) ? DEFAULT_JA25_TILESET : GENERIC_1;
```

lynxlynxlynx: I think it'd be clea[nr]est if it was outsourced to a function, since we do these lookups in three places.

PR #1056, opened 2020-05-01
Change char* to ST::string

flaviojs:
This PR aims to convert most uses of `char*` with `ST::string` and fix string related issues.

Refactoring is not the aim, but will probably happen. I hope I don't change too much...

PR #1211, opened 2020-09-08
Refactor TileEngine

ja2-derek:
1. Replaced `const char*` with `ST::string`. Changes cascade into sgp/ and editor/ (#1056)
2. Organized imports (#939)
3. Extracted method for the tileset defaulting logic (`GetAdjustedTilesetResource`) (#1210)


lynxlynxlynx:
Looks fine, merge as you will (not sure which order you prefer).

PR #1212, opened 2020-09-08
Use fixed Windows SDK version on appveyor

selaux:
Currently all builds on appveyor fail because fltk cannot be compiled. There are a bunch of errors of the following kind, within `winbase.h`.

```
C:\Program Files (x86)\Windows Kits\8.1\Include\um\winbase.h(9036,31): error C2065: '_InterlockedOr64': undeclared identifier [C:\projects\ja2-stracciatella\ci-build\dependencies\lib-fltk\builder\fltk-prefix\src\fltk-build\src\fltk.vcxproj] [C:\projects\ja2-stracciatella\ci-build\dependencies\lib-fltk\builder\fltk.vcxproj]
```

Since we didn't change anything about fltk or how it is built, my guess this is due to external changes.

According to https://developercommunity.visualstudio.com/content/problem/1139955/c-builds-with-windows-sdk-81-fail-after-upgrade-to.html there is an issue with windows sdk in version 8.1 and Visual Studio 16.7. Since those are the versions currently in use on appveyor (see https://www.appveyor.com/docs/windows-images-software/), my hope is that this does not happen with windows sdk 10.0, as written in the linked issue. But it is still a shot in the dark...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -67,7 +67,7 @@ init:
         }
         true;
 
-  - ps: $env:BUILD_SWITCHES = "-DCMAKE_BUILD_TYPE=%BUILD_TYPE% -DVERSION_TAG=%VERSION_TAG% -DCPACK_GENERATOR=ZIP;NSIS -DCMAKE_TOOLCHAIN_FILE=./cmake/toolchain-msvc.cmake"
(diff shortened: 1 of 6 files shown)


ja2-derek:
Thank you and LGTM.

This undoes https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1197 but I think that's needed until the next VS release.

PR #1214, opened 2020-09-09
Better lua debugging functionality

selaux:
- Errors in lua scripts now include the filename where the error happened
- Log messages from lua now print the lua filename where they are triggered from (unfortunately not the exact location as it only prints the vfs filename right now, e.g. `main.lua` instead of `mods/mod_name/scripts/main.lua`).

References: #1205

### DIFF
--- a/src/externalized/scripting/ScriptingExtensionsLua.cc
+++ b/src/externalized/scripting/ScriptingExtensionsLua.cc
@@ -143,16 +143,25 @@ static void RegisterGlobals()
 	lua.set_function("loadfile", [](void) { throw std::logic_error("loadfile is not allowed. Use JA2Require instead"); });
 }
 
+static void LogLuaMessage(LogLevel level, std::string msg) {
+		lua_Debug info;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
LGTM!

ja2-derek:
It works great! Thank you!

-----

Example of log:
```
2020-09-09T12:02:58 [ERROR] externalized\scripting\ScriptingExtensionsLua.cc: Lua script engine has failed to initialize:
 sol: runtime error: sol: syntax error: enums.lua:22: syntax error near 'end'
stack traceback:
        [C]: in function 'base.JA2Require'
        main.lua:1: in main chunk
2020-09-09T12:03:09 [ERROR] externalized\scripting\ScriptingExtensionsLua.cc: Lua script had an error. Scripting engine is now DISABLED. The error was:
enums.lua:21: attempt to index a number value (local 'fc')
stack traceback:
        enums.lua:21: in function 'base.call_error'
        main.lua:7: in function 'base.BeforePrepareSector'
```

PR #1215, opened 2020-09-11
Build AppImage for linux

### LINKED ISSUE #138: Create packages for Linux

---

It would be great to create packages for popular Linux distributions, so Linux users don't have to compile the game.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/138

### LINKED ISSUE #1213: Proposal: Build distro-agnostic packages for Linux releases instead of Debian Package
### What is wrong

Currently we only build a debian package for linux releases. This has the disadvantage that users of non debian-based systems have to compile the source for themselves, making the barrier to actually try out the game higher than necessary (e.g. I run arch linux, so i cannot test the PR builds as easily as I want to). Additionally it is not entirely clear with which debian-based system the deb package is actually compatible, although I personally never experienced issues with that.

### Proposal

We should have a look at more modern packaging options such as

- Flatpak
- AppImage
- Snap

I think the easiest route for us to go down is building AppImages, as there is no additional software required to run an AppImage on any Linux distribution, which makes the installation process quite easy (or even give the user the option not to install anything). Also we can keep the existing way of distributing releases via the github releases page. A disadvantage might be that we'll actually need some small code changes to support the game running from the AppImage. Also the adoption of AppImage is less than Snaps or Flatpaks respecitvely. What do you think?

PS: Here is a good comparison of the three (although it might be a little biased): https://github.com/AppImage/AppImageKit/wiki/Similar-projects#comparison

selaux:
Some changes to currrent logic were necessary:

- Always search EXTRA_DATA_DIR relative to binary (makes binary relocatable)
- Install linuxdeploy in CI, to build the AppImage
- Add cmake task `package-appimage` that is executed instead of `package` for linux

Closes #1213. Closes #138.

I will wait for build and test the resulting AppImage on multiple distros before merging.

### DIFF
--- a/.ci/ci-publish.sh
+++ b/.ci/ci-publish.sh
@@ -43,8 +43,9 @@ cd ci-build
 echo "## publish ##"
 for file in ja2-stracciatella_*; do
   echo "$file"
-  if [[ "$file" == *".deb" ]]; then
-    dpkg -c "$file"
+  if [[ "$file" == *".AppImage" ]]; then
+    export PATH=$PATH:$HOME/linuxdeploy
+    appimagelint $file
   elif [[ "$file" == *".zip" ]]; then
     unzip -l "$file"
   elif [[ "$file" == *".dmg" ]]; then
--- a/.github/scripts/ci-build.sh
+++ b/.github/scripts/ci-build.sh
@@ -42,10 +42,9 @@ fi
 
 export RUN_TESTS=true
 export RUN_RUST_CHECKS=false
-export RUN_INSTALL_TEST=true
 export RUSTUP_INIT_ARGS="-y --no-modify-path --default-toolchain=$(cat ./rust-toolchain) --profile=minimal"
 if [[ "$CI_TARGET" == "linux" ]]; then
-  export CONFIGURE_CMD="${CONFIGURE_CMD} -DCMAKE_INSTALL_PREFIX=/usr -DEXTRA_DATA_DIR=/usr/share/ja2 -DCPACK_GENERATOR=DEB"
+  export CONFIGURE_CMD="${CONFIGURE_CMD} -DCMAKE_INSTALL_PREFIX=AppDir/usr -DEXTRA_DATA_DIR=../share/ja2"
   export RUN_RUST_CHECKS=true
 
 elif [[ "$CI_TARGET" == "linux-mingw64" ]]; then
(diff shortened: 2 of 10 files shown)


ja2-derek:
EXTRA_DATA_DIR defaults to the executable directory. Is it simpler just to package data files under `./externalized` and skip EXTRA_DATA_DIR totally? Windows package already does that.

selaux:
I think it is working now (lets see how Travis handles the master build). I tested the AppImage on:

- Fedora 32
- Ubuntu 20.04
- Debian 10
- Manjaro

Regarding removing EXTERNAL_DATA_DIR: I'd still like for the usual `make install` under Linux/BSD to work nicely. So as far as I can see this is still necessary.

ja2-derek:
It complains of missing `editor.slf` when I try to start the editor. And can I start ja2 directly skipping the launcher?

selaux:
I think the editor slf is not included with Pull Request builds as they are debug builds.

Currently it is not possible to start ja2 directly, we would need to write a wrapper script around `ja2` and `ja2-launcher` in order to do that.

ja2-derek:
> I think the editor slf is not included with Pull Request builds as they are debug builds.


Ah right. So that's not a problem.

> Currently it is not possible to start ja2 directly, we would need to write a wrapper script around `ja2` and `ja2-launcher` in order to do that.

Not really a problem either. It's just one click on the JA2 Launcher.

ja2-derek:
Travis build is failing:

https://travis-ci.org/github/ja2-stracciatella/ja2-stracciatella/jobs/727108794. Logs:
```
 if [ "$TRAVIS_MINGW" != "true" ]; then if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./AppDir/usr/bin/ja2 -unittests && ./AppDir/usr/bin/ja2-launcher -help; else ./ja2 -unittests && ./ja2-launcher -help; fi fi

/home/travis/.travis/functions: line 109: ./AppDir/usr/bin/ja2: No such file or directory

The command "if [ "$TRAVIS_MINGW" != "true" ]; then if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ./AppDir/usr/bin/ja2 -unittests && ./AppDir/usr/bin/ja2-launcher -help; else ./ja2 -unittests && ./ja2-launcher -help; fi fi" exited with 127.
```

selaux:
I will have a look later...

PR #1218, opened 2020-09-13
Fix the Tileset defaulting logic

ja2-derek:
The comparison is on the TileSetID not the TileType. It should be some copy-and-paste mistake I made during the refactor.

This commit has been tested and the game loads Vanilla maps correctly.


selaux:
:+1:

PR #1221, opened 2020-09-14
Updating cargo deps

ja2-derek:
Mac build is currently broken. Also noticed there is a new crate release of serde. It might have caused inconsistency in the cargo lock.

Trying `cargo update` to see if it helps with the Mac build problem.

<details>
<summary>List of crates updated</summary>
<pre>
    Updating aho-corasick v0.7.10 -> v0.7.13
    Updating autocfg v1.0.0 -> v1.0.1
    Updating base64 v0.11.0 -> v0.12.3
    Updating chrono v0.4.11 -> v0.4.15
    Updating clap v2.33.0 -> v2.33.3
      Adding crossbeam-channel v0.4.4
    Removing crossbeam-queue v0.2.1
    Updating dunce v1.0.0 -> v1.0.1
    Updating either v1.5.3 -> v1.6.0
    Updating getrandom v0.1.14 -> v0.1.15
      Adding hashbrown v0.9.0
    Updating hermit-abi v0.1.10 -> v0.1.15
    Updating indexmap v1.3.2 -> v1.6.0
    Updating itoa v0.4.5 -> v0.4.6
    Updating libc v0.2.68 -> v0.2.77
    Updating log v0.4.8 -> v0.4.11
    Updating memoffset v0.5.4 -> v0.5.5
    Updating num-integer v0.1.42 -> v0.1.43
    Updating num-traits v0.2.11 -> v0.2.12
    Updating num_cpus v1.12.0 -> v1.13.0
    Updating ppv-lite86 v0.2.6 -> v0.2.9
    Updating proc-macro2 v1.0.10 -> v1.0.21
    Updating quote v1.0.3 -> v1.0.7
    Updating rayon v1.3.0 -> v1.4.0
    Updating rayon-core v1.7.0 -> v1.8.0
    Updating redox_syscall v0.1.56 -> v0.1.57
    Updating redox_users v0.3.4 -> v0.3.5
    Updating regex v1.3.6 -> v1.3.9
    Updating regex-syntax v0.6.17 -> v0.6.18
    Updating remove_dir_all v0.5.2 -> v0.5.3
    Updating rust-argon2 v0.7.0 -> v0.8.2
    Updating ryu v1.0.3 -> v1.0.5
    Updating serde v1.0.105 -> v1.0.116
    Updating serde_derive v1.0.105 -> v1.0.116
    Updating serde_json v1.0.50 -> v1.0.57
    Removing smallvec v1.2.0
    Updating syn v1.0.17 -> v1.0.40
[...]

PR #1222, opened 2020-09-15
Clean up dead code

ja2-derek:
## Unused variables

Removed unused variables in Map_Screen_Interface_Map, flagged by GCC.

## Remove Air Raid module

Removing the air raid logic which is never used in game (see this [video](https://www.youtube.com/watch?v=0HQFDAUn908)). Air raid data are persisted in saves, as a 132 bytes struct. We will keep a 132 zero'ed bytes placeholder so saves are still compatible with old versions.


lynxlynxlynx:
The build is failing from the recent bump:
> error[E0463]: can't find crate for `serde_derive` which `cbindgen` depends on

ja2-derek:
#1221 was an attempt to solve this exact same error... Now I suspect it might be the `.cargo` caching...

PR #1223, opened 2020-09-16
Dependency inversion with Observable and notify

ja2-derek:
Further to the discussion at https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1205#issuecomment-689418616. This is a proposal to implement the "Observer" pattern, for 1) dependency inversion, and 2) allow registering callback at runtime.

## Overview

This is conceptually similar to event/delegate of C#. The subject ("observable") allows registering of typed callback functions, and the subject need no knowledge of what's inside the callbacks. Typically it is the core engine calling back a campaign or sector specific logic.

### Limitations

In contrast to the Strategic_Event_Handler implementation, the Observable class is strong-typed. But that means each callback must be registered explicitly, you cannot have a single handler function switching on event type. This could mean quite a lot of boilerplate code in the scripting extension.

Currently it only support void-functions. Return values could be complicated, because at runtime there might be multiple callbacks. One way is to pass a pointer argument (e.g. `int*`), which can be read and modified by every callback function.

Also there is no clear point of callback registration (ReStartingGame? GameInit?). We need to be careful to register exactly once, at the correct place.

### Drawbacks

Tracing of code is harder because of the added indirection.

## Example implementation

An example is that TileEngine has logic about SAM site but it shouldn't. Using the observer pattern we can move the logic back to Strategic layer.

This is just a proposal to illustrate my idea. I have only briefly tested.

### DIFF
--- a/src/game/Strategic/StrategicMap.cc
+++ b/src/game/Strategic/StrategicMap.cc
@@ -286,6 +286,7 @@ void BeginLoadScreen( )
 
 static void InitializeMapStructure(void);
 
+static void UpdateAndDamageSAMIfFound(INT16 sSectorX, INT16 sSectorY, INT16 sSectorZ, INT16 sGridNo, UINT8 ubDamage);
 
 void InitStrategicEngine()
 {
@@ -296,6 +297,9 @@ void InitStrategicEngine()
 
 	// set up town stuff
 	BuildListOfTownSectors( );
+
+	// handle SAM site damages
+	OnStructureDamaged.addListener("default", UpdateAndDamageSAMIfFound);
 }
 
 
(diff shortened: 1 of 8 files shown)


lynxlynxlynx:
Looks pretty clean, modern c++ has come a long way. :)

selaux:
I agree, this looks quite nice. One improvement would be that it would be nice if it would be possible to un-register a single callback, but then the chaining syntax is probably not possible. Then (looking at the modding use-case) we could just expose the callback-ids for vanilla and mods would be able to un-register to their liking.

Regarding this specific callback, I think it would make sense to add the structure to the callback signature as well, not only gridNo.

Regarding return values, passing a pointer sounds like an working solution to me.

ja2-derek:
> I agree, this looks quite nice. One improvement would be that it would be nice if it would be possible to un-register a single callback, but then the chaining syntax is probably not possible. Then (looking at the modding use-case) we could just expose the callback-ids for vanilla and mods would be able to un-register to their liking.

`callback-id` is a great idea! But then I am thinking if we should impose any ordering of the callbacks. 

If we use a naive `std::map` implementation, then the callbacks will be called in the lexical ordering of the given `callback-id`. However, implementing another numeric priority, or allowing pre-/append listeners may make this unnecessarily complex.

The fluent interface is not a big deal. We can give it up, if it gets in our way.


> Regarding this specific callback, I think it would make sense to add the structure to the callback signature as well, not only gridNo.

I will check that again, but the function does not need anything from the structure

selaux:
> I will check that again, but the function does not need anything from the structure

Yes I think it is not relevant for this functionality, but again coming from a mods perspecive I might want to know in the `OnStructureDamaged` callback which structure was actually damaged, so I can react accordingly (think of some "destroy this structure" quest). I think getting the callback signatures right from the start will help compatibility between stracciatella versions and mod versions later, so mods don't have to update their code because we change callback signatures. As PR this is currently more focussed on cleaning up code than modding support (which can come later), feel free to just ignore my comment :wink:.

> If we use a naive std::map implementation, then the callbacks will be called in the lexical ordering of the given callback-id.
> However, implementing another numeric priority, or allowing pre-/append listeners may make this unnecessarily complex.

Yes for now it should be fine to use callback-id as ordering.

ja2-derek:
Maybe some convention like `mod-01-MyMod`. Just need to somehow get the mod ordering from VFS, and mod probably shouldn't register the same callback more than once.

ja2-derek:
I have tidied up the code slightly.

I will leave the `OnStructureDamaged` signature unchanged for now. I plan to do a round of full refactoring in TileEngine. Then we can look at it as a whole, to see what APIs are needed.

lynxlynxlynx:
If we were worried about the stability of the signatures, we could go a layer deeper and basically use just an argv argument, like they often do in js. But that just moves the problem, so i don't see a point.

PR #1226, opened 2020-09-17
Try to fix Mac build

ja2-derek:
Changes so far:

- Update cargo.lock - `cargo update` again
- Changing the paths to cache
   - Cache only some directories under `~/.cargo`, as per [docs](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci)
   - Do not cache lib-stracciatella/target/, we rebuild every time

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -57,13 +57,15 @@ jobs:
       uses: actions/cache@v2
       with:
         path: |
-          ~/.cargo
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Works, merge it if you consider it done.

ja2-derek:
It always works with a clean `.cargo`, but fails on the 2nd run. Only fails on Mac :-\

PR #1227, opened 2020-09-17
Build caching on AppVeyor

ja2-derek:
Enabling build cache on AppVeyor so we can re-run stuff between builds (#1102). It should save up to 3 minutes per build 

- Caching the [cargo home directory](https://doc.rust-lang.org/cargo/guide/cargo-home.html#caching-the-cargo-home-in-ci)
- Caching rustup install (`~/.rustup`)
- Caching and re-using the lib-fltk working directories
- Use [sccache](https://github.com/mozilla/sccache) to speed up rustc ~and VC++~ compilations

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,8 +17,13 @@ endmacro()
 
 ## Project Setup
 
+# compiler caching
 find_program(CCACHE_PROGRAM "ccache")
-if(CCACHE_PROGRAM)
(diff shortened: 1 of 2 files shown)


ja2-derek:
I couldn't make Visual C++ use `sccache`. But if you are on Linux (maybe also Mac), you can install `sccache`, then do something like:

`RUSTC_WRAPPER=sccache cmake --build .`

Then all your Rust and C builds will be cached.

PR #1232, opened 2020-09-20
Refactor SAM sites logic

ja2-derek:
# Summary

Refactoring - only moving code around and no functional changes

* Remove all Repair SAM site code and comments - feature was never in the released JA2
* Added Strategic/SAM_Site.(h|cc)
* Moved SAM sites logic away from StrategicMap and ExplosionControl


ja2-derek:
Noticed that you can repair a vehicle but not a robot in tactical. You can repair Robot in strategic though.

It is a (trivial) bug since initial import (https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/game/Tactical/Structure_Wrap.cc#L305-L307). I guess nobody ever starts a repair in tactical.

lynxlynxlynx:
It takes too long to be useful during combat.

PR #1236, opened 2020-09-23
More TileEngine refactoring

ja2-derek:
Follow up after our discussion in [#1223](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1223#issuecomment-694161770). Further cleaning up TileEngine by removing sector-specific code.

## Summary

- TileEngine to do callback just before and right after a structure being damaged
    - Added `STRUCTURE*` and `fIsDestroyed` to `OnStructureDamaged` callback
- Moved away SAM sites, brothel and statue logic from TileEngine
- Replaced an abused BOOLEAN return value with enum

### DIFF
--- a/src/game/Strategic/SAM_Sites.cc
+++ b/src/game/Strategic/SAM_Sites.cc
@@ -23,7 +23,7 @@ BOOLEAN fSamSiteFound[NUMBER_OF_SAMS] = {
 
 Observable<> OnAirspaceControlUpdated = {};
 
-static void UpdateAndDamageSAMIfFound(INT16 sSectorX, INT16 sSectorY, INT16 sSectorZ, INT16 sGridNo, UINT8 ubDamage);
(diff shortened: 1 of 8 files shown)


selaux:
Looks good. One thing we have to keep in mind is that those observables are only called for explosive damage right now. So maybe we could rename them that way, or maybe call them from within `DamageStructure`...

ja2-derek:
Agreed. Those are specific to explosives, the functions should be renamed accordingly.

I didn't know gun fire (and what else?) can damage a structure too.

selaux:
According to `DamageStructure` damage reasons are defined as:

```
enum StructureDamageReason
{
	STRUCTURE_DAMAGE_EXPLOSION = 1,
	STRUCTURE_DAMAGE_GUNFIRE   = 2
};
```

PR #1238, opened 2020-09-24
Observable listeners from Lua scripts

ja2-derek:
Continuing to flesh out the Lua scripting extension.

## Summary

- Lua scripts can now use `RegisterListener` and `UnregisterListener`
    - `BeforePrepareSector` is now implemented with Observable
- Observables need to be "declared" in `RegisterListener`, we need to "declare" for every instance of the  template class
- Added support for using variables and struct members of type `ST::string` from Lua
- Mapped additional structures and enums for Lua, such as `STRUCTURE`, `SOLDIERTYPE`, `MercProfiles`,
- Added new Observable `OnSoldierCreated` 
    - I used this to test ST::string, by overriding the string variables `VestPal` and `PantPal` on `SOLDIERTYPE` (to be added to Wildfire mod)
    - Can be used for UB [PATCH136](https://github.com/ja2-derek/ja2-ub-comparison/commit/1bfb4e0d)
- `ScriptExtensionLua.cc` is now too big for the mingw64 compiler. We have to add an `-Wa,-mbig-obj` option
## Open issues

A couple of things to be sorted out (in future PRs):
- The scripting engine uses only a single entrypoint `main.lua` via VFS. So it does not allow multiple mods using Lua yet.
- Need a convention/rules of the keys identifying Observable listeners

### DIFF
--- a/assets/externalized/scripts/enums.lua
+++ b/assets/externalized/scripts/enums.lua
@@ -1,17 +1,195 @@
 Items = {
     SUNGOGGLE = 212,
-    CANTEEN   = 214
+    CANTEEN   = 214,
 }
 
 Visibility = {
 	BURIED               =  -3,
 	HIDDEN_IN_OBJECT     =  -2,
 	INVISIBLE            =  -1,
-	VISIBLE              =   1
+	VISIBLE              =   1,
 }
 
 SectorFlags = {
     SF_ALREADY_VISITED = 32,
     SF_USE_ALTERNATE_MAP = 64,
-    SF_HAS_ENTERED_TACTICAL = 512
+    SF_HAS_ENTERED_TACTICAL = 512,
+}
+
+CivGroups = {
+    NON_CIV_GROUP = 0,
+    REBEL_CIV_GROUP = 1,
+    KINGPIN_CIV_GROUP = 2,
+    SANMONA_ARMS_GROUP = 3,
+    ALMA_MILITARY_CIV_GROUP = 7,
+    DOCTORS_CIV_GROUP = 8,
+    HICKS_CIV_GROUP = 10,
+    WARDEN_CIV_GROUP = 11,
+}
+
+Teams = {
+    OUR_TEAM = 0,
+    ENEMY_TEAM = 1,
+    CREATURE_TEAM = 2,
+    MILITIA_TEAM = 3,
+    CIV_TEAM = 4,
+}
+
+SoldierClasses = {
+	SOLDIER_CLASS_NONE = 0,
+	SOLDIER_CLASS_ADMINISTRATOR = 1,
+	SOLDIER_CLASS_ELITE = 2,
+	SOLDIER_CLASS_ARMY = 3,
+	SOLDIER_CLASS_GREEN_MILITIA = 4,
+	SOLDIER_CLASS_REG_MILITIA = 5,
+	SOLDIER_CLASS_ELITE_MILITIA = 6,
+	SOLDIER_CLASS_CREATURE = 7,
+	SOLDIER_CLASS_MINER = 8,
+}
+
+MilitiaRanks = {
+    GREEN_MILITIA = 0,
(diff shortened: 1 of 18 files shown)


assets/mods/generous-rebels/data/scripts/main.lua
```diff
@@ -1,9 +1,10 @@
 JA2Require("enums.lua")
+RegisterListener("BeforePrepareSector", "mod:generous-rebels", "addItemsToSector")
```
lynxlynxlynx, line 2: This is odd, why do we hardcode mod names?

ja2-derek: My thinking is that, there may eventually be multiple mods listening to this same Observable. So that we need a unique listener key to prevent them from overriding each other.

lynxlynxlynx: This partly defeats the point of externalization. Can we not use random keys or let the mods specify their namespace somewhere?

ja2-derek: A mod may needs to:
- Disable vanilla listeners
- Register multiple listeners with the same Observables 

So yes, perhaps we can make vanilla listeners have well-known names, then the engine decides the random and/or namespaced keys for the mods'.

assets/externalized/scripts/enums.lua
```diff
@@ -1,17 +1,195 @@
...
+
+Teams = {
+    OUR_TEAM = 0,
+    ENEMY_TEAM = 1,
+    CREATURE_TEAM = 2,
+    MILITIA_TEAM = 3,
+    CIV_TEAM = 4,
+}
+
+SoldierClasses = {
```
selaux, line 38: I wonder if there is a better way to define enums that are already defined in C++. This looks quite error prone...

ja2-derek: I would think about code generation. Do you know of any good ways and tools (that understands C code) for that?

selaux: Unfortunately not. But I found there is a `new_enum` function in sol, which we could use from C++ to create the enum: https://sol2.readthedocs.io/en/latest/api/table.html#new-enum

selaux:
> The scripting engine uses only a single entrypoint main.lua via VFS. So it does not allow multiple mods using Lua yet.

I think for this we could extend the VFS (in another PR) with a function that returns all `main.lua` files within the VFS and call all of them.

assets/externalized/scripts/enums.lua
ja2-derek: If I am not mistaken, with the `new_enum` function we still need to provide each and every enum name and value. This is just as error-prone and copy-and-pasting.

I am thinking if we can get away with a simple'ish regex scripts, that specifically handles our usage of `enum {` and `#define SOME_PREFIX_` - oh yes, some of the "enums" are a series of `define` sharing the same prefix.

ja2-derek: Just found this: https://github.com/Neargye/magic_enum

But requires C++17, and we also need to change those `#define` into `enum`.

selaux: That looks quite nice, but probably requires us to migrate all enums to enum classes, so I think it is out of scope for this PR. What is keeping us from using C++17? Only MinGW on travis? Because I wanted to suggest to retire Travis as we already have Github Actions which give us excactly the same functionality. Only nightly builds would have to be migrated, which is trivial imho.

ja2-derek: One problem is `rapidjson`. The library gives warnings on C++14 and C++ 17 (hard errors on VC++ 2019). There is a [fix](https://github.com/Tencent/rapidjson/issues/1131) on master already but there has been no new releases.

We might need to pin to a newer commit on master, or switch to another library.

PR #1239, opened 2020-09-25
More flexible NPC faces handling

ja2-derek:
We have layers of mapping, from merc or head miners to an external faces enum, then to the actual `FACETYPE` struct, and the list of external NPC faces is static. An example of an "external NPC faces" is Skyrider's dialogues at the Strategic map screen.

This PR simplifies the merc-to-face mappings and allows mods to load additional NPC faces.

## Summary

- Direct mapping from merc profile ID to `FACETYPE*` struct. The mapping is dynamically sized (using `std::map`)
- Removed the explicit mappings from `HEAD_MINER_TYPE` struct to face IDs
- Automatically loads if an NPC face is requested but not yet loaded


src/game/Tactical/Dialogue_Control.cc
```diff
@@ -170,36 +170,52 @@ void ShutdownDialogueControl()
...
 {
-	INT32 iCounter = 0;
 	// go and grab all external NPC faces that are needed for the game who won't exist as soldiertypes
 
 	if (fExternFacesLoaded) return;
 
 	fExternFacesLoaded = TRUE;
 
-	for( iCounter = 0; iCounter < NUMBER_OF_EXTERNAL_NPC_FACES; iCounter++ )
+	for (int i = 0; i < lengthof(preloadedExternalNPCFaces); i++)
```
lynxlynxlynx, line 202: Why not just use .size()?

ja2-derek: `.size()` is a method in `std::array<ProfileID, 6>`. 

So we can either: 1) switch to `std::array<>` or 2) use `lengthof()`- which is sizeof(array) / sizeof(element).

lynxlynxlynx: Ah, I see now I was looking at the definition of externalNPCFaces instead. Nevermind.

PR #1240, opened 2020-09-25
Android build

selaux:
I'm having another take at this, I really want to be able play this game on my tablet. This Draft should just ensure I dont break other platforms while implementing. When this PR is finished we should have:

- Minimal Android Launcher to select data dir
- A running android version (starting up, finding all the resources, being able to start a new game, etc.)
- A CI Task that builds this version automatically as APK, similar to the other releases

Most of the Android Project is just copy pasted from the SDL2 example project. I intend to keep it that way so we can just update by copy pasting the newer example :slightly_smiling_face:. I think it should be possible to inherit from their SDL activity.

Other things that need to be taken care of afterwards:
- Some mouse emulation system (im thinking about reusing the current mouse regions and implement some ShortTouch -> Left Click, LongTouch -> Right Click behavior in the beginning) and check where to go from there

Feel free to try it out and comment. Canonical way to build should be:

- Install rust targets:
  - `rustup target add armv7-linux-androideabi`
  - `rustup target add aarch64-linux-android`
  - `rustup target add i686-linux-android`
  - `rustup target add x86_64-linux-android`
- Install and open android studio
- Load project in `android` folder
- Build & Run in emulator

Known Issues:

- [x] Data dir is static: Internal Storage/Downloads
- [x] Configuration is not restored
- [ ] Mods from `assets` dir are not supported
- [ ] Android Assets VFS is not case insensitive
- [x] armv7 architecture missing due to https://github.com/bazelbuild/bazel/issues/8753 -> applied patch from current SDL2 master
- [x] Default SDLActivity needs quite a lot of permissions and android version: 6.0 -> Current required version is 4.3
- [x] Permissions need to be given manually
- [x] Android 10 cannot read files that are not written by the application -> fixed by using target api version 26
- [ ] Uses fork of ndk && ndk_sys because of linking issues
- [x] C++ Errors do not bubble up to Android Application
- [x] Test on a real device

Initial Status: APK builds fine on my local machine and starts up in emulator. Game dir is selectable but the setting is not restored on App restart.


ja2-derek:
I don't have Android Studio or Android SDK, but I look forward to installing the CI build on my phone!

selaux:
Here it is: https://storage.googleapis.com/ja2-builds/pull-requests/1240/ja2-stracciatella_-1240pullrequest+08f98b5_android.apk

selaux:
First iteration is done for me now. Please review. I will create follow up tickets for things that I noticed (or left unfinished in this PR).

src/sgp/SGP.cc
```diff
@@ -542,9 +572,53 @@ int main(int argc, char* argv[])
 	delete cm;
 	GCM = NULL;
 
+#ifdef __ANDROID__
```
ja2-derek, line 575: Probably all other platforms could use a try-catch in the main function

android/app/build.gradle
```diff
@@ -0,0 +1,87 @@
...
+    sourceSets {
+        main {
+            assets {
+                srcDirs = [ "../../assets" ]
+            }
+        }
+    }
+}
+
+dependencies {
```
ja2-derek, line 66: Can we use the pre-packaged SDL_android library from Maven? Or do we have some specific patches in the SDL source code?

src/sgp/SGP.cc
```diff
@@ -342,15 +349,35 @@ int main(int argc, char* argv[])
 		}
 	}
 
+	#ifdef __ANDROID__
+	JNIEnv* jniEnv = (JNIEnv*)SDL_AndroidGetJNIEnv();
+	if (jniEnv == NULL) {
+		SLOGE("Failed to get jni env");
+	}
+	RustPointer<char> configFolderPath(EngineOptions_getStracciatellaHome(jniEnv));
```
ja2-derek, line 357: Can we have something like `InitJNI` function at Rust side, then the Rust side remembers the jniEnv pointer. That way we don't need to have whole set of new functions or pass the pointer every call?

dependencies/lib-sdl2/builder/Bug5100.patch
```diff
@@ -0,0 +1,33 @@
+--- ./CMakeLists.txt	2020-03-10 18:36:18.000000000 -0700
```
ja2-derek, line 1: What a shame there is still no new SDL release!

ja2-derek:
Great work! I know nothing about Android nor Kotlin, though.

It runs quite smoothly. There are only very minor slowdowns (less than a second, before entering turn based mode). The graphics and sounds work well. Probably just need to work out the right click and maybe screen scrolling. It took me a while to learn to scroll.

lynxlynxlynx:
Yeah, good stuff. Can't be of much use with review though.

dependencies/lib-sdl2/builder/Bug5100.patch
selaux: True. But we can change easily to the new release when it arrives.

android/app/build.gradle
selaux: I think SDL_android on maven is a different SDL (SmartDeviceLink). And yes we apply a single patch to SDL.

src/sgp/SGP.cc
selaux: I changed it to apply to all platforms.

src/sgp/SGP.cc
selaux: That would be nice, but unfortunately it is not that easy as in rust the JNIEnv has an internal lifetime. But I'll take another shot at this when refactoring the Android asset VFS layer...

ja2-derek:
LGTM!

selaux:
Will merge when build is done. I also created three follow up tickets:

- https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1248
- https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1249
- https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1250

PR #1243, opened 2020-09-27
Cleanup ci-setup.sh

selaux:


.ci/ci-setup.sh
```diff
@@ -0,0 +1,84 @@
...
+env
+which rustc
+rustc -V
+which cargo
+cargo -V
+cargo clippy -- -V
+cargo fmt -- -V
+which cmake
+cmake --version
+command -v sccache && sccache -V || echo "sccache not available"
```
ja2-derek, line 78: Now that we have `sccache` on all platforms, we probably don't need the existence check.

selaux: :+1: Removed.

selaux:
Giving this up, I dont have the patience for this anymore....

ja2-derek:
I like your idea here. Mind if I try to continue from here?

selaux:
Nope, go for it

ja2-derek:
This is done now.

PR #1244, opened 2020-09-29
Externalization and generalization of M.E.R.C.

ja2-derek:
This PR externalizes (#665) and generalizes the merc profile handling in M.E.R.C. The goal is to allow mods to add new mercs in M.E.R.C., and let them change the availabilities (e.g. make everybody available right from the start). Quite a few parts of the M.E.R.C. code was re-written.

## What's changed

- Soldiers' availability on M.E.R.C. (`CONTITION_FOR_MERC_AVAILABLE`, not my typo) are fully controlled from JSON data
- Speck's merc-specific quotes are also controlled from JSON data
- The save game data field `ubLastMercAvailableId` is no longer used

## Larry handling

Larry (`LARRY_NORMAL` and `LARRY_DRUNK`) handling has been quite ugly. The old code has 2 Larry profiles in the MERC array, and will use only of them depending on whether Larry has relapse or not. In some places (e.g. Mercs_Files page), `LARRY_DRUNK` is always skipped.

This PR changes Larry handling to keep the special logic away from the data. Now the MERC data array only have a single entry pointing at `LARRY_NORMAL`, but the code will check for Larry's relapse, and return `LARRY_DRUNK` if we are asking for Larry and he has relapsed.

However, this also means the saved index will be off by one. The later M.E.R.C. hires, `NUMB` and `COUGAR` had indices 9 and 10 in old code. After this PR, their indices will become 8 and 9. This PR also included a load game handler to address this off-by-one difference. However, it cannot handle loading of new saves from old versions. In that case, the old version game will not crash, but the later mercs might become unavailable and but should be available again in a few days again.

## Testing 

This is going to take me some time to test all the M.E.R.C. functions. I am drafting this PR first while I am still testing. Let me know if you object to this, or spot some stopper issues.

### DIFF
--- a/assets/externalized/mercs-MERC-listings.json
+++ b/assets/externalized/mercs-MERC-listings.json
@@ -1,0 +1,128 @@
+/**
+ * This file defines the profile listings on the M.E.R.C. website, and also some of the things Speck says.
+ *
+ * This only controls how the profiles show up on M.E.R.C. and not the soldier profiles themselves.
+ *
+ * Fields:
+ *  - profileID:        The profile ID of the merc
+ *  - minTotalSpending: The player must spend this amount before this merc is available for hire; Defaults to 0
+ *  - minDays:          The day (in-game days) after which this merc may be available; Defaults to 0
+ *  - quotes:           A list of quotes that Speck make regarding this merc
+ *     - quoteID:       ID of the quote. See Speck_Quotes.h
+ *     - profileID:     ID of the other merc. Only used in CROSS_SELL quotes
+ *     - type:          Type of quote:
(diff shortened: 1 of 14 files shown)


ja2-derek:
Tested OK and also tried saves before and after this PR. Quick recap:

- This PR fixes an vanilla bug where `SPECK_QUOTE_ALTERNATE_OPENING_4_TRYING_TO_RECRUIT ` check was not correct. It should be checking for the next available merc.
- Loading old saves in new version works fine
- Loading new saves in old version when `NUMB` or `COUGAR` is already available - the last available will be unavailable right after loading, but will be available again in a couple of days.

lynxlynxlynx:
I did a cursory read and it looks fine. 👍🏾

PR #1137, opened 2020-07-04
Fix time consuming logging

selaux:
Fixes #1136 

Should optimize logging performance in general:

- Dont build log record when it is filtered anyways
- Pass filename via Macro to Rust instead of building it from the full path in rust

PR #1246, opened 2020-10-02
Cleaning up some C++ code

ja2-derek:
Fixes several compiler warnings, and apply the correct practices.

## Highlights

- Fixes the compiler warnings in Mac builds (https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1137#discussion_r453317787)
- Replaces `std::shared_ptr` with `std::unique_ptr` whenever applicable. `std::unique_ptr` should be used if pointer is to be owned by one. It also have much less overhead
- Fixes a number of compiler warnings


selaux:
:+1: Thanks.

PR #1247, opened 2020-10-02
Supplementary data to Merc Profiles

ja2-derek:
This is a proposal to externalize more merc data from code to JSON. (#665)(https://github.com/ja2-stracciatella/ja2-stracciatella/issues/289#issuecomment-207977449)

# Motivation

Most of the merc profile info is in `PROF.dat` and that is fine. But there quite a lot more hard-coded information scattered in the code, for example the [FIRST_RPC](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/7d66aa47554c125a250db5d27dd4f77acb513ae8/src/externalized/content/Dialogs.cc#L26), [mercID < BIFF](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/614ec847c9aca357b49308b0dd60f61997b12c6f/src/game/Tactical/Merc_Hiring.cc#L176), [mercID <= BUBBA](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/614ec847c9aca357b49308b0dd60f61997b12c6f/src/game/Tactical/Merc_Hiring.cc#L203) tests.

We need to generalize and externalize it if we were to add new mercs. For example, Unfinished Business re-purposes the profile num of `IRA` and `CARLOS` as new M.E.R.C. soldiers, and a new IMP voice is added but the ID is greater than `LAST_IMP_MERC`.

# Proposal

- Add a JSON data entry for each merc profile. It does not overlap with `PROF.DAT`, but has more information extracted from codebase. This data is static and never changes at run-time. 
- Do not touch `MERCPROFILESTRUCT` as it maps directly to `PROF.DAT` and save games.
- Use the [MercProfile](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/f9e420497266833dd4018350bcd9dd2fe6c68435/src/externalized/mercs/MercProfile.h) class to provide access to the supplementary info
- Add an implicit cast from `MercProfile` to `MERCPROFILESTRUCT&`, so that we can return `MercProfile` by default at places like [GetProfile](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/614ec847c9aca357b49308b0dd60f61997b12c6f/src/game/Tactical/Soldier_Profile.h#L14), and it will not break anything that expects `MERCPROFILESTRUCT`
-  At the very first stage, there is only one data field "type", which is one of AIM/MERC/RPC/NPC.  We  can replace [checks](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/162a4fbc7926d55abe8d35fdbfd37d73d67191f1/src/game/Tactical/Soldier_Profile.cc#L1073) that relies on `<= BIFF`, `FIRST_NPC`, `FIRST_RPC`, etc
- Later on, we can add more attributes, such as [isTerrorist](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/162a4fbc7926d55abe8d35fdbfd37d73d67191f1/src/game/Tactical/Soldier_Profile.cc#L889), `isAssassin`, [isHeadMiner](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/162a4fbc7926d55abe8d35fdbfd37d73d67191f1/src/game/Tactical/Soldier_Profile.cc#L899) etc


ja2-derek:
This is ready for review.

selaux (approved):
:+1: Looks good! I found a few nitpicks, but this is way better than before.

One thing I'm not sure of is the naming of `type` which is kind of a mixup of `type` as in Human, Robot, Vehicle and the recruitment part as in AIM, MERC, Dialog, etc. Might make sense to split them in the future, but for now I'm also fine with `type`.

lynxlynxlynx:
Thanks!

PR #1251, opened 2020-10-06
Externalizing M.E.R.C. profile bio index

ja2-derek:
The index/offset of M.E.R.C. profile text in `mercbios.edt` was determined by `profileID - BIFF`. It assumes the mercs are in a continuous range after `BIFF`. 

This PR externalizes the bio offset in JSON.

### DIFF
--- a/assets/externalized/mercs-MERC-listings.json
+++ b/assets/externalized/mercs-MERC-listings.json
@@ -5,6 +5,7 @@
  *
  * Fields:
  *  - profileID:        The profile ID of the merc
+ *  - bioIndex:         The start location offset to the merc's bio in the mercbios.edt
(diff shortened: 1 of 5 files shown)


lynxlynxlynx:
LGTM, but one question: why is there no bioIndex of 7?

ja2-derek:
`bioIndex` 7 would have been `LARRY_DRUNK`, but Mercs_Files.cc had [special logic](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c470204ef067529a30eb426745e9431b0a834cfb/src/game/Laptop/Mercs_Files.cc#L247) to skip LARRY_DRUNK and it is never used.

Issue #1248, opened 2020-10-03
Android Asset VFS layer improvements


selaux:
- [x] Make VFS case insensitive
- [x] Support mods from assets dir
- [ ] Get rid of ndk fork
- [x] Store JNI pointer internally

lynxlynxlynx:
Is this / the ndk thing still relevant?

selaux:
Yes. I'm actually currently looking into this. Looks like its working now after the linking update for SDL 2.0.20. Will open a PR soon.

PR #1256, opened 2020-10-09
Android asset manager rehault - Part 1

selaux:
This fixes 2 out of 4 issues in #1248.

- Makes Android Asset VFS case insensitive
- Stores pointer to JNI globally in Rust

One thing to note is thatI needed to implement case insensitive matching via JNI, as it is not possible to list asset directories through the NDK asset manager. It can only list files in directories not directories.

Supporting mods from assets is gonna be a little trickier, as mods may also contain SLF files, so we have an SLF VFS within another VFS.

### DIFF
--- a/src/sgp/SGP.cc
+++ b/src/sgp/SGP.cc
@@ -349,13 +349,12 @@ int main(int argc, char* argv[])
 
 		#ifdef __ANDROID__
 		JNIEnv* jniEnv = (JNIEnv*)SDL_AndroidGetJNIEnv();
-		if (jniEnv == NULL) {
-			SLOGE("Failed to get jni env");
+		
+		if (setGlobalJniEnv(jniEnv) == FALSE) {
+			SLOGE("Failed to set global jni env");
 		}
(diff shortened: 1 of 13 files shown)


src/sgp/SGP.cc
```diff
@@ -349,13 +349,12 @@ int main(int argc, char* argv[])
 
 		#ifdef __ANDROID__
 		JNIEnv* jniEnv = (JNIEnv*)SDL_AndroidGetJNIEnv();
-		if (jniEnv == NULL) {
-			SLOGE("Failed to get jni env");
+		
+		if (setGlobalJniEnv(jniEnv) == FALSE) {
+			SLOGE("Failed to set global jni env");
```
ja2-derek, line 354: What would happen if we continue we `jniEnv` being null? I mean should we be throwing an exception if it is not going to work?

selaux: Good point. Fixed now (and also when failing to find home directory).

selaux:
Any remaining comments about this or can I merge?

lynxlynxlynx:
LGTM!

PR #1260, opened 2020-10-10
Externalizing I.M.P. initial level

ja2-derek:
For #665. 

Allows I.M.P. mercs to start at higher levels, by changing the settings at `imp.json`. 

Added a field `initial_level` which has default of 1.


selaux:
:+1: LGTM!

PR #1261, opened 2020-10-11
Implement primitive value boxing for Lua listeners

ja2-derek:
# Motivation

The observable [BeforeStructureDamaged](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/63eb7a5b1a066db795dcb314ae72184539043f0b/src/game/TileEngine/Explosion_Control.cc#L363) is an example of sharing a variable with all listeners via a pass-by-pointer argument.

But such a simple pointer to primitive does not work as intended with Lua. In Lua, all primitives arguments (including pointers to primitive types) are always passed by value and all complex types are passed by reference. To implement pass-by-reference, I have created a simple struct to wrap around the primitive type. You can think of it like the variable boxing in [Java](https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html) or [C#](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/types/boxing-and-unboxing).

# Usage

The wrapper struct (e.g. `BOOLEAN_S` wraps a `BOOLEAN`) is implicitly convertible to and from the underlying data type. So it can used just like the original type. e.g.

```c++
BOOLEAN_S fSomeBoolFlag = true;
if (fSomeBoolFlag) DoSomething();
```

In Lua, it is passed as an user object, and it must be accessed via the `.val` member field.

```lua
function (bool_s_arg)            -- listener to Observable<BOOLEAN_S*> 
    if bool_s_arg.val == 1 then  -- cannot use true/false, since BOOLEAN is uint8_t / unsigned char
        do_something()
    end
    bool_s_arg.val = 0           -- set the value to FALSE and pass back to C++
end
```

### DIFF
--- a/src/externalized/scripting/FunctionsLibrary.h
+++ b/src/externalized/scripting/FunctionsLibrary.h
@@ -51,7 +51,7 @@ extern Observable<> BeforePrepareSector;
  * @param fSkipDamage damage processing will be skipped if it is set to TRUE
  * @ingroup observables
  */
-extern Observable<INT16, INT16, INT8, INT16, STRUCTURE*, UINT32, BOOLEAN*> BeforeStructureDamaged;
+extern Observable<INT16, INT16, INT8, INT16, STRUCTURE*, UINT32, BOOLEAN_S*> BeforeStructureDamaged;
 
 /**
   * Callback just after a structure has just been damaged by explosives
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
Nice work, you're creating quite a framework.

ja2-derek:
If you are interested, you can check out my branch for [unfinished business](https://github.com/ja2-derek/ja2-stracciatella/tree/unfinished-business/assets/mods/unfinished-business/data/scripts) mod. I pick the commits into master if it is general enough.

PR #1262, opened 2020-10-14
Externalizing strategic map secrets

ja2-derek:
This PR externalizes (#665) map secrets to JSON data files. Map secrets in base game are: Tixa, Orta and the 4 SAM sites. They are hidden at the start of game, and revealed to the player by NPCs or when arriving at the sectors.

# Summary

- Moving the traversibility mapping (from string name to enum ID) to its own file 
- Extracting map secrets information from code:
    - Sector (as primary key); Map icon; Terrain land type string
- Refactoring the sector description function (e.g. "Cambria Hospital", "Tixa Dungeon", "Farm, road")
    - Using the newly externalized data to determine land type for secret sectors
- Generalizing map secrets and save games:
    - `fFoundTixa` is generalized as "1st non-SAM site secret"
    - `fFoundOrta` is generalized as "2nd non-SAM site secret"
    - SAM site secrets are now dynamically defined by JSON, so that a mod can have any number of hidden SAM sites
    - NPC actions and quests still have direct references to TIXA and ORTA

## Limitations and assumptions

- The "is secret found?" state of only the first 2 secrets and SAM sites are saved - to fit `fFoundTixa` and `fFoundOrta`. A mod can define more but state persistence will not handled by the base game.
    - Some possibilities are: changing field from BOOLEAN to bitset, or allowing Lua scripts to write extra data to saves.
- There is at most one secret per sector
- A secret town can only have 1 town sector

# Possible use-cases 

_These should be feasible, but not tested. Testing was done with vanilla data_

- Replacing map secrets with other secret facilities on the map, and with new icons
- SAM sites that are known on game start (SAM sites without corresponding secret)
- Hide Estoni, until some NPC tells you about it
- Make the random weapon cache a map secret - that can be revealed by quests or NPCs
- Enabling [PATCH106](https://github.com/ja2-derek/ja2-ub-comparison/commit/84c0858f) 

# Testing

- Save/load of secret revealed states (Tixa, Orta and SAM sites)
- Revealing of secrets by NPC
- Revealing of secrets by visiting
- Sector string in bottom-right, saves and strategic map pop-up

### DIFF
--- a/assets/externalized/strategic-map-movement-costs.json
+++ b/assets/externalized/strategic-map-movement-costs.json
@@ -84,32 +84,5 @@
 	/* N */ [  0,  0, 80, 80, 80, 40, 20, 10,  5,  5,  0,  0,  0,  0,  0,  0 ],
 	/* O */ [  0,  0, 90, 90,  0,  0,  0,  5,  5,  0,  0,  0,  0,  0,  0,  0 ],
 	/* P */ [  0,  0,100,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0,  0 ]
-	],
-	// Maps the traversability rating "short names" into codebase enums. You probably don't need to alter this.
-	"mapping": {
-		"A":  3,   // SAND	
-		"AR": 23,  // SAND_ROAD
-		"C":  20,  // COASTAL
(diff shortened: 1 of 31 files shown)


selaux:
LGTM. Thanks.

PR #1264, opened 2020-10-15
Externalize sector land types

ja2-derek:
Data externalization (#665). Some sectors have special descriptions such as "Cambria Hospital" and "Orta Basement". This PR externalizes these data so mods can add sectors with new descriptions.

- Moving the translated land type strings (`pLandTypeStrings`) to JSON  
- Adding mapping of sectors to special land types in JSON
- Updating the sector description function to use JSON data


lynxlynxlynx:
LGTM, thanks!

PR #1265, opened 2020-10-16
Android VFS Rework Part 2

selaux:
Next step for proper AssetManagerFs support: Allow to load mods from Android assets.

This kind of got out of hand because now we have to support to create an `SlfFs` from another Vfs layer. I fixed this by introducing a trait for `VfsLayer` which allows to abstract away the `VfsLayer` (which also allowed me to remove a bunch of wrapper code). Currently it allows `open` for files and `read_dir` for directories. This allows to use `read_dir` on a VfsLayer to list all SLF files and `open` to open the `SlfFs` from the other VfsLayer. This is currently used for `DirFs` and `AndroidAssetFs`.

From the C++ side there were no changes needed, although I needed to wrap the `Box<VfsFile>` into another struct in the API, because it seems like we need to return concrete types to C++. But since this is an opaque pointer on C++ side anyways, it worked out fine.

Unfortunately there is no UI for this yet, so I tested by just adding the mods to `ja2.json` manually in the LauncherActivity. This would probably be the next step for proper Android support.

### DIFF
--- a/rust/stracciatella/tests/test_vfs.rs
+++ b/rust/stracciatella/tests/test_vfs.rs
@@ -4,11 +4,11 @@
 mod vfs {
     #[test]
     fn read() {
-        let (temp, dir) = create_temp_dir();
+        let (temp, dir, dir_fs) = create_temp_dir();
         create_data_slf(&dir); // data.slf
 
         let mut vfs = Vfs::new();
-        vfs.add_slf(&dir.join("data.slf")).unwrap();
+        add_slf(&mut vfs, &dir_fs, "data.slf");
         let data = read_file_data(&vfs, "foo.txt");
         assert_eq!(&data, b"data.slf");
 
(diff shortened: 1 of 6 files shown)


ja2-derek:
Hijacking the thread... sorry. But have you have tried it on an actual mod and how does it perform?

(Using the current master,) I have a mod (UB) with 20K+ files. Reading a game res is quite slow on my Linux laptop. It takes up to 60ms to `openGameResForReading` a BATTLESND file. When you have a team of 6 mercs, it is a 360ms freeze on click which is very noticable. I am yet to do a proper profiling, but I would guess it is the case-insensitive file search.

selaux:
When this is merged I will have a look at the remaining performance optimizations for the VFS, so feel free to merge if it looks good.

rust/stracciatella/tests/test_vfs.rs
```diff
@@ -80,25 +80,85 @@ mod vfs {
...
+        create_file(&dir.join("foo/foo2.txt"));
+        create_file(&dir.join("foo/foo3.txt"));
+
+        let mut vfs = Vfs::new();
+        vfs.add_dir(&dir).expect("dir");
+        add_slf(&mut vfs, &dir_fs, "foo.slf");
+        add_slf(&mut vfs, &dir_fs, "foobar.slf");
+
+        let root_paths = ["foo", "foo.slf", "foobar.slf"];
+        let foo_paths = ["foo3.txt", "foo2.txt", "foo1.txt", "bar.txt", "bar"];
```
derek9gag, line 98: Should we add a few test cases for non-ASCII directory names?

ja2-derek, line 98: Should we add a few test cases for non-ASCII directory names?

selaux: Good call. I added some more tests and prompty found a bug in the SLF directory listing logic. Nevertheless that should be fixed now and tests for non-ASCII directory (and file) names are there.

ja2-derek:
LGTM!

Issue #1268, opened 2020-10-17
VfsFile_open performance


ja2-derek:
Expanding https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1265#issuecomment-710732174 with profiling data. `VfsFile_open` is slow for me - I am on Linux (ext4 on SSD), with a mod that has over 20K files (extracted data files from UB). It freezes up to 60ms for every BATTLESND played (e.g. soldier acks).

## Findings

Below is a profiling session at the Tactical screen. I selected all mercs (`=`) and ordered them to walk about a few times. The majority of the time is spent on rendering (expected), and loading the sound (unexpected). One bottleneck is `DirFs` looking up files, at `stracciatella::unicode::Nfc::caseless`:

![image](https://user-images.githubusercontent.com/63151803/96328457-e2f72d80-1075-11eb-93d4-9998f8869c88.png)

---

`caseless` is doing the proper unicode "case folding", and it is understandably complex as it is supposed to work on every language. But isn't it an overkill for directory listing? I think we can just assume it's always ASCII and go for [eq_ignore_ascii_case](https://doc.rust-lang.org/std/string/struct.String.html#method.eq_ignore_ascii_case).

ja2-derek:
There is also this [unicase](https://docs.rs/unicase/2.6.0/unicase/) crate which claims to use the faster ASCII comparison when applicable

ja2-derek:
I did a quick test with UniCase. The latency reduced from 30-60 ms to 1-5 ms. 

```patch
diff --git a/rust/stracciatella/src/vfs/dir.rs b/rust/stracciatella/src/vfs/dir.rs
index 8c7b3cb37..71a35edde 100644
--- a/rust/stracciatella/src/vfs/dir.rs
+++ b/rust/stracciatella/src/vfs/dir.rs
@@ -5,6 +5,7 @@ use std::fmt;
 use std::io;
 use std::io::SeekFrom;
 use std::path::{Path, PathBuf};
+use unicase::UniCase;
 
 use crate::fs;
 use crate::fs::File;
@@ -48,6 +49,7 @@ impl DirFs {
                     "special path components are not supported",
                 ));
             }
+            let want_u = UniCase::new(want);
             for candidate in candidates {
                 if !candidate.is_dir() {
                     continue;
@@ -57,7 +59,7 @@ impl DirFs {
                     let is_match = path
                         .file_name()
                         .and_then(|x| x.to_str())
[...]

PR #1270, opened 2020-10-17
Only do unicode normalization on strings where it is necessary

selaux:
See  #1268

### DIFF
--- a/rust/stracciatella/src/unicode.rs
+++ b/rust/stracciatella/src/unicode.rs
@@ -61,7 +61,7 @@ use std::fmt;
 use std::ops;
 
 use caseless::Caseless;
-use unicode_normalization::{is_nfc, UnicodeNormalization};
+use unicode_normalization::{is_nfc, is_nfc_quick, IsNormalized, UnicodeNormalization};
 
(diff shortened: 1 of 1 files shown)


ja2-derek:
Thanks! For me, it halved the latency, from 60ms to about 30ms. It is a good improvement. `default_case_fold` might account for the remaining half of the time spent.  

With a naive string-to-lower, the latency can be down to 1ms or less.

ja2-derek:
How about checking [is_ascii](https://doc.rust-lang.org/std/string/struct.String.html#method.is_ascii), and then do the naive `to_ascii_lowercase`? I believe this can optimize the performance for 99% of the cases.

selaux:
Added to_ascii_lowercase. I'd still not close #1268 though. I would still want to check how it behaves on a HDD instead of an SSD, as there listing directories might be the slowest part.

ja2-derek:
Great! It is now down to 1ms!

selaux:
Yeah I actually was thinking about the same (cache), but lets optimize all the acesses first and then check how fast we are. Maybe we can avoid caching...

PR #1271, opened 2020-10-20
Some more Android fixes

selaux:
I tested the Android build again on all my devices. This resulted in some fixes that were required. Most of them are self-explanatory, but there are two notable ones:

- Current Stracciatella Version is read from file: This keeps versions in android and cmake builds in sync. For android builds that are released it is still necessary to increase `versionCode` for releases though. This might require an uninstall and reinstall for the Android app after this PR is merged (because versionCode is now lower than before).
- JNI Local References are cleaned up properly: This caused issues for me on older SDKs (e.g. SDKv19). According to the [JNI Reference](https://docs.oracle.com/en/java/javase/12/docs/specs/jni/design.html#referencing-java-objects) all references created by native code are considered local references and are only cleaned up when a foreign method exits. In our case this only happens when the native application quits. Older Android versions seem to have a limit of 512 local references, which is not remotely enough for all our local references, so we need to ensure that they are cleaned up. jni-rs gives us [AutoLocal](https://docs.rs/jni/0.17.0/jni/objects/struct.AutoLocal.html) and [.with_local_frame](https://docs.rs/jni/0.17.0/jni/struct.JNIEnv.html#method.with_local_frame) for this, which I added to the JNI functions.

The apk works now again for me on all tested systems, but https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1268 is really noticeable on Android now as well...


android/app/build.gradle
```diff
@@ -17,8 +17,10 @@ android {
         applicationId "io.github.ja2stracciatella"
         minSdkVersion 19
         targetSdkVersion 26
-        versionCode 18
-        versionName "0.18.0"
+
+        File versionFile = file('../../version')
+        versionCode 17
```
ja2-derek, line 22: Shouldn't `versionCode` be just an ever incrementing number? Could we use the CI build number for this (something like `GITHUB_RUN_NUMBER`)?

selaux: Kind of, yes. From Android docs:

> A positive integer used as an internal version number. This number is used only to determine whether one version is more recent than another, with higher numbers indicating more recent versions. This is not the version number shown to users; that number is set by the versionName setting, below. The Android system uses the versionCode value to protect against downgrades by preventing users from installing an APK with a lower versionCode than the version currently installed on their device.

> The value is a positive integer so that other apps can programmatically evaluate it, for example to check an upgrade or downgrade relationship. You can set the value to any positive integer you want, however you should make sure that each successive release of your app uses a greater value. You cannot upload an APK to the Play Store with a versionCode you have already used for a previous version.

I think using `GITHUB_RUN_NUMBER` would not suit this very well, e.g. if I want to install an older nightly version of the same minor release I'd have the problem that its version number is smaller than the one installed, so I'd have to remove it first. I think its better to just increase it for each release.

What I can think of is to just build it from the version as follows:
- Pad each version (MAJOR, MINOR, PATCH) with some Zeros (lets say 4) and concatenate them: `000000170001`
- Parse that into a number: `170001`

That would return e.g.:
- `0.17.1` -> `170001`
- `0.17.2` -> `170002`
- `0.17.1000` -> `171000`
- `0.18.0` -> `180000`
- `1.0.0` -> `100000000`
- `1.1.1` -> `00010001`

So it would be inceasing for each release.

selaux: NB:

> Warning: The greatest value Google Play allows for versionCode is 2100000000. 

So the example above would work til version 21.

selaux: Added this logic, but with padding to 3 characters, so max semver version is `999.999.999` and version code `999999999`.

lynxlynxlynx:
Sounds good to me.

android/app/build.gradle
ja2-derek: LTGM!

PR #1273, opened 2020-10-23
Fix some memory leaks

selaux:
Fixes some memory leaks found with the `valgrind` tool. This is especially important for Android, as we use `ja2-stracciatella` as a library there, so memory leaks will accumulate over time. I omitted leaks in `SoundMan.cc`, as I have another branch (that I wanted to actually test for memory leaks) that does a big refactoring there...

### DIFF
--- a/src/game/TileEngine/Render_Dirty.cc
+++ b/src/game/TileEngine/Render_Dirty.cc
@@ -350,11 +350,13 @@ void InvalidateBackgroundRects(void)
 
 void ShutdownBackgroundRects(void)
 {
-	for (UINT32 i = 0; i < guiNumBackSaves; ++i)
+	for (auto backgroundSave : gBackSaves)
 	{
-		BACKGROUND_SAVE* const b = gBackSaves[i];
-		if (b->fAllocated) FreeBackgroundRectNow(b);
+		if (backgroundSave->fAllocated) FreeBackgroundRectNow(backgroundSave);
+		delete backgroundSave;
 	}
+	gBackSaves.clear();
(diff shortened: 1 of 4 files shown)


src/game/TileEngine/Render_Dirty.cc
```diff
@@ -355,6 +355,12 @@ void ShutdownBackgroundRects(void)
 		BACKGROUND_SAVE* const b = gBackSaves[i];
 		if (b->fAllocated) FreeBackgroundRectNow(b);
 	}
+	for (auto elem : gBackSaves)
+	{
+		delete elem;
```
ja2-derek, line 360: Why don't we just do `delete b` right after line 356?

selaux: Because the size of the vector might be larger than `guiNumBackSaves`. `guiNumBackSaves` is basically the number of background rects that the game is currently using, while the vector is resized in increments of 100 [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/1dabffb17d4d732a8279597b613fbe6e35b4d54e/src/game/TileEngine/Render_Dirty.cc#L116).

ja2-derek: I see. In that case, should we move the `FreeBacgroundRectNow(b)` to right before line 360? There is a check for `b->fAllocated` so it should  be safe to run on all of gBackSaves.

selaux: Sounds good. Done.

Issue #664, opened 2018-01-04
moddable music error - not enough memory


JAsmine-ja2:
![music](https://user-images.githubusercontent.com/17213795/34560737-6115e0e0-f147-11e7-9a8e-8a655134dbff.jpg)

Adding custom music works - thanks to #387. AFAIK, we're restricted to the wav format right now. Adding some tracks, I ran into the error as seen in the screenshot.

EDIT: I can upload the wav file if it helps. It is around 16MB.

Manuel-K:
Argh.

Related to #609.

Manuel-K:
Can you confirm that you have SDL 2.0.7?

Manuel-K:
On the other hand – the SDL version does not matter that much. The main problem is that we load all sound files (even music) completely. 

Solutions: 
 - Increase the sound buffer once more. Ugly. Easy workaround.
 - Stream large sound files. Tedious.

JAsmine-ja2:
I have replaced sdl2.dll with this one: 
https://www.libsdl.org/download-2.0.php
which is version 2.0.7 which is -not- the same file as in the stracciatella folder. I have not messed with this file before.
However, it does not change anything. Same error.

JAsmine-ja2:
Another idea: If we had support for OGG and/or MP3, filesizes would be much smaller...

Manuel-K:
No, that would not help. The problem is not the size on the disc, but the size after conversion to wav with correct sample rate.

Issue #64, opened 2013-08-08
OGG audio files support


lynxlynxlynx:
Originally reported by: **Gennady Trafimenkov (Bitbucket: [gennady](http://bitbucket.org/gennady), GitHub: [gennady](http://github.com/gennady))**

---

It would be great to implement OGG audio files support.
It might come very handy for ModPack project (see wiki about ModPack).

OGG is open-source friendly format.  It should be a better choice than MP3.

OGG decoder preferably should be licensed under one of these licenses: LGPL, MIT, BSD 3-clause.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/64

flaviojs:
SDL_mixer supports OGG.

PR #1274, opened 2020-10-23
Rework SoundMan.cc to stream audio if necessary

### LINKED ISSUE #1249: Game hangs when loading a new music track
### What is wrong

Currently whenever a new piece of music is loaded. This is not very noticeable on a desktop device, but on Android this hang lasts about 1-2 seconds on some devices.

My Guess is that this is caused by the `SoundConvertBuffer` logic: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/sgp/SoundMan.cc#L561.

We should think about how to implement sound streaming for music pieces.

|  0.17.0/git | any  | no    |    yes     |

selaux:
So here is another big one that should close #1249, #664 and #64. THB C++ is not my strong suit, but I tried to write everything as cleanly as possible, so I'm keen to get some feedback. I chose not to use any different framework for playback, but for decoding and converting file formats I went for miniaudio, because this was the easiest to include in the source code and does not have any additional dependencies required.

Basically I reworked the whole sound loading / callback logic:

- We still have 16 channels and 128 possible loaded sound samples as  before
- A sample can now be one of the following:
  - A handle to a file and a miniaudio decoder to decode this file into our desired output format
  - A memory buffer and a miniaudio decoder to decode this buffer into our desired output format. This is used for files `< 1MiB`, so we dont do too much file io if not really necessary
  - A memory buffer (which now is raw audio data) and a miniaudio converter to convert into our desired output format. This is currently only used for the playback of Smacker files
- A channel now contains a lockfree ring buffer that holds decoded and converted data (size 256kb). So data from this ring buffer can simply be mixed and written out to the sound callback buffer, which simplifies the `SoundCallback` (no need to check for channels anymore)
- There is a new thread in order to offload the decoding / converting computation from the main thread: `SoundManBufferServiceThread`
  - This thread will be woken up, when the `SoundCallback` notices that one of the buffers of the channels it read from is less than half filled
  - It will then iterate over all channels and fill their ring buffers back up, based on the samples of the channels
  - When no servicing of the buffers is needed the thread should sleep idly in the background
- I removed the bookkeeping of the sound memory. Currenltly memory is limited by the number of loadable sound samples and the threshold between in-memory and streaming of files. So the maximum size should be `128MB`. This is way higher than the previous limit of `32MB`, but from the profiling I have done, the previous implementation went way above this limit during decoding / conversion of the data. A short test with `valgrind` showed me peak memory of `155MB` when starting a game and entering and exiting tactical. Now memory allocation is way more stable and there are no high peaks anymore. Same goes for CPU cycles used. The required work is stretched over a larger timespan with the RingBuffer.
- Miniaudio Decoding supports the following formats: `wav`, `mp3` , `ogg` and `flac`. So those formats can now be used by mods. I updated the `o-fortuna` mod to use a `mp3` file to demonstrate this.

NB: I dislike the current structure of `SoundMan.cc` quite a lot, but did not decide to rewrite it a lot, as I'm not the best C++ dev. If it was up to me alone, I'd like to try to rewrite the whole thing in Rust, just to see if its possible to do a whole game subsystem in Rust, but I'm not sure how much you would like that :wink:


ja2-derek:
So now that we determine whether to stream by file size, is `SoundPlayStreamFile` no longer necessary?

Also, can we and should we unify the API for streaming and non-streaming modes? Can we stream everything, or is there some other concerns like performance?

selaux:
Yes. `SoundPlayStreamFile` is no longer necessary. It is just an alias to `PlaySound` (which it also was before). So the API could be cleaned up...

Regarding streaming all sounds: Yes that would be possible (see commit https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1274/commits/fad19bf7b14687d445dbe92793cb9226853bc4a0, where that is actually the case). In the end I decided against it, because for small audio files that are played quite often (e.g. sounds for button clicks) that would mean file access for each play of the sound. To me file access is expensive and memory is mostly a non-issue on newer systems (at least if we look at the 128MB limit on the sound manager). I did not do any exhaustive profiling to validate that assumption though...

selaux:
Regarding the build failure: MinGW does not seem to support synchronization primitives provided by C++11 (although newer versions might have it, but I am not quite sure which). My suggestion would be to conditionally add https://github.com/meganz/mingw-std-threads/tree/master/cmake_stdheaders_generator for MinGW for the time-being.

src/sgp/SoundMan.cc
```diff
@@ -467,6 +463,111 @@ void SoundStopAllRandom(void)
...
+			for (UINT32 i = 0; i < lengthof(pSoundList); i++)
+			{
+				SOUNDTAG* Sound = &pSoundList[i];
+				if (Sound->State == CHANNEL_PLAY) {
+					FillRingBuffer(Sound);
+				}
+			}
+			fBuffersNeedService = FALSE;
+		}
+		lk.unlock();
```
ja2-derek, line 568: Do we know, in practice, how long is this critical section?  If I am understanding correctly, the main thread will wait on this mutex to update the `fBuffersNeedService` flag. 

My next question is - do we really need a strict mutex lock here? For example, is a simpler `std::atomic` flag enough? Or can we release the lock earlier? The worst case is that the main thread will update the `fBuffersNeedService` during the servicing, and that will trigger another servicing right away.

selaux: Small correction: The main thread (where the game logic runs) is not involved in this anymore. It is only the `SoundManBufferServiceThread` and the audio callback, which afaik also runs in a separate thread (but I actually cannot find the documentation for this fact anymore :slightly_frowning_face:).

I think you are right with the assumption that, if this section takes too long, we block the audio callback thread, which is bad an can result in distorted audio. And the runtime depends quite a lot on the hardware used, as e.g. file reading is involved. We could use `mutex.try_lock` in the sound callback in order to mitigate this issue, because if the mutex is locked already, that just means we already are servicing the buffers and dont need to change `fBuffersNeedService` anyways.

Regarding whether we need the mutex: I think it is currently only required for the `condition_variable`, so we can make the `SoundManBufferServiceThread` wait (without using CPU resources) until there is actually something to service. I am not sure how to achieve this using other means, but if you have suggestions on how to make the thread wait without `std::condition_variable`, let me know...

ja2-derek: I tried digging into the documentation and Google, and came to the same conclusion as yours. It looks like `std::condition_varable` is the only way to do this (up to C++ 14), and thus a mutex must be involved. 

Then I think the current implementation is good as long as  the critical section is made as short as possible.

There is actually no multi-thread locking and synchronization. It may be a use-case of `std::async`, so that we make an async future to service the buffer whenever buffer is low.

selaux: I changed the sound callback to use `try_lock`, so the length of the critical section should not matter anymore...

src/sgp/SGP.cc
```diff
@@ -629,6 +631,7 @@ void TerminationHandler()
 			SLOGE("Game has been terminated due to an unknown error");
 		}
 	}
+	shutdownGame();
```
ja2-derek, line 634: Just curious - what happens if it is not explicitly shut down? I thought the OS would automatically clean up everything

selaux: I the case of any other system than Android: Yes. In the case of Android: As the game is used as a library there, the OS does not clean up anything. So that means that the SoundCallback thread and FillBuffer thread continue to run and thus the nice JA2 music continues to play 😉 .

But right now I missed some cases where it is not cleaned up yet (because the `TerminationHandler` returns early).

selaux: But in general I think it makes sense to try shut down in this case, because it e.g. also cleans the temp dir and things like that. It just leaves everything in a cleaner state...

lynxlynxlynx: And memory checkers don't complain about still bound memory on exit.

selaux: True. I refactored the `TerminationHandler` a little, now it should clean up in all situations...

ja2-derek (approved):
Briefly tested on Windows with OGG-vorbis and MP3 music. Everything seems to work fine.

lynxlynxlynx:
Then I think we have all platforms covered already. :)

I'm not fully up to speed atm, but regarding the not-started refactoring possibilities in rust: surely it's possible, since there are even rust-only game engines out there: https://veloren.net/

selaux:
Will merge now as is and maybe refactor some more in a separate PR.

Issue #1255, opened 2020-10-08
Excessive temp file usage in .ja2/tmp/temp


misanthropos:
Working on an gentoo ebuild for ja2-stracciatella-0.17.0 I stumbled over a rather old issue - and hoped it would be fixed by
[PR#1220](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1220) - it is not.

Loading a SaveGame takes quite some time - if your home directory is a NFS share in Linux.

I forgot about that because one can simply link .ja2/tmp/temp -> /tmp and the problem is gone.

Reading about this fix however lets me hope it should not be too much trouble to reduce loading time because my believe is this may a related issue.

ja2-derek:
I tried with some added debug logging - unfortunately it might not be that straight-forward... this is what I get:

<details>
<summary>Debug logs at each file open</summary>
<pre>
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/i_A2.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/r_A2.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/m_A2.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/d_A2.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/v_A2.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/ds_A2.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/c_A2.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/i_A9.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/r_A9.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/m_A9.dat at FileMan::openForWriting
2020-10-09T01:33:04 [ INFO] sgp\FileMan.cc: Opening temp/d_A9.dat at FileMan::openForWriting
[...]

PR #1275, opened 2020-10-24
Add LRU cache for directory listings in canonicalize

selaux:
This implements a caching layer for file path canonicalization in the VFS. Basically it cached all `fs::read_dir`/`android::list_asset_dir` calls in their original and NFC variant. I decided to cache the listings and not actually the resolved files, as with a single `read_dir` this allows us to resolve multiple files via cache already.

For me this reduces the time to read a file from ~100000ns to ~5000ns (given the cache has been filled to an extent).

@ja2-derek Can you test this as well and let me know how much that improves the situation for you?

NB: One thing I'm not quite sure about yet is the `Temp/` dir within the `Data` dir. I still have to find out what exactly is written there and why...That is why this is still a draft.

References #1268


ja2-derek:
Briefly tested on my set up and it is in line with your results. The `VfsFile_open` used to take up to 3000 μs, and with this PR it is down to 100 μs. So with a squad of 6 soldiers, the improvement is ~20ms per click down to 0.6ms per click.

selaux:
So this got bigger again. I cleaned up all the file accesses, because they were kind of a mess: Sometimes stuff was accessed using absolute paths, sometimes relative. Files in ja2 home were accessed using relative paths to the current working directory (e.g. `../ja2.set`). In this PR I introduced three ways to open files with associated sets of functions in the content manager:

- Private files (reside in the home directory)
- Temp files
- Resource files

This should make it easy to e.g. implement https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1255 in the future.

There are a big open question though: Previously when saving stuff in the editor it was going to the vanilla game directory. This seems wrong to me as the user probably does not want to overwrite vanilla stuff, but rather create a mod or something similar. How should we handle this? A few options come to mind:

- Store in the mod directory with highest priority
- Let the user choose freely where to store
- Store somewhere in home directory, so user can copy to mod

lynxlynxlynx:
I propose two extra settings for the editor:
- boolean whether to save in the game dir, changing file directly, or a separate one
- file path for option 2, with a default to a new mod folder in the home directory

lynxlynxlynx:
So, how much more is missing here?

selaux:
I kind of got lost while implementing the editor changes.

I think I need to split this one up into:

- Making all file reads/writes happen through GCM
- Adding LRU cache

Some of the functionality in this PR has also already been introduced in master, so I have to merge them into these ones (e.g. https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1324). So this PR still has some way to go. Maybe I can find some time this week or next week to finalize.

selaux:
Sorry I didn't get around to it yet. Too many merge conflicts, which need resolving...

selaux:
Close in favor of https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1402

Issue #1276, opened 2020-11-07
TravisCI Mac builds no longer free


ja2-derek:
Heads up to the new TravisCI pricing model. I understand they no longer run Mac builds for free.

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

Maybe we should look at building nightly with GitHub action again 

[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

lynxlynxlynx:
Appears so, let's migrate.

ja2-derek:
I believe the easiest way is to make the `nightly` push with a Personal Access Token (see also https://github.com/ja2-stracciatella/ja2-stracciatella/issues/873#issuecomment-723438614). To proceed, I need an org admin to add the tokens as [encrypted secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets):

1. [Personal Access Token](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token) from any user with write repo access (maybe @madlab-bot ?). This token is used to push `nightly` branch, or directly run the workflow via GitHub API if push doesn't work.

2. Coverity scan token. We can migrate the coverity build into another scheduled GitHub Action. Something like [this](https://github.com/ruby/actions-coverity-scan/blob/master/.github/workflows/coverity-scan.yml).

ja2-derek:
A Coverity scan was successfully submitted earlier today ([build](https://github.com/ja2-stracciatella/ja2-stracciatella/actions/runs/390820585)). But the Coverity dashboard is still not updated, and the [badge](https://scan.coverity.com/projects/8431/badge.svg?flat=1) still says "Coverity Failed".

Are you able to get any further information from Coverity side?

lynxlynxlynx:
You should have the same access, but anyway, it says the last ok build was in october and anything later failed due to compilation errors:
```
Last Build Status: Failed. Your build has failed due to the following reason. Please fix the error and upload the build again.

Error details: 
invalid literal for int() with base 10: "tail: cannot open '/opt/workspace/ja2-stracciatella_ja2-stracciatella/cov-int/build-log.txt' for reading: No such file or directory
```
Like the results weren't successfully submitted, even though the upload was fine.

ja2-derek:
[...]

PR #1277, opened 2020-11-15
Nightly build with GitHub actions

ja2-derek:
Drafting this to share the work for #1276 that I have done so far. It does not work yet but I have no way to test it right now.

The coverity build flow is taken from  https://scan.coverity.com/scripts/travisci_build_coverity_scan.sh; I can't find any official instructions for GitHub actions.

----

This PR assumes the presence of 2 repository/org secrets:
- `BOT_GITHUB_TOKEN` which is a user access token belonging to @madlab-bot with permissions to merge commits into `nightly`
- `COVERITY_SCAN_TOKEN` which is the token to submit Coverity builds

Can @selaux please to add those as [organization secrets](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets#creating-encrypted-secrets-for-an-organization)?


.github/workflows/coverity-scan.yml
```diff
@@ -0,0 +1,68 @@
...
+                -O cov-analysis-linux64.tar.gz
+          mkdir cov-analysis-linux64
+          tar xzf cov-analysis-linux64.tar.gz --strip 1 -C cov-analysis-linux64
+          export PATH=$(pwd)/cov-analysis-linux64/bin:$PATH
+
+      - name: Fixed world writable dirs
+        run: |
+          chmod go-w $HOME
+          sudo chmod -R go-w /usr/share
+
```
lynxlynxlynx, line 37: Why is this needed?

lynxlynxlynx:
The coverity stuff looks fine. I wrote this last year and it pretty much matches:
https://github.com/gemrb/gemrb/blob/master/.github/workflows/coverity.yml

.github/workflows/coverity-scan.yml
ja2-derek: I just copied it somewhere without thinking too much. I will remove that.

selaux:
I added the coverity token. For `BOT_GITHUB_TOKEN` this might take some time as I have to remember the credentials first.

selaux:
Allright. The required secrets should be there now.

ja2-derek:
> Allright. The required secrets should be there now.

@selaux can you check if there is an extra space in the `BOT_GITHUB_TOKEN`?

selaux:
@ja2-derek There was a leading space, should be fixed now.

ja2-derek:
The new build workflows are working - [nightly trigger](https://github.com/ja2-stracciatella/ja2-stracciatella/actions/runs/376026603) and [Coverity scan](https://github.com/ja2-stracciatella/ja2-stracciatella/runs/1435856212?).

A couple more changes:
- Added a CMake variable to skip `sccache`, which is required for Coverity build to work
- No longer using the `ad-m/github-push-action` action. We are doing a straight-forward `git-push` but with custom credentials. It is easier to do it with just shell commands.

----

But I don't understand why the Mac CI build is now failing the locate `gcloud`. I will take a look tomorrow.

lynxlynxlynx:
LGTM. I know how annoying it can be to get these working, so kudos!

ja2-derek:
The Mac build failure is due to a [recent change](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) from GitHub. It should be fixed now.

This PR is ready for merge. I will create another PR to remove `.travis.yml` after confirming builds works in master.

PR #1279, opened 2020-11-24
Replacing const char* in model classes

ja2-derek:
Continuing #1056.

Replacing `const char *` with `ST::string` in all model classes in `src/externalized`. Also a little clean up in WeaponModel.

## Example of changes

- `const char*` => `ST::string`
- `if (!strcmp(s1, "STR"))` => `if (s1 == "STR")`
- `strlen(str)` => `str.size()`
- `#include "game/Tactical/Item_Types.h"` => `#include "Item_Types.h"`
- `ST::string str` (constructor args) => `std::move(str)` (see also [performance-unnecessary-value-param](https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html))
- `Class::~Class() {}` => `Class::~Class() = default;`
- `std::map<ST::string, const ModelClass*>::const_iterator` => `auto`

### DIFF
--- a/src/externalized/AmmoTypeModel.cc
+++ b/src/externalized/AmmoTypeModel.cc
@@ -1,19 +1,16 @@
 #include "AmmoTypeModel.h"
-
 #include "JsonObject.h"
-
-#include <string_theory/format>
-
 #include <stdexcept>
+#include <string_theory/format>
+#include <utility>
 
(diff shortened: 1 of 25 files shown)


lynxlynxlynx:
Nice, thanks!

PR #1280, opened 2020-11-25
Externalizing I.M.P. activation code

ja2-derek:
This PR externalizes the activiation code to `imp.json`. Player can proceed to I.M.P. creation if the given code matches any.

You usually do not want to change the activation code, because it is fixed in the email text. Unless you are playing [Unfinished Business](https://github.com/ja2-derek/ja2-ub-comparison/commit/9770aba9), in which the `email.edt` is changed as well.

And also, if you insist (I don't know why), you can name your IMP character `XEP624` and still use [I.M.P. import](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1133).

### DIFF
--- a/assets/externalized/imp.json
+++ b/assets/externalized/imp.json
@@ -1,4 +1,7 @@
 {
+    // I.M.P. homepage activation code
+    "activation_codes": ["XEP624", "xep624"],
+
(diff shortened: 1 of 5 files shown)


lynxlynxlynx:
Thanks!

Issue #1272, opened 2020-10-21
Supporting extra save game data from Lua scripts


ja2-derek:
There is a need for mod Lua scripts to persist game states. We need a way for the mod to keep track of the game state, for anything outside of  the basic facilities (e.g. Facts, Quests) provided by the base game.

## Example use-cases

- Mappings - e.g. quote being said per soldier. It is much better and readable to use a real map than a long list of `FACT_X_HAS_DONE_Y`.Also there are only 500 `NUM_FACTS` slots and not enough for multiple mappings.
- Persisting non-boolean values like sector IDs, counters or amounts, though we might not need this often. Facts are UINT8, but they usually take `enum` or `BOOLEAN` value.
- Mods can opt to use this instead of Facts for everything, so as not to worry about assigning Fact IDs and the limit of `NUM_FACTS`

## Summary

- Allow scripts to save arbitrary, key-value pairs into save files.
- Using the same files so we don't need to manage another file (e.g. if we created an extra `.json` alongside the `.SAV` files.
- Re-purpose a few bytes in the [Air_Raid data range](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/fd8e3f44719c59ff36228ddf29942fd6b862ad52/src/game/SaveLoadGame.cc#L398-L399) to indicate the size of the extra data. Avoid the first bytes which is should be always `0` but would be used by older versions and vanilla.
- Append the data from scripts to the end of the save files.
- Increment [SAVE_GAME_VERSION](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/fd8e3f44719c59ff36228ddf29942fd6b862ad52/src/game/GameVersion.cc#L23) to `101`. This can be done but not necessary for this to work.

## Compatability

- Only to be activated on request by mod scripts (via an Observable)
- If no mod has requested to save data, we do not append to the end of the save, and the file will look exactly the same as before

## Details

- C++ does not allow dynamic typing, so the types must be fixed. We can use a `std::map<ST::string, int32_t>` and this should give us enough flexibility.
- Lua side can do nested tables, but on save, the tables must be flattened to a map from string to integer. I don't see a real need for string and float values, but it can be worked around if really needed. For example:
```lua
save_data_nested = {
    FavouriteMerc = MercsEnum.BARRY,
    QuoteUsed = {
      MercsEnum.BARRY = {
        QUOTE_101 = true,
        QUOTE_102 = false,
      },
    },
[...]

Issue #823, opened 2019-05-03
Support an arbitrary amount of saved games


lynxlynxlynx:
We should add a scrollbar or scrollbuttons to allow for more than 10 save slots to be shown.

Sort of a blocker for #822.

PR #1282, opened 2020-11-29
Appending extra game states at the end of Saves

ja2-derek:
Providing the basis for #1272.

## Summary

- Renamed `GameState` to `GameMode`, because the class has nothing to do with the current state of the game
- Game save version incremented from `100` to `101`
- Saved game is compatible with current master (Not yet tested with base vanilla)
- Extra key-value data is appended to the end of the save
- Key are strings; Values are `std::variant`, can be any one of  primitives (`int32_t`, `float`, `bool` and `string`) or, a map or vector of primitives.

## Usage

It is accessed via the `g_gameStates` in SaveLoadGameStates.h.

- Loading a primitive value (one of `int32_t`, `bool`, `ST::string` or `float`)
```c
#include "SaveLoadGameStates.h"

INT32 intval = g_gameStates.Get<INT32>("key1");
ST::string st = g_gameStates.Get<ST::string>("key2");
```
- Saving a typed container 
```c
#include "SaveLoadGameStates.h"

std::map<ST::string, INT32> intmap = g_gameStates.GetMap<INT32>("some-key");
g_gameStates.SetMap<INT32>("some-key", intmap);

std::vector<ST::string> strings = g_gameStates.GetVector<ST::string>("my-strings");
g_gameStates.SetVector<ST::string>("my-strings", strings);
```
- Using saved game states from Lua
```lua
function handle_before_game_saved()
  local states = {
    what_happened = "nothing"
  }
  PutGameStates("my-awesome-mod", states)
end

function handle_on_game_loaded()
  local states = GetGameStates("my-awesome-mod")
  log.info(states.what_happended)
end
```

## Details

With this PR, loading saves across versions will give the "load an older version save" prompt. Nothing uses the states yet, so it will always write `2;{}` (data length and then the JSON data, in binary).

~The usage of VARIANT should match `std::variant` in C++ 17, so we can migrate later on~. Although the values are variant, we still need to pre-determine the keys and type to make use of them. But JSON and variant serialization should be good for future extensions. 

Data is written as raw JSON and not compressed. Even in plain text it is not that big, and if we were to compress, we should layer it over the whole file.

------

Could you please review and give me some feedbacks? If it is OK, we can start talking about what data can be included - e.g. states from mod scripts, list of mod names the game was started with.

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -10,7 +10,7 @@
 #include "game/GameRes.h"
 
 // XXX
-#include "game/GameState.h"
+#include "game/GameMode.h"
 
 #include "sgp/FileMan.h"
 #include "sgp/MemMan.h"
--- a/src/externalized/scripting/FunctionsLibrary.cc
+++ b/src/externalized/scripting/FunctionsLibrary.cc
@@ -4,6 +4,7 @@
 #include "Item_Types.h"
 #include "Items.h"
 #include "Queen_Command.h"
+#include "SaveLoadGameStates.h"
 #include "StrategicMap.h"
 #include <stdexcept>
 #include <string>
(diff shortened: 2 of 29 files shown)


lynxlynxlynx:
Looks fine to me.

Ideally we'd store the used mod list here and ensure they're loaded, so the choices at game/launcher startup aren't important. I think our VFS can handle that. Would also mean that we don't need to have per-mod save dirs, but for that also #823 is important. And we could improve the save game list display to also show this mod info.

selaux:
Looks good to me as well. It would be nice to be able to store lists (`std::vector`) and maps (`std::map`) in there as well. Would that be possible with the variant approach?

ja2-derek:
Thank you guys.

Given that C++ is static-typed, we can support vectors and maps of primitives with templates/generics, but not anything nested. And we still need to pre-determine the allowed types in `std::variant` and then implement all (de-)serialization with RapidJSON. I don't see any blocking issues with that.

Re C++ 17, I can try to make the switch, but as I remember there are some warnings or errors with RapidJSON, which hasn't had a formal release since 2016.

ja2-derek:
`std::variant`, `std::vector` and `std::map` for primitive types are done.

lynxlynxlynx:
Is this indeed still a draft?

ja2-derek:
I don't have anything to add to this PR. I hesitate to get this merged because it introduce a save game format break without adding any real value to the vanilla game play.

My original plan was to incorporate this in my Unfinished Business mod work before merging. However I don't have much progress in the UB support lately.

lynxlynxlynx:
Hmm, ok. I don't want it to needlessly bit rot, but it sounds like we should wait for some work that needs it or maybe do it right after a release.

ja2-derek:
The branch was rebased to resolve a merge conflict. Only these two commits are new 774359f and 38349b0 .

I am going to verify the save game compatibility with 0.18. After that this should be good for merge.

ja2-derek:
Tested the save game compatibility, both forward and backwards, between 0.18 and master, also between vanilla (1.12 GOG) and master. In all cases it prompts "Attempting to load an older version save..." even when the save is made by the newer version.

lynxlynxlynx:
Reread the diffs and all looks fine to me. 👍🏾

PR #1283, opened 2020-11-30
Fixing Coverity scans

ja2-derek:
Coverity expects analysis results under `cov-int/`. See https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1276#issuecomment-735913970


ja2-derek:
The Coverity badge now says "passed".

lynxlynxlynx:
Looks like it will be fine:
> Last Build Status: Running. Your build is currently being analyzed
Merge once it's done.

PR #1284, opened 2020-12-01
Upgrading to C++ 17

ja2-derek:
For new language features such as `std::variant`.

## Summary

- Upgraded `CMAKE_CXX_STANDARD` to 17
- Upgraded sol2 to v3.2.2 (which requires C++ 17)
- Changed Lua to use CMake-getter


ja2-derek:
Note: MinGW build fails on Ubuntu 20.04, with errors very similar to [rust-lang/rust#48272](https://github.com/rust-lang/rust/issues?q=is%3Aissue+is%3Aclosed+Latest+MSYS2+mingw-w64+build+failure++__onexitbegin).

selaux:
LGTM. Thanks a lot.

PR #1294, opened 2020-12-17
Add new logging macros for ST format strings

JRF-github:
In many places of the code base, the old C-Style formatting strings in the logging macros were replaced with string_theory formatting strings and while the improved type safety and not having to remember the correct conversion specifier for size_t is nice, the current syntax is a bit clunky. This change introduces new STLOGx macros, where STLOGx(args) is equivalent to SLOGx(ST::format(args)). 

As a demonstration this changeset includes an updated DefaultContentManager.cc which was created with sed -E 's/SLOG(.)\(ST\:\:format\((.*)\)\);/STLOG\1\(\2\);/

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -388,7 +388,7 @@ ST::string DefaultContentManager::getMapPath(const ST::string& mapName) const
 	result += '/';
 	result += mapName.c_str();
 
-	SLOGD(ST::format("map file {}", result));
+	STLOGD("map file {}", result);
 
 	return result;
 }
@@ -400,7 +400,7 @@ ST::string DefaultContentManager::getRadarMapResourceName(const ST::string &mapN
 	result += "/";
 	result += mapName;
 
-	SLOGD(ST::format("map file {}", result));
+	STLOGD("map file {}", result);
 
 	return result;
 }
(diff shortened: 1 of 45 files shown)


lynxlynxlynx:
Nice idea, thanks!

lynxlynxlynx:
I wonder, could we just dump the old versions and channel everything to the new ones? I don't know if ST::format insists on having tokens in the format string, if the pointless conversion would make logging noticeably slower or if we need to log somewhere where ST is not available yet.

ja2-derek:
I don't think ST::format supports sprintf-'%s%d'-style formatting. Those need to be converted one-by-one.

Currently [LogMessage](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/996ee9a74f95a49878d80c00d40273a89bbb11d4/src/sgp/Logger.cc#L24) already converts the formatted `char[]` to `ST::string`.

JRF-github:
Yes, that's right, all you get when you try something like format("%d", 20) is "%d". I believe ST is already required not just for logging but to play the game, for example the laptop texts are now formatted with ST. Performance shouldn't really be much of a concern, after all, the game doesn't dump dozens of logging messages every second, not even with -debug.

lynxlynxlynx:
Ah, I forgot #1056 is not done yet. Also, perhaps modifiers are in use somewhere and I don't know if ST can handle that.

Thanks, I'll just merge.

PR #1297, opened 2020-12-21
Remove reinterpret_casts in LoadSaveData.cc

JRF-github:
There are currently several reinterpret_casts from void* to uint8_t* in the DataReader and DataWriter classes, all of which can easily be avoided by declaring the member variables as uint8_t*.

I decided to change char[] to BYTE[], not uint8_t[] in LoadSaveData_unittest.cc because that's what all the users of DataReader and DataWriter currently use.

### DIFF
--- a/src/sgp/LoadSaveData.cc
+++ b/src/sgp/LoadSaveData.cc
@@ -12,7 +12,7 @@
 
 /** Constructor.
  * @param buf Pointer to the buffer for writing data. */
-DataWriter::DataWriter(void* buf)
+DataWriter::DataWriter(uint8_t* buf)
 	:m_buf(buf), m_original(buf)
 {
 }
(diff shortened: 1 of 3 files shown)


ja2-derek:
Should we use also `BYTE` in DataWriter? This is only stylistic difference, but to me `BYTE*` (and to some extent, `void*`) means 'raw data bytes' , and that we don't care about the signed'ness at all.

JRF-github:
Well, I prefer standard types over the JA2 types, but it's just a mild preference. If the consensus says to use BYTE* I certainly wouldn't object.

lynxlynxlynx:
No preference on my side.

selaux:
I have no preference either, feel free to use standard types.

ja2-derek:
Let's merge it then. Thank you for the PR!

PR #1300, opened 2020-12-22
Remove the unused MapUtility screen

JRF-github:
If you didn't like the AniView screen, you are not going to like this screen either. The descriptions says:

> // Utililty file for sub-sampling/creating our radar screen maps
>// Loops though our maps directory and reads all .map files, subsamples an area, color
>/ quantizes it into an 8-bit image ans writes it to an sti file in radarmaps.

That does not sound like a thing stracciatella users are very interested in. The screen can only be reached with help from the debugger, too, and when you try, it just exits the game because for some reason I didn't investigate GCM->getAllMaps() returns an empty list.


lynxlynxlynx:
Thanks!

PR #1302, opened 2020-12-28
Refactor the event pump with std::variant

JRF-github:
The event pump manages six different event types in three event queues. Currently this is implemented with void pointers: the callers have to pass an argument indicating the underlying type. The data is then copied to a vector<BYTE> and copied back from that vector to the right struct when the event is processed.

Now that the project requires C++17, this logic can be implemented more cleanly with std::variant.

Pros:
- Less code, improved readability
- No more bit twiddling with void *, vector<BYTE> and memcpy
- Less arguments to pass around
- Mismatches between claimed and actual type are no longer possible. This nonsense currently compiles without an error: `EV_S_FIREWEAPON efw; AddGameEvent(S_NOISE, 0, &efw);`

Cons:
- Replaces proven working code with unproven new code. I have tested this in two combat situations so far without problems, which is not saying much.


src/game/Utils/Event_Pump.cc
```diff
@@ -150,15 +206,28 @@ static SOLDIERTYPE* GetSoldier(const UINT16 soldier_idx)
...
 
-static BOOLEAN ExecuteGameEvent(EVENT* pEvent)
+// Copied from https://stackoverflow.com/posts/52303671/revisions
+// Author: Barry
+template<typename VariantType, typename T, std::size_t index = 0>
+    constexpr std::size_t variant_index() {
+        if constexpr (index == std::variant_size_v<VariantType>) {
+            return index;
+        } else if constexpr (std::is_same_v<std::variant_alternative_t<index, VariantType>, T>) {
+            return index;
```
ja2-derek, line 216: Can you remove the extra indent to better align the code?

ja2-derek:
I think we seldom change or extend anything with the game event pump. So using `std::variant` for stricter compile time checks is a good change to me.

src/game/Utils/Event_Pump.cc
```diff
@@ -150,15 +206,28 @@ static SOLDIERTYPE* GetSoldier(const UINT16 soldier_idx)
...
+		} else {
```
ja2-derek, line 217: Can you remove the extra indent to better align the code?

ja2-derek: Or alternatively, would you consider using `std::visit` instead? So it is more idiomatic and will look like:

```c++
    std::visit(overload{
        [](EV_S_GETNEWPATH& enp) { /* do something */ },
        [](EV_S_SETDESIREDDIRECTION& edd) {  /* do something else */ },
        // and so on...
    }, pEvent->gameEvent);
```

JRF-github: Another avid cppreference reader, I see :+1:. Sure, I had at first decided against it because it wanted to make it easier to verify that the actual logic in ExecuteGameEvent is unchanged, but overall I like the std::visit solution more, too.

src/game/Utils/Event_Pump.cc
```diff
@@ -10,62 +7,121 @@
...
-		SLOGD("AddGameEvent: Sending Local #%d", uiEvent);
-		AddGameEventToQueue(uiEvent, usDelay, pEventData, PRIMARY_EVENT_QUEUE);
-	}
+	PRIMARY_EVENT_QUEUE,
+	SECONDARY_EVENT_QUEUE,
+	DEMAND_EVENT_QUEUE
+};
+
+
+typedef std::vector<EVENT*>EventList;
```
selaux, line 29: Would [std::deque](http://www.cplusplus.com/reference/deque/deque/) be a better candidate for the data structure here?

src/game/Utils/Event_Pump.cc
```diff
@@ -10,62 +7,121 @@
...
+	pEvent->TimeStamp  = GetJA2Clock();
+	pEvent->usDelay    = usDelay;
+	pEvent->eventExpired = false;
+	pEvent->gameEvent  = gameEvent;
+	GetQueue(ubQueueID).push_back(pEvent);
 }
 
 
-static void AddGameEventToQueue(UINT32 const uiEvent, UINT16 const usDelay, PTR const pEventData, EventQueueID const ubQueueID)
+static EVENT* RemoveEvent(UINT32 uiIndex, EventQueueID ubQueueID)
```
selaux, line 49: Is this ever called with a `uiIndex != 0`? If not does it make sense to make this function a `PopEvent` function?

selaux (approved):
Some minor questions, but generally looking good. Thanks.

src/game/Utils/Event_Pump.cc
JRF-github: You mean because we usually remove the first container element? Maybe, but the largest queue size I've seen was 20 and I had to burst fire an entire clip to get the number that high. OTOH [cppreference](https://en.cppreference.com/w/cpp/container/deque) has this to say about deque:

> typical implementations use a sequence of individually allocated fixed-size arrays, with additional bookkeeping, which means indexed access to deque must perform two pointer dereferences, compared to vector's indexed access which performs only one. 
>....  deques typically have large minimal memory cost; a deque holding just one element has to allocate its full internal array (e.g. 8 times the object size on 64-bit libstdc++; 16 times the object size or 4096 bytes, whichever is larger, on 64-bit libc++). 

Overall I don't believe deque is the better choice here.

src/game/Utils/Event_Pump.cc
JRF-github: I've never actually seen it happening but at least theoretically there is a code path in `DequeAllGameEvents()` where an expired event could be not the first in the secondary queue. But now that I've had a closer look at that code I think there is a bug here: if that ever happens we'd have an out of bound access to the vector because uiQueueSize is not updated. I'll post a fix tomorrow.

JRF-github: At least theoretically there is a code path in `DequeAllGameEvents` where this could happen if an expired event is not the first in the secondary queue. But now that I've looked closer at that code I've noticed that we'd have an out of bounds access to the vector in that case. I'll post a fix later.

PR #1304, opened 2020-12-30
Do not throw pointers to std::runtime_error in Cinematics.cc::SmkOpenFlic

JRF-github:
This code throws pointers to runtime_error and then tries to catch them by reference, which just doesn't work. Fortunately there errors are very unlikely to occur, otherwise the exceptions would be caught by the catchall in main() and the game be terminated.


ja2-derek:
Thank you!

PR #1305, opened 2021-01-05
Clean up the event queue management

JRF-github:
After selaux's question in #1302 about `RemoveEvent` I had a closer look at the rest of the event pump code and found quite a bit of questionable logic. I am unsure whether any of these issues could cause actual problems in the game but since a more robust implementation isn't that hard and also more concise and more performant, why not?

Copy and paste of my commit message:

- `PeekEvent` and `RemoveEvent` were using try-catch blocks to return NULL in case they were called with an out-of-bounds-index but vector::operator[] does not throw an exception in that case.
- So, all null checks after `PeekEvent` and `RemoveEvent` were pointless.
- Removing all expired events in `DequeAllGameEvents` can be written more concisely and more performant with an STL algorithm.
- This eliminates the only use of `RemoveEvent` with an index other than 0, so that function can be simplified. I renamed it `PopFrontEvent`.
- `PeekEvent` had only one user left, which also was better implemented with a range-based for loop so it got removed.
- `DequeAllGameEvents`, `DequeueAllDemandGameEvents` and `ClearEventQueue` never fail, so their return type got changed to void.
- `FreeEvent` also had useless null checks, replaced it with direct delete statements.

### DIFF
--- a/src/game/Tactical/OppList.cc
+++ b/src/game/Tactical/OppList.cc
@@ -3272,20 +3272,16 @@ void DebugSoldierPage4()
 // Noise stuff
 //
 
-#define MAX_MOVEMENT_NOISE			9
-
-UINT8 MovementNoise(SOLDIERTYPE* pSoldier) // XXX TODO000B
+UINT8 MovementNoise(SOLDIERTYPE const * const pSoldier)
 {
-	INT32 iStealthSkill, iRoll;
-	UINT8 ubMaxVolume, ubVolume, ubBandaged, ubEffLife;
-	INT8  bInWater = Water( pSoldier->sGridNo );
+	constexpr UINT8 MAX_MOVEMENT_NOISE = 9;
 
 	if ( pSoldier->bTeam == ENEMY_TEAM )
 	{
 		return( (UINT8) (MAX_MOVEMENT_NOISE - PreRandom( 2 )) );
 	}
 
-	iStealthSkill = 20 + 4 * EffectiveExpLevel( pSoldier ) + ((EffectiveDexterity( pSoldier ) * 4) / 10); // 24-100
+	INT32 iStealthSkill = 20 + 4 * EffectiveExpLevel( pSoldier ) + ((EffectiveDexterity( pSoldier ) * 4) / 10); // 24-100
 
 	// big bonus for those "extra stealthy" mercs
 	if ( pSoldier->ubBodyType == BLOODCAT )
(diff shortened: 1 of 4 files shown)


src/game/Utils/Event_Pump.cc
```diff
@@ -45,33 +45,15 @@ static void AddEvent(GAMEEVENT const& gameEvent, UINT16 const usDelay, EventQueu
 	GetQueue(ubQueueID).push_back(pEvent);
 }
 
-
-static EVENT* RemoveEvent(UINT32 uiIndex, EventQueueID ubQueueID)
-try
+// Remove the first element from a queue and return it.
+// Calling this function on an empty queue is undefined behavior, don't do it.
```
lynxlynxlynx, line 49: Why not assert it?

lynxlynxlynx:
Other than that it looks good to me, thanks!

src/game/Utils/Event_Pump.cc
JRF-github: I'd like to reverse the question: why should we put an assertion here? From a code review POV this function has only two callsites, both in the same file, so it's easy enough to verify they're ok. Why burden correct code with extra checks when I've just removed unnecessary null checks in this patch?

So an assert would only protect against someone calling this function incorrectly sometime in the future. But if you don't trust your contributors to call a simple four line function with one parameter correctly you'd have to place asserts in every function and forbid the use of the entire standard library because these functions don't babysit their users either.

And finally, asserts are ugly. IMO this code base has too much debug messages, not too few. If you compare the current master against the original JA2 code release you'll notice that many of the debug messages have been there for more than twenty years. I've even seen one instance (in Rotting_Corpses.cc) where the debug messages were never in a JA2 release, only in the JA2 beta/test versions, but they still live on in Stracciatella. I hope these messages helped fix whatever problem they were supposed to find, but if not then maybe it's time to admit they've failed to do their job? In both cases it would make sense to remove them.

lynxlynxlynx: Because assumptions are just asking for trouble. Someone adds a new user down the line or changes the internals and things can go south. It doesn't hurt to be explicit, but neither did I say we should be 100% defensively programming.

An assert is a single line of code ensuring consistency in debug builds and has nothing to do with logging. But yeah, debug calls can surely be cleaned up, and if we don't expose the log levels to users yet, that's a simple way to add filtering.

JRF-github (commented):
Ignore this one, it's unrelated.

ja2-derek:
> 
> 
> Ignore the first one, it's unrelated.

Or would you force-push the second commit without the first?

src/game/Tactical/OppList.cc
```diff
@@ -3315,45 +3311,37 @@ UINT8 MovementNoise(SOLDIERTYPE* pSoldier) // XXX TODO000B
...
-		if (bInWater)
+		if (Water(pSoldier->sGridNo))
 		{
 			ubMaxVolume++; // in water, can be even louder
 		}
 		switch (pSoldier->usAnimState)
 		{
 			case CRAWLING:
-				ubMaxVolume -= 2;
+				ubMaxVolume = ubMaxVolume <= 2 ? 0 : ubMaxVolume - 2;
```
selaux, line 3341: Would it make sense to use `INT32` for `ubVolume` and `ubMaxVolume` as well and just use [`std::clamp`](https://en.cppreference.com/w/cpp/algorithm/clamp) in the end to limit the result to a `UINT8` value again?

selaux:
So, this is ready to merge right? I think we can leave the MovementNoise commit, no need to create a separate PR...

src/game/Tactical/OppList.cc
lynxlynxlynx: A simple `std::max(0, ubMaxVolume - 2)` would work as well.

lynxlynxlynx:
Yep, fine by me.

PR #1306, opened 2021-01-07
Externalize Strategic AI policy

ja2-derek:
Moving the tunable parameters in `Strategic_AI.cc` to JSON. It should be a "policy" as it controls game logic based on difficulty. I created a separate file because the `game.json` is already quite big.

The detailed comments in the original code are moved to the JSON.

----

I wanted to experiment with different Strategic AI settings to tune the game to suit my skill level. 

On the other hand, we might be able to implement something like the infamous Drassen counterattack with this and some Lua scripts.


lynxlynxlynx:
Nice, LGTM and almost makes me want to play again. :)

Looks like you killed one include too many though.

ja2-derek:
> Looks like you killed one include too many though.

Fixed!

Issue #1309, opened 2021-01-14
Fredo refuses to repair extended ears


JRF-github:
With the attached saved game, ask Fredo to repair Steroid's extended ears and he'll refuse, telling you that he only works on electronics.

git bisect points the finger at commit 296189384a378af94ceb4d31a0196e6a98d66f31.
[QuickSave.sav.gz](https://github.com/ja2-stracciatella/ja2-stracciatella/files/5816171/QuickSave.sav.gz)

ja2-derek:
I will take a look. Thank you for reporting and tracking this down.

PR #1312, opened 2021-01-15
Fix: Buttons stay disabled after leaving the shop keeper interface

JRF-github:
The saved game from #1309 can be used to see this bug: give Fredo or Arnold something to repair and finish the transaction. The buttons that were disabled when you enter the interface stay disabled.

git bisect says this bug was introduced by 97e943d9b01692eb305f14e43f669223d312911b, so this bug is quite old and in the 0.17.0 release. Since that commit supposedly was needed to fix other bugs I took another approach to fix this one.

Btw., ShopKeeperInterface_SetSMpanelButtonsState isn't the prettiest function name I've ever written. Have your considered adding some namespaces?

### DIFF
--- a/src/game/Editor/Sector_Summary.cc
+++ b/src/game/Editor/Sector_Summary.cc
@@ -47,7 +47,6 @@
 #include <string_theory/format>
 #include <string_theory/string>
 
-#include <algorithm>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


selaux:
Thanks.

PR #1314, opened 2021-01-27
Move two debug-only global variables

JRF-github:
These two variables are only used by DebugLevelNodePage() and therefore it is not necessary to have them as globals. This also avoids constructing the string objects immediately and should help improving the startup time a tiny bit.


lynxlynxlynx:
LGTM, thanks!

PR #1315, opened 2021-01-28
Avoid unnecessary string copies in Tile_Surface.cc and WorldDef.cc

JRF-github:
Several locations made string copies where a reference could have been used instead.

The most egregious case is `LoadTileSurfaces(), which copies an array of 151 strings it does not even use. I was surprised what a difference ST_LITERAL makes, that change alone reduced the size of a release build by about 8 KB.

### DIFF
--- a/src/game/TileEngine/Tile_Surface.cc
+++ b/src/game/TileEngine/Tile_Surface.cc
@@ -20,7 +20,7 @@
 TILE_IMAGERY				*gTileSurfaceArray[ NUMBEROFTILETYPES ];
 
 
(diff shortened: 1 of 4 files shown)


ja2-derek:
> The most egregious case is `LoadTileSurfaces(), which copies an array of 151 strings it does not even use. I was surprised what a difference ST_LITERAL makes, that change alone reduced the size of a release build by about 8 KB.

I would have never expected that.. that could save a lot if we use that in the translation string files.

ja2-derek:
The changes LGTM

lynxlynxlynx:
Thanks!

PR #1323, opened 2021-02-02
Turn BAR macro in LoadSavedGame() into a lambda.

JRF-github:
This reduces the code size. In my case (Linux x64, GCC 9.3.0) the executable shrinks by about 3800 bytes.


selaux:
LGTM. Thanks.

PR #1324, opened 2021-02-04
Change signature of GetMapTempFileName()

JRF-github:
This function constructs a filename in an out char array. Since almost all of its users then pass that char array to functions taking ST::string it makes sense to construct a string directly.

This also fixes a format-overflow warning where GCC complains about the fact that we try to `sprintf()` a 512 char buffer into a 128 char buffer. It's not really a problem because the filenames cannot get that long but ironically the warnings generated are so huge they overflow my terminal window buffer.


ja2-derek:
LGTM. Thanks!

PR #1325, opened 2021-02-06
Fixing game stuck after halting soldier

### LINKED ISSUE #1318: Game hanging after an attack
To reproduce:
1. Load the attached save.
2. Johny runs in the direction of an enemy. Run to hit him with a crowbar.
3. As Johny proceeds with the attack, he notices another enemy farther away and stops to ask ask what to do. Continue and hit the enemy.
4. The enemy gets damage, but the cursor turns to a clock and stays like that.

Note that if I save the game after Johny stops to ask what to do and then reload - the issue doesn't appear.

[s.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/5900397/s.zip)

|  310e240 | JA2 EN  |    ?    |    ?     |

ja2-derek:
Fixes #1318. Also includes a part of Unfinished Business [PATCH201](https://github.com/ja2-derek/ja2-ub-comparison/commit/f523161d).

## Context

"Attack busy count" is incremented whenever an attack sequence starts to freeze the UI temporarily, and decrements when the attack finishes. It should be zero at the normal state or else the game UI gets stuck.

In combat mode, soldiers with pending actions are halted when they see new enemy. The "attack busy count" should be decremented if the cancelled action is an attack. Firing guns are [handled](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/843e20b372f1c2955561ae94bbe9d7a27a13e7e8/src/game/Tactical/Soldier_Control.cc#L7500-L7506) via a `fTurningToShoot` flag, and throwing objects (except throwing knifes) are checked via [the existence of](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/843e20b372f1c2955561ae94bbe9d7a27a13e7e8/src/game/Tactical/Soldier_Control.cc#L7399-L7400) `pTempObject`.

These existing checks do not cover throwing knives or melee attacks. The fix here is to use the pending animation ID to determine if we need to decrement the busy-count.

## See also

JA2 1.13 takes this [one step further](https://github.com/aimnas/ja2-1.13-source-mirror/blob/1256639fe564b20562f060919ed0237b38907a8e/Tactical/Soldier%20Control.cpp#L4352-L4354) with a new `ANIM_ATTACK` flag on all attack animations.

UB PATCH201 includes fixes for issues with throwing knives and jumping fence. This PR only includes the former.


selaux:
Thanks.

Issue #1292, opened 2020-12-14
Urban Chaos support


ja2-derek:
Continuing from https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1290, where @brechtm suggested Urban Chaos.

We are able to get in game with the Urban Chaos ("classic") data files. There are issues to be investigated, and in this issue we will do the analysis and track any work done:

### Game crashes and freezes

- [ ] Intros .SMK cause game to crash
- [ ] Game may hang after intro movie at lower difficulty level. It is a dead loop due to insufficient terrorists for slots (at [DecideActiveTerrorists](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/bee3fe37530d618b7c391bc7e85cf955369a8626/src/game/Tactical/Soldier_Profile.cc#L247))
- [ ] Pre-placed militia has have higher than usual ranks (causes [exception](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/b413e4d66eefec9c98db4bb28e4b22d88d0bb726/src/game/Tactical/Militia_Control.cc#L77)) 

### Data differences (that we know so far)

- [ ] Items and weapons uses different data, and also:
    - [ ] Weapons and armors have the wrong inventory graphics - e.g. the machete has the image of a sniper rifle
    - [ ] There are "backpack" items. Not sure how it was implemented in original "classic" UC
- [ ] Email senders (no more Enrico)
- [ ] More towns, different town names and more town sectors
- [ ] Sector descriptions
- [ ] (Most probably) Garrison groups and patrol groups
- [ ] No Sci-fi mode

[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

----

[...]

PR #1326, opened 2021-02-07
Address #1292 item 1: support Urban Chaos quasi-empty smacker files

JRF-github:
- Fix crash in libsmacker
- Change Intro.cc to continue with the next video after an invalid one

With this commit the Urban Chaos intro sequence plays as intended.


selaux:
LGTM. Thanks.

PR #1327, opened 2021-02-09
Proposal: Remove the screenshot and video capture functionality

JRF-github:
- JA2's screenshots are always the full window in TGA format. Modern OSs can do better.
- The 'video capture' function is not even hooked up to the UI and can only be activated by setting a variable with the debugger. It then simply starts to take screenshots automatically, no audio capture.

This was the only user of PlatformIO.h, so that file can go as well.


ja2-derek:
Agreed. I can't think of any reasons to prefer the in-game screenshot functions.

lynxlynxlynx:
Less is more, thanks! :)

PR #1328, opened 2021-02-13
Fix the hot spot of the hand cursor

### LINKED ISSUE #683: arrows in IMP attribute allocation screen not exact
because it came up in #644

screen001

The right arrow "more points" has a small area to the left, where it is not clickable (red area).
The left arrow "less points" has a small area to the right, where it is still clickable, although the graphic for the arrow ends earlier (green area).
We're talking about just a few pixels on each side, but it is noticeable (and annoying ;)

JRF-github:
Put the hot spot of CURSOR_WWW where you would expect it to be, at the top of the index finger.

Fixes #683


lynxlynxlynx:
Thanks!

PR #1334, opened 2021-03-06
Adding NO_PROFILE checks before creating MercProfile object

### LINKED ISSUE #1332: Quest Debug System crashes most of the time

### What is wrong
If you try to start the Quest Debug System and there is anyone or anything present in the sector without a merc profile the QDS crashes with the message "_Game has been terminated due to an unrecoverable error: invalid m_profileID '200'_". As you probably know, NO_PROFILE is defined as 200 in `Soldier_Control.h`

Since the QDS is actually useful when it's working and is mentioned on the [features page](https://ja2-stracciatella.github.io/features/) it'd be nice to get at least the basic functionality working again. It's also quite crashy when you let characters speak, but that has been the case since vanilla.

|  v0.17.1-git+866226846 | JA2 Gold Eng.  |    no    |    yes    |

ja2-derek:
Creating MercProfile object with invalid ID will result in an exception. This has caused problems in the QDS (and maybe some other places).

Resolves #1332.

### DIFF
--- a/src/game/Strategic/Strategic_Merc_Handler.cc
+++ b/src/game/Strategic/Strategic_Merc_Handler.cc
@@ -177,6 +177,7 @@ void MercDailyUpdate()
 			if (--s->bCorpseQuoteTolerance < 0) s->bCorpseQuoteTolerance = 0;
 
 			MERCPROFILESTRUCT& p = GetProfile(s->ubProfile);
+			MercProfile const profile(s->ubProfile);
 
 			// CJC: For some personalities, reset personality quote said flag
 			switch (p.bPersonalityTrait)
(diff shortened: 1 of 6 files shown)


ja2-derek:
This fixes the crash on opening the QDS.

src/game/Strategic/Strategic_Merc_Handler.cc
```diff
@@ -177,6 +177,7 @@ void MercDailyUpdate()
 			if (--s->bCorpseQuoteTolerance < 0) s->bCorpseQuoteTolerance = 0;
 
 			MERCPROFILESTRUCT& p = GetProfile(s->ubProfile);
+			MercProfile const profile(s->ubProfile);
```
lynxlynxlynx, line 180: Does this internalize the check or why isn't it needed here?

ja2-derek: [GetProfile](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/bee3fe37530d618b7c391bc7e85cf955369a8626/src/game/Tactical/Soldier_Profile.h#L14-L20) checks for a valid profile ID, and aborts on invalid ID.

lynxlynxlynx: Still quite strict, but it's unlikely to be hit in this place.

lynxlynxlynx:
Thanks!

Issue #1257, opened 2020-10-09
ja2.log should go to temp


misanthropos:
### wherever I start ja2 it will create the ja2.log in the current working directory.

I think we should create it in temp - or maybe .ja2/

PR #1335, opened 2021-03-08
Create ja2.log in temp directory

JRF-github:
Addresses #1257

Quoting [cppreference](https://en.cppreference.com/w/cpp/filesystem/temp_directory_path):

> On POSIX systems, the path may be the one specified in the environment variables TMPDIR, TMP, TEMP, TEMPDIR, and, if none of them are specified, the path "/tmp" is returned.
> 
> On Windows systems, the path is typically the one returned by GetTempPath

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -273,6 +273,10 @@ if (MINGW)
     LIST(APPEND JA2_LIBRARIES mingw_stdthreads)
 endif()
 
+if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    LIST(APPEND JA2_LIBRARIES stdc++fs)
+endif()
+
 if(ANDROID)
     add_library(${JA2_BINARY} SHARED ${JA2_SOURCES})
 else()
--- a/src/sgp/SGP.cc
+++ b/src/sgp/SGP.cc
@@ -55,6 +55,26 @@
 #include <new>
 #include <utility>
 
+#ifdef __ANDROID__
+	static inline ST::string get_temp_filename(void)
+	{
+		return "ja2.log";
+	}
+#else
+	#if defined(__GNUC__) && __GNUC__ < 8
+	#include <experimental/filesystem>
+	using std::experimental::filesystem::temp_directory_path;
+	#else
+	#include <filesystem>
+	using std::filesystem::temp_directory_path;
+	#endif
+
+	static inline ST::string get_temp_filename(void)
+	{
+		return ST::string{temp_directory_path().append("ja2.log")};
+	}
+#endif
+
 extern BOOLEAN gfPauseDueToPlayerGamePause;
 
 ////////////////////////////////////////////////////////////////////////////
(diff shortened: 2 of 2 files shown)


lynxlynxlynx:
It looks fine, but it doesn't build. GCC supported c++17 since version 5, so something is fishy (7.5 is installed).

selaux:
You could also use tempdir logic that is already exposed by rust:

```cpp
/// Attempts to create a temporary directory inside the temporary directory of the operating system.
/// Returns null on error.
/// Sets the rust error.
/// @see https://doc.rust-lang.org/std/env/fn.temp_dir.html
TempDir *TempDir_create();

/// Destroys the temporary directory.
/// This is a best effort approach, in case of error the directory might not be deleted.
/// Sets the rust error.
bool TempDir_destroy(TempDir *tempdir);

/// Gets the path of the temporary directory.
char *TempDir_path(TempDir *tempdir);
```

JRF-github:
selaux, if I understand this correctly this interface will create a temporary directory under the normal temp directory which will get removed when the RustPointer is destroyed, right? I'd like the file to be in /tmp (or $TEMP, %TEMP%, whatever) and stay around after the game exits. Is that possible with this API?

On the C++ side all that seems to be missing now is the correct incantation for the Android linker. Does anyone know what is missing here and where I'd need to add it?

selaux:
Yes, the rust functions should work as you described. Regarding where it creates the temp dir, it should be according to rust docs:
> Unix:
> Returns the value of the TMPDIR environment variable if it is set, otherwise for non-Android it returns /tmp. If Android, since there is no global temporary folder (it is usually allocated per-app), it returns /data/local/tmp.
> Windows:
> Returns the value of, in order, the TMP, TEMP, USERPROFILE environment variable if any are set and not the empty string. Otherwise, temp_dir returns the path of the Windows directory. This behavior is identical to that of GetTempPath, which this function uses internally.

So, regarding Android this seems to become an issue. It should be done through [getCacheDir](https://developer.android.com/reference/android/content/Context.html#getCacheDir%28%29) (via JN) although I read on StackOverflow somewhere that `/data/local/tmp` should be possible as well to use as temp dir.

Also please take note of my description [here](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1275#issuecomment-731989203) about how JA2S currently locates some of its paths. Basically it changes its working directory to the temp directory on startup and assumes that some of the configuration files are located one level above.

JRF-github:
So I finally decided to use the C++ way because I didn't like how Rust's tempdir is a hidden directory with a random name under /tmp. IMHO that makes it too difficult to locate the log file, it's bad to tell users who want to report a problem to go hunt for a file named ja2.log somewhere.

This leaves the situation under Android unchanged. I know nothing about about JNI and quite honestly have no interest in learning it. `/data/local/tmp` was allegedly locked down under later Android versions and is not guaranteed to be world writable.

lynxlynxlynx:
LGTM, thanks. I wouldn't worry too much about the android log location.

selaux:
Can we do the same for the launcher then: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/launcher/main.cc#L7?

ja2-derek:
Can we merge this first? The launcher change can be done in a separate PR.

selaux:
Sure.

PR #1345, opened 2021-04-21
Create ja2-launcher.log in temp directory

ja2-derek:
For #1257, see also https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1335#issuecomment-795058271.

We have changed log location of the main process. This updates the same for the `ja2-launcher`.

### DIFF
--- a/.ci/ci-functions.sh
+++ b/.ci/ci-functions.sh
@@ -42,6 +42,13 @@ linux-install-via-android-sdkmanager () {
     done
 }
 
+linux-set-gcc-version () {
+    sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-9 9
+    sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-9 9
+    sudo update-alternatives --set gcc "/usr/bin/gcc-9"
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
I guess gcc9 needs to be installed explicitly?

ja2-derek:
> I guess gcc9 needs to be installed explicitly?

No, image has gcc/g++ both 7.5 and 9.0 installed already - https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md

ja2-derek:
The MingW build is still on Ubuntu 18.04, in which [gcc-mingw-w64-x86-64](https://packages.ubuntu.com/bionic/gcc-mingw-w64-x86-64) is at v7.3, and do not have the `std::filesystem`.

There is a problem with building with newer the MingW [package](https://packages.ubuntu.com/focal/gcc-mingw-w64-x86-64) in Ubuntu 20.04. Later on when we get that sorted, we can clean up all usages of `std::experimental::filesystem`.

selaux:
Thanks for the PR and fighting the builds 😉 .

Issue #1348, opened 2021-04-22
Game crash after trying to choose JA2 directory on Android 11


Srandista:
### What is wrong
When I try (after successful apk installation) choose JA2 directory, game immediately crash (before showing directory listing). There's no error message, app immediately crash right after I touch the "Choose" button.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.18 | JA2 1.06  |    no    |    no     |

[//]: # (Please also include:)
[//]: # Click on "Choose" button after running JA2S
[//]: # Android 11 on Pixel 3a
[//]: # (Just for information, same build works fine on Android 10 with Huawei P20 Pro)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

selaux:
I can reproduce this in the Android emulator. This seems to be related to https://github.com/codekidX/storage-chooser/issues/127

Should be easily fixable by changing our storage-chooser dependency to `com.github.noelchew:storage-chooser:2.0.4.4a`. Although I would prefer the original library fixing that and not using some patched fork.

selaux:
Should be fixed in the next nightly and eventually in the `0.18.0` release. Thanks for the report.

Srandista:
I can confirm, that this new build works as expected. Thank you for super quick fix!

PR #1349, opened 2021-04-22
Fix directory chooser on Android 11

selaux:
com.github.noelchew:storage-chooser is a fork of the old dependency with a patch for Android 11 applied.

Closing #1348.


lynxlynxlynx:
Thanks!

PR #1350, opened 2021-04-22
Fix fetching and install location of editor slf

### LINKED ISSUE #1347: Editor does not work in AppImage
### What is wrong

With the current pre-release AppImage of `0.18.0` the editor does not work out of the box. Seems like it is not included in the build at least that is what the logs are reporting:

```
2021-04-22T06:33:19 [ WARN] stracciatella::vfs: Free editor.slf not found in DirFs { "/tmp/.mount_ja2-stnHlTNd/usr/share/ja2/externalized" }, the editor might not work
```

| 0.18.0 prerelease | English |    no    |    yes     |

selaux:
Changes the installed `editor.slf` location to match the one expected in the code. I'm not sure where this actually broke, probably with refactoring the VFS at some point, but I'm fine putting the slf into `externalized`. If you have objections, please let me know.

I tested the resulting AppImage and it works for me now. Not 100% sure about Windows builds, but they should behave the same.

Regarding rate limiting: Using a token makes rate limiting user specific so we use the one issued for CI, which should give us 5000 request / h.

Fixes #1347.


lynxlynxlynx:
Great, thanks! The auth fix should also make further releases smoother. :)

PR #1358, opened 2021-04-24
Remove explicit paths from DefaultContentManager startup

selaux:
This is a first step to splitting up https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1275.

I needed to add an empty slf for unittests as they use the unittests directory as vanilla game dir and we check for the existence of slf files there.


lynxlynxlynx:
Can't really audit the rust part, but a negative diffstat is always good in my book. :)

ja2-derek:
I think all the `m_externalizedDataPath` and `getExternalizedDataDir()` in DefaultContentManager should be removed now. The file system(s) should all be internal to Rust-lib and opaque the C side.

https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1358/files#diff-940ecd5e7134e247ed4b6ffbf4863581279344ee27e7470394b3540e26d8c926R540-R544


And I like the now properly named `assetDir` - which was `extraDataDir`, defaults to `.` (on Windows), or `/usr/share/ja2` (if .DEB).

selaux:
Yep, that is the eventual goal. I will stretch this over a few PRs though, as removing it (and `m_dataDir`) completely needs quite a bit of changes.

PR #1359, opened 2021-04-25
Externalizing starting sector

ja2-derek:
This adds 2 new configuration in `game.json`. These only takes effect for new campaigns 

It may not make much sense on a vanilla campaign, and this is intended for full map mods. Mercs are arrives on heli at the designated sector, so the sector must also be outside of SAM site coverage.

### DIFF
--- a/src/game/Tactical/Merc_Hiring.cc
+++ b/src/game/Tactical/Merc_Hiring.cc
@@ -42,6 +42,9 @@
 #include "Map_Screen_Interface_Bottom.h"
 #include "Quests.h"
 #include "Logger.h"
+#include "GamePolicy.h"
+#include "GameInstance.h"
+#include "ContentManager.h"
 
 #include <string_theory/string>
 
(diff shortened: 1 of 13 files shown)


src/game/Tactical/Merc_Hiring.cc
```diff
@@ -54,7 +57,7 @@ extern BOOLEAN gfFirstHeliRun;
 // ATE: Globals that dictate where the mercs will land once being hired
 // Default to start sector
 // Saved in general saved game structure
-INT16 g_merc_arrive_sector = START_SECTOR;
+INT16 g_merc_arrive_sector = 0;
```
ja2-derek, line 60: This is always set at `SetupNewStrategicGame()` when a game starts.

selaux:
LGTM. Maybe we could put it into a separate `campaign.json` file if more externalizable values for the campaign pop up.

Issue #1360, opened 2021-04-30
AppImage does not start in Debian 10 => libstdc++.so.6: version `GLIBCXX_3.4.26' not found


christianbaun:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
When trying to start the AppImage of ja2-stracciatella 0.18 in Debian 10 (stable), it results in the error message that the library libstdc++.so.6 lacks support for GLIBCXX_3.4.26.

```
$ ./ja2-stracciatella_0.18.0-git+ebc73ce_x86-64.AppImage 
./ja2-stracciatella_0.18.0-git+ebc73ce_x86-64.AppImage: 
/usr/lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.26' 
not found (required by ./ja2-stracciatella_0.18.0-git+ebc73ce_x86-64.AppImage)
```

### Some system information:

```
$ cat /etc/debian_version 
10.9
```

I use Debian stable (buster = 10). Just the kernel is from the buster backports:

```
$ uname -r
5.10.0-0.bpo.5-amd64
```

```
$ gcc -v
...
gcc version 8.3.0 (Debian 8.3.0-6) 
```

```
$ dpkg -l | grep libstdc++6 | grep amd64
ii  libstdc++6:amd64          8.3.0-6          amd64        GNU Standard C++ Library v3
```

```
$ strings /usr/lib/x86_64-linux-gnu/libstdc++.so.6 | grep GLIBCXX
GLIBCXX_3.4
GLIBCXX_3.4.1
GLIBCXX_3.4.2
GLIBCXX_3.4.3
...
GLIBCXX_3.4.22
GLIBCXX_3.4.23
GLIBCXX_3.4.24
GLIBCXX_3.4.25
GLIBCXX_DEBUG_MESSAGE_LENGTH
```

lynxlynxlynx:
As you have found out, your stack is too old. Installing gcc9 should fix it, while on our end we could consider building the image with an older gcc. We're using c++17 features, so that may not be ideal though.

ja2-derek:
But that is AppImage not the .DEB. Shouldn't everything be self-contained?

And we may need to check if the .deb has the correct libc++ version requirements.

[...]

PR #1361, opened 2021-05-01
CI-build with gcc-8

ja2-derek:
For #1360.

The problem is that JA2:S, being a maintained runtime of an old video game, does not run on a relatively old Linux (Debian 10). The current CI build on GitHub Action uses Ubuntu 18:04 and gcc-9.

We do not have much choice over the build environment on GitHub Action. Ubuntu-16.04 is [scheduled to be decommissioned later this year](https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/), so the only reasonable choice now is Ubuntu-18.04.

If the build still doesn't work, we may need to do the build inside a docker container of an old-stable Debian or CentOS.

## What's done so far

- Downgraded CI build to use `gcc-8` (on Ubuntu 18.04, GLIBC 2.27)
- Smoke test to run `./ja2 -help` on Debian 10
- `gcc` major version can be set via the environment variable `TARGET_GCC_MAJOR_VERSION`

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -89,6 +89,13 @@ if [[ "$RUN_TESTS" == "true" ]]; then
   if [[ "$CI_TARGET" == "linux" ]]; then
     ./AppDir/usr/bin/ja2 -unittests
     ./AppDir/usr/bin/ja2-launcher -help
+
+    # Smoke test to check if the binary can run on older distros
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
LGTM. Might be handy using a global var for the major version, so it can all be bumped with a single line next time.

PR #1363, opened 2021-05-07
Read tilecache names from VFS

selaux:
Also a part of #1275. This reads tile cache files from VFS, not by listing the data directory directly.


selaux:
Can I merge this? Then I'll continue with the rest...

lynxlynxlynx:
It's rust, so I'm not so quick to accept, but since it's blocking your work ... :)

PR #1372, opened 2021-05-21
Added option for mouse cursor to always be visible in tactical view

FluffyQuack:
Under certain circumstances, the mouse cursor is not rendered. Ie, when highlighting a tile for movement in real-time mode or hovering mouse cursor over a mercenary. It still behaves the same behind the scenes, so I feel it's nice to actually see the exact location of it at all times.

This adds a toggle ("always_show_cursor_in_tactical" in game.json) to make mouse cursor always be visible.

Examples of what the change looks like (with default JA2 behaviour, the cursor would be invisible in all of these samples):
ja2 2021-05-21 18-12-12-683
ja2 2021-05-21 18-12-18-582
ja2 2021-05-21 18-12-23-939
ja2 2021-05-21 18-12-35-720

I haven't done exhaustive testing of this, but based on the code changes, I think this is unlikely to cause any issues.

### DIFF
--- a/src/sgp/Cursor_Control.cc
+++ b/src/sgp/Cursor_Control.cc
@@ -1,12 +1,14 @@
+#include "ContentManager.h"
 #include "Cursor_Control.h"
+#include "Cursors.h"
 #include "Debug.h"
+#include "GameInstance.h"
 #include "HImage.h"
 #include "Timer.h"
 #include "VObject.h"
 #include "Video.h"
 #include "VSurface.h"
-#include "ContentManager.h"
-#include "GameInstance.h"
+
 #include "policy/GamePolicy.h"
 
 
@@ -192,7 +194,7 @@ void CursorDatabaseClear(void)
 
 BOOLEAN SetCurrentCursorFromDatabase(UINT32 uiCursorIndex)
 {
-	uiCursorIndex = ReturnCursorIndex(uiCursorIndex);
+	uiCursorIndex = ModifyCursorIndex(uiCursorIndex);
 	
 	if (uiCursorIndex == VIDEO_NO_CURSOR)
 	{
@@ -364,12 +366,12 @@ void SetExternMouseCursor(SGPVObject const& vo, UINT16 const region_idx)
 }
 
 // Checks if uiCursorIndex is VIDEO_DEFAULT_TO_NO_CURSOR. If so, we parse it as VIDEO_NO_CURSOR or 0 (generic cursor)
-UINT32 ReturnCursorIndex(UINT32 uiCursorIndex)
+UINT32 ModifyCursorIndex(UINT32 uiCursorIndex)
 {
 	if (uiCursorIndex == VIDEO_DEFAULT_TO_NO_CURSOR)
 	{
 		if (gamepolicy(always_show_cursor_in_tactical))
-			return 0;
+			return CURSOR_NORMAL;
 		else
 			return VIDEO_NO_CURSOR;
 	}
--- a/src/sgp/Cursor_Control.cc
+++ b/src/sgp/Cursor_Control.cc
@@ -5,6 +5,9 @@
 #include "VObject.h"
 #include "Video.h"
 #include "VSurface.h"
+#include "ContentManager.h"
+#include "GameInstance.h"
+#include "policy/GamePolicy.h"
 
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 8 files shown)


src/sgp/Cursor_Control.cc
```diff
@@ -357,3 +362,16 @@ void SetExternMouseCursor(SGPVObject const& vo, UINT16 const region_idx)
...
 	gusExternVoSubIndex = region_idx;
 }
+
+// Checks if uiCursorIndex is VIDEO_DEFAULT_TO_NO_CURSOR. If so, we parse it as VIDEO_NO_CURSOR or 0 (generic cursor)
+UINT32 ReturnCursorIndex(UINT32 uiCursorIndex)
+{
+	if (uiCursorIndex == VIDEO_DEFAULT_TO_NO_CURSOR)
+	{
+		if (gamepolicy(always_show_cursor_in_tactical))
+			return 0;
```
lynxlynxlynx, line 372: This is NO_UICURSOR, right? Would be cleaner with a named constant.

src/sgp/Cursor_Control.cc
```diff
@@ -357,3 +362,16 @@ void SetExternMouseCursor(SGPVObject const& vo, UINT16 const region_idx)
 	guiExternVo         = &vo;
 	gusExternVoSubIndex = region_idx;
 }
+
+// Checks if uiCursorIndex is VIDEO_DEFAULT_TO_NO_CURSOR. If so, we parse it as VIDEO_NO_CURSOR or 0 (generic cursor)
+UINT32 ReturnCursorIndex(UINT32 uiCursorIndex)
```
lynxlynxlynx, line 367: I think we can come up with a better name — to me ReturnCursorIndex with a uiCursorIndex sounds like something that will get inlined due to triviality. Maybe just ModifyCursorIndex?

src/sgp/Cursor_Control.cc
```diff
@@ -5,6 +5,9 @@
 #include "VObject.h"
 #include "Video.h"
 #include "VSurface.h"
+#include "ContentManager.h"
+#include "GameInstance.h"
+#include "policy/GamePolicy.h"
```
lynxlynxlynx, line 10: Please move these to be in sorted order like the rest. See https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/CONTRIBUTING.md#axioms-of-style

FluffyQuack: > Please move these to be in sorted order like the rest. See https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/CONTRIBUTING.md#axioms-of-style

I just looked at oher examples of `#include "policy/GamePolicy.h` and it seems to vary how it's included. What's most common is that it's separated by a linebreak from the main list of includes and it has ContentManager.h and GameInstance.h right above it.

Should I combine the new entries into the main list of includes, or do like most other files and add a linebreak before ContentManager.h?

lynxlynxlynx: You can separate the subdir one, yeah.

src/sgp/Cursor_Control.cc
FluffyQuack: It's a variant of VIDEO_NO_CURSOR (-1). I wasn't sure what to call it. Depending on the new config var, it's parsed as a generic cursor (0) or VIDEO_NO_CURSOR.

src/sgp/Cursor_Control.cc
FluffyQuack: I can change it to that. I'll wait for feedback from the other questions, and I'll update PR.

src/sgp/Cursor_Control.cc
lynxlynxlynx: But that's what I'm saying, isn't the generic cursor NO_UICURSOR?

src/sgp/Cursor_Control.cc
FluffyQuack: Okay, I'll just separate the subdir one.

src/sgp/Cursor_Control.cc
FluffyQuack: Oh, sorry, I misunderstood your first question.

I wrote 0 as that's what I noticed the list in gUICursors[] used for generic cursor. Looking again, that list randomly uses either CURSOR_NORMAL or 0.

I'll change my 0 to CURSOR_NORMAL.

FluffyQuack:
Okay, pushed an update to the PR based on the above discussions.

lynxlynxlynx:
Thanks!

PR #1373, opened 2021-05-21
make sure to expire break lights/smoke when trashing the area

### LINKED ISSUE #95: Artifacts when entering a map

---

Sometimes, when entering a map via "merc placement" (before a battle), there are "artifacts" present on the map. Mostly lights (maybe from break lights), sometimes patches of gas (tear/smoke). Here is an example:

Note, that I've not entered the sector before in this campaign. Also, I'm entering from the north and the light is somewhere in the south. When I save the game (after merc placement) and then quit and load again, the light is gone.

This has happened at least 3 times in my running campaign. I have not yet made out a pattern behind this...

Forgot to mention: It's v0.12.189 with 800x600 resolution.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/95

lynxlynxlynx:
My guess was right, they didn't get a chance to expire and the Reset* functions just clear some internal state.

fixes #95

### DIFF
--- a/src/game/TileEngine/Overhead_Map.cc
+++ b/src/game/TileEngine/Overhead_Map.cc
@@ -19,6 +19,7 @@
 #include "Interface_Items.h"
 #include "Interface_Panels.h"
 #include "Isometric_Utils.h"
+#include "LightEffects.h"
 #include "Line.h"
 #include "Map_Information.h"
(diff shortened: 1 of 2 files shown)


selaux:
Thanks. Looks good to me.

PR #1374, opened 2021-05-24
Read and write to a proper temporary directory

### LINKED ISSUE #135: Refactor operations with temporary files

---

It is necessary:

1) use unique temporary folders for every process, not "temp" in the config_folder/tmp as it now;

2) avoid using game resources file operations (e.g. GCM->openGameResForReading) for working with temporary files.  Use new functions (e.g. GCM->openTempFileForReading instead).

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/135

### LINKED ISSUE #1255: Excessive temp file usage in .ja2/tmp/temp

selaux:
Split out of #1275. Closes #1255. Closes #135.

This changes temp file interaction logic to use a "proper" temporary directory in the OS-specified location of the temporary files. The temporary directory is unique per process, so multiple instances of the game can run in parallel. An example on linux would be `/tmp/ja2-stracciatella-hufmWU3N`. Some special handling needed to be introduced for Android where there is no temp dir, only an app specific cache dir, which is then used as temp dir.

All interactions with this temporary directory are now done through `ContentManager`. When the content manager deallocates, the temp directory is automatically removed. Some adaptions were necessary here to ensure that the `ContentManager` is always properly deallocated when exiting the game.

The whole mechanism is using the rust `tempfile` crate which was until now only used by tests.

Temp File Operations are done by:
- Saving/Loading laptop information: History, Finances and Files
- Saving/Loading some information about maps, such as Door States, Rotting Corpses, Civilian States, Map Modifications
- Changing Tileset in Editor

Switching the working directory to the old temp dir still happens on startup, because private files are still acessed using relative paths. This would be the next step that needs to be split out of #1275. After that some final cleanup should be done and the original purpose of #1275, the LRU cache, can be implemented safely.

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -74,11 +74,6 @@
 
 #define DIALOGUESIZE 240
 
-// XXX: Issue #135
-// We need a safe way to create temporary directories - unique and random for every process
-// Boost probably provides this functionality
-#define NEW_TEMP_DIR "temp"
-
 const MercProfileInfo EMPTY_MERC_PROFILE_INFO;
 
 static ST::string LoadEncryptedData(ST::string& err_msg, STRING_ENC_TYPE encType, SGPFile* File, UINT32 seek_chars, UINT32 read_chars)
@@ -203,11 +198,22 @@ DefaultContentManager::DefaultContentManager(RustPointer<EngineOptions> engineOp
 	m_loadingScreenModel = NULL;
 	m_samSitesAirControl = NULL;
 
+	// Initialize temp dir
+	RustPointer<TempDir> tempDir(TempDir_create());
+	if (tempDir.get() == NULL) {
+		RustPointer<char> err{ getRustError() };
+		auto error = ST::format("Failed to create temporary directory: {}", err.get());
+		throw std::runtime_error(error.c_str());
+	}
+	m_tempDir = move(tempDir);
+	RustPointer<char> tempDirPath(TempDir_path(m_tempDir.get()));
+	m_tempDirPath = tempDirPath.get();
+
+	// Initialize VFS
 	auto succeeded = Vfs_init_from_engine_options(m_vfs.get(), m_engineOptions.get());
 	if (!succeeded) {
 		RustPointer<char> err{ getRustError() };
 		auto error = ST::format("Failed to build virtual file system (VFS): {}", err.get());
-		SLOGE(error);
 		throw std::runtime_error(error.c_str());
 	}
 }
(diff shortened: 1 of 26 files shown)


src/externalized/DefaultContentManager.cc
```diff
@@ -203,6 +198,19 @@ DefaultContentManager::DefaultContentManager(RustPointer<EngineOptions> engineOp
 	m_loadingScreenModel = NULL;
 	m_samSitesAirControl = NULL;
 
+	// Initialize temp dir
+	RustPointer<TempDir> tempDir(TempDir_create());
+	if (tempDir.get() == NULL) {
+		RustPointer<char> err{ getRustError() };
+		auto error = ST::format("Failed to create temporary directory: {}", err.get());
+		SLOGE(error);
+		throw std::runtime_error(error.c_str());
```
lynxlynxlynx, line 207: Do we really still not have a function that does both (or all four)?

ja2-derek: Now we have the termination handler printing exception messages, we might not need the `SLOGE` line.

I have started creating custom exception classes that takes `ST::string` (https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/externalized/Exceptions.h). Let me know your thoughts...

src/game/Laptop/Laptop.cc
```diff
@@ -3263,10 +3265,9 @@ static void HandleWebBookMarkNotifyTimer(void)
 void ClearOutTempLaptopFiles(void)
 {
 	// clear out all temp files from laptop
-	FileDelete("files.dat");
-	FileDelete("finances.dat");
-	FileDelete("email.dat");
```
lynxlynxlynx, line 3268: just for anyone else reading this: the file is never created.

lynxlynxlynx:
LGTM!

src/game/Laptop/Laptop.cc
selaux: Yes. I have looked through the code for quite some time, because I thought the path might be created by building a string or something, but I have not seen any indications of that, or the file ever being created...

src/externalized/DefaultContentManager.cc
selaux: I removed the explicit logging here.  :+1: for the custom exception classes.

Currenly it is quite inconsistent which errors are handled through the TerminationHandler, which are catched and logged and which are just ignored with `catch (...)`. Personally I'd say all fatal exceptions should be handled by the TerminationHandler an non-fatal ones should be catched and in most cases logged with warning log level.

Issue #1368, opened 2021-05-12
Feature Request: Chance-to-hit feature improvements


DSeyka:
Hello again! :) Glad to see the project is in active development.
Regarding the new CTH on 'f' hotkey feature (huge thanks for implementing that, @dariusk!), I humbly propose it being enhanced slightly.

1) Is it being tied to a hotkey really necessary? I mean, if user enables it, they probably want to see their hit chances regularly. I'll also admit that I've pressed 'd' (end turn) instead of 'f' more than once while playing with this feature, too. :D So I propose just displaying your chance-to-hit next to the aiming cursor when it is enabled in json.
2) While we're at it, why not take aiming clicks into account while displaying CTH? It's pretty counter-intuitive that it doesn't happen already.

Thanks for considering it, and again -- for not abandoning the project, playing good old JA2 has been helping me out quite a bit lately.

ja2-derek:
My opinions:

I personally do not like the idea of the persistent text on screen as that would distract me. Of course we can implement it as an option.

Instead I think the aiming crosshair should be doing that. We had a half-done implementation of the "accurate aim circle" (#1178), which the size of the crosshair circle indicates the chance-to-hit.

Or, maybe can we steal the OCTH crosshair from 1.13? It has a bar that displays the CTH, like this:
![image](https://user-images.githubusercontent.com/63151803/117984141-272b8180-b36a-11eb-98e4-4e2baefeae90.png)

DSeyka:
> persistent text on screen
It could only show up when aiming at the enemy, much like that bar thing in 1.13 (which is better than nothing, but obscures enemy sprites, as seen in the screenshot). Honestly, I'd prefer chance % being shown instead of circles/bars since that gives you much more precise readout, plus we already have these calculations done; I just don't like pressing an extra button and looking at text in the bottom left, is all. :)

MrMaggot:
Is it possible to disable the cth% entirely in the current build?

lynxlynxlynx:
What do you mean?

MrMaggot:
I mean the text that gives you the hit percentage chance when you press F. The first post mentions a user enabling it as if you can opt out but I didn't see any options to toggle it in the configs.

DSeyka:
It can be disabled with ```"extra_hotkeys": false``` in game.json.

lynxlynxlynx:
The extra keys are documented here btw:
[...]

PR #1375, opened 2021-05-28
Show chance-to-hit next to mouse cursor

FluffyQuack:
Here's my work-in-progress implementation for showing chance-to-hit when aiming (this resolves https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1368).

Some notes:

- I'm using `st_format_printf("%d%%", uiHitChance)` to generate the string with the hit chance. I haven't looked into how st_format_printf works, but I noticed other code did something similar, so I figure and hope this is the way to go.
- gusMouseCursorHeight contains the size of the mouse image data. This isn't big enough to contain body location string and hit chance string. I ended up using a rather hacky solution (whenever the hit chance string is used, then a value gsMouseSizeYModifier is increased so that the mouse rendering routine include the hit chance string). Another solution I'm considering is when SetMouseCursorProperties() is called SetCurrentCursorFromDatabase() is to set gusMouseCursorHeight  to a bigger value if the mouse cursor is a targeting cursor. I think that solution would be cleaner.
- I made it so 'F' only shows hit chance if the new config var is true. I saw someone request that, and then it's also closer to original game behaviour (which never shows the hit chance).
- There's some code I need to remove related to the position of the hit chance string. Once we're completely sure where we want it positioned, then I'll remove the extra code.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -117,6 +117,10 @@
 	//  Always show mouse cursor during tactical view (if false, no mourse cursor is shown when moving in real-time mode, selecting a merc, etc)
 	//  vanilla setting: false
 	"always_show_cursor_in_tactical": false,
+	
+	//  Show chance-to-hit next to mouse cursor when preparing an attack
+	//  vanilla setting: false
+	"show_hit_chance": false,
 
     //  IMP creation tweaks
     "imp": {
--- a/src/game/Tactical/DisplayCover.cc
+++ b/src/game/Tactical/DisplayCover.cc
@@ -24,6 +24,7 @@
 #include "ContentManager.h"
 #include "GameInstance.h"
 #include "WeaponModels.h"
+#include "policy/GamePolicy.h"
 
 #include <algorithm>
 
@@ -415,14 +416,6 @@ void DisplayRangeToTarget(SOLDIERTYPE* s, INT16 const sTargetGridNo)
 		ScreenMsg(FONT_MCOLOR_LTYELLOW, MSG_INTERFACE,
 				st_format_printf(zNewTacticalMessages[TCTL_MSG__RANGE_TO_TARGET_AND_GUN_RANGE],
 				GCM->getWeapon(s->inv[HANDPOS].usItem)->usRange / 10, usRange));
-		// Get the chance to hit
-		UINT32 const uiHitChance = CalcChanceToHitGun(s, sTargetGridNo, s->bAimTime, s->bAimShotLocation, false );
-		// Get the cover modifier chance from a soldier to a grid location
-		UINT8 ubChanceToGetThrough = SoldierToLocationChanceToGetThrough(s, sTargetGridNo, s->bTargetLevel, s->bTargetCubeLevel, 0);
-		//display a string with the weapons range, then range to target
-		ScreenMsg(FONT_MCOLOR_LTYELLOW, MSG_INTERFACE,
-				st_format_printf(zNewTacticalMessages[TCTL_MSG__CHANCE_TO_HIT_TARGET],
-				uiHitChance, uiHitChance*ubChanceToGetThrough/100.0f));
 	}
 	else
 	{
--- a/src/game/Tactical/UI_Cursors.cc
+++ b/src/game/Tactical/UI_Cursors.cc
@@ -28,6 +28,7 @@
 
 #include "ContentManager.h"
 #include "GameInstance.h"
+#include "policy/GamePolicy.h"
 
 #include <string_theory/string>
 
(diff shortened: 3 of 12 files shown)


assets/externalized/game.json
```diff
@@ -117,6 +117,10 @@
 	//  Always show mouse cursor during tactical view (if false, no mourse cursor is shown when moving in real-time mode, selecting a merc, etc)
 	//  vanilla setting: false
 	"always_show_cursor_in_tactical": false,
+	
+	//  Show chance-to-hit when pressing 'F' and next to mouse cursor when preparing an attack
+	//  vanilla setting: false
+	"show_hit_chance": true,
```
lynxlynxlynx, line 123: If we're adding a new config key, then I'm for removing the hotkey. Or at least making it a toggle like caps lock.

src/game/Tactical/DisplayCover.cc
```diff
@@ -415,14 +416,18 @@ void DisplayRangeToTarget(SOLDIERTYPE* s, INT16 const sTargetGridNo)
...
-		// Get the chance to hit
-		UINT32 const uiHitChance = CalcChanceToHitGun(s, sTargetGridNo, s->bAimTime, s->bAimShotLocation, false );
-		// Get the cover modifier chance from a soldier to a grid location
-		UINT8 ubChanceToGetThrough = SoldierToLocationChanceToGetThrough(s, sTargetGridNo, s->bTargetLevel, s->bTargetCubeLevel, 0);
-		//display a string with the weapons range, then range to target
-		ScreenMsg(FONT_MCOLOR_LTYELLOW, MSG_INTERFACE,
-				st_format_printf(zNewTacticalMessages[TCTL_MSG__CHANCE_TO_HIT_TARGET],
-				uiHitChance, uiHitChance*ubChanceToGetThrough/100.0f));
+
+		if (gamepolicy(show_hit_chance)) // Only show chance-to-hit if the setting for it is turned on
```
lynxlynxlynx, line 420: This comment is redundant.

src/game/Tactical/UI_Cursors.cc
```diff
@@ -241,6 +244,26 @@ static UICursorID HandleActivatedTargetCursor(SOLDIERTYPE* const s, GridNo const
...
+			GridNo targetTile = gUIFullTarget ? gUIFullTarget->sGridNo : map_pos;
+			giHitChance = is_throwing_knife ? CalcThrownChanceToHit(s, targetTile, s->bShownAimTime / 2, s->bAimShotLocation) :
+				CalcChanceToHitGun(s, targetTile, s->bShownAimTime / 2, s->bAimShotLocation, false);
+			giHitChance *= SoldierToLocationChanceToGetThrough(s, targetTile, s->bTargetLevel, s->bTargetCubeLevel, 0) / 100.0f;
+		}
+	
+		// Attach chance-to-hit to mouse cursor
+		if(giHitChance != -1)
+		{
+			SetChanceToHitText(st_format_printf("%d%%", giHitChance));
```
lynxlynxlynx, line 261: Since you asked: there are also the native ST variants that just use `{}` for placeholders when you don't need any modifiers. Random example:
`ST::format("Invalid mineId {}", lair->associatedMineId);`

lynxlynxlynx:
Looks mostly fine to me, thanks!

The bot failure is an upstream problem that I reported earlier today: https://github.com/TheAssassin/appimagelint/issues/22

assets/externalized/game.json
FluffyQuack: You mean removing 'F' hot key? That also shows range information, if I recall. I think that was the intended usage of the hot key.

lynxlynxlynx: Just the extra functionality.

FluffyQuack: I'm okay with that. I doubt anyone would want to only see hit chance by pressing 'F' when they can make it visible on the mouse cursor.

lynxlynxlynx:
Considering the wide support for this placement, please just remove the leftovers, so we can merge.

https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1368#issuecomment-847352822

FluffyQuack:
Okay, I removed leftover code and did some minor tweaks (like making the config var false by default so it matches normal JA2 behaviour, and adding a comment explaining the point of the "hacky" gsMouseSizeYModifier value).

I ended up with a ton of minor commits, so I tried to squash them into one to clean up Git history... but that just ended up making Git history looking even messier. I'm new-ish to using Git so I'm not sure what the intended way of doing that is.

lynxlynxlynx:
Usually one does that with git rebase and then force pushing (since it rewrites history). Here I'll just do it through the GUI.

PR #1376, opened 2021-05-28
Config var for scaling website loading time on laptop

FluffyQuack:
Adds "website_loading_speed_scale" config var for scaling the speed of loading time of websites on the laptop.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -122,9 +122,9 @@
   //  vanilla setting: false
   "show_hit_chance": false,
 
-  //  Scale the loading speed of websites on the laptop. Setting this to 0.0 removes the loading entirely.
+  //  Scales the loading time of websites on the laptop. Lower value means faster loading. Setting this to 0.0 removes the loading entirely.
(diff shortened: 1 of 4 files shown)


ja2-derek:
Minor nitpick - I think it is scaling the loading time rather speed? I got a little confused reading the json comments. 

Maybe we can just say something like higher value means longer load time.

FluffyQuack:
> Minor nitpick - I think it is scaling the loading time rather than speed? I got a little confused reading the json comments.
> 
> Maybe we can just say something like higher value means longer load time.

Just committed a change that renamed the variable and changed the comments to make it clearer what it does.

lynxlynxlynx:
Thanks! Not sure why the needless whitespace change at the last minute though — try to refrain from that.

FluffyQuack:
> Thanks! Not sure why the needless whitespace change at the last minute though — try to refrain from that.

Oh, sorry, I didn't even notice that happened, and I'm not sure how it happened. I edited game.json using notepad++ but I didn't expect that to result in a formating change.

Issue #1382, opened 2021-08-26
Game is not using ICC color profile on windows


selaux:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong

Reported via Discord. When a color profile setting is used on Windows, ja2-stracciatella is not using it. Not sure if that also applies to other platforms.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.18.0 | -  |    no    |    unknown     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

PR #1383, opened 2021-08-27
Do not adjust screen brightness unless requested

ja2-derek:
During the investigation of #1382, I found that the ICC profile is reset when JA2 sets the brightness. 

The [SDL_SetWindowBrightness](https://wiki.libsdl.org/SDL_SetWindowBrightness) function adjusts gamma of the whole screen. On Windows it also resets the system's active color profile and undoes some other display options like those available on the NVIDIA Control Panel.

The launcher does not allow changing the brightness. Users may be better off adjusting the brightness at the device or OS / driver. The EngineOption of brightness defaults to 1.0, meaning no adjustment is requested (since #734):

```
    -brightness GAMMA_VALUE
                        Screen brightness (gamma multiplier) value to set
                        where 0.0 is completely dark and 1.0 is normal
                        brightness. Default value is 1.0
```

Or, should we just remove the `-brightness` option?

### DIFF
--- a/assets/distr-files-linux/ja2_manpage
+++ b/assets/distr-files-linux/ja2_manpage
@@ -21,7 +21,7 @@ Path to the original JA2 game
 Screen resolution, e.g. 800x600. Default value is 640x480
 .TP
 \fB\-brightness GAMMA\fR
-Screen brightness (gamma multiplier) value to set where 0.0 is completely dark and 1.0 is normal brightness. Default value is 1.0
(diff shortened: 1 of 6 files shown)


ja2-derek:
I have updated the implementation to avoid the special handling of an otherwise valid gamma value `1.00f`.

I have also patched the launcher and EngineOptions to set the brightness to `-1.0` by default.

ja2-derek:
And BTW, `-brightness` is not taking effect on my Linux laptop (Pop_OS!)

As per SDL docs:
> Many platforms will refuse to set the display brightness in modern times. You are better off using a shader to adjust gamma during rendering, or something similar.

Issue #1388, opened 2021-09-01
Incorrect adjacent teammates search


sevenfm:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/TacticalAI/FindLocations.cc#L480-L500
NumberOfTeamMatesAdjacent function is supposed to find if spot has friendly soldiers in adjacent spots, and is used in FindBestNearbyCover to penalize cover value for tiles to prevent overcrowding
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/TacticalAI/FindLocations.cc#L728
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/TacticalAI/FindLocations.cc#L885-L895
but with current code it just checks the tile itself.

Should be
```if ( sTempGridNo != sGridNo )
{
	const SOLDIERTYPE* const tgt = WhoIsThere2(sTempGridNo, pSoldier->bLevel);
	...
```
As sTempGridNo is the adjacent tile we need to check and sGridNo is the checked tile where we plan to move.

This fix should make AI less crowded and less vulnerable to grenades when in combat.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|    |   |    yes    |    no     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

Issue #1385, opened 2021-08-29
AI Order/Attitude incorrect comparison


sevenfm:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/TacticalAI/AIUtils.cc#L354-L360

Should be
```
pSoldier->bAttitude == DEFENSIVE || pSoldier->bAttitude == CUNNINGSOLO || pSoldier->bAttitude == CUNNINGAID 
```
since `DEFENSIVE` , `CUNNINGSOLO` and `CUNNINGAID` constants can only be used with `bAttitude`.
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/Tactical/Overhead_Types.h#L105-L116

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  |   |    yes    |    no     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

Issue #1384, opened 2021-08-27
Throw search ends if one the checked spots was too far


ja2-derek:
Seven reported a bug (vanilla bug, [fixed](https://github.com/aimnas/ja2-1.13-source-mirror/commit/bc7fe70fb75e8959b7ea79019397d5fd538fb383#diff-3d3284d19aee5ddb71be59fbf005357adceb07cb33c4a12d508f4c14f6d61b99) on 1.13) on [Discord](https://discord.com/channels/348449397308391426/842331863560618035/880875435108479026).

### What is wrong

> to stracciatella team: looks like Attacks.cc has a vanilla bug which was fixed in 1.13:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/TacticalAI/Attacks.cc#L702-L706
> should be
```c++
  // can't throw there!
  continue;
```
> since we need to check all spots and not just randomly exit if one of the spots it too far for a throw

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  master | -  |    yes|    no     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

PR #1281, opened 2020-11-26
Tactical AI patches from Unfinished Business

ja2-derek:
Closes #1198.

This PR ports the remaining UB patches mentioned in #1198. These are 2 small improvements to the tactical AI. Both are togglable via `game.json` and disabled by default.

I have tested this in game and it works fine. If you also want to test it, you can enable both options and monitor these [two](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/5dd6d95d72a68b3e510f35e3361184540601231d/src/game/TacticalAI/DecideAction.cc#L1994) [lines](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/5e26678e99be074cc7926079449905765797d343/src/game/TacticalAI/AIUtils.cc#L1206) of logging.

I would recommend enabling both options for more challenging tactical encounters especially in Wildfire and NightOps maps. These two new patches make the AI more defensive, and the modded maps have more many defensible locations. And you can no longer ambush enemies that easily -  also enemies on all the "guard towers" will stay where they should be.

## Rotting corpse mechanism

[...]

PR #1390, opened 2021-09-03
Tactical AI fixes

ja2-derek:
Including the simple and (seemingly) straight-forward fixes as suggested in #1384, #1385 and #1388. May add more commits if we have more of the bugs.

I propose to gate all these bug-fixes (together with #1281) behind one option in `game.json`. Defaults to enabled, but it can still be disabled to allow a true "vanilla" (and buggy) experience.

I plan to do a play-through with all fixes enabled, and see how it goes. I don't really have a systematic way to verify the AI fixes.

### DIFF
--- a/src/game/TacticalAI/AIUtils.cc
+++ b/src/game/TacticalAI/AIUtils.cc
@@ -357,7 +357,7 @@ void NewDest(SOLDIERTYPE *pSoldier, UINT16 usGridNo)
 	//pSoldier->sDestination = usGridNo;
 	BOOLEAN fSet = FALSE;
 
-	if ( IS_MERC_BODY_TYPE( pSoldier ) && pSoldier->bAction == AI_ACTION_TAKE_COVER && (pSoldier->bOrders == DEFENSIVE || pSoldier->bOrders == CUNNINGSOLO || pSoldier->bOrders == CUNNINGAID ) && (SoldierDifficultyLevel( pSoldier ) >= 2) )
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
You can use the existing main ai key, since that was added in a similar situation.

ja2-derek:
I have played for some time with the fixes. I don't notice anything changed, but that doesn't break anything either.

As you can see these fixes some obvious coding errors, I think we can just apply them directly.

selaux:
:+1: Sounds good to me.

ja2-derek:
New defects reported by Coverity - 

```
*** CID 240335:  Code maintainability issues  (UNUSED_VALUE)
/home/runner/work/ja2-stracciatella/ja2-stracciatella/src/game/TacticalAI/FindLocations.cc: 1627 in SearchForItems(SOLDIERTYPE &, ItemSearchReason, unsigned short)()
1621     							{
1622     								INT8 const new_rating = EffectiveArmour(&o);
1623     								INT8 current_rating = EffectiveArmour(&cur_armour);
1624     								if (current_rating >= new_rating) continue; // the new armour isn't better, skip
1625     								if (current_rating == 0) current_rating = 1; // avoid division by zero
1626     
>>>     CID 240335:  Code maintainability issues  (UNUSED_VALUE)
>>>     Assigning value from "100 * new_rating / current_rating" to "temp_value" here, but that stored value is overwritten before it can be used.
1627     								temp_value = 100 * new_rating / current_rating;
1628     							}
1629     						}
1630     						/* FALLTHROUGH */
1631     
1632     					case SEARCH_WEAPONS:
```

PR #1396, opened 2021-09-09
Add OpenBSD toolchain file and update build instructions

oldlaptop:
The documentation in COMPILATION.md for OpenBSD builds no longer works as written for me:

 * Upstream/bundled GTest doesn't seem to build for me:
```
/home/peter/ja2-build/lib-gtest/src/googletest/src/gtest-port.cc:200:19: error: implicit conversion changes signedness: 'int' to 'unsigned long' [-Werror,-Wsign-conversion]
  mib[5] = size / mib[4];
                ~ ^~~~~~
/home/peter/ja2-build/lib-gtest/src/googletest/src/gtest-port.cc:210:30: error: implicit conversion changes signedness: 'int' to 'unsigned long' [-Werror,-Wsign-conversion]
  for (int i = 0; i < size / mib[4]; i++) {
                           ~ ^~~~~~
/home/peter/ja2-build/lib-gtest/src/googletest/src/gtest-port.cc:214:10: error: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Werror,-Wsign-conversion]
  return nthreads;
  ~~~~~~ ^~~~~~~~
/home/peter/ja2-build/lib-gtest/src/googletest/src/gtest-port.cc:200:17: error: implicit conversion loses integer precision: 'unsigned long' to 'int' [-Werror,-Wshorten-64-to-32]
  mib[5] = size / mib[4];
         ~ ~~~~~^~~~~~~~
/home/peter/ja2-build/lib-gtest/src/googletest/src/gtest-port.cc:210:21: error: comparison of integers of different signs: 'int' and 'unsigned long' [-Werror,-Wsign-compare]
  for (int i = 0; i < size / mib[4]; i++) {
                  ~ ^ ~~~~~~~~~~~~~
5 errors generated.
```

 * miniaudio seems to need _POSIX_C_SOURCE defined:
```
In file included from /home/peter/src/ja2-stracciatella/src/sgp/SoundMan.cc:44:
/fritzchen-dump/lib-miniaudio/src/miniaudio.h:42410:5: error: expected unqualified-id                                                                                                                                                                              
int fileno(FILE *stream);                                                                                                        
    ^                                                                                                                            
/usr/include/stdio.h:410:20: note: expanded from macro 'fileno'
#define fileno(p)       (!__isthreaded ? __sfileno(p) : (fileno)(p))
[more along these lines]
```

These can all be solved with some CMake options at configure time (as documented in the procedure in the patch), though perhaps the miniaudio issue ought to get fixed more permanently. (Note that one can't just define _POSIX_C_SOURCE since that gets some BSD extensions hidden and that breaks other things, you need _BSD_SOURCE too.)


selaux:
Thanks. Could you put those CMake parameters into a `toolchain-openbsd.cmake` toolchain file? Then people can just do `cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/toolchain-openbsd.cmake ..`. For an example see: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/cmake/toolchain-macos.cmake

oldlaptop:
Possibly; I'll see if I remember to get around to that over the weekend.

oldlaptop:
As of the most recent commit there's a (very small) toolchain file that Works For Me (TM) on recent -current on amd64.

selaux:
Thanks for updating. I think Works For Me is preferable to having nothing at all 😉 .

PR #1400, opened 2021-09-18
Externalize vehicles

ja2-derek:
For #665. This externalizes the table data of [g_vehicle_type_info](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/Tactical/Vehicles.cc#L68) into JSON and model class. 

### Intended use-cases

- Change the number of seats, to fit in more soldiers
- Change the movement type, so the vehicle can traverse more terrains types other than roads (e.g. plains).
- Change the armour type, to make it tougher or easier to destroy

### Notes

- Adding new vehicles or re-ordering elements is likely not to work without code support
- `class VehicleModel` is a direct replacement of the old `struct VehicleTypeInfo`
- The helicopter does not seem to use the `enterSound` or `moveSound`
- Adding the `Eldorado` might work if it is also placed in a sector. I haven't tried.

### DIFF
--- a/assets/externalized/vehicles.json
+++ b/assets/externalized/vehicles.json
@@ -1,0 +1,72 @@
+/**
+ * Definitions of player-controllable transport vehicles.
+ *
+ * The ordering in this list is important, and the game has hard-coded references to the
+ * vehicles. For example, the game expects the HELICOPTER to be the 6th vehicle.
+ *
+ * In base vanilla game, only the Hummer, Ice-cream truck and Helicopter are available.
+ *
+ * Fields:
+ * - enterSound: Sound to play when a soldier enters the vehicle. The value must be a valid SoundID.
+ * - moveSound: Sound to play when the vehicle moves. The value must be a valid SoundID.
+ * - profileID: Profile ID associated to the vehicle.
+ * - movementType: Determines if the vehicle can travel through map sectors and how fast. Must be one of FOOT, CAR, TRUCK, TRACKED or AIR.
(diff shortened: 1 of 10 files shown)


assets/externalized/vehicles.json
```diff
@@ -0,0 +1,72 @@
...
+ * - moveSound: Sound to play when the vehicle moves. The value must be a valid SoundID.
+ * - profileID: Profile ID associated to the vehicle.
+ * - movementType: Determines if the vehicle can travel through map sectors and how fast. Must be one of FOOT, CAR, TRUCK, TRACKED or AIR.
+ * - armourType: Determines the vehicle's protection against damage. Use the internal name of any armour items.
+ * - seats: Maximum number of soldiers allowed in the vehicle. Must not be greater than 10.
+ */
+[
+    // Eldorado (unused)
+    {
+        "enterSound": 85, // S_VECH1_INTO
```
selaux, line 20: Can we just write the sound string here? Other wise we limit mods to use the sound files that are included with the game.

selaux: It's also done that way in `weapons.json`:

```
"Sound": "sounds/weapons/9mm single shot.wav",
```

selaux (commented):
LGTM. Except for the `soundId` things.

assets/externalized/vehicles.json
ja2-derek: Updated. Thanks!

PR #1401, opened 2021-09-18
Refactor the rest of file accesses to go through GCM where applicable

selaux:
This got quite big again, but mostly it's search and replace operations for file opening / reading / writing. It took me quite some time as well, so I hope I remembered all the changes correctly. This is the last step before fixing https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1268 / getting https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1275 up-to-date.

### Biggest changes

- `FileMan`:
  - Now a namespace
  - Renamed some functions
  - Extracted file related functions into `SGPFile`
 - `SGPFile`:
   - Is now a class instead of a struct and provides file related functionality (`FileRead(...)`) as methods `file->read(...)`
 - `DirFs` in C++:
   - Similarily to the VFS `DirFs` this provides all functionality that `FileMan` provides but within a subdirectory
   - Used in `ContentManager` to provide access to tempDirs (`GCM->tempFiles()->openForReading(...)`) and files in ja2 home dir ((`GCM->userPrivateFiles()->openForReading(...)`))

### Removed

- `LoadShadeTablesFromTextFile`: The related boolean was never set to true, so the function was never called
- `WriteOutCurrentImpCharacter`: The current imp character was always written out to `imp.dat`, but this file was never used. Instead we now write out the current character on every save regardless of the `load_saved_merc_by_nickname` policy. This way they can be reused once the switch is enabled, without requiring doing a new save
- `Fs_read`, `Fs_write` was replaced by just opening files and saving to them.


src/externalized/DefaultContentManager.cc
```diff
@@ -180,10 +178,8 @@ DefaultContentManager::DefaultContentManager(RustPointer<EngineOptions> engineOp
 	RustPointer<char> vanillaGameDir{EngineOptions_getVanillaGameDir(m_engineOptions.get())};
 
 	RustPointer<char> stracciatellaHome{EngineOptions_getStracciatellaHome()};
-	m_userHomeDir = stracciatellaHome.get();
 
-	RustPointer<char> dataDir{Fs_resolveExistingComponents(BASEDATADIR, vanillaGameDir.get(), true)};
-	m_dataDir = dataDir.get();
+	m_userPrivateFiles = std::unique_ptr<DirFs>(new DirFs(stracciatellaHome.get()));
```
ja2-derek, line 182: I think it's better and safer to use `std::make_unique<DirFs>(stracciatellaHome.get())`. As a bonus, it makes the line shorter.

src/sgp/SGPFile.cc
```diff
@@ -0,0 +1,297 @@
...
+#include "RustInterface.h"
+#include "Exceptions.h"
+#include "Logger.h"
+
+#include <string_theory/string>
+#include <string_theory/format>
+
+#define SDL_RWOPS_SGP 222
+
+void DeleteSGPFile(SGPFile *file)
```
ja2-derek, line 11: Maybe not this time, but next time we can refactor away the `AutoSGPFile` class and this function.

src/sgp/FileMan.cc
```diff
@@ -1,344 +1,63 @@
...
 }
 
-
-File* GetRealFileHandleFromFileManFileHandle(const SGPFile* f)
-{
-	return f->flags & SGPFILE_REAL ? f->u.file : nullptr;
-}
-
-uint64_t GetFreeSpaceOnHardDriveWhereGameIsRunningFrom(void)
+uint64_t FileMan::getFreeSpaceOnHardDriveWhereGameIsRunningFrom(void)
```
ja2-derek, line 47: Out of scope of this PR and probably only on rare cases - 

This check is not accurate. We should check the disk free on temp and userPrivate directories, not the game executable.

ja2-derek: Is this the purpose / side-effect of the `FileMan::switchTmpFolder` line that was just removed?

src/externalized/DefaultContentManager.cc
selaux: Fixed.

src/sgp/SGPFile.cc
selaux: Yes, I had started with that, but that would have bloated the PR even more.

src/sgp/FileMan.cc
selaux: Good catch. Yes before it checked the user directory, with `switchTmpFolder`. I added a commit that adapts this check to check for min(userDir, tempDir).

ja2-derek:
LGTM! The Rust interfacing code is now wrapped neatly.

PR #1402, opened 2021-09-19
File open speedup

selaux:
Implements the changes from: https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1275.

Closes: https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1268

Also implements automated buffered file reads when opening files.


rust/stracciatella/src/vfs/android.rs
```diff
@@ -75,27 +93,42 @@ impl AssetManagerFs {
...
+                        .collect();
+                    canonicalization_cache.put(candidate.clone(), entries);
+                    canonicalization_cache
+                        .get(&candidate)
+                        .expect("we should be able to get a cache key that was just set")
+                };
+
+                for (nfc, os_string) in entries {
+                    if want == nfc.as_str() {
+                        next.push(candidate.join(&os_string));
```
lynxlynxlynx, line 131: Can `entries` validly contain duplicates or can we continue here?

lynxlynxlynx, line 131: Can `entries` validly contain duplicates or can we continue here?

rust/stracciatella/src/vfs/android.rs
```diff
@@ -75,27 +93,42 @@ impl AssetManagerFs {
...
+                        .flat_map(|path_buf| {
+                            path_buf
+                                .to_str()
+                                .map(|s| (Nfc::caseless(s), path_buf.into()))
+                        })
+                        .collect();
+                    canonicalization_cache.put(candidate.clone(), entries);
+                    canonicalization_cache
+                        .get(&candidate)
+                        .expect("we should be able to get a cache key that was just set")
```
ja2-derek, line 126: I don't understand this part. Is this testing code, or what use-cases does it cover?

rust/stracciatella/src/vfs/android.rs
```diff
@@ -57,6 +65,7 @@ impl AssetManagerFs {
         Ok(Rc::new(AssetManagerFs {
             base_path: base_path.to_owned(),
             asset_manager,
+            canonicalization_cache: Mutex::new(LruCache::new(CANONICALIZATION_CACHE_SIZE)),
```
ja2-derek, line 68: What does the Mutex protects us from? If there are concurrent access (do we really do so?) to the same key, isn't the worst case just false cache-miss?

Can we simplify the code and tolerate race conditions?

rust/stracciatella/src/vfs/android.rs
```diff
@@ -3,25 +3,33 @@
...
 
+use lru::LruCache;
 use ndk::asset::{Asset, AssetManager};
 
 use crate::android::get_asset_manager;
 use crate::unicode::Nfc;
 use crate::vfs::{VfsFile, VfsLayer};
 
+/// The size of the cache used for canonicalization
+const CANONICALIZATION_CACHE_SIZE: usize = 256;
```
ja2-derek, line 22: Is this a bit on the low side? 

My thinking is: the strategic map has more than 100 sectors and each visited sector can have several temp files. We can have up to 20 soldiers in team and each soldier have more than 10 BATTLESNDs.

This is text values, so the memory consumption should be nothing when compared to the graphics.

Of course, we could always load a late-game save and measure the actual cache usage.

ja2-derek: It is actually fine for mobile devices which could have stricter memory limits. So this question should have been on the same line in `dir.rs`

rust/stracciatella/src/vfs/android.rs
selaux: It's needed because we move the ownership into `canonicalization_cache`, so we cannot access the `entries` variable anymore. This is why it is necessary to use `.get` to get a reference to it.

rust/stracciatella/src/vfs/android.rs
selaux: In theory it can contain updates, though I am not sure if we need to handle it (e.g. to paths resolve to the same NFC). We could also just continue...

rust/stracciatella/src/vfs/android.rs
selaux: This was more or less chosen arbitrarily. I need to check actual cache usage, since I wrote the original code long ago.

rust/stracciatella/src/vfs/android.rs
selaux: It's hard to simplify this in rust, as there are strict interior mutablility rules. We could use some lock-free cache instead of `lru`, I just opted with lru because it is the more widely used choice. Or I could try to use `Cell` or `RefCell` to move ownership checks to runtime.

rust/stracciatella/src/vfs/android.rs
ja2-derek: I get it now. I am not used to the implicit `return`

rust/stracciatella/src/vfs/android.rs
ja2-derek: I see. Using lock-free sounds good, but I don't think we really need that.  Let's just keep what we have.

rust/stracciatella/src/vfs/android.rs
selaux: So looking at the actual cache usage, the current cache size never gets full, even when having all data dumped from the `*.slf` files. This is because it contains directory listings, so each directory within the DirFS only occurs once in the cache.

rust/stracciatella/src/vfs/android.rs
selaux: I looked into it a little more. This case handles the issue when there are two paths e.g. `test/foo/baz` and `test/FOO/bar` that have the same base directory in different casing. Always choosing the first match here, would not allow us to find `test/FOO/bar` because `test/FOO` is always normalized to `test/foo`.

PR #1406, opened 2021-09-22
Use require function instead of custom require

selaux:
A bit of cleanup in the scripting area:

- Require the use of `require` function instead of custom `JA2Require`
- Errors that happen when `main.lua` is executed are catched correctly

### DIFF
--- a/src/externalized/scripting/ScriptingExtensionsLua.cc
+++ b/src/externalized/scripting/ScriptingExtensionsLua.cc
@@ -37,7 +37,6 @@ struct GAME_OPTIONS;
     */
 struct TacticalStatusType;
 
-static std::set<std::string> loadedScripts;
 static bool isLuaInitialized = false;
 static bool isLuaDisabled = false;
 static sol::state lua;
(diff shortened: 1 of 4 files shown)


src/externalized/scripting/ScriptingExtensionsLua.cc
```diff
@@ -51,28 +50,49 @@ static void RegisterLogger();
...
+		sol::stack::push(L, errorMessage.c_str());
+		return 1;
+	}
+}
+
+void RunEntryPoint()
 {
 	if (isLuaInitialized)
 	{
 		throw std::runtime_error("JA2Require is not allowed after initialization");
```
ja2-derek, line 75: This is probably useless now.

ja2-derek:
Clean up looks good.

Do you know if module-require is supported as well? Like this - https://www.tutorialspoint.com/lua/lua_modules.htm

The GR mod example should be updated too. 

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/assets/mods/generous-rebels/data/scripts/main.lua#L1

src/externalized/scripting/ScriptingExtensionsLua.cc
ja2-derek: BTW, my original thinking was to force the requiring, parsing and validation of all script files at init stage, so it fails early instead of bugging out in the middle of a game.

selaux: Good point, we could still keep this, but I don't see the necessity to restrict things like that. I would rather add some documentation for conventions in Lua.

Another thing I noticed is that it probably makes sense to namespace scripts such as `/example-mod/data/scripts/example-mod/script.lua` (by convention). So that

- the chance that requires accidentially require things from other mods is minimized
- multiple mods can still require each others lua scripts, so a mod that provides basic lua functionality can be built upon (similarily to for example [HugsLib](https://steamcommunity.com/workshop/filedetails/?id=818773962) for Rimworld)

selaux:
Yes module-require works. I removed the remaining references to Ja2Require.

src/externalized/scripting/ScriptingExtensionsLua.cc
ja2-derek: The namespace convention is a good idea. Otherwise there will likely be filename clashes.

This should also apply to the entrypoint script, so each mod has a distinct entrypoint script like `$MODNAME/main.lua` or `$MODNAME.lua`.

PR #1409, opened 2021-09-24
Dynamic tactical bottom bar

ja2-derek:
To dynamically draw the bottom bar panels, to support different squad sizes.

# Summary

When squad size is 6 or below it is same as before. When the squad size is 7 or above, we will "assemble" the panel with multiple blits at init. Below shows the panels for squads of 8:

Screenshot from 2021-09-24 11-12-39

Screenshot from 2021-09-24 11-16-59

# Caveats

Squad size externalization is yet to be implemented. For now you can only do it with a custom build from source with a changed `NUMBER_OF_SOLDIERS_PER_SQUAD` (and that will break save game too).

Squad size of `n` requires horizontal resolution of `640 + (n - 6) * 83`. Squad size of 8 requires screen width of at least 806 px.

If the screen is not wide enough to show all squad slots, the last few mercs will not have a slot in the team panel. The merc can only be selected with mouse click in tactical, but works otherwise normally. The game warns about this at start up.

### DIFF
--- a/src/game/Tactical/Interface_Panels.cc
+++ b/src/game/Tactical/Interface_Panels.cc
@@ -6,6 +6,7 @@
 #include "Soldier_Ani.h"
 #include "Timer_Control.h"
 #include "VObject.h"
+#include "VSurface.h"
 #include "SysUtil.h"
 #include "Overhead.h"
 #include "MouseSystem.h"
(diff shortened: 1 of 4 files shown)


selaux:
:+1: Good idea! LGTM.

PR #1411, opened 2021-09-25
Change the squad save format

ja2-derek:
This is a proposal to update the Squad save format, to allow squad sizes other than 6 in the save and to allow loading saves done with different squad sizes. It retains some compatibility to all old versions down to 1.12 vanilla.

In the current (old) save format, squad size is always 6. The full squads list is a simple list of 120 (20 squads * 6) slots, each containing one `pSoldier->ubID` followed by 10 bytes of unused space.

# New squad data format

To support dynamic squad sizes, we store also the squad ID with each slot in the unused space, instead of assuming fixed-size lists.

At squad size of 6 or lower, the data layout is unchanged, and so it is fully compatible with all previous versions; At higher squad sizes, there are not enough space in the data region to store all unused the squad slots.

For backwards compatability, we keep the data to 120 (6 soldiers * 20 squads) records and pad each squad to multiples of 6. The 120 slots is more than enough for the max number (18) of soldiers, with any squad sizes including padding.

## Examples

Below are some illustrations when the squad size is different.

At squad size of 7 or higher, not all squads will be saved. The data layout is different are illustrated below - `S` and `s` are soldiers in different squads and `.` are empty slots in squads.

Saving squads of size 8, in the new format:

```
Squad 1 | S S S S S S
Squad 1 | S S . . . .
Squad 2 | s s . . . .
        | . . . . . .
```

In all old game versions, it is interpreted as:
```
Squad 1 | S S S S S S
Squad 2 | s s . . . .
Squad 3 | S S . . . .
Squad 4 | . . . . . .
```

When we open the save with squad size 4 in new game version, then it is re-interpreted as:
```
Squad 1 (S) | S S S S s s
Squad 2 (s) | s s . . . .
Squad 3     | S S . . . .
            | . . . . . .
```
## Problems

There are problems only when a new save with big squads is loaded in old versions. There are no problems when the saves are loaded in the same JA2:S version, no matter how the squad size has changed.

The save version is already incremented for the next release in #1282, so players will get the generic save version warning popup.

### When loading big squads in old versions

When squad size is higher than 6 and the squads are not in continuous numbering, in the new save format:
```
Squad 3 | S S S . . .
Squad 5 | s s . . . .
        | . . . . . .
```

Old versions interprets as Squads 1 and 2:

```
Squad 1 | S S S . . .
Squad 2 | s s . . . .
Squad 3 | . . . . . .
```

Once loaded, the soldiers will still have `bAssignment` of Squads 3 and 5. It doesn't affect much other than display. The player can re-assign the squads to fix.

### Dead mercs

There is an `sDeadMercs` array storing dead soldiers on each squad. The list is kept at size 6. In case of a squad wipe, that more than 6 mercs on squad are killed, only the first 6 are kept in the team panel slot. The rest will still "die normally" but get dropped from the squad right after.

### DIFF
--- a/src/game/SaveLoadGame.cc
+++ b/src/game/SaveLoadGame.cc
@@ -130,7 +130,7 @@ static const ST::string g_savegame_ext   = "sav";
 //Global variable used
 
 extern		INT32					giSortStateForMapScreenList;
-extern		INT16					sDeadMercs[ NUMBER_OF_SQUADS ][ NUMBER_OF_SOLDIERS_PER_SQUAD ];
+extern		INT16					sDeadMercs[ NUMBER_OF_SQUADS ][ NUMBER_OF_DEAD_SOLDIERS_ON_SQUAD ];
 extern		INT32					giRTAILastUpdateTime;
 extern		BOOLEAN				gfRedrawSaveLoadScreen;
 extern		UINT8					gubScreenCount;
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Sound good and a quick look at the diff didn't leave me any questions.

selaux:
LGTM as well.

PR #1414, opened 2021-09-26
Fix UI locking up when item is spotted during movement

### LINKED ISSUE #1367: Game temporarily hangs (stopwatch) when spotting an item mid movement

### What is wrong
When spotting an item during a squad member's movement, you are forced to wait an extra 10-20 seconds as a stopwatch appears before you can continue issuing any orders. Playing on Windows 10 and I have uploaded a save where the issue can be replicated by sending Razor north past the corpse at the corner of the hospital about 10 tiles above him.  https://cdn.discordapp.com/attachments/615653794079965204/841833910820012062/QuickSave.sav

|  release | JA2 Gold   |    no    |   unknown     |

ja2-derek:
Fixes #1367. 

This is a logic error introduced in #1325([diff](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1325/files#diff-5c7a7fd92593a4c65764acb3fec8e25342b7d717db783924d012b1761cf8f9baR7464)), the original condition to unlock the UI (`UnSetUIBusy`) checks for:
```c++
pSoldier->pTempObject == NULL
```
And `fIsThrowing` is defined as:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c75a2773eb1905c82111b9bb294ca7ecf5fb5bb0/src/game/Tactical/Soldier_Control.cc#L7383

### DIFF
--- a/src/game/Tactical/Soldier_Control.cc
+++ b/src/game/Tactical/Soldier_Control.cc
@@ -7461,7 +7461,7 @@ void HaultSoldierFromSighting( SOLDIERTYPE *pSoldier, BOOLEAN fFromSightingEnemy
 	}
 
 	// Unset UI!
-	if ( fFromSightingEnemy || ( fIsThrowing && !pSoldier->fTurningToShoot ) )
+	if ( fFromSightingEnemy || ( !fIsThrowing && !pSoldier->fTurningToShoot ) )
 	{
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Yep, thanks!

PR #1417, opened 2021-09-28
Add NULL check when counting shots hit

### LINKED ISSUE #1365: Crash when doing aimed burst to certain NPCs

### What is wrong
I consistently get a crash when using the aimed burst (hold B then drag mouse while holding LMB) in a line to damage civilians. In vanilla 1.12, there is no crash and NPCs go hostile. I get the rust error `Command_execute "D:\\Program Files\\JA2 Stracciatella\\ja2.exe" []: exited with status code -1073741819`
This has been present for years in Stracciatella in various release/nightly builds and makes the aimed burst unusable. Playing on Windows 10 and here is a link to a save file where the issue can be replicated quickly. https://cdn.discordapp.com/attachments/615653794079965204/841232788539572234/SaveGame03.sav

|  release |  JA2 Gold   |    no    |    no     |

ja2-derek:
Fixes #1365. 

Adds a NULL check, and skips the `usShotsHit++` if the current `pFirer` does not have a target. In this case the firing soldier is not aiming at people but at grids, and the `tgt` just happens to be standing there.

### DIFF
--- a/src/game/Tactical/LOS.cc
+++ b/src/game/Tactical/LOS.cc
@@ -1996,10 +1996,10 @@ static BOOLEAN BulletHitMerc(BULLET* pBullet, STRUCTURE* pStructure, BOOLEAN fIn
 		}
 		iDamage = BulletImpact(pFirer, &tgt, ubHitLocation, iImpact, sHitBy, &ubSpecial);
 		// handle hit here...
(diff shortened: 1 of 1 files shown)


selaux:
:+1: Thanks

Issue #1386, opened 2021-08-30
guiBaseJA2Clock signed/unsigned inconsistency


sevenfm:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
guiBaseJA2Clock is defined as UINT32 in Timer_Control.cc, but later it is assigned or compared to INT32 variables, which may sometimes result in quest related bugs.

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/game/Utils/Timer_Control.cc#L20

Possible solution: reset guiBaseJA2Clock once it increments over max INT32 value in Timer_Control.cc:
```c
if (!gfPauseClock)
{
	guiBaseJA2Clock += BASETIMESLICE;

	// fix possible signed/unsigned inconsistency in assignment or comparison later
	if ((INT32)guiBaseJA2Clock < 0)
		guiBaseJA2Clock = 0;

	for (UINT32 i = 0; i != NUMTIMERS; i++)
...
```

A more complex fix would be to check all instances in the code where timer is used and make sure all used variables are of the same signed/unsigned type.

This fix was originally suggested by Terapevt on Ja2.su forum:
http://forum.ja2.su/cgi-bin/yabb/YaBB.pl?board=ja2v113;action=display;num=1221160473;start=4113#4113

> There is such a cyclic bug in vanilla. Affects all NPCs, but most often they bump into it with the Boss. Only it does not depend on progress, but simply on the passage of time (a very specific overflow of one of the timer variables). In a day or two, it passes by itself.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|     |   |    yes   |    no     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[...]

PR #1418, opened 2021-09-28
Use UINT32 to handle GetJA2Clock time

ja2-derek:
This is a find-and-replace update to ensure we use `UINT32` variables to handle clock time. I searched for keywords including `GetJA2Clock` and `timer`, and replaced the usage of `INT32` if it is not appropriate.

None of the variables are persisted in saves, so it is safe to change the variable type.

Related to #1386.


lynxlynxlynx:
I think it causes more signedness warnings, but I didn't notice any obvious underflows or things that would be practically important. 👍🏾

ja2-derek:
Thank you for catching that. 

The new commit reverted the change on that variable (`INT32 iLastFileClicked`) flagged by compiler. It does not store clock time and should not be changed to UINT32.

Issue #1419, opened 2021-09-30
Bugs with Strategic AI


ja2-derek:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)
The strategic AI felt a bit too easy no matter how I configure it, but I don't think I am such a good player... So I instead found a couple of suspicious places in the code.

### 1. Garrison groups not properly re-filled

This one is caused by https://github.com/ja2-stracciatella/ja2-stracciatella/commit/1321c50c2af04149cea30ff0d152ac9205a553ec. In Strategic_AI.cc:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c75a2773eb1905c82111b9bb294ca7ecf5fb5bb0/src/game/Strategic/Strategic_AI.cc#L2005-L2006

`iWeight` is now unsigned UINT32, but in later usage, it is assigned from GARRISON_GROUP.bWeight, which may be negative when the garrison has too many soldiers (see [here](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c75a2773eb1905c82111b9bb294ca7ecf5fb5bb0/src/game/Strategic/Strategic_AI.cc#L499-L518)). `iWeight` underflows to very large positive number, and then all garrison reinforcements are denied.

As a result, the Strategic AI seems to be doing much less than it should.

### 2. Defeated patrol groups not re-created

This seems to be the vanilla behaviour (bug?).

In Stracciatella codebase, which has not been changed since initial import, the function [PatrolRequestingMinimumReinforcements](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/c75a2773eb1905c82111b9bb294ca7ecf5fb5bb0/src/game/Strategic/Strategic_AI.cc#L3264-L3285) always return `FALSE` for defeated Patrol Groups. Compare to the JA2-1.13 version [here](https://github.com/aimnas/ja2-1.13-source-mirror/blob/40c5e115ed23a0da0d1e3bae26393f8be059a562/Strategic/Strategic%20AI.cpp#L6113-L6125).

Maybe this should be yet another Strategic AI policy option?

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
[...]

PR #1420, opened 2021-10-01
Fixing signedness errors in Strategic AI

ja2-derek:
As mentioned in #1419, there are some regression in Strategic AI behaviour since 0.17.0. It is caused by underflow of unsigned integers.

## Changes

 - Variables dealing with group weights are changed back to signed (`INT32`)
 - Renamed some variables to be consistent with the type, so it is easier to spot similar issues going forward
 - Reverted a change (`UINT32` back to `INT32`)in LoadStrategicAI. This does not seems to have caused any problems, but it is an implicit cast from unsigned to signed. It might have an effect if we have platforms of different endianness.


lynxlynxlynx:
Thanks!

PR #1421, opened 2021-10-01
Option to refill totally defeated patrol groups

ja2-derek:
The second issue in #1419.

The vanilla Strategic AI seems not replacing a totally defeated patrol group. It is able to reinforce if the group has some of the soldiers killed but the group is still active. But if the group is destroyed, it will never get replaced.

Here we have a new option `refill_defeated_patrol_groups` in the Strategic AI policy. I set it to disabled by default for consistency with vanilla, though I would personally always play with it enabled.


selaux:
:+1: Thanks.

PR #1422, opened 2021-10-01
Add JSON schema for all externalized files

selaux:
I'm looking out for feedback for this one. It adds JSON schema files for all JSON files in the `externalized` directory. This allows the game to validate the JSON before reading it in C++, ideally providing useful error messages when the format does not match (though this is not really implemented yet). JSON schema files also would allow for:

- A better editing experience in editors that support it, e.g. by getting autocomplete information from the JSON schema
- Autogenerated reference documentation generated from the JSON schema

Do you think this could be useful? If so I would add more descriptions etc. to the JSON schema or use the ones included in the JSON files already.

A related things I noticed while extracting the schemas:

- The naming in the JSON files is inconsistent. Some use underscores (e.g. `game.json`), some use camel case with type hints (e.g. `weapons.json`) and some use plain camel case (e.g. `army-compositions.json`)
  Proposal: Always use unhinted camel case
- Sometimes entities are refererred to by string ids, sometimes numerical ids
   Proposal: Always use string ids when referring to entities. All entities should have a string id, also the ones currently not referred to from anywhere. This way we can pave the road to allowing mods to include partial JSON files, which can be merged with those of other mods / stracciatella
- Often the top level of the JSON file is an array. While this seems quite intuitive, it does not allow for the [$schema](http://json-schema.org/understanding-json-schema/reference/schema.html#id4) property to be used to define the schema for that particular file.

Doing adaptations here will lead to existing mods breaking. Any ideas how to avoid this?

TODO:
- [x] Improve error message when json schema validation fails
- [x] Add descriptions and comments to json schema


ja2-derek:
I like the idea of having schemas and I hope it can help us with:

### 1. Better editing experience for modders

Is doc root `$schema` a VS Code thing? The official spec does not require `$schema` in the data. It is supposed to be out-of-band metadata. That said, it is great to have a recommended editor that works out of the box, but I just wonder if there are something more generic.

I am not against changing the data format to allow metadata. We still have an excuse that we are still on version 0.x. We can then change to a consistent document structure for all JSONs, e.g. for files with table data:
```json
{ 
    "_schema": "./schemas/schema.json",
    "_tableData": [
        { /* entities.*/. },
         ...
    ]	
}
```

Or, are there any well-known / standard way to provide a mapping from documents to schemas? Something like the [Schema Store](https://www.schemastore.org/json/)?


### 2. Code generation and validation

I want code-gen to replace the model classes and deserialization. It would be even better if we can generate Lua mappings too.

Also, we can clean up the comments in the JSON.

----

Did you write the JSON schema by hand? How much maintenance overhead will it take to maintain the schemas?

Re string keys, it can be difficult as the codebase and map files uses numeric indices everywhere.

selaux:
1. Yes, I thought this would be generic, but it seems to be a VSCode thing. It's also possible for some editors (including VSCode again) to do a manual mapping from JSON to it's schema. I don't think there is a standard way though. There is also the possibility to generate editor UIs, e.g. [with json-editor](https://json-editor.github.io/json-editor/)
2. That is the primary function, which is already implemented in this PR, although the error messages still need improvement
3. Regarding codegen for parsing the only thing I found is [json-schema-codegen](https://github.com/pearmaster/json-schema-codegen), which would require `boost` again. There is also [schemafy](https://github.com/Marwes/schemafy) for Rust, but then we would need to write wrappers for the rust code, so I don't know if that really reduces maintenance overhead.

---

I wrote all schemas by hand for this PR. I would say they are quite easy to maintain once you get used to write json schemas. There are also possibilities to automatically extract the schema from JSON, but they dont care for e.g. limited integer types such as `uint8`. They will just extract `"type": "integer"`

Regarding using IDs I want to emphasize that I mean mostly references from JSON to JSON. So I have no issues with numeric internal ids hanging around for now, but IDs that reference other JSON objects should use strings, so we have the ability to phase out the numeric ids, once we have the ability to get rid of them in code.

selaux:
Here is an example for some generated documentation from JSON schema: https://storage.googleapis.com/ja2-builds/modding/schema.html

lynxlynxlynx:
How is the user experience? We used to crash in ugly ways when we had problems in game.json. This could be a way to also help users upgrade their local config copies on upgrades.

selaux:
This should be done now.

lynxlynxlynx:
Just took a cursory look, so it doesn't mean much, but great work!

rust/stracciatella/src/schemas/yaml/types/sector-id.schema.yaml
```diff
@@ -0,0 +1,4 @@
+type: string
+title: Sector ID
+description: String representation of sector ids, such as `A1`, `A2`, `B1` etc.
+pattern: ^[A-Z][0-9]+$
```
ja2-derek, line 4: Should we use a stricter regex? The valid range is only `A1` - `P16`. 

I mean like `^[A-P][1-9][0-6]?$` or even `^[A-P]([1-9]|1[0-6])$` but the latter might be unnecessarily overly complicated.

assets/externalized/strategic-map-creature-lairs.json
```diff
@@ -4,10 +4,10 @@
  * Field definitions:
  *  - lairId:    A unique numeric ID for the lair. Must be an integer greater than zero.
  *  - associatedMineId:    Creature lair must be located and associated with one mine.
- *  - entranceSector:    Where the lair entrance is located. There must be an alternate map file (*_a.dat), which will be activated at lair initialization.
+ *  - entranceSector:    Where the lair entrance is located. There must be an alternate map file (_a.dat), which will be activated at lair initialization.
```
ja2-derek, line 7: Why is this needed? Is `*` invalid in comments / schema?

rust/stracciatella/src/schemas/yaml/tactical-map-item-replacements.schema.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+    This is intended for compatibility and for maps that was originally built 
+    for different item sets (e.g. Wildfire maps).
+
+    Items are identified by either item index (if given integer), or internal 
+    name (if given string). Items mapped to 0 ("Nothing") are removed from map 
+    placements.
+  properties:
+    from:
+      $ref: types/id.schema.yaml
```
ja2-derek, line 20: Both `from` and `to` values can be numeric ID or internal name. It is because we need to replace items which does not even exist in our config or with an invalid ID. This helps supporting maps ported from 1.13.

See also: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/src/externalized/tactical/MapItemReplacementModel.cc#L3

rust/stracciatella/src/schemas/yaml/vehicles.schema.yaml
```diff
@@ -0,0 +1,51 @@
...
+    enterSound:
+      title: Enter sound
+      description: Sound file to play when a soldier enters the vehicle.
+      $ref: types/resource-path.schema.yaml
+    moveSound:
+      title: Move sound
+      description: Sound file to play when the vehicle moves.
+      $ref: types/resource-path.schema.yaml
+    profileID:
+      title: Profile ID
```
ja2-derek, line 24: Should we use a different type for merc profile ID? It should be always be `0 < n <= 200`.

rust/stracciatella/src/schemas/yaml/strategic-map-underground-sectors.schema.yaml
```diff
@@ -0,0 +1,84 @@
...
+        type: string
+        enum:
+        - N
+        - S
+        - W
+        - E
+    numTroops:
+      title: Number of troops
+      description: |
+        The numbers of enemy troops in sector, as an array of 3 integers, e.g. [1,2,3] meaning: 1 enemy on DIF_LEVEL_EASY; 2 on DIF_LEVEL_MEDIUM; 3 on DIF_LEVEL_HARD
```
ja2-derek, line 48: Should we also add the number of item to the schema? There should be exactly 3 items in the array.

assets/externalized/strategic-map-creature-lairs.json
selaux: You catched me :grin:. It should be perfectly valid but there is a [bug with the rust json reader](https://github.com/tmccombs/json-comments-rs/pull/9), which I am currently using to build an editor based on the JSON data: https://github.com/selaux/stracciatella-toolset

rust/stracciatella/src/schemas/yaml/vehicles.schema.yaml
selaux: As one of my goals is to replace all occurences of `profileID` with a string id pointing to `mercs-profile-info.json` I will add it to `mercs-profile-info.json` and leave it at that.

rust/stracciatella/src/schemas/yaml/tactical-map-item-replacements.schema.yaml
selaux: Fixed. Thanks for the catch!

rust/stracciatella/src/schemas/yaml/types/sector-id.schema.yaml
selaux: Makes sense. Added the strictest one.

ja2-derek:
Looks good to me. Thank you @selaux!

rust/stracciatella/src/schemas/yaml/strategic-map-underground-sectors.schema.yaml
selaux: Done.

ja2-derek:
We've got a few of new gcc warnings in DefaultContentManager.cc:

```
2022-01-04T11:21:19.1989694Z In file included from /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.cc:1:
2022-01-04T11:21:19.1994282Z /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.h: In constructor ‘DefaultContentManager::DefaultContentManager(RustPointer<EngineOptions>)’:
2022-01-04T11:21:19.1999822Z /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.h:241:19: warning: ‘DefaultContentManager::m_vfs’ will be initialized after [-Wreorder]
2022-01-04T11:21:19.2007436Z   RustPointer<Vfs> m_vfs;
2022-01-04T11:21:19.2008173Z                    ^~~~~
```

```
2022-01-04T11:21:19.2010450Z In file included from /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.cc:1:
2022-01-04T11:21:19.2013455Z /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.h:162:29: warning:   ‘RustPointer<SchemaManager> DefaultContentManager::m_schemaManager’ [-Wreorder]
2022-01-04T11:21:19.2016128Z   RustPointer<SchemaManager> m_schemaManager;
2022-01-04T11:21:19.2016969Z                              ^~~~~~~~~~~~~~~
2022-01-04T11:21:19.2018840Z /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.cc:170:1: warning:   when initialized here [-Wreorder]
2022-01-04T11:21:19.2021100Z  DefaultContentManager::DefaultContentManager(RustPointer<EngineOptions> engineOptions)
2022-01-04T11:21:19.2022170Z  ^~~~~~~~~~~~~~~~~~~~~
```

```
2022-01-04T11:21:19.2024771Z /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.cc: In member function ‘std::unique_ptr<rapidjson::GenericDocument<rapidjson::UTF8<> > > DefaultContentManager::readJsonDataFileWithSchema(const ST::string&) const’:
2022-01-04T11:21:19.2048012Z /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.cc:1108:14: warning: comparison with string literal results in unspecified behavior [-Waddress]
2022-01-04T11:21:19.2059536Z     if (ty == "integer") {
2022-01-04T11:21:19.2060421Z               ^~~~~~~~~
2022-01-04T11:21:19.2062783Z /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.cc:1120:14: warning: comparison with string literal results in unspecified behavior [-Waddress]
2022-01-04T11:21:19.2065256Z     if (ty == "integer") {
2022-01-04T11:21:19.2066059Z               ^~~~~~~~~
2022-01-04T11:21:19.2067365Z In file included from /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/IGameDataLoader.h:3,
2022-01-04T11:21:19.2069423Z                  from /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.h:7,
2022-01-04T11:21:19.2071850Z                  from /home/runner/work/ja2-stracciatella/ja2-stracciatella/src/externalized/DefaultContentManager.cc:1:
```

Issue #1337, opened 2021-03-18
Make save game location configurable on Android


nedsed:
I have tested JA2S version 0.17.1 on Android 9 on a Galaxy S8 and while it ran pretty well, I noted that there is no way to access the savegames on a non-rooted device. This is kind of unfortunate because that way you cannot move your game progress back and forth between PC and smartphone, old savegames cannot be continued etc.

I'm not a Android developer but couldn't you add an option for a user defined folder for savegames just like the chooser for the game data directory? That way savegames could be stored and loaded from an unprotected location.

Thanks for keeping this game alive!

selaux:
Yes I think making it configurable on Android should be possible.

PR #1429, opened 2021-10-17
Add separate tab for mods to launcher

### LINKED ISSUE #1352: Launcher: preload all mods
Now it looks like only one mod is packaged, but actually all are available, one just has to click on Add first.

EDIT: hmm, that affects the save dir name, even if the mods are not enabled. Not sure if it matters.

selaux:
This should fix the confusion about enabled vs available mods in the launcher. It adds a new mod tab to the launcher that shows clearly which mods are available and which are enabled in two separate lists. Mods can then be enabled/disabled by moving them between lists using the buttons in the middle.

Closes #1352

Screenshot from 2021-10-17 15-06-28

I needed to increase the launcher size a little in order to fit all mod names. For longer mod names it will show a horizontal scroll bar. There is some behavior implemented to lead the user to correctly use the interface (e.g. buttons are enabled/disabled depending on which list is selected)

I intentionally left some space below the mods to be able to display mod information (e.g. from a mod manifest) later.


ja2-derek:
I like the new tab. It is much more intuitive, to me.

But I wonder if `mods-available` and `mods-enabled` are clear enough for people who never did Linux admin before. Maybe we can say something like "mods disabled" instead; or, is it easy to add mouse-over tooltip?

selaux:
I adapted the naming to Enabled/Disabled mods. I also tried to add some explanation for priority to the tooltips of the respective buttons.

Regarding the "Data" tab I would leave it as is right now. I think there might be new content coming in the future there (e.g. #1337).

ja2-derek:
@lynxlynxlynx, do you know if the build failure is a problem with AppImageLint, or something to be changed in build config?

```
  File "/tmp/.mount_appimas4L9EQ/usr/conda/lib/python3.7/site-packages/appimagelint/checks/gnu_abi_check_base.py", line 131, in _check_ubuntu_compat
    max_supported_version = versions_map[release]
KeyError: 'impish'
```

lynxlynxlynx:
Pretty sure it's the same cache issue it had before. We were discussing it upstream, but since it went away on its own back then, perhaps nothing was done to mitigate in the future, even though it was discussed. 
https://github.com/TheAssassin/appimagelint/issues/22

Issue #551, opened 2017-06-08
graphical issues when moving camera position during character movement


Tercus:
I'm getting heavy graphical artefacts when I'm moving the camera, while a character is moving on the screen. It gets a lot worse, if "Merc Lights during Movement" is enabled. If it is disabled, the Mercenary is simply not visible until he has reached his destination

Attached are two videos, because it is hard to explain and it was too difficult to create it as gif. The first video shows how it is without the option enabled and the second one with it enabled.
[JA2 Debug.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1062560/JA2.Debug.zip)

JAsmine-ja2:
"Merc Lights during Movement" is known to cause a lot of trouble. This problem is known to v1.13, too. I don't think, they came up with a better solution than deactivating this option.

chmallon:
I think, this is solvable -- mostly by getting rid of unnecessary surfaces for rendering.
When I originally ported JA2, I tried to translate the low-level rendering mechanics faithfully.
It seems, this was faithful enough to also retain this vanilla display bug.
Scrolling is done by copying the screen buffer (aka back buffer) to itself (this also causes the blitting problems with SDL < 2.0.5 and < 1.2.13) while rendering is done on the frame buffer.
Usually (part of) the frame buffer is copied to the screen buffer.
I think the problem is as follows:
When merc lighting during movement is on, this causes parts from the frame buffer, which were not updated nor scrolled, to be copied to the screen buffer.
Probably this copying is triggered by the call chain ExecuteOverhead() -> DeleteSoldierLight() -> LightSpriteDestroy() -> LightSpriteDirty().
This low-level rendering is unnecessarily complicated (and was even more complicated originally when it used DirectX), so by simplifying it, the scroll+moving light glitch could be solved.

Fun experiment:
Try non-video scrolling, i.e. scrolling by re-renderding instead of copying.
Hit ALT+V. (if you did not remove this functionality)
Then scrolling while a luminous merc moves should not jump around anymore.
Though the merc itself will not be visible while scrolling due to a bug in ScrollBackground().

flaviojs:
I was considering creating another issue, but it's probably the same underlying cause so I'll post it here.

After entering cheat mode (CTRL+GABBI) and enabling the display of FPS (CTRL+F), you will see traces of the text while you scroll around. This resembles what you show in the second video.

[...]

PR #1437, opened 2021-11-06
Fix graphical glitching during scrolling when merc lights are enabled

rebb:
This seems to fix the bad glitches during camera scrolling while mercs are moving and the "merc lights during movement" option is enabled.

It's kind of a workaround-y way to do it i guess, until a better fix comes along.

Related issues : #551

### DIFF
--- a/src/game/GameScreen.cc
+++ b/src/game/GameScreen.cc
@@ -61,6 +61,7 @@
 #include "GameMode.h"
 #include "EditScreen.h"
 #include "Logger.h"
+#include "GameSettings.h"
 
 #define ARE_IN_FADE_IN( )		( gfFadeIn || gfFadeInitialized )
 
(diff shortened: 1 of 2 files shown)


selaux:
For some reason, there are no lights at night at all for me currently. Can anyone else confirm / review this one? @ja2-derek @lynxlynxlynx

ja2-derek:
I never noticed any glitches about the merc lights, so I have no idea what this PR fixes.

I built this PR and it seems to work fine. I tried it briefly with "merc lights during movement" on and off, and it seems to work well. There are lights on the merc body and around.

rebb:
The issue pops up when a merc ( merclights on ) at night is moving, and you scroll the camera on him at the same time.
It causes really bad graphical corruption.

rebb:
Here's a video showing the problem and how it looks when it is fixed. ( Finally an excuse to try out GitHub Video support ;) )

https://user-images.githubusercontent.com/3278522/141314118-8dc9f4b2-eba9-49b7-be3f-243ee7190f14.mp4

Should i make a new Issue for this ? I figured #551 would cover it pretty well.

lynxlynxlynx:
Thanks for the extra effort! I can't build right now, but it looks clear enough.

selaux:
I think so too. Lets merge.

PR #1438, opened 2021-11-06
Fix camera focus when clicking on overhead map

rebb:
Clicking the overhead map was broken and often ended up in strange places. This should fix it.


rebb:
Thanks :)

PR #1439, opened 2021-11-06
Highlight compatible items in sector inventory

rebb:
Normally the map inventory screen only highlights compatibles when hovering on an item that is in the map inventory, showing compatibles in the map and on the current merc.

This patch makes it so that hovering over a merc item shows compatibles in the map inventory as well.


selaux:
Hi, thanks for your PR. I think this definitely makes sense to include. Please ensure to make PRs against `master` (the development branch) rather than `nightly` in the future.

rebb:
Ok, i will - Thanks and sorry :D

PR #1445, opened 2021-11-29
Chinese Language Support

noword:
* Modify the codes of font system, let it support Chinese.
* Add the Chinese translation in src/game/Utils/_ChineseText.cc, it's base on #810 ChinaSakhalin's work.
* Add a Chinese Localization mod.

I create another project for managing the resource of Chinese Localization.
https://github.com/noword/ja2-chinese-resource


noword:
The check of Linux mingw64  failed.
```
CMake Error at /usr/local/share/cmake-3.22/Modules/BundleUtilities.cmake:471 (file):
  file READ_ELF given FILE
  "/home/runner/work/ja2-stracciatella/ja2-stracciatella/ci-build/_CPack_Packages/win64/ZIP/ja2-stracciatella_0.19.0-1445pullrequest+0b55377_win-mingw64-cross/ja2.exe"
  that is not a valid ELF file.
```
I don't get it, why READ_ELF  a exe file.

assets/mods/chinese-localization/manifest.json
```diff
@@ -0,0 +1,5 @@
+{
+    "name": "Chinese Localization",
+    "description": "Chinese Localization",
```
lynxlynxlynx, line 3: Traditional or simplified?

src/externalized/DefaultContentManager.cc
```diff
@@ -898,6 +898,7 @@ void DefaultContentManager::loadStringRes(const ST::string& name, std::vector<co
 	case GameVersion::POLISH:       fullName += "-pl";    break;
 	case GameVersion::RUSSIAN:
 	case GameVersion::RUSSIAN_GOLD: fullName += "-rus";   break;
+    case GameVersion::CHINESE: fullName += "-chs";   break;
```
lynxlynxlynx, line 901: tabs for indentation

src/game/Utils/WordWrap.cc
```diff
@@ -422,19 +491,35 @@ UINT16 IanWrappedStringHeight(UINT16 max_w, UINT8 gap, SGPFont font, const ST::u
...
+					i++;
+				}
 				break;
 
 			case TEXT_CODE_NEWCOLOR:
 			case TEXT_CODE_DEFCOLOR:
+				if (*(i + 1) == TEXT_SPACE)
+				{
+					i++;
+				}
```
lynxlynxlynx, line 506: This repeats a lot, so I suggest you add a static function to replace the calls. `SkipSpace` sounds like an appropriate name.

src/launcher/Launcher.cc
```diff
@@ -34,7 +34,8 @@ const std::vector<GameVersion> predefinedVersions = {
 	GameVersion::ITALIAN,
 	GameVersion::POLISH,
 	GameVersion::RUSSIAN,
-	GameVersion::RUSSIAN_GOLD
+	GameVersion::RUSSIAN_GOLD,
+    GameVersion::CHINESE
```
lynxlynxlynx, line 38: tab

src/sgp/TranslationTable.h
```diff
@@ -10,17 +10,18 @@
 class CTranslationTable
 {
 public:
-	unsigned char m_table[TRANSLATION_TABLE_SIZE];
+	UINT16 m_table[TRANSLATION_TABLE_SIZE];
```
lynxlynxlynx, line 13: With the bigger size and type this will always use over 40x more memory. It's still in the kilo range, so likely it's cleaner to make everyone suffer than to ifdef chinese.

lynxlynxlynx:
Thanks! I have some further questions:
- what is the license of the new (mod) data and fonts?
- how gracefully is choosing chinese, but not enabling the mod handled?

src/sgp/TranslationTable.cc
```diff
@@ -909,15 +909,7131 @@ class CTranslationTableRussianGold : public CTranslationTable
...
+		m_table[40682] = 6967; // 黪
+		m_table[40687] = 6968; // 黯
+		m_table[40738] = 6969; // 鼢
+		m_table[40748] = 6970; // 鼬
+		m_table[40751] = 6971; // 鼯
+		m_table[40761] = 6972; // 鼹
+		m_table[40759] = 6973; // 鼷
+		m_table[40765] = 6974; // 鼽
+		m_table[40766] = 6975; // 鼾
+		m_table[40772] = 6976; // 齄
```
ja2-derek, line 7894: I guess this is generated from https://github.com/noword/ja2-chinese-resource/blob/master/GB2312_Level1.txt

We should read it from a resource file rather than hard-coding it. I suspect adding 8000 lines could slow down the builds. 

That said, this table just work fine, maybe further improvements can be made in a future PR.

src/externalized/DefaultContentManager.cc
noword: Thanks for replay.
Which indent tool and config should I use in this project.

ja2-derek:
I have tried the build. The translations and text display look fine. This is a big chunk of work and thank you!

src/externalized/DefaultContentManager.cc
lynxlynxlynx: For rust just use rustfmt, while for the inherited c++ code, we don't have a formalized standard yet. There was some discussion in a still open issue, but nothing has been codified. Standardizing indentation was a big first step a few years back, since we got the code in a mixed state.

Rule of thumb: match the indentation style in the surrounding code.

ja2-derek: Some guidance here: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/CONTRIBUTING.md#axioms-of-style

noword:
> Thanks! I have some further questions:
> 
> * what is the license of the new (mod) data and fonts?
> * how gracefully is handled choosing chinese, but not enabling the mod?
The translation texts is from the JA2 Gold 1.12 Chinese localization project, it was long time ago at bbs.tbsgame.net( This forum has been closed now). 
Actually, I'm the initiator/coder of this project. GPL or MIT is all fine with me.

The font is use Zfull-GB.ttf, I got it from https://github.com/andot/zfull-for-yosemite
This font was merged from many other fonts, so the license issue is complex.
But this font circulated on the Internet for a long time, I guess no one will be sued.

I think use the mod is the most graceful way to support Chinese, otherwise the resource files should be copied to the game directory, after doing that users can no longer switch language.

assets/mods/chinese-localization/manifest.json
noword: It's simplified.
I will replace all 'Chinese' to 'Simplified Chinese'.

src/sgp/TranslationTable.h
noword: May we could malloc the `CTranslationTable::m_table` dynamically.

`UINT16` m_table[0x452] increased 0x452 bytes only，compared to the original.

src/externalized/DefaultContentManager.cc
noword: Thanks, I will fix them.

lynxlynxlynx:
License: ok, that's fine then for the code. Anything merged uses the project license unless the author provides a different one. The font upstream should be mentioned somewhere though.

Gracefulness: I meant how the error handling works. Choosing Chinese should automatically enable the required mod and choosing a different language should automatically disable the mod.

noword:
> License: ok, that's fine then for the code. Anything merged uses the project license unless the author provides a different one. The font upstream should be mentioned somewhere though.
> 
> Gracefulness: I meant how the error handling works. Choosing Chinese should automatically enable the required mod and choosing a different language should automatically disable the mod.

I see, you are right. 
But rust is a new program language for me, I don't known how to use it.
Maybe in the future I will learn it.

src/sgp/TranslationTable.cc
noword: Yes, it's very ugly.

lynxlynxlynx:
I think it's actually a tcl thing or maybe c++, unless we want to force it also for command line users.

assets/mods/chinese-localization/manifest.json
ja2-derek: How about also adding a short usage instructions to tell players to choose "Chinese" for game version?

src/sgp/TranslationTable.h
noword: 75febe0f50387ab5b29391e460cbe71ecc9e39c9
I change the type of m_table to std::vector<UINT16>.
Hope it can be solved the memory size problem.

noword:
> I think it's actually a tcl thing or maybe c++, unless we want to force it also for command line users.

335321f3a9c0410078578672a1e872e7f1ffbe69
Now it will enable Chinese mod automatically. If mean while a message dialog poped up, I think it will be better. But I don't known how to realize that.

src/launcher/Launcher.cc
```diff
@@ -673,3 +673,41 @@ void Launcher::moveDownMods(Fl_Widget* widget, void* userdata) {
...
+			if (strcmp(modId, SIMPLIFIED_CHINESE_MOD_NAME) == 0)
+			{
+				window->enabledModsBrowser->insert(1, window->availableModsBrowser->text(i));
+				window->enabledModsBrowser->data(1, window->availableModsBrowser->data(i));
+				window->enabledModsBrowser->select(1, 1);
+				window->availableModsBrowser->hide(i);
+				break;
+			}
+		}
+	}
```
lynxlynxlynx, line 697: This should also do the opposite, disable the mod when a non-chinese language is chosen.

noword: Done.

src/sgp/TranslationTable.cc
noword: If we use some resource files, where should them be placed?

Or just provide 5 full arrays in cc file for initializing vector/array. 
It's more efficient when running, not set array's data one by one compared with before.

ja2-derek: > If we use some resource files, where should them be placed?

I would say put it in a new directory under `assets/externalized`.

Run-time efficiency should not be a big concern. It might take slightly longer (if at all noticeable) to initialize at launch, but we save the compilation time and memory for people not using Chinese mods.

selaux:
One more question from me: Would it make sense to use JSON for the `mercedt/*` files (e.g. like the [test-json-dialogs](https://github.com/ja2-stracciatella/ja2-stracciatella/tree/master/assets/mods/test-json-dialogs/data/mercedt) mod does)? This would reduce the amount of binary data in the mod and make easier to change later if we need to.

As a sidenote this PR is actually useful to find cases where we are missing exactly that functionality, so maybe it makes sense to create a bug out of it to externalize some more texts, so the mod can become a JSON only mod in the future.

src/sgp/TranslationTable.cc
noword: Make 5 new files:
```
assets/externalized/EnglishTranslationTable.json
assets/externalized/FrenchTranslationTable.json
assets/externalized/RussianTranslationTable.json
assets/externalized/RussianGoldTranslationTable.json
assets/externalized/SimplifiedChineseTranslationTable.json
```
Then read them dynamically depending on the selected language.
Am I  right?
If we do so, where the `TranslationTable.h` and `TranslationTable.cc` should be placed?
Still at `src/sgp`,  or `src/externalized`?

ja2-derek:
The mingw build is working on my laptop (Ubuntu hirsute, gcc 10). The CI build failure might be related to GCC/MingW versions.

src/sgp/TranslationTable.cc
lynxlynxlynx: Sounds good to me and I don't think the source particular placement matters much.

ja2-derek:
Can you clean up the commit history (maybe with a interactive git-rebase)? Thanks!

noword:
> Can you clean up the commit history (maybe with a interactive git-rebase)? Thanks!

Looks like I make a mistake, didn't create a branch before  submitting a pull request.
I'm a newbie in git, hope i won't mess up.

src/sgp/TranslationTable.cc
noword: Done.
Now, the code for reading translation table from json files is in DefaultContentManager.cc.

src/externalized/DefaultContentManager.cc
```diff
@@ -1254,6 +1256,56 @@ void DefaultContentManager::loadVehicles()
...
+	}
+	fullName += "-translation-table.json";
+
+	m_translationTable.assign(tableSize, 0);
+
+	auto json = readJsonDataFile(fullName);
+	int count = 0;
+	for (auto& element : json->GetArray())
+	{
+		m_translationTable[ST::string(element.GetString()).to_utf16()[0]] = count++;
```
lynxlynxlynx, line 1305: Is all this conversion jumping through hoops really needed?

noword: Use numbers in json?
```
[
65, /* A */
66, /* B */
67, /* C */
.....
]
```
Json does not support comments. but rapidjson has no complaints about it.

lynxlynxlynx: That or a map, but I don't know if we need to bother, since I imagine it's still a fast conversion.

lynxlynxlynx:
Are you otherwise done with the changes or is this it?

src/externalized/DefaultContentManager.cc
noword: The original implementation used an array, I think it's fine.
Array is faster than map when running, but the advantage is very small,  can be ignored.

noword:
> Are you otherwise done with the changes or is this it?

No, I write a script comparing files one by one.

> Are you otherwise done with the changes or is this it?

I think it's done.

lynxlynxlynx:
That's a bit of an ambiguous set of answers, but let's just merge and if anyone wants to further improve it, it can be done via a separate PR. Thanks a lot!

PR #1453, opened 2022-01-06
Replace references to profileID in externalized JSON with profile name

selaux:
This replaces references to merc profiles by their string id. It helps with understandability of the json to make modding easier and more consistent. The references are always cross-checked with `mercs-profile-info.json`.

Additionally it fixes the new compiler warnings introduced with https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1422.

### Breaking changes

This is a breaking change for mods that modify the following files. I'm planning to to some more breaking changes for `0.19.0`, so mod authors only have to adapt once.

- `dealers.json`
- `mercs-MERC-listings.json`
- `strategic-map-npc-placements.json`
- `mercs-rpc-small-faces.json`
- `vehicles.json`

Also some of the dealer inventory files have changed filenames, as the dealer itself now uses the profile id instead of its separate id.


src/externalized/mercs/MERCListingModel.cc
```diff
@@ -36,24 +36,51 @@ static SpeckQuoteType SpeckQuoteTypefromString(std::string s)
 	throw std::runtime_error("unsupported quote type: " + s);
 }
 
-MERCListingModel* MERCListingModel::deserialize(uint8_t index, const rapidjson::Value& json)
+MERCListingModel* MERCListingModel::deserialize(uint8_t index, const rapidjson::Value& json, const MercSystem* mercSystem)
 {
+	ST::string profileName = json["profile"].GetString();
+	auto profile = mercSystem->getMercProfileInfoByName(profileName);
+	if (profile == NULL) {
+		throw std::runtime_error(ST::format("'{}' does not refer to a valid profile", profile).c_str());
```
ja2-derek, line 44: Should we use `DataError` here and below?

selaux: Fixed.

Issue #1451, opened 2022-01-04
MinGW cmake packaging failure


ja2-derek:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong

MinGW build has been failing for quite a while. It still builds on my laptop. By trial-and-error, I found it works on older CMake and breaks on 3.22. It last worked with CMake 3.21.4.

I haven't figured out which commit on CMake broke it. We might need to isolate the problem and file a bug on CMake.

Quick work-around is to install CMake 3.21 in `ci-setup.sh`:

```sh
    ## package target is broken on newer CMake versions
    curl -sL https://cmake.org/files/v3.21/cmake-3.21.4-linux-x86_64.sh -o cmakeinstall.sh \
        && chmod +x cmakeinstall.sh \
        && sudo ./cmakeinstall.sh --prefix=/usr/local --exclude-subdir \
        && rm cmakeinstall.sh
```
This is based on: https://github.com/actions/virtual-environments/blob/b7f276c003aea42575b52247bdb2183e355fca2f/images/linux/scripts/installers/cmake.sh

-----

When the build fails, the logs are like:

```
CPack: Install projects
CPack: - Run preinstall target for: ja2-stracciatella
CPack: - Install project: ja2-stracciatella []
CMake Error at /usr/local/share/cmake-3.22/Modules/BundleUtilities.cmake:471 (file):
  file READ_ELF given FILE
  "/home/runner/work/ja2-stracciatella/ja2-stracciatella/ci-build/_CPack_Packages/win64/ZIP/ja2-stracciatella_0.19.0-git+f7dfca6_win-mingw64-cross/ja2.exe"
  that is not a valid ELF file.
Call Stack (most recent call first):
  /usr/local/share/cmake-3.22/Modules/BundleUtilities.cmake:527 (get_item_rpaths)
  /usr/local/share/cmake-3.22/Modules/BundleUtilities.cmake:593 (set_bundle_key_values)
  /usr/local/share/cmake-3.22/Modules/BundleUtilities.cmake:934 (get_bundle_keys)
  /home/runner/work/ja2-stracciatella/ja2-stracciatella/ci-build/install-dlls-mingw.cmake:104 (fixup_bundle)
[...]

PR #1455, opened 2022-01-09
Install CMake 3.21 for MinGW builds

ja2-derek:
The GitHub Action ubuntu image always [installs](https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/cmake.sh) the latest CMake release, but the MinGW packaging target fails since CMake 3.22. Check #1451 for details.   

This is a temporary work-around and should fix the builds for now.  If CMake can address this problem in a future release, we should revert changes in this PR.


lynxlynxlynx:
Thanks!

PR #1456, opened 2022-01-09
Externalize all weapon shooting sounds properly

selaux:
Previous status was that weapon sounds could be modded to some degree, but it was complicated.

Example: `weapons.json` contained both `Sound` and `BurstSound` as paths. Sound was actually used, but in most cases the sound was only dependent on the calibre. `BurstSound` from `weapons.json` was only used to check wether it is not empty and then the burst sound from `calibre.json` was taken, so the actual value didn't matter.

Some weapon sounds were defined in `calibres.json` and some in `weapons.json`. Now both calibres and weapons have the same properties, where the weapon properties override the calibre ones:

- `sound`
- `burstSound`: Should contain a `%d` substring which is replaced by the number of shots fired by the burst
- `silencedSound`
- `silencedBurstSound`: Should contain a `%d` substring which is replaced by the number of shots fired by the burst

NB: All of the sounds are now optional. If there is no sound for a weapon or its calibre, the sound simply does not play.

## Breaking changes

- `calibres.json`
- `weapons.json`


lynxlynxlynx:
Nice and LGTM, through I didn't recheck the data.
It's a bit shocking how many times we have to look them up though, but of course that's inherited.

PR #1459, opened 2022-01-21
Externalize item graphics

selaux:
This changes externalized data, so images for items can be changed more easily. Before `ubGraphicType` and `ubGraphicNum` were used to determine in code which file to use for small/big inventory graphics as well as tile graphics. Now these graphics are referenced explicitly by the `inventoryGraphics` and `tileGraphic` properties. This allows them to be changed independent of each other and without having to overwrite the original item graphics.

I tried to implement the `inventoryGraphics` property as robust as possible. So when one of the images is not found, it simply shows a fallback image (see screenshots). The fallback images were created using free-ja2-resources here: https://github.com/ja2-stracciatella/free-ja2-resources/pull/8

Screenshot from 2022-01-21 12-00-55
Screenshot from 2022-01-21 12-01-00


src/externalized/DefaultContentManager.cc
```diff
@@ -1190,6 +1190,21 @@ const ItemModel* DefaultContentManager::getItemByName(const ST::string &internal
...
 
+std::vector<ST::string> DefaultContentManager::getAllSmallInventoryGraphicPaths() const
+{
+	std::vector<ST::string> v = {};
+
+	for (auto item : m_items) {
+		auto& path = item->getInventoryGraphicSmall().getPath();
+		auto existing = std::find(v.begin(), v.end(), path);
+		if (existing == v.end()) {
+			v.push_back(item->getInventoryGraphicSmall().getPath());
```
lynxlynxlynx, line 1201: Why not just push `path`?

src/externalized/InventoryGraphicsModel.cc
```diff
@@ -0,0 +1,21 @@
...
+
+JsonObject InventoryGraphicsModel::serialize(rapidjson::Document::AllocatorType& allocator) const {
+    JsonObject v(allocator);
+
+    auto s = small.serialize(allocator);
+    v.AddMember("small", s.getValue());
+    v.AddMember("big", big);
+
+    return v;
+}
```
lynxlynxlynx, line 21: Didn't we standardize on tabs? Either way line 9 is misaligned.

lynxlynxlynx:
Other than that, LGTM. But I didn't double check all the jsons. Is there any particular reason you made `small` an object and not just adding everything to the same level as `big`?

src/externalized/InventoryGraphicsModel.cc
selaux: Yep we did. But my editor didn't know. Reindented the new files with tabs.

src/externalized/DefaultContentManager.cc
selaux: Good point.

selaux:
The JSONs were adapted programatically, so I don't think there should be anything broken in there.

You have a good point about the `small` and `large` graphic variants. Originally in the game the `small` variant was always an STI with images for multiple items, while the `big` variant only had a single image. That is why the structure is was as you described. But I don't see why we should enforce that. Let people use whatever images they choose. I changed it to have the same structure.

I also changed a little bit of stuff in `Interface_Items` again, so images which have an out-of-bound `subImageIndex` are handled gracefully as well.

src/externalized/InventoryGraphicsModel.cc
ja2-derek: Maybe time for an EditorConfig!

selaux:
Adapted it once more to match https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1460

PR #1460, opened 2022-01-23
Add editorconfig

selaux:
This allows editors that support [editorconfig](https://editorconfig.org/) to automatically configure encoding, indentation and line endings.

From what I can see the editorconfig reflects what we have right now.


lynxlynxlynx:
Thanks!

PR #1462, opened 2022-01-23
Reduce direct references to item ids

selaux:
This removes the `FIRST_*` defines from `Item_Types.h` (and some other unused ones) and thus reduces direct dependencies on numeric item ids, with the goal to eventually get rid of them in code.

Sorry for the whitespace changes in `DefaultContentManager.cc`, but this is what my editor does now with the `editorconfig`. If you want I can remove them.


lynxlynxlynx:
Thanks and sorry, somehow this flew under everyone's radar.

PR #1466, opened 2022-02-09
Fix case-insensitive mod data directory not being found

### LINKED ISSUE #1465: Minor regression of VFS case-insensitivity

### What is wrong

We used to allow mods to have data directory of either `$MOD_NAME/Data` or `$MOD_NAME/data`. Just tried on master build and the former is no longer working on case-sensitive filesystems.

> sgp/SGP.cc: Game has been terminated due to an unrecoverable error: Failed to build virtual file system (VFS): Vfs_init_from_engine_options: Error initializing VFS for "$HOME/.ja2/mods/unfinished-business/data": No such file or directory (os error 2) (St13runtime_error).

|  master |   |    no    |    yes     |

selaux:
Should fix #1465.

I think I want some regression tests for that so this is still a draft.


lynxlynxlynx:
LGTM!

PR #1467, opened 2022-02-11
Remove condition for checking terrain

### LINKED ISSUE #1463: Incorrect condition, || instead of &&

### What is wrong

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/432ee5b6398b5dca72529129c7cfd0476a5689dc/src/game/Tactical/OppList.cc#L3517

Probably should be
if ((ubSourceTerrType >= FLAT_GROUND) && (ubSourceTerrType <= DEEP_WATER))
or else any ubSourceTerrType will pass the check, including NO_TERRAIN and NUM_TERRAIN_TYPES:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/432ee5b6398b5dca72529129c7cfd0476a5689dc/src/game/TileEngine/TileDef.h#L41-L56

|    |    |    yes    |    no     |

selaux:
Closes #1463.

I decided to remove the condition completely as later `ubSourceTerrType` is compared against `FLAT_FLOOR` anyways, so the condition does not really need `bCheckTerrain`, to know whether it has to check for the terrain.


lynxlynxlynx:
Ideally we'd use enum types to always ensure we have a valid value.

PR #1469, opened 2022-03-04
Make save game directory configurable

selaux:
This allows changing the save game location (closes https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1337). E.g. you can put it on a Cloud Storage and synchronize the Saves between devices.

The default is the same save game directory as before, which will now be created by `EngineOptions` instead of `ContentManager` if it does not exist. It adds options for those to both the  normal and Android launchers. It also changes the way how filenames are generated to ASCII only when saving, as some characters in file names are invalid on Android.


lynxlynxlynx:
Is this finished?

selaux:
Yes. This should be done.

PR #1471, opened 2022-03-19
removed remaining explicit guiBaseJA2Clock casts to int #1386

lynxlynxlynx:


src/game/Laptop/Laptop.cc
```diff
@@ -1990,9 +1987,10 @@ static void DisplayLoadPending(void)
 	// check to see if we are only updating screen, but not passed a new element in the load pending display
 
 	// decide how many time units are to be displayed, based on amount of time passed
-	for (INT32 i = 0, iTempTime = iTotalTime; i <= 30 && iTempTime > 0; ++i, iTempTime -= iUnitTime)
+	for (UINT32 i = 0, uiTempTime = uiTotalTime; i <= 30 && uiTempTime > 0; ++i, uiTempTime -= uiUnitTime)
```
ja2-derek, line 1990: This breaks the check for `iTempTime < 0`. `uiTempTime` is now unsigned, it will underflow instead of going negative.

ja2-derek: I see! Then should we remove the check `uiTempTime > 0` from this line?

lynxlynxlynx: No, it covers the case of it being exactly 0, which is still representable.

PR #1474, opened 2022-03-22
Encapsulte sectors

lynxlynxlynx:
- see description of first commit
- check the SGPSector class addition to Types.h
- this is far from finished, but works

Sorry for the barely reviewable commits 1 & 3, I just realized yesterday that there is a better way to do it. First I'll convert individual variables to the member ones and only later tackle converting functions that that enables. Should be able to separate changes better that way.


lynxlynxlynx:
Crap, accidentally pushed to master. Will revert, force pushing sounds risky.

PR #1479, opened 2022-03-23
Encapsulate sectors v2

lynxlynxlynx:
-   see description of first commit
-   check the SGPSector class addition to Types.h

Sorry for the barely reviewable commits 1 & 3, I just realized yesterday that there is a better way to do it. First I'll convert individual variables to the member ones and only later tackle converting functions that that enables. Should be able to separate changes better that way.

TODO:
- [x] finish migrating
- [x] remove ST::String constructor hack
- [ ] address new FIXME/TODO that are unrelated to the changes

The remaining TODO/FIXME notes might as well go into separate PRs, since if they are actually problems, they're there from before:
```
 // FIXME: suspicious, this ensures gfGotoSectorTransition is set only if the world is not underground — but why would it matter?
 // FIXME: useless check, always false
// TODO: verify, it's suspicious we pass a z coord instead of gridno
// TODO: verify, it's suspicious we pass a z coord instead of gridno
```

Two extra notes:
- didn't touch BeforeStructureDamaged, OnStructureDamaged, GetCurrentSector, since they're used by the Lua bindings
- iCurrentMapSectorZ and sSelMap are very often used together in mostly the same files; perhaps they could be merged (before this PR, sSelMap didn't have a Z coordinate), but it requires a bit deeper dive.


lynxlynxlynx:
Ok, this is pretty much complete now. I've rereviewed the whole work (that's what the force pushes were for) and the remaining TODO/FIXME notes might as well go into separate PRs, since if they are actually problems, they're there from before:
```
 // FIXME: suspicious, this ensures gfGotoSectorTransition is set only if the world is not underground — but why would it matter?
 // FIXME: useless check, always false
// TODO: verify, it's suspicious we pass a z coord instead of gridno
// TODO: verify, it's suspicious we pass a z coord instead of gridno
```

Two extra notes:
- didn't touch BeforeStructureDamaged, OnStructureDamaged, GetCurrentSector, since they're used by the Lua bindings
- iCurrentMapSectorZ and sSelMap are very often used together in mostly the same files; perhaps they could be merged (before this PR, sSelMap didn't have a Z coordinate), but it requires a bit deeper dive.

lynxlynxlynx:
I've tested this as far as rebuking the attack on the first SAM site and all seems well. Two things to recheck with a new game, as I've fixed some regressions during play, but in general, it seems a-ok and we can look at it again post-release.

lynxlynxlynx:
Ok, tested some more, underground sectors and transitions work fine. So this is ready for final review before merging.

src/game/Strategic/Assignments.cc
```diff
@@ -932,24 +923,25 @@ void UpdateAssignments()
 	// check for mercs tired enough go to sleep, and wake up well-rested mercs
 	HandleRestFatigueAndSleepStatus( );
 
+	SGPSector sector;
 	// run through sectors and handle each type in sector
-	for(sX = 0 ; sX < MAP_WORLD_X; sX++ )
+	for (sector.x = 0; sector.x < MAP_WORLD_X; sector.x++)
 	{
-		for( sY =0; sY < MAP_WORLD_X; sY++ )
+		for (sector.y = 0; sector.y < MAP_WORLD_X; sector.y++)
```
selaux, line 930: We should probably use `MAP_WORLD_Y` here, just to be clean.

src/sgp/Types.cc
```diff
@@ -0,0 +1,140 @@
...
+{
+	return !(*this == sector);
+}
+
+bool SGPSector::operator<(const SGPSector& rhs) const noexcept
+{
+	return x < rhs.x || y < rhs.y;
+}
+
+SGPSector SGPSector::operator+(const SGPSector& rhs) const noexcept
```
selaux, line 70: Should we also consider the z coordinate when adding / removing?

src/sgp/Types.cc
```diff
@@ -0,0 +1,140 @@
...
+{
+	return x == sector.x && y == sector.y && z == sector.z;
+}
+
+bool SGPSector::operator!=(const SGPSector& sector) const noexcept
+{
+	return !(*this == sector);
+}
+
+bool SGPSector::operator<(const SGPSector& rhs) const noexcept
```
selaux, line 65: I'm not sure about defining `<` for `SGPSector`. I think from a usage perspective it might not be clear what `sector1 < sector2` means.

src/game/TileEngine/Radar_Screen.cc
```diff
@@ -79,7 +79,7 @@ void InitRadarScreen()
 }
 
 
-void LoadRadarScreenBitmap(const char* const filename)
+void LoadRadarScreenBitmap(ST::string filename)
```
selaux, line 82: Should this rather be `const ST::string&`

src/game/Tactical/Overhead.cc
```diff
@@ -1981,10 +1981,14 @@ static BOOLEAN HandleAtNewGridNo(SOLDIERTYPE* pSoldier, BOOLEAN* pfKeepMoving)
 
 	if (pSoldier->bTeam == OUR_TEAM)
 	{
+		static const SGPSector drassen(13, MAP_ROW_B);
 		if (pSoldier->ubWhatKindOfMercAmI == MERC_TYPE__EPC)
 		{
 			// are we there yet?
-			if (pSoldier->sSectorX == 13 && pSoldier->sSectorY == MAP_ROW_B && pSoldier->bSectorZ == 0)
+			static const SGPSector sMap(13, MAP_ROW_B);
```
selaux, line 1988: Better name: `drassenAirport`.

selaux (approved):
Looked through everything and looks like it has been very well done :+1: (though my brain probably could not ingest everything). I added a few minor comments, nothing that would keep us from merging though.

src/game/Editor/EditorMapInfo.cc
```diff
@@ -215,9 +215,9 @@ BOOLEAN ApplyNewExitGridValuesToTextFields()
 	//exit grid input fields
 	if( iCurrentTaskbar != TASK_MAPINFO )
 		return FALSE;
-	str = ST::format("{c}{}", gExitGrid.ubGotoSectorY + 'A' - 1, gExitGrid.ubGotoSectorX);
+	str = ST::format("{}", gExitGrid.ubGotoSector.AsShortString());
```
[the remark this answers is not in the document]
lynxlynxlynx: True, this can be simplified — good catch.

src/sgp/Types.cc
lynxlynxlynx: It seems quite intuitive to me, in two ways. I imagine the sector grid and basically anything that's:
- within the bigger rectangle formed by the current sector as one vertex and 1,1 the other is *less* than the current sector (ignoring equality here)
- to the left or to the top of the current sector is less

lynxlynxlynx:
Thanks for taking a look, I know it's tedious.

lynxlynxlynx:
Applied suggestions. :)

selaux:
Thanks. There are conflicts now unfortunately.

PR #1498, opened 2022-04-09
Properly print CLI usage on CLI error

selaux:
The previous approach was limited by the 256 character limit for log messages. I decided to move the error logging to Rust and  use error types in `EngineOptions` initialization to get the information whether CLI parsing failed.

The second commit is what includes the change. The first commit is only restructuring for tests and removing some redundant functions (I couln't help myself :wink:).

Closes #1498


ja2-derek (commented):
LGTM. Thanks!

ja2-derek (approved):

Issue #1356, opened 2021-04-23
0.19 release checklist


lynxlynxlynx:
Lest I forget, we have a cleaner version of the release checklist in docs/.

## Pre-release development
- [x] Fix or postpone issues in the [milestone](https://github.com/ja2-stracciatella/ja2-stracciatella/milestone/7)
- [x] Triage open pull requests and recent bugs
- [x] Finalise choice of version number
  - [x] Decide if this version should be 1.0 #443 
- [x] Add new contributors to contributors.txt (make rebuild-contributors-list)
- [x] Update changes.md (run docs/draft-changelog.sh to get started)
- [x] Update man page if necessary

## Release candidate
- [x] Optionally update version number to `-rc` and tag
- [x] Mark the tag as a Github prerelease
- [x] Runtime tests of prerelease packages
  - [x] Linux
  - [x] Windows
  - [x] OS X
  
Runtime tests:
 * IMP creation
 * AIM and extending contracts
 * taking Drassen airport or more
 * buying from Bobby Ray's
 * visual inspection of laptop screens

If you want to build the packages manually:

    $ make clean
    $ make build-releases # to build all releases
    $ make build-win-release-on-linux # crosscompile Windows release on Linux
    $ make build-release-on-mac # build the mac bundle
    $ make package-appimage # build the appimage

# Release
- [x] Attach built versions to GitHub release
  - [x] Linux, Windows
  - [x] OS X
- [x] Document new features on the website
- [x] Make GitHub prerelase the release, update texts
- [x] Announce
  - [x] Write main announcement
    - mention c0281336fda
  - [x] Website download page (update frontmatter variables as needed)
  - [x] Website news
  - [x] Bear's pit
  - [x] Moddb

## Post-release
- [x] Set version to +1 and add the `-dev` suffix
- [x] Close this bug and GitHub milestone
- [x] Create new milestone if none exists yet

selaux:
As it was mentioned in Discord: Should we prioritize the issues currently in the `0.19` milestone and focus on the next release?

lynxlynxlynx:
Yep. I got distracted while waiting for feedback.

lynxlynxlynx:
Time flies! I thought it's been 5 years since the project has been revived in the current form (first commit here on 5.2.2016) and that it'd be nice to mention it in the release notes, but that's actually 6 already. 😃

lynxlynxlynx:
I wrote a draft of the announcement here:
[...]

PR #1505, opened 2022-04-12
Update SDL2 to 2.0.20 for macOS

selaux:
As mentioned in https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1356. Lets see if the builds work and behave better than the release ones before merging. After release it probably makes sense to update the library for other OSs as well.


lynxlynxlynx:
Cool, it builds now. Any more luck with this and newer fltk?

selaux:
Yep. Works better. Windowed mode and mouse issues are fixed. Pixel perfect scaling still does not work, but this might just be due to my display configuration.

Issue #1508, opened 2022-04-15
Small issues with the new Save Game screen


ja2-derek:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
Details, details, details.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.19.0  |  1.12  |    no    |    yes     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)
1. Open the save game screen
2. Click on any empty space in the main panel, to deselect any selected game slot
3. Click "Save"
4. Crashes with error:
   ```
   terminate called after throwing an instance of 'ST::unicode_error'
     what():  Invalid UTF-8 sequence byte
   Aborted (core dumped)
   ```

   The crash should be coming from this line, `header().sSavedGameDesc` being NULL:

    https://github.com/ja2-stracciatella/ja2-stracciatella/blob/8769f9bce12a8f2a9ff8f00ed974dae7c7ac6902/src/game/SaveLoadScreen.cc#L725

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

ja2-derek:
A couple more small issues with the new save screen:

1. There should be a button to delete old saves.
2. Small issue with empty save game description - when you save new file without description, the "Save Game" mouse button does not respond, but you can still save with "enter" on the keyboard.

lynxlynxlynx:
1. is an enhancement and 3. I wouldn't consider a bug.

selaux:
1. Should be already implemented (it's the delete key)
2. Agreed
3. I agree that it's different than before. Do you think it is better to hide it? I think it's good to give the players the possibility to pick up where they left, even if it sometimes does not work (maybe highlighting via a message that it sometimes might not work would be good).

ja2-derek:
[...]

PR #1509, opened 2022-04-16
Fix save load screen issues

selaux:
This should address the issues in the save load screen mentioned in #1508.

- No crashes anymore when deleting / saving without having a save game slot selected
- Make save button work while entering text in new save game


lynxlynxlynx:
So I guess something else enables the button once there's no more reason for it to be disabled?

selaux:
Yep. It is reenabled a few lines below when an existing save is selected (and probably in some more places to be honest).

PR #1514, opened 2022-04-21
Instructions and credits for the simplified chinese mod

ja2-derek:
The `simplified-chinese-localization` mod requires a little more instructions. The `manifest.json` is a nice place to put it.


lynxlynxlynx:
Not sure if it's really needed, since choosing Chinese enables the mod, but it doesn't hurt either.

ja2-derek:
Oh right, I forgot about that. But it doesn't (yet) work the other way - enabling the mod does not select the Game Version. Maybe we should be doing that instead.

PR #1515, opened 2022-04-22
Upgrade included SDL to 2.0.20

selaux:
Follow up of https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1505 for non-macOS systems.


selaux:
The Android version needed some convincing, as

- It seems SDL implicitly took care previously of linking against the android ndk, which we now have to do manually
- There were some changes to `hidapi`, so there is no separate library for that anymore. I updated the Android code by simply copying the [example files from SDL](https://github.com/libsdl-org/SDL/tree/main/android-project/app/src/main/java/org/libsdl/app) from the newest version.

I tested starting up the Android version, so everything seems to be fine now.

PR #1520, opened 2022-04-23
cmake: rewrite the contributor tracking target

lynxlynxlynx:
- simplifies the code by letting git deduplicate and removing legacy logic
- handles coauthorship (why Seven was uncaught in 0.19)
- minor cleanup afterwards


lynxlynxlynx:
Git can't do mixed sorting either, so that's the only part that wasn't redundant.

PR #1526, opened 2022-04-26
LUA and SOL2 may be taken from system now

timp87:
Ability to take LUA and SOL2 dependencies from the system


selaux:
Thanks. Ignore the build error. It seems like the Android signing key does not work properly for external contributors. I'll have a look into it.

PR #1527, opened 2022-04-26
Add read/write functionality for STCI to rust

selaux:
I've had this code lying around a bit and don't want to let it go to waste. It adds support for reading and writing STCI files in Rust.

If you have concerns regarding compile times (as the code is currently not used by the engine) let me know. I can adapt it to not compile by default, only if explicitly enabled.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -87,9 +87,9 @@ checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
 name = "backtrace"
-version = "0.3.64"
+version = "0.3.65"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5e121dee8023ce33ab248d9ce1493df03c3b38a659b240096fcbd7048ff9c31f"
+checksum = "11a17d453482a265fd5f8479f2a3f405566e6ca627837aaddb85af8b1ab8ef61"
 dependencies = [
  "addr2line",
  "cc",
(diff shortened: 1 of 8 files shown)


lynxlynxlynx:
What are these files used for? Does it slow down compiling significantly or why the mention?

selaux:
Initially I wanted to replace the image loading in stracciatella to also support other file formats, which was too much work. Now I'm using it outside of this project sometimes, by referencing this repo. I was mentioning that because it's 1800 lines of currently unused code for this repo and I was not sure whether we want that.

lynxlynxlynx:
We could leave it in a branch. It sounds like it would fit better in the toolset repo, but I guess that requires a bit more work.

ja2-derek:
I am not too concerned about compilation times. I would worry about unused code rotting and getting in the way in the future - say a refactoring.

Maybe can this be put in a gist?

selaux:
To make it more clear on how this is used, I've moved a prototype that I made recently into our organization: [stracciatella-toolset](https://github.com/ja2-stracciatella/stracciatella-toolset). It's an editor for JSON files based on our JSON schema definitions (of course as always unfinished :wink:).

The idea was that you can just reference the stracciatella Rust crate from this (or other) repositories and have all necessary functionality to be able to create/read/modify assets for a specific version of Stracciatella. In the prototype I use the STCI code to render images. I also use the JSON schemas to generate forms as well as the VFS.  For this usecase it would be helpful to have this in master, so I could basically reference a tag (e.g. `v0.19.0`) and have all code present. Otherwise I would need to create a specific branch for each release version. Of course it would also be possible to just have the code in the stracciatella-toolset repo, but since we already have a file formats module it just seemed better to me to put it there.

With Gist the disadvantage would be that I cannot include it as a rust dependency AFAIK. Unit tests are actually already included in each file (they are usually right below the code in rust).

Long story short, I'm fine with either putting this here, or into the prototype repo.

lynxlynxlynx:
With a public user now, I don't have any problems with including this.

PR #1529, opened 2022-04-27
Add 0.19.x to branch filter for appveyor

selaux:
It seems like otherwise the builds for `0.19.1` are not triggered on AppVeyor. I hope this fixes it. We can remove it once `0.20.0` is released.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,6 +25,7 @@ branches:
   only:
     - master
     - nightly
+    - /\d+\.\d+\.[\dx]+/
 
 matrix:
   fast_finish: true


lynxlynxlynx:
Cherry-picked into the branch and retagged.

appveyor.yml
```diff
@@ -25,6 +25,7 @@ branches:
   only:
     - master
     - nightly
+    - 0.19.x
```
lynxlynxlynx, line 28: More generic for future needs:
```suggestion
    - /\d+\.\d+\.[\dx]+/
```

selaux: Applied.

PR #1478, opened 2022-03-23
Fix renaming files on different file systems for all rust versions

selaux:
Closes #1472.

PR #1530, opened 2022-04-30
Handle the EXDEV error on all OS's where it can happen

NorwegianRockCat:
This EXDEV error can happen on all Unix-like OS's that have a rename(2) that
conform to POSIX.1. So, go with the more inclusive "target_family"
instead of target_os.

Makes JA2 work with FreeBSD (and likely all the other OS's that may have their
tmp on another file system).

This is the more general fix to #1478.


selaux:
Thanks! Sometimes its hard to get all platforms right...

Issue #1377, opened 2021-06-02
Gentoo ebuilds impossible if code gets downloaded at compile time. (v.0.18.0 and later)


misanthropos:
Emerging (download, compile and install) software on Gentoo runs in a sandbox with no internet access. Therefore no ebuilds can be made, since compilation will fail.

e.g.
```
[1/9] Creating directories for 'get-lua'

[2/9] Performing download step (download, verify and extract) for 'get-lua'
FAILED: get-lua-prefix/src/get-lua-stamp/get-lua-download 
```
The same is true for dependencies which download software at compile time (like lib-miniaudio).

Package maintainers should be able to set variables for software packages to indicate they are locally installed (like e.g. libsmacker). In that case a maintainer can write a gentoo-ebuild for that package. (like in 0.17.0).

An alternative could be to include all the sources needed in releases.

lynxlynxlynx:
Isn't it all here?
https://github.com/ja2-stracciatella/ja2-stracciatella/tree/master/dependencies

But yeah, it looks like system lua is not supported yet.

misanthropos:
If you look closely, lib-miniaudio is not there. I needs to be downloaded first.

selaux:
[Here](https://github.com/NixOS/nixpkgs/blob/master/pkgs/games/ja2-stracciatella/default.nix) is how I've done it for Nix for 0.17.0, which also runs in a no-network sandbox and requires you to download and verify all files beforehand.

Basically I ended up downloading the required dependencies separately and patching the respective CMake files to point to the separately downloaded file. I've no idea how Gentoo ebuilds work, but I guess something similar should be possible.

misanthropos:
Sorry - I forgot to mention the version I was talking about: Its 0.18.0

The ebuild for 0.17.0 is up and part of gentoos repository. I have problems with the creation of an ebuild for 0.18.0

misanthropos:
Great! Thank you for those changes.
However lib-miniaudio will still be downloaded at compile time.

ja2-derek:
Maybe also same for `magic_enum` now.

timp87:
Hi,
I'm currently maintaining this game port for FreeBSD, so I'm facing with the same problems as the @misanthropos 
The linked issue I did has fixed Lua and sol2.
[...]

PR #1531, opened 2022-04-30
Upgrade miniaudio dependency to latest version in 0.10.x branch

timp87:
In general I'd like to have an ability to pick up miniaudio dependency from the system during ja2 build process.
See also https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1377

And the first step I'd like to be done is to update the needed miniaudio version to latest in 0.10.x branch (0.11.x has broken a lot of APIs).
Besides that provide a comment in the cmake file telling what miniaudio version ja2 needs now.
Also provide small patch to make ja2 compatible with miniaudio 0.10.36+


selaux:
Thanks.

PR #1541, opened 2022-05-15
Externalize more hardcoded values

lynxlynxlynx:
Lifted, adapted and extended the (IMHO) last useful things from #1002.

### DIFF
--- a/rust/stracciatella/src/schemas/yaml/game.schema.yaml
+++ b/rust/stracciatella/src/schemas/yaml/game.schema.yaml
@@ -189,6 +189,13 @@ properties:
       - Progress bars for the mercs stats in tactical mode.
 
       Vanilla setting: `false`
+  skip_sleep_explanation:
+    type: boolean
+    title: Skip sleep explanation
+    description: |
+      Skips the `cannot sleep` popup when driving or when the merc is already rested.
+
+      Vanilla setting: `false`
   extra_attachments:
     type: boolean
     title: Extra attachments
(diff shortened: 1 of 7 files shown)


rust/stracciatella/src/schemas/yaml/game.schema.yaml
```diff
@@ -234,6 +241,41 @@ properties:
       Multiplier for damage that is inflicted on the legs
 
       Vanilla setting: `0.5`
+  aim_bonus_per_std_ap:
+    type: number
```
ja2-derek, line 245: I think we should use `$ref: types/int8.schema.yaml` for these, to guard against overflow - if somebody decides they want a really high value for these.

lynxlynxlynx: That's a design question to me: where do we draw the line? If they want to break the file, the game, they have countless options already.

ja2-derek: Maybe always use the same numeric value as the backing C++ code?

I was looking at `range_bonus_barrel_extender`. The default is `100`. It is not unreasonable to raise it to say `150`, and that already causes overflow.

lynxlynxlynx: Good point, it should be bigger on the engine side already.

ja2-derek:
It looks good to me now.

Now that I think more about it, the next thing I want (in the future in another proposal) is to have it generalized. The aim penalties and bonuses should have been properties of the item, so a mod have more varieties of weapon attachments.

PR #1544, opened 2022-05-16
several more coverity fixes

lynxlynxlynx:
there was a fresh report right after the previous PR was merged, so I picked some more off the list.

The big chunk is just an indentation reduction, easiest to look at with whitespace changes disabled.


selaux:
Thanks for cleaning up! Should we use `AssertMsg(conditon, "message");` for assertions instead of `assert(condition)`? At least that seems more consistent with the way we are doing asserts in other places. Other than that looks good.

selaux:
Allright.

PR #1546, opened 2022-05-20
Fix error messages when commiting read pointer in sound system

selaux:
The newer version of miniaudio we introduced seems to [return MA_AT_END](https://github.com/mackron/miniaudio/commit/9be681b84829d0247ea00dd23db22c95d71b56aa) when the ringbuffer is fully read. This PR fixes the annoying log when a sound has finished:

> 2022-05-20T09:44:25 [ERROR] sgp/SoundMan.cc: Could not commit read pointer for channel 1: At end

Functionally there should be no differences.


lynxlynxlynx:
Cool, didn't notice.

Issue #1548, opened 2022-05-24
Assertion failed FPS drop with display cover key


MrMaggot:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
Framerate drops to single digits when using the display cover key directly south/southwest of the player's squad. The error ```[ERROR] game\Tactical\DisplayCover.cc: Assertion failed in C:\projects\ja2-stracciatella\src\game\Tactical\DisplayCover.cc, line 170``` repeats in the log for as long as the key is held. Bug encountered on Windows 10 and the save can be downloaded at https://drive.google.com/file/d/1UYMNWIP1Ir5DWAauXfx5QMv-1DTA05NY/view?usp=sharing

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  v0.20.0-20220519+12a919a | JA2 Gold  |    don't know    |    don't know     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

MrMaggot:
Tested this bug on Arch instead of Windows 10 and there was no framerate drop, but the assertion errors still occur. The first couple of cover values being checked by the assert on line 141 are within the bounds, but afterwards they go into the negatives and fail the assertion. I also noticed that some of the cover tiles being drawn were different between Windows 10 and Linux as some cover tiles were not being drawn on W10 that are present on Arch.

momoko-h:
DisplayCover.cc:332 was `bPercentCoverForGridno = 100 - __min(iTemp, 100);` but is now `bPercentCoverForGridno = std::min(0, 100 - iTemp);` which is simply not the same thing when iTemp is greater than 100.

MrMaggot:
[...]

PR #1549, opened 2022-05-26
Fix bPercentCoverForGridno values going out of 0-100 bound

MrMaggot:
bPercentCoverForGridno was set up in such a way that the minimum between 0 and 100 - iTemp could be a negative number, causing the assertion errors mentioned in #1548

### DIFF
--- a/src/game/Tactical/DisplayCover.cc
+++ b/src/game/Tactical/DisplayCover.cc
@@ -329,7 +329,7 @@ static INT8 CalcCoverForGridNoBasedOnTeamKnownEnemies(SOLDIERTYPE const* const p
 	{
 		bPercentCoverForGridno = iTotalCoverPoints / bNumEnemies;
 		INT32 const iTemp = bPercentCoverForGridno - (iHighestValue / bNumEnemies) + iHighestValue;
-		bPercentCoverForGridno = std::min(0, 100 - iTemp);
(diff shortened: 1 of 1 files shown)


src/game/Tactical/DisplayCover.cc
```diff
@@ -329,7 +329,7 @@ static INT8 CalcCoverForGridNoBasedOnTeamKnownEnemies(SOLDIERTYPE const* const p
 	{
 		bPercentCoverForGridno = iTotalCoverPoints / bNumEnemies;
 		INT32 const iTemp = bPercentCoverForGridno - (iHighestValue / bNumEnemies) + iHighestValue;
-		bPercentCoverForGridno = std::min(0, 100 - iTemp);
+		bPercentCoverForGridno = std::min(0, std::clamp(100 - iTemp, 0, 100));
```
lynxlynxlynx, line 332: This will always be 0, no?

Either go with a working clamp:
```suggestion
		bPercentCoverForGridno = std::clamp(100 - iTemp, 0, 100);
```

Or like it was before:
```suggestion
		bPercentCoverForGridno = 100 - std::min(iTemp, 100);
```

MrMaggot: Total mixup on the logic there, my bad.

lynxlynxlynx:
Thanks!

lynxlynxlynx:
I've rechecked the original PR for any other similar traps / mistakes, but found none. :)

PR #1551, opened 2022-05-27
Small cleanups

momoko-h:
The unused variable in WordWrap.cc is currently one of two warnings I get while building apart from the deprecated declarations. The other one is in miniaudio.

SelectWin.cc stores 16-bit values in an array of 32-bit integers for no good reason.

LocateMeanWhileGrid is used just once and does nothing but call another function.


lynxlynxlynx:
LGTM, thanks!

PR #1552, opened 2022-05-27
Touch controls

selaux:
This PR adds touch controls (primarily targeted at Android, but they should work on other touchscreen devices as well). Initially I just wanted to cleanup the Input system and add minor features such as supporting additional mouse buttons. But I got carried away...

Highlights:
- Mouse and ButtonSystem now use `std::function` for callbacks
- Screens that want keyboard events only dequeue keyboard events now, not all events
- Callback reasons are now consistently `UINT32`
- Several new callback reasons (`X1` and `X2` mouse buttons, `TFINGER` events, composite `POINTER` events for left mouse button and touch)
- Messages that could previously be dismissed with left AND right mouse button can now be dismissed by every mouse button
- Of course **touch controls**:
  - In principle: Touch = Left Mouse Button, except for the tactical viewport
  - One gesture is currently implemented: Double Tap, which is mapped to left mouse button double click
  - For `MouseRegions` where we have an alternative action though a right click, the alternative action is triggered by a long touch. This is implemented via the `MouseCallbackPrimarySecondary` function in `MouseSystem.h`. Exceptions:
    - Editor (I didn't bother)
    - Quest Debug Screen (I didn't bother)
    - Bobby Rays, where adding/removing items to/from cart works by tapping upper/lower half of the items picture respectively
  - Default cursors and fast help are not displayed on touch, as they are only distracting

Note: Main changes are in `Input.(h/cc)`, `MouseSystem.(h/cc)`, `Cursor_Control.(h/cc)` and wherever `MouseCallbackPrimarySecondary` was used. A lot of the other stuff was search and replace (sorry for the PR size).


src/sgp/Input.cc
```diff
@@ -19,17 +19,19 @@ static BOOLEAN fCursorWasClipped = FALSE;
...
 static UINT32 guiMiddleButtonRepeatTimer;
+static UINT32 guiX1ButtonRepeatTimer;
+static UINT32 guiX2ButtonRepeatTimer;
 
 BOOLEAN gfLeftButtonState;  // TRUE = Pressed, FALSE = Not Pressed
 BOOLEAN gfRightButtonState; // TRUE = Pressed, FALSE = Not Pressed
 BOOLEAN gfMiddleButtonState;// TRUE = Pressed, FALSE = Not Pressed
+BOOLEAN gfX1ButtonState;// TRUE = Pressed, FALSE = Not Pressed
+BOOLEAN gfX2ButtonState;// TRUE = Pressed, FALSE = Not Pressed
+
```
lynxlynxlynx, line 35: There's a lot of duplication the way we handle keys now. Button state could be a joint bitfield (potentially including modifier keys) and the repeat timer could be a single timer.

selaux: I have reduced duplication for this and also for double click handling. Button state is a `std::bitset` now (without modifier keys, as I haven't touched them in this PR yet). Each button still has its own repeat timer.

src/sgp/Input.cc
```diff
@@ -11,34 +11,79 @@
...
-BOOLEAN gfRightButtonState; // TRUE = Pressed, FALSE = Not Pressed
-BOOLEAN gfMiddleButtonState;// TRUE = Pressed, FALSE = Not Pressed
-BOOLEAN gfX1ButtonState;// TRUE = Pressed, FALSE = Not Pressed
-BOOLEAN gfX2ButtonState;// TRUE = Pressed, FALSE = Not Pressed
+static ButtonState<0, MOUSE_BUTTON_LEFT> gLeftButtonState;
+static ButtonState<1, MOUSE_BUTTON_RIGHT> gRightButtonState;
+static ButtonState<2, MOUSE_BUTTON_MIDDLE> gMiddleButtonState;
+static ButtonState<3, MOUSE_BUTTON_X1> gX1ButtonState;
+static ButtonState<4, MOUSE_BUTTON_X2> gX2ButtonState;
+static ButtonState<5, 0, TOUCH_FINGER_DOWN, TOUCH_FINGER_UP, TOUCH_FINGER_REPEAT, TOUCH_REPEAT_TIMEOUT, TOUCH_REPEAT_TIME> gMainFingerState;
```
lynxlynxlynx, line 80: The mouse ones could still be grouped into an array or somesuch, indexed by button type. Then IsMouseButtonDown wouldn't need an extra mapping.
Ideally it would be in one object, so we could use the bitset capabilities, eg. merging `IsMouseButtonDown(MOUSE_BUTTON_LEFT) || (IsMouseButtonDown(MOUSE_BUTTON_RIGHT)` to `IsMouseButtonDown(MOUSE_BUTTON_LEFT | MOUSE_BUTTON_RIGHT)`.

selaux: While I agree that using bitflags for the individual mouse buttons would be nicer from the user perspective, I would not optimize this further right now. I didn't find any occurence of checking multiple mouse buttons in a single condition in our current codebase (that didn't also check for other things).

From my current understanding of the input/mouse system, using `IsMouseButtonDown` indicates code smell anyways (didn't want to setup a mouse region or didn't use the information of the mouse region). Same for usage of `gusMouseXPos`, `gusMouseYPos`. All of this information is propagated through the mouse regions already, it just needs to be used from there:

- `ButtonState` contains mouse button state using bitflags (so there we can already check multiple buttons)
- `MouseXPos`, `MouseYPos` contains absolute mouse coordinates

lynxlynxlynx: That (shortened) example is from WaitForHelpScreenResponse. But yeah, it doesn't have to be done within the same PR, another issue could be opened as a reminder that we have needlessly competing implementations and only one should stay.

selaux: I give you this one :grinning:. THB I excluded searching through the editor files, there is a bunch of weird input handling in there. I have opened https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1557. I will also include other things that I notice along the way...

src/game/Credits.cc
```diff
@@ -195,7 +195,7 @@ ScreenID CreditScreenHandle(void)
 
 static void InitCreditEyeBlinking(void);
 static void RenderCreditScreen(void);
-static void SelectCreditFaceMovementRegionCallBack(MOUSE_REGION* pRegion, INT32 iReason);
+static void SelectCreditFaceMovementRegionCallBack(MOUSE_REGION* pRegion, UINT32 iReason);
```
ja2-derek, line 198: Should the argument be renamed `uiReason` too?

ja2-derek:
The "exit sector" cursor does not appear in tactical view on the east and south (right and bottom) edges. I am using a resolution of 900x510. It looks like `gfScrolledToBottom` and `gfScrolledToRight` are still false when I scroll to the right and to the bottom.

I will report back if I find other issues.

src/game/Credits.cc
selaux: Yes, but I dont know how to easily achieve this without spending a lot of time manually renaming all of the occurences, so I left it.

selaux:
The exit cursor should be fixed.

ja2-derek:
> The exit cursor should be fixed. It was one tile off...

I can confirm it's fixed. Thanks!

ja2-derek:
I am getting a debug assert message at the "enter sector" screen:

```
2022-10-12T19:48:48 [ERROR] sgp/Debug.h: Assertion failed in ~/ja2-stracciatella/src/sgp/MouseSystem.cc, line 685:
Attempting to define a mouse region that already exists.
```

But other than the message, everything seems to work fine.

src/game/Tactical/Interface_Panels.cc
```diff
@@ -3620,9 +3624,9 @@ static bool IsMouseInRegion(MOUSE_REGION const& r)
 static void ConfirmationToDepositMoneyToPlayersAccount(MessageBoxReturnValue);
 
 
-static void SMInvMoneyButtonCallback(MOUSE_REGION* pRegion, INT32 iReason)
+static void SMInvMoneyButtonCallback(MOUSE_REGION* pRegion, UINT32 iReason)
 {
-	if (iReason == MSYS_CALLBACK_REASON_LBUTTON_DWN )
+	if (iReason == MSYS_CALLBACK_REASON_POINTER_DWN )
```
ja2-derek, line 3629: @selaux: this should have been a `&` operator.

momoko-h:
I've seen that assertion as well. IMHO it should be removed, `MSYS_DefineRegion` should not require its callers to provide a zero'd memory block. It initializes all fields anyway.

PR #1554, opened 2022-05-28
Fix #1538 Possible INT8 overflow in bLockDamage

### LINKED ISSUE #1538: Possible INT8 overflow in bLockDamage

### What is wrong
bLockDamage is defined as INT8:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/37b9b08cb55c1b66313402cd2eafd293111651be/src/game/Tactical/Keys.h#L62

No check for INT8 overflow in bLockDamage, resulting in negative values after several attacks on lock.
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/37b9b08cb55c1b66313402cd2eafd293111651be/src/game/Tactical/LOS.cc#L2303
Suggested fix from 1.13:
```
//Madd: catch the overflow
if (sLockDamage + pDoor->bLockDamage > 127)
	pDoor->bLockDamage = 127;
else
	pDoor->bLockDamage+= sLockDamage;
```
Similar issue when blowing up locks with shaped charge
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/37b9b08cb55c1b66313402cd2eafd293111651be/src/game/Tactical/Keys.cc#L649

| |  |    yes    |    no     |

momoko-h:
Small behaviour change, previously you got the 'Lock destroyed' message only when you shot the lock. Now it's displayed when you blow the lock with an explosive, too.

### DIFF
--- a/src/game/Tactical/Keys.cc
+++ b/src/game/Tactical/Keys.cc
@@ -646,12 +646,9 @@ BOOLEAN AttemptToBlowUpLock( SOLDIERTYPE * pSoldier, DOOR * pDoor )
 
 		// Not sure if this makes sense, but the explosive is small.
 		// Double the damage here as we are damaging a lock rather than a person
-		pDoor->bLockDamage += Explosive[GCM->getItem(SHAPED_CHARGE)->getClassIndex()].ubDamage * 2;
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Thanks!

Issue #1559, opened 2022-05-29
Skyrider crashes the game on my current progress


Trokinos:
Using Stracciatella 0.19.1 on window 10

On my current game, i ran into a crash that can be reproduced everytime : 

Here is the save : 
[2022-05-29t12-12-34z-helo-crash.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/8793336/2022-05-29t12-12-34z-helo-crash.zip)

Squad 1 is already in the helicopter.
Double Click on C6 to send Skyrider to that part of San Mona 

![err1](https://user-images.githubusercontent.com/14024168/170868043-a08da960-7d05-47df-b7f3-06fe925bfed7.jpg)

Press time acceleration -> it will always crash (at least it does for me) to desktop

Can't manage to capture the error message in the log window unfortunately as it is closing at the same time as the game crash (and close), i spotted something about "soundloaddisk" in the short interval before the log window closes itself

momoko-h:
This only crashes for me right after a fresh restart of the game when I immediately load the provided save. The destination of the helicopter doesn't matter.

The reason is that `HandleNewSectorAmbience` hasn't been called yet, so `gAmbData` is empty. Usually `LoadSavedGame` takes care of that, but for this save the fWorldLoaded flag is false and we end up in a branch where the call to `HandleNewSectorAmbience` does not happen.

Issue #1537, opened 2022-05-07
Ice Cream truck out of fuel crash?


Bdan720:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
I think there could be a bug with the icecream truck when it runs out of fuel. (I notice the XX warning) When reaching its destination, the game hard crashes. Refueling the van delays this crash. Leaving the truck and travel on foot is stable.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.19.1 | JA2 from GOG  |    yes?    |    ???     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)
To reproduce bug? Move the truck around.

Windows 10 Home.

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

2022-05-07T09:15:22 [WARN] game\Tactical\Tactical_Save.cc: SynchronizeItemTempFile() Reported 0, should be 45
2022-05-07T09:15:24 [ERROR] sgp\SoundMan.cc: Could not commit read pointer for channel 1: At end
2022-05-07T09:15:27 [ERROR] sgp\SGP.cc: Game has been terminated due to an unrecoverable error: Parameter index out of range
Creating an emergency savegame succeeded (Error.sav).
Please report this error with a description of the circumstances. Do not forget to attach the savegame. (class std::runtime_error)
2022-05-07T09:15:29 [ERROR] launcher\Launcher.cc: Command_execute "L:\\Games\\JA2 Stracciatella\\ja2.exe" []: exited with status code -1073740791

Here are some of my save files. Hope this is ok. 
[2022-05-07-d38 save files.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/8644800/2022-05-07-d38.save.files.zip)

momoko-h:
[...]

PR #1560, opened 2022-05-29
Fix crashes #1537 and #1559

momoko-h:
• Ensure ambient sounds are always loaded after loading a game
• Pass all required arguments to st_format_printf

### DIFF
--- a/src/game/Strategic/Strategic_Movement.cc
+++ b/src/game/Strategic/Strategic_Movement.cc
@@ -3109,7 +3109,7 @@ void AddFuelToVehicle(SOLDIERTYPE* pSoldier, SOLDIERTYPE* pVehicle)
 static void ReportVehicleOutOfGas(VEHICLETYPE const& v, const SGPSector& sMap)
 {
 	// Report that the vehicle that just arrived is out of gas
-	ST::string str = st_format_printf(gzLateLocalizedString[STR_LATE_05], pVehicleStrings[v.ubVehicleType], sMap.AsShortString());
+	ST::string str = ST::format(gzLateLocalizedString[STR_LATE_05].c_str(), pVehicleStrings[v.ubVehicleType], sMap.AsShortString());
 	DoScreenIndependantMessageBox(str, MSG_BOX_FLAG_OK, 0);
 }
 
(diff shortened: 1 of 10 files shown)


src/game/Strategic/Strategic_Movement.cc
```diff
@@ -3109,7 +3109,7 @@ void AddFuelToVehicle(SOLDIERTYPE* pSoldier, SOLDIERTYPE* pVehicle)
 static void ReportVehicleOutOfGas(VEHICLETYPE const& v, const SGPSector& sMap)
 {
 	// Report that the vehicle that just arrived is out of gas
-	ST::string str = st_format_printf(gzLateLocalizedString[STR_LATE_05], pVehicleStrings[v.ubVehicleType], sMap.AsShortString());
+	ST::string str = st_format_printf(gzLateLocalizedString[STR_LATE_05], pVehicleStrings[v.ubVehicleType], sMap.y + 'A' - 1, sMap.x);
```
lynxlynxlynx, line 3112: Let's rather fix the format strings, so this can remain simple. Eg. in `src/game/Utils/_EnglishText.cc:3242` and the rest. If you want, you can convert it to ST syntax and then this line can also be changed to use `ST::format`, or just extract the c cstring from the short string to remain in the printf world.

src/game/Utils/_ChineseText.cc
```diff
@@ -3266,7 +3266,7 @@ static const ST::string s_cn_gzLateLocalizedString[gzLateLocalizedString_SIZE] =
 	"%s精力不足, 无法改变姿势.",	//"%s does not have enough energy to change stance.",
 
 	//A message that pops up when a vehicle runs out of gas.
-	"%s汽油耗尽, 在%c%d抛锚了.",	//"The %s has run out of gas and is now stranded in %c%d.",
+	"{}汽油耗尽, 在{}d抛锚了.",	//Originally "The %s has run out of gas and is now stranded in %c%d.",
```
selaux, line 3269: I think the `d` needs to be removed as well, right?

momoko-h: Yes, of course.

selaux:
Works for me. Thanks.

PR #1561, opened 2022-06-02
Cleanup KEY_ON_RING related code

momoko-h:
Cleans up a bit of copy & paste coding, removes the function `RemoveKeyFromSlot` and uses the generic `RemoveKeysFromSlot` instead. 

Adds a simple method to make the key ring slot checks easier to read.

### DIFF
--- a/src/game/Tactical/Interface_Items.cc
+++ b/src/game/Tactical/Interface_Items.cc
@@ -2862,30 +2862,20 @@ void BeginItemPointer( SOLDIERTYPE *pSoldier, UINT8 ubHandPos )
 
 void BeginKeyRingItemPointer( SOLDIERTYPE *pSoldier, UINT8 ubKeyRingPosition )
 {
-	BOOLEAN fOk;
-
 	// If not null return
 	if ( gpItemPointer != NULL )
 	{
 		return;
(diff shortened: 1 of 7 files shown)


src/game/Tactical/Interface_Items.cc
```diff
@@ -2877,7 +2877,7 @@ void BeginKeyRingItemPointer( SOLDIERTYPE *pSoldier, UINT8 ubKeyRingPosition )
 	}
 	else
 	{
-		RemoveKeyFromSlot( pSoldier, ubKeyRingPosition, &gItemPointer );
+		RemoveKeysFromSlot( pSoldier, ubKeyRingPosition, 1, &gItemPointer );
 		fOk = (gItemPointer.ubNumberOfObjects == 1);
 	}
```
lynxlynxlynx, line 2882: Can't this else block be merged with the previous one now? From what I can tell there is only the difference in the count of items.

momoko-h: Sure. I'll also remove the nonsensical comment about a bloodied knife while I am here.

lynxlynxlynx:
Thanks!

Issue #1136, opened 2020-07-04
Game gets choppy in tactical real-time mode


ja2-derek:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
Details, details, details.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  master | Steam version |    no    |    not sure     |

I notice the game gets choppy, and scrolls and UI responses are noticably delayed in certain situations. The delay starts in the middle of a fight, and when multiple enemies are on high alert. But it does not always happen.

I am running master build on Win 10. I am not sure if this issue is specific to me, as I haven't seen any other reports about performance. I remember seeing this is also in 0.16.1 in late game sectors, but I am not able to reproduce this issue from scratch - e.g. starting fresh game, teleporting to P3 does not reproduce it.

I did a profiling during the stuttering, and the results points to excessive debug logging. Game was not in debug mode during profiling.

Function breakdown by CPU time used:
![Perf01](https://user-images.githubusercontent.com/63151803/86505746-29908700-bdfb-11ea-87e1-9e18512b7f66.png)

Call stack of LogMessage:
![PERF02](https://user-images.githubusercontent.com/63151803/86505801-b89d9f00-bdfb-11ea-81b0-a6cc3f17b9d2.png)

As seen above, the problem is the debug logging the `RadioSighting` method. It takes about a third of the total CPU time spent. The choppiness seems to be gone when I commented out all those debug logs.

While I am not 100% certain about the root cause (as it is not easily reproducible), there should be no harm reducing debug logs. I suggest that we:

- [ ] Reduce or remove debug logging in the inner loops of `RadioSighting`
- [ ] ~Update `SLOGD`, to not do anything unless debug log is enabled~
- [ ] Improve efficiency of Logger_log in lib-stracc (#1137)

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[...]

PR #1564, opened 2022-06-03
Remove Logger.cc

momoko-h:
One of the two overloads in this file is not used any more since the removal
of the old logging functions and the other one just passes its arguments on
to Logger_log.

Btw, is it really not possible to get the current log level from C++? I tried profiling why an opponent's turn takes so long and a significant amount of time is spent just formatting all those debug messages even when they are not even printed, so it would be great if `LogMessageST` could check the log level before it calls ST::format and Logger_log.


selaux:
Thanks. Getting the log level from rust is currently not implemented. Here is a commit that does it: https://github.com/ja2-stracciatella/ja2-stracciatella/commit/39528ce39f9b81e98570d2bcde3cce7cf567bf60. Do you want to include it in this PR?

momoko-h:
Yes, I think these two issues are closely related, so we might as well combine them.

selaux:
Is the performance difference significant now? Would you mind sharing a flamegraph or something?

rust/stracciatella/src/logger.rs
```diff
@@ -165,7 +181,12 @@ impl Logger {
 
     /// Sets the global log level to a specific value
     pub fn set_level(level: LogLevel) {
-        GLOBAL_LOG_LEVEL.store(level.into(), Ordering::Relaxed);
+        RuntimeLevelFilter::set_global_log_level(level.into())
+    }
+
+    /// Sets the global log level to a specific value
```
selaux, line 187: I know this came from me, but there seems to be a copy-paste-error in the comment...

rust/stracciatella_c_api/src/c/logger.rs
```diff
@@ -28,6 +28,12 @@ pub extern "C" fn Logger_setLevel(level: LogLevel) {
     Logger::set_level(level)
 }
 
+/// Set log level
```
selaux, line 31: I know this came from me, but there seems to be a copy-paste-error in the comment

ja2-derek:
> Is the performance difference significant now? Would you mind sharing a flamegraph or something?

I also want to know how things have changed since 0.17.. The last time we did this was:  https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1136

momoko-h:
This is what current master looks like:
KCG_Master

`CalcChanceToGetThrough` spends more than half of its time logging.

This is the same turn, as much as the RNG allows, with this patch applied:
KCG_RemoveLogger

Even though the number of LogFormatST is even higher, `CalcChanceToGetThrough` doesn't even make it to the top of the list.

ja2-derek:
The call graph indeed looks much better. Thanks!

selaux:
Looks great. Thanks a lot!

PR #1565, opened 2022-06-05
New implementation of atan8

momoko-h:
The current one is not that great, to put it mildly. It looks like the author couldn't quite figure out the math (that is the commented out version), so he chose brute force and used eight `if` statements.

Three of the eight if blocks are missing the `break` statement so we are doing unnecessary comparisons after we already have the correct result.
x and y are swapped in the atan2 call so the ranges look weird, north is where east should be, etc.
Finally, all these instances of 0.623 should be 0.625.

Due to the last bug, this implementation does not always return the same result. I wrote a short test program to compare the results of the two functions and the output is exactly what I expected:

> 201850880 comparisons, 392136 different.

So in 0.2% of all cases the result differs. I doubt this has any negative influence on the users of this function, it's more likely that getting the correct result benefits them.

### DIFF
--- a/src/game/Tactical/Soldier_Control.cc
+++ b/src/game/Tactical/Soldier_Control.cc
@@ -9031,4 +9031,107 @@ TEST(SoldierControl, asserts)
 	EXPECT_EQ(sizeof(KEY_ON_RING), 2u);
 }
 
+UINT8 oldatan8( INT16 sXPos, INT16 sYPos, INT16 sXPos2, INT16 sYPos2 )
+{
+	DOUBLE  test_x =  sXPos2 - sXPos;
+	DOUBLE  test_y =  sYPos2 - sYPos;
+	UINT8	  mFacing = WEST;
+	DOUBLE angle;
+
+	if ( test_x == 0 )
+	{
+		test_x = 0.04;
+	}
+
+	angle = atan2( test_x, test_y );
+
+	do
+	{
+		if ( angle >=-PI*.375 && angle <= -PI*.125 )
+		{
+			mFacing = SOUTHWEST;
+			break;
+		}
+
+		if ( angle <= PI*.375 && angle >= PI*.125 )
+		{
+			mFacing = SOUTHEAST;
+			break;
+		}
+
+		if ( angle >=PI*.623 && angle <= PI*.875 )
+		{
+			mFacing = NORTHEAST;
+			break;
+		}
+
+		if ( angle <=-PI*.623 && angle >= -PI*.875 )
+		{
+			mFacing = NORTHWEST;
+			break;
+		}
+
+		if ( angle >-PI*0.125 && angle < PI*0.125 )
+		{
+			mFacing = SOUTH;
+		}
+		if ( angle > PI*0.375 && angle < PI*0.623 )
+		{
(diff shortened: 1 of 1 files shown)


src/game/Tactical/Soldier_Control.cc
```diff
@@ -6364,90 +6364,29 @@ INT16 GetDirectionToGridNoFromGridNo( INT16 sGridNoDest, INT16 sGridNoSrc )
...
+	// return the same result for this case.
+	if (x == 0.0 && y == 0.0) return EAST;
 
-		if ( angle <= PI*.375 && angle >= PI*.125 )
-		{
-			mFacing = SOUTHEAST;
-			break;
-		}
+	double theta = std::atan2(y, x);  // theta now ∈ [-π, +π] = [-180°, 180°]
+	if (theta < 0) theta += 2 * M_PI; // theta now ∈ [0, +2π] = [0°, 360°]
```
lynxlynxlynx, line 6381: Isn't theta now in [π, 3π]? It still has the same range of values, just shifted, which I guess explains the extra M_PI below.

src/game/Tactical/Soldier_Control.cc
```diff
@@ -6364,90 +6364,29 @@ INT16 GetDirectionToGridNoFromGridNo( INT16 sGridNoDest, INT16 sGridNoSrc )
...
-		{
-			mFacing = NORTH;
-		}
-		if ( angle > -PI*0.623 && angle < -PI*0.375 )
-		{
-			mFacing = WEST;
-		}
-
-	} while( FALSE );
+	const int directionIndex = static_cast<int>((8 * theta + M_PI) / (2 * M_PI)) % 8;
```
lynxlynxlynx, line 6383: Just curious: how did you come up with this? My immediate thought would be to divide by π/4 and round, which should be slower.

lynxlynxlynx:
Hmm, would it make sense to include some key parameter pairs in our test suite?

src/game/Tactical/Soldier_Control.cc
momoko-h: No, if theta is negative it's in the range -π, 0. Add 2 * π to that and you arrive at π, 2π. Basically this line adds the part that was negative on top of the positive part.

lynxlynxlynx: My comment was aimed at the comment, but I see now that it's referring to all cases, not just the negative one.

src/game/Tactical/Soldier_Control.cc
momoko-h: Just a bit of math: to get the octant you could add 22,5° to the atan2 result like this:
`theta += M_PI / 8;`

Then the formula for directionIndex would look like this:
`8 * theta / (2 * M_PI)`

Substitute theta + M_PI /8 for theta in that equation:
`8 * (theta + M_PI / 8) / (2 * M_PI)`

And `8 * (theta + M_PI / 8) = 8 * theta + M_PI`

momoko-h:
Maybe, but as I found out while writing this code incorrect results become quite obvious when you just play the game: everyone looks, runs and shoots in the wrong direction.

PR #1566, opened 2022-06-06
Fix FindNumTurnsBetweenDirs

momoko-h:
This function is an abomination. It has one really simple job but manages to make a complete mess out of it: its far too complicated algorithm produces the wrong result for 56 out of the 64 possible combinations of its arguments.

It even has a safeguard against its own brokenness which is itself really silly: why would you abort the algorithm at 101 when you know perfectly well that the greatest possible result is 4?

Luckily, the three users of this function do not care that much, as long as `FindNumTurnsBetweenDirs` is consistently wrong they're happy.

### DIFF
--- a/src/game/TileEngine/Isometric_Utils.h
+++ b/src/game/TileEngine/Isometric_Utils.h
@@ -82,10 +82,7 @@ INT16 CardinalSpacesAway(INT16 sOrigin, INT16 sDest);
 bool FindHigherLevel(SOLDIERTYPE const*, UINT8* out_direction = 0);
 bool FindLowerLevel(SOLDIERTYPE const*, UINT8* out_direction = 0);
 
-template<UINT8 maxDistance = 4>
-INT8 QuickestDirection(UINT8 origin, UINT8 dest);
-INT8 ExtQuickestDirection(UINT8 origin, UINT8 dest);
-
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
LGTM and looks like ExtQuickestDirection and QuickestDirection could also be simplified and merged.

src/game/TileEngine/Isometric_Utils.h
```diff
@@ -82,8 +82,9 @@ INT16 CardinalSpacesAway(INT16 sOrigin, INT16 sDest);
 bool FindHigherLevel(SOLDIERTYPE const*, UINT8* out_direction = 0);
 bool FindLowerLevel(SOLDIERTYPE const*, UINT8* out_direction = 0);
 
-INT16 QuickestDirection(INT16 origin, INT16 dest);
-INT16 ExtQuickestDirection(INT16 origin, INT16 dest);
+template<UINT8 maxDistance = 4>
+INT8 QuickestDirection(UINT8 origin, UINT8 dest);
+INT8 ExtQuickestDirection(UINT8 origin, UINT8 dest);
```
lynxlynxlynx, line 87: ExtQuickestDirection has only one user, so I suggest we add a defaulted maxDistance parameter to QuickestDirection instead of templating it and remove ExtQuickestDirection altogether.

lynxlynxlynx:
Great, thanks!

PR #1571, opened 2022-06-10
Overhaul Animation_Cache

momoko-h:
• Do not dynamically allocated 6 byte arrays and then use an 8 byte pointer
  to store their addresses.
• Convert the free functions to class methods.
• Don't let functions other than init pass in the soldier's profile id
  because the AnimCache and the profile id must come from the same soldier.
• The cache size was not helpful


ja2-derek:
This looks fine to me. Thanks!

PR #1575, opened 2022-06-17
Externalize item names and descriptions

selaux:
This PR moves item names and description into the `ItemModel` and makes them overridable via JSON. No need to edit `itemdesc.edt` anymore to adapt names and descriptions.

Logic:
- Read `itemdesc.edt` when its present and extract the names / descriptions, if available
- Read properties on the items JSON object and overwrite vanilla name if a property is available

Advantages:
- Robustness: Game can start without `itemdesc.edt` even being present, it will just show empty names and descriptions
- Robustness: Game can read longer `itemdesc.edt` files, when there are more items than in vanilla (there are other limitations still forcing item number to be less or equal to vanilla though)
- Code Cleanup: No more accessing global `ItemNames` and `ShortItemNames` variables to print the names, just use `item->getName()` / `item->getShortName()`

Disadvantages:
- We now additionally store the description in memory, so we need more memory
- Currently its not possible to have translatable names / descriptions when overriding via JSON. This could be fixed in the future by also allowing to reference string resources in JSON


lynxlynxlynx:
LGTM and it gave me a small idea. :)

PR #1576, opened 2022-06-18
Remove MemMan.h and MemMan.cc

### LINKED ISSUE #1574: Does MemMan still have a purpose?
In [its original version](https://github.com/dariusk/ja2/blob/master/Standard%20Gaming%20Platform/MemMan.c) the memory manager was a fairly elaborate set of functions designed to help the developers detect memory leaks.

Today the situation is very different: much of our memory is managed by C++ and the standard library and for the rest we have tools like valgrind. The current code reflects this: all that is left is a rather sad collection of unused functions and macros, initialization and shutdown functions that don't really do anything and one function template that just does `delete[] ptr; ptr=NULL;`

Strangest of all, however, is the replacement `operator new` where I see absolutely no advantage over the default version.

For a quick test I simply replaced both files with empty ones and it took just very few changes to make the build succeed. If you are interested, these changes are [here](https://github.com/ja2-stracciatella/ja2-stracciatella/compare/master...momoko-h:NoMM).

To put it bluntly, I think MemMan doesn't add anything useful, it just increases our build times.

momoko-h:
Resolves #1574


lynxlynxlynx:
Thanks!

PR #1579, opened 2022-06-21
Fix Structure not defined warnings

momoko-h:
• Due to some fishy calculations in BulletHitStructure(), StructureHit()
occasionally ended up with the wrong GridNo.
• Two places used 0 instead of INVALID_STRUCTURE_ID

 I checked the other code locations that use BULLET's qCurrX, Y, Z but could not find any other place that also add 0.5 to these values.

I can only assume that this is not intentional but an actual bug, but it would be good to get more testing, that's why I changed the warning to an assert: it makes it less easy to ignore any remaining issues, if there are any.

### DIFF
--- a/src/game/Tactical/LOS.cc
+++ b/src/game/Tactical/LOS.cc
@@ -2183,18 +2183,6 @@ static BOOLEAN BulletHitMerc(BULLET* pBullet, STRUCTURE* pStructure, BOOLEAN fIn
 }
 
 
-static void BulletHitStructure(BULLET* pBullet, UINT16 usStructureID, INT32 iImpact, BOOLEAN fStopped)
-{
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Looks tricky indeed! I guess shooting at buildings from various positions and angles is the main way to test?

momoko-h:
Yes, especially trying to shoot at different heights, through roofs, at the ground, at walls, basically everything that has its own structure type in [Structure_Internals](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/52f4c0456903417339d882cff45479978b816936/src/game/TileEngine/Structure_Internals.h#L69).

lynxlynxlynx:
I tested a bunch of destruction and hit no assertions. :)

PR #1582, opened 2022-06-24
Fix fast help active checks

momoko-h:
• Several locations in Map_Screen_Interface_Bottom.cc were testing the wrong uiFlags.
• MSYS_RemoveRegion was testing a different flag than four other locations before
  calling FreeBackgroundRectPending.

The first bug (testing `GUI_BUTTON`'s uiFlags instead of `MOUSE_REGION`'s) seems harmless, we either redraw too often or miss one redraw, I didn't notice any display errors.

The second bug, however is more serious: there are five places in `MouseSystem.cc` that call `FreeBackgroundRectPending`, four of them tested the `MSYS_GOT_BACKGROUND` flag before and cleared this flag afterwards.

Only MSYS_RemoveRegion used `MSYS_HAS_BACKRECT` instead, which could theoretically lead to a second `FreeBackgroundRectPending` call for this `BACKGROUND_SAVE`. Worst case scenario would be that this structure was already allocated somewhere else between these two calls and we'd delete someone else's `BACKGROUND_SAVE`. I can't say whether this can actually happen in practice.

AFAICT we only have this flags because `FastHelpRect` was an INT32 in the original code base and you couldn't simple compare it to 0. Now that it is a pointer these flags are not necessary any more and, as seen here, just complicate things.

I hope this does not clash with #1552, but I think I'm touching different lines.

### DIFF
--- a/src/game/Strategic/MapScreen.cc
+++ b/src/game/Strategic/MapScreen.cc
@@ -4386,20 +4386,7 @@ static void ContractButtonCallback(GUI_BUTTON* btn, INT32 reason)
 {
 	if (g_dialogue_box) return;
 
-	if (reason & MSYS_CALLBACK_REASON_LBUTTON_DWN)
-	{
-#if 0 // XXX was commented out
-		if (bSelectedDestChar != -1 || fPlotForHelicopter)
-		{
-			AbortMovementPlottingMode();
-			return;
-		}
-#endif
-
-		// redraw region
-		if (btn->HasFastHelp()) fCharacterInfoPanelDirty = TRUE;
-	}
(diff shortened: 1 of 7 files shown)


momoko-h:
I had a second look and it turns out this is what this _JA2_RENDER_DIRTY code is supposed to do: free mouse region clients from restoring their contents themselves when they use fast help. All these checks can probably go entirely, I need to test this a bit more.

momoko-h:
Yes, looks like these checks and redrawing the entire bottom of the map screen weren't necessary the entire time.

lynxlynxlynx:
LGTM, thanks!

PR #1583, opened 2022-06-25
Do no update sight values in DecaySmokeEffects and DecayLightEffects

momoko-h:
Let the callers of these functions call AllTeamsLookForAll themselves when
they need to.

### DIFF
--- a/src/game/TileEngine/LightEffects.cc
+++ b/src/game/TileEngine/LightEffects.cc
@@ -121,7 +121,7 @@ LIGHTEFFECT* NewLightEffect(const INT16 sGridNo, const INT8 bType)
 }
 
 
-void DecayLightEffects( UINT32 uiTime )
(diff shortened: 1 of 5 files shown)


lynxlynxlynx:
I think it would be more robust and DRY if you added a defaulted optional param and called with it set to false from TrashWorld, the only other real user.

momoko-h:
Here is the alternative version, in case you prefer this way.

lynxlynxlynx:
I don't feel strongly about it, but using an argument would also make the extra work explicit. It's a pretty much finished product though, so it's hard to gauge whether we'll need to use these functions in any new places later, increasing the duplication. @selaux @ja2-derek any preferences?

selaux:
I like the explicit version slighly better, but no strong feelings as well.

PR #1587, opened 2022-06-30
merge bugfixes into 0.19.x

guijan:
Could you guys merge these bugfixes into 0.19.x for a future point release?
0.19.x crashes on OpenBSD a lot without them, but now that I've been playing 0.19.x with these cherry-picked commits it seems to be good.


lynxlynxlynx:
By now we've accrued enough work to just release 0.20 earlier. Is the package pulling sources from git or via the official tarballs?

lynxlynxlynx:
In that case we could also just merge this, tag and ignore it, not bother with a proper release. You'd have something to target in the update and we'd have more time. :)

selaux:
I think it's fine to merge. If we ever want to release a `0.19.2` those bugfixes will be good to have.

PR #1590, opened 2022-07-01
Fix incorrect schemas for weapons and item replacements

selaux:
- Weapons ids are item ids and should be uint16 (although they can only go from 0 - 70)
- Some weapon sound properties were missing their type
- Item ids should be uint16 for tactical item replacements

### DIFF
--- a/rust/stracciatella/src/schemas/yaml/tactical-map-item-replacements.schema.yaml
+++ b/rust/stracciatella/src/schemas/yaml/tactical-map-item-replacements.schema.yaml
@@ -9,11 +9,11 @@ items:
     replace items with something else, or ignore totally. This does not affect
     items created from other means.
 
-    This is intended for compatibility and for maps that was originally built 
+    This is intended for compatibility and for maps that were originally built
     for different item sets (e.g. Wildfire maps).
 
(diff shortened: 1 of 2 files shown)


rust/stracciatella/src/schemas/yaml/tactical-map-item-replacements.schema.yaml
```diff
@@ -9,25 +9,25 @@ items:
     replace items with something else, or ignore totally. This does not affect
     items created from other means.
 
-    This is intended for compatibility and for maps that was originally built 
+    This is intended for compatibility and for maps that was originally built
```
lynxlynxlynx, line 12:
```suggestion
    This is intended for compatibility and for maps that were originally built
```

lynxlynxlynx:
Looks fine otherwise, the windows build failed due to a network error.

rust/stracciatella/src/schemas/yaml/tactical-map-item-replacements.schema.yaml
selaux: Fixed.

Issue #1589, opened 2022-06-30
Another crash while trying to go to Drassen Airport


guijan:
### What is wrong
The game crashes with SIGABRT

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  58b880d261ee36013025e63e2ee1392820d3aa56 | JA2 1.12 GoG  |    don't know    |    don't know     |

- how to reproduce the bug and, if possible, a save game

[Error.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/9023042/Error.sav.zip)
Load my error save, it crashes immediately.

- the operating system

OpenBSD 7.1 amd64

- more information

I get this error message when the game crashes:
```Text
libc++abi: terminating with uncaught exception of type ST::unicode_error: Invalid UTF-8 sequence byte
```
Here's a backtrace:
```gdb
(gdb) bt
#0  thrkill () at /tmp/-:3
#1  0x000000eebb2176ee in _libc_abort () at /usr/src/lib/libc/stdlib/abort.c:51
#2  0x000000ee06287cea in abort_message (format=<optimized out>) at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/abort_message.cpp:78
#3  0x000000ee0626477b in demangling_terminate_handler () at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_default_handlers.cpp:66
#4  0x000000ee062686ef in std::__terminate (func=0x0) at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_handlers.cpp:59
#5  0x000000ee06282e5c in __cxxabiv1::failed_throw (exception_header=0xee1f4b9f00) at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_exception.cpp:152
#6  0x000000ee06282dbd in __cxa_throw (thrown_object=0xee1f4b9f80, tinfo=0xebdb65f940 <typeinfo for ST::unicode_error>, dest=0xebda5f9bb0 <ST::unicode_error::~unicode_error()>)
    at /usr/src/gnu/lib/libcxxabi/../../../gnu/llvm/libcxxabi/src/cxa_exception.cpp:283
[...]

PR #1596, opened 2022-07-10
Fix indexing into m_sectorLandTypes

### LINKED ISSUE #1588: RUNTIME ERROR in Drassen Airport
### What is wrong
The game gives me a "RUNTIME ERROR" message.

|  58b880d261ee36013025e63e2ee1392820d3aa56 | JA2 1.12 GoG  |   don't know    |    Yes, works just fine with 0.19.x    |

- how to reproduce the bug

[2022-06-30t17-58-12z-j.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/9022678/2022-06-30t17-58-12z-j.sav.zip)

Load my save, take Squad 1 to tile B13 (Drassen Airport).

- the operating system

OpenBSD 7.1 amd64

- extra information

2022-06-30-185018_640x480_scrot

I'm a bit more clueless about this one. I set up a breakpoint on ErrorScreenHandle and this is the backtrace but that's the backtrace of ErrorScreenHandle() and not of what led to calling it, again I can't help much because I don't know C++.
```gdb
Thread 1 hit Breakpoint 1, 0x00000d1727ed6cb0 in ErrorScreenHandle() ()
(gdb) bt
#0  0x00000d1727ed6cb0 in ErrorScreenHandle() ()
#1  0x00000d1727ecddcf in GameLoop() ()
#2  0x00000d1727f3e96f in main ()
```

I see the current master can create an error save, but it didn't create one for this.

selaux:
Should fix #1588 and #1589.

There might be some underlying issues with `SGPSector`. The `std::map` container we use expects the elements to be [Compare](https://en.cppreference.com/w/cpp/named_req/Compare). I'm not sure which of the listed constraints we currently violate with `SGPSector`, but to me it seems like we violate something, otherwise the error would probably not occur.

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -1510,11 +1510,12 @@ const MovementCostsModel* DefaultContentManager::getMovementCosts() const
 int16_t DefaultContentManager::getSectorLandType(uint8_t const sectorID, uint8_t const sectorLevel) const
 {
 	SGPSector key = SGPSector::FromSectorID(sectorID, sectorLevel);
-	if (m_sectorLandTypes.find(key) == m_sectorLandTypes.end())
+	auto result = m_sectorLandTypes.find(key);
+	if (result == m_sectorLandTypes.end())
 	{
 		return -1;
 	}
-	return m_sectorLandTypes.at(key);
+	return result->second;
 }
 
 const CacheSectorsModel* DefaultContentManager::getCacheSectors() const


momoko-h:
This looks like a compiler or library bug to me. I don't see a legitimate reason why `find()` could return a valid iterator but a subsequent `at()` with the same key throws an exception.

Which compiler version are you using? I don't get this crash with GCC 12.1.1

SGPSector implements `operator<`, that should be sufficient to satisfy Compare.

In any case I like your patch better than the current code simply because it avoids the unnecessary second search of the map.

ja2-derek:
How about some unit tests to confirm and check the behaviours? I don't understand how it would fix the issue either.

lynxlynxlynx:
The only way I could see this fail is if there was a data race, but we don't multithread this anyway.

I don't see anything useful in the concepts or metaprogramming headers, so it looks like we can't use an out-of-the-box assert.

selaux:
I did some testing and found some cases where the `comp(a, b)` requirements are not fulfilled for `SGPSector`. Both of those cases should succeed in my opinion:

```cxx
TEST(sectorTest, test1)
{
	auto sector1 = SGPSector::FromShortString("B2");
	auto sector2 = SGPSector::FromShortString("A3");

	EXPECT_TRUE(sector1 < sector2);
	EXPECT_FALSE(sector2 < sector1);
}

TEST(sectorTest, test2)
{
	auto sector1 = SGPSector::FromShortString("F4");
	auto sector2 = SGPSector::FromShortString("B12");
	auto sector3 = SGPSector::FromShortString("C2");

	EXPECT_TRUE(sector1 < sector2);
	EXPECT_TRUE(sector2 < sector3);
	EXPECT_TRUE(sector1 < sector3);
}
```

So my suggestion would be to change to `std::unordered_map`, so we don't need to fulfill Compare for `SGPSector`.

selaux:
I agree. Basically changing `||` into `&&` in  the comparison function would be an option as well, we just need to check whether we use that function in any other place, so we don't break anything. I am fine with either way.

momoko-h:
Just changing || to && would fail if we'd ever have to compare SGPSectors where Z matters, too. I would suggest something like
`	return x + 32 * y + 1024 * z < rhs.x + 32 * rhs.y + 1024 * rhs.z;`

lynxlynxlynx:
Just using && would fail too often. I remember adding the operator for one specific case initially, so I have to look up what that code expected. If we linearize, we'll consider more sectors as matching. I doubt Z matters though.

momoko-h:
Maybe you remember TownModel::getBaseSector()? I just tested compiling with `bool operator<(const SGPSector&) = delete;`, this breaks only the following files:

- externalized/strategic/SectorLandTypes.cc.o
- externalized/strategic/TownModel.cc.o
- externalized/DefaultContentManager.cc.o

selaux:
So I think I have adapted the comparison function now in a way where it will order all sectors regardless of the z value. Rationale behind this would be that for me the comparison operator should be generic, not tailored to a specific usecase. As `TownModel` only uses surface sectors, the comparison there should still work.

PR #1602, opened 2022-07-13
[0.19.x] update miniaudio to 0.10.43

guijan:
Somewhere between versions 0.10.20 and 0.10.43, miniaudio changed the
type of the ma_decoder_seek_proc function type's 2nd parameter from int
to ma_int64. That's an ABI breakage warranting a major version bump, but
the project only bumped the patch version.

The latest Stracciatella master already runs on the latest miniaudio
version which is 0.11.9, but Stracciatella 0.19.x needs to update to
0.10.43 because package maintainers will try to package the latest
0.10.x miniaudio to satisfy the dependency and that causes a compile
error.


guijan:
Sorry for the force-push, I didn't notice the indentation was spaces and not tabs.

PR #1603, opened 2022-07-14
Do not use the c_str() result of an already destroyed ST::string

momoko-h:
This is currently preventing the game from properly reporting errors (see for example 1589), instead of

> Game has been terminated due to an unrecoverable error: ...

we're getting 

> terminate called after throwing an instance of 'ST::unicode_error'
  what():  Invalid UTF-8 sequence byte


lynxlynxlynx:
Nice catch, thanks!

Issue #1584, opened 2022-06-26
Crash while defending Drassen Airport


guijan:
| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  40577cfdc3679eee48a4028712da0830d75b036d | Latest from GoG (1.12 apparently) |   don't know    |    don't know     |

- how to reproduce the bug

Load my save, enter tactical view, and end the turn without taking any actions.
[2022-06-25t19-32-15z-j.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/8985936/2022-06-25t19-32-15z-j.sav.zip)

- the operating system

OpenBSD 7.1 amd64

- backtrace

```gdb
Thread 1 received signal SIGSEGV, Segmentation fault.
0x00000c6c7b0ffa9c in GetTerrainType (grid_no=25601) at ../src/game/TileEngine/WorldMan.cc:1716
1716            return gpWorldLevelData[grid_no].ubTerrainID;
(gdb) bt
#0  0x00000c6c7b0ffa9c in GetTerrainType (grid_no=25601)
    at ../src/game/TileEngine/WorldMan.cc:1716
#1  0x00000c6c7afbc767 in PtsToMoveDirection (pSoldier=0xc6c7bb82530 <Menptr+15216>,
    bDirection=5 '\005') at ../src/game/Tactical/Points.cc:1244
#2  0x00000c6c7b021e68 in EndTurn (ubNextTeam=1 '\001')
    at ../src/game/Tactical/TeamTurns.cc:217
#3  0x00000c6c7af04be2 in UIHandleEndTurn (pUIEvent=0x0)
    at ../src/game/Tactical/Handle_UI.cc:1046
#4  0x00000c6c7b24f429 in MainGameScreenHandle () at ../src/game/GameScreen.cc:354
#5  0x00000c6c7b24937c in GameLoop () at ../src/game/GameLoop.cc:205
#6  0x00000c6c7b304981 in MainLoop (msPerGameCycle=13) at ../src/sgp/SGP.cc:262
#7  0x00000c6c7b3045f1 in main (argc=1, argv=0x7f7ffffd7ec8) at ../src/sgp/SGP.cc:527
```

The backtrace above is from a build made using this command line:
```console
[...]

Issue #1581, opened 2022-06-22
Crash in Omerta when Fatima talks to Dmitri


guijan:
### What is wrong
The game crashes.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  40577cfdc3679eee48a4028712da0830d75b036d | GoG version downloaded today |    don't know    |    don't know    |

- how to reproduce the bug

Load my save, wait and let Fatima talk to Dmitri and bring the player down the basement. You can click through the dialogue or not press anything, it doesn't seem to make a difference. The game crashes as soon as the loading screen shows up.
[2022-06-22t19-43-58z-j.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/8961052/2022-06-22t19-43-58z-j.sav.zip)

- the operating system

OpenBSD 7.1 amd64

- more information

This is the command I used to compile the game:
```console
$ cmake -DCMAKE_TOOLCHAIN_FILE=cmake/toolchain-openbsd.cmake -DCMAKE_BUILD_TYPE=Debug -B build -G Ninja && cmake --build build -j4
```
Compiler flags appear to make a difference, so if you can't reproduce, try making a release build or specifying different combinations of optimization flags like `-O3`, `-O2`, `-flto`, and `-march=native`.

I can't help because I don't know C++ but here's a backtrace:
```gdb
(gdb) bt
#0  0x000007f0b35238d5 in DistanceVisible (pSoldier=0x7f0b41189c0 <Menptr>,
    bFacingDir=9 '\t', bSubjectDir=0 '\000', sSubjectGridNo=16028, bLevel=0 '\000')
    at ../src/game/Tactical/OppList.cc:1022
#1  0x000007f0b352c2c7 in ManLooksForMan (pSoldier=0x7f0b41189c0 <Menptr>,
    pOpponent=0x7f0b41606e0 <Menptr+294176>, ubCaller=1 '\001')
    at ../src/game/Tactical/OppList.cc:1460
#2  0x000007f0b3522666 in OtherTeamsLookForMan (pOpponent=0x7f0b41606e0 <Menptr+294176>)
    at ../src/game/Tactical/OppList.cc:2010
#3  0x000007f0b3521db2 in HandleSight (s=..., sight_flags=SIGHT_LOOK)
[...]

PR #1604, opened 2022-07-14
Enable bounds checking of world level data in debug builds

momoko-h:
I think it is unfortunate that one apparently has to use OpenBSD to find bugs like #1581 and #1584 while the rest of us just gets to continue with a possibly corrupted game state. That could cause mysterious and almost impossible to diagnose problems much later.
With this path, if there are any more instances of `gpWorldLevelData[NOWHERE]` or similar bugs hiding somewhere, everyone running a debug build would now get an exception reported immediately.

### DIFF
--- a/src/game/TileEngine/WorldDef.cc
+++ b/src/game/TileEngine/WorldDef.cc
@@ -97,7 +97,12 @@ void SetAllNewTileSurfacesLoaded( BOOLEAN fNew )
 
 
 // Global Variables
+#ifdef NDEBUG
 MAP_ELEMENT			*gpWorldLevelData;
+#else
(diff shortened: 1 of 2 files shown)


momoko-h:
I tried building with -fsanitize=address but the problem is that an invalid memory read in the Z blitters makes the game crash almost immediately after entering the tactical screen. As long as no one can figure out the cause of that bug I can't really set this compiler option.

The coverity page is [here](https://scan.coverity.com/projects/ja2-stracciatella-ja2-stracciatella), right? Is the detailed report publicly available?

momoko-h:
Thought about it a bit more and I now agree that this isn't the right approach. Why have bounds checking for this array but not all the others?

PR #1605, opened 2022-07-14
Fix OOB in AIMHistory

momoko-h:
NUM_AIM_HISTORY_PAGES includes only the actual history pages without the TOC
but the 'visited' flag is also set for the TOC page.

Turns out ASAN is quite useful even when I can't enter the tactical screen. This problem was already found but not fixed correctly in #892, and then got exposed again by #1578.


selaux:
Thanks!

PR #1609, opened 2022-07-16
Add integration points to change shop inventories and prices

ja2-derek:
# Summary

- Adding a couple of observables so mods can update dealers items and prices
- Exposing a few functions to manipulate shop keeper behaviour and their inventories
- Added support for passing UINT32 values to and from Lua

These should do nothing on its own, but allow Lua scripts to override items and prices. They allow mod support for things like Unfinished Business ([PATCH132](https://github.com/ja2-derek/ja2-ub-comparison/commit/70809a14)), and also #1205.

# Usage example

```lua
RegisterListener("OnDealerInventoryUpdated", "adjust_franz_inventory")

ARMS_DEALER_FRANZ = 2

-- Check every morning, and ensure Franz has a VideoCam to sell if the robot quest is started
function adjust_franz_inventory()
	if gubQuest[Quests.QUEST_DELIVER_VIDEO_CAMERA] == QuestStatuses.QUESTINPROGRESS then
		GuaranteeAtLeastXItemsOfIndex(ARMS_DEALER_FRANZ, Items.VIDEO_CAMERA, 1)
	end
end
```

### DIFF
--- a/src/externalized/scripting/FunctionsLibrary.cc
+++ b/src/externalized/scripting/FunctionsLibrary.cc
@@ -1,4 +1,5 @@
 #include "FunctionsLibrary.h"
+#include "Arms_Dealer_Init.h"
 #include "Campaign_Types.h"
 #include "Handle_Items.h"
 #include "Item_Types.h"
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
LGTM!

PR #1610, opened 2022-07-17
Try to avoid false positives from coverity scans

momoko-h:
Using Assert and AssertMsg can lead coverity scans to assume that the condition
we asserted could be false. For example

Assert(ptr); if (ptr->something == 0) ....

could create a defect report.

If this works as well as I hope it does, it should eliminate reduce the number of open defects quite a bit.

### DIFF
--- a/src/sgp/Debug.h
+++ b/src/sgp/Debug.h
@@ -7,7 +7,21 @@
 #define DEBUG_PRINT_FPS                         (0)             /**< Flag telling to print FPS (Frames per second) counter. */
 #define DEBUG_PRINT_GAME_CYCLE_TIME             (0)             /**< Flag telling to print how much time every game cycle takes. */
 
+// Don't ever define __COVERITY__. It is needed to avoid false positives during
+// a coverity scan from patterns like
+//
+//   Assert(p);
+//   p->something = 0;
+//
+// https://community.synopsys.com/s/question/0D534000046YuzbCAC/suppressing-assertsideeffect-for-functions-that-allow-for-sideeffects
+
+#ifndef __COVERITY__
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
I think it would be cleaner to just extend our Coverity model file. It already handles LogMessage (due to the assert param).

You can see it at the bottom here:
https://scan.coverity.com/projects/ja2-stracciatella-ja2-stracciatella?tab=analysis_settings
https://scan.coverity.com/tune

momoko-h:
Sorry, but I cannot access the page, it's for project admins only. Could you make these changes?

lynxlynxlynx:
Ah, didn't know — will do.

lynxlynxlynx:
Ah, it's a regression from the logging unification, we even have the model versioned: src/coverity_model.cpp.

momoko-h:
I believe those 7 are the bugs fixed in #1608. If this had worked as planned 4 of the reports first detected on 6/6/22 and several others should be gone now.

lynxlynxlynx:
Yeah, you're right. My session expired, so I didn't look at the case I had open before. But following it up, it's true, 254710 is still open. Not sure if the link will work:
https://scan9.scan.coverity.com/reports.htm#v28877/p12081/fileInstanceId=98711492&defectInstanceId=8820756&mergedDefectId=254710

lynxlynxlynx:
I'll try an approach where the payload is not in the macro itself. Just have to fix my setup, since sdl2 headers got uninstalled due to conflicts. :/

lynxlynxlynx:
Some are surely redundant or even detrimental, but there's way too many users to just hand-wave them away. So unless we want to inline all of them, we have to find a better solution.

PR #1612, opened 2022-07-17
coverity_model: there's no more LogMessage, another attempt at #1610

lynxlynxlynx:
Will try it on coverity first.


lynxlynxlynx:
No luck with the poorly documented `#nodef` stuff either. Perhaps those and the declarations would have to be near the original macros. Which is then similar to what was proposed in #1610.

PR #1613, opened 2022-07-18
More coverity fixes

momoko-h:
More stuff for your kind consideration. Don't wory, I won't have much free time for the rest of the week and won't bother you any more. :)

Some of this is mainly to make coverity happy, but the strategic movement stuff seems useful. If we're lucky it's the cause of those spurious movement groups warnings that start to appear after a couple of game days.

### DIFF
--- a/src/game/Strategic/Strategic_Movement.cc
+++ b/src/game/Strategic/Strategic_Movement.cc
@@ -1822,8 +1822,8 @@ void RemoveGroup(GROUP& g)
 	if (g.fPersistant && !gfRemovingAllGroups)
 	{
 		CancelEmptyPersistentGroupMovement(g);
+		SLOGW("Strategic Info Warning: Attempting to delete a persistant group.");
 		return;
-		DoScreenIndependantMessageBox("Strategic Info Warning:  Attempting to delete a persistant group.", MSG_BOX_FLAG_OK, NULL);
 	}
 
 	RemoveGroupWaypoints(g);
@@ -2720,8 +2720,13 @@ void RetreatGroupToPreviousSector(GROUP& g)
 {
 	AssertMsg(!g.fBetweenSectors, "Can't retreat a group when between sectors!");
 
-	UINT8 direction = 255;
-	if (g.ubPrev.x != 16 || g.ubPrev.y != 16)
+	if (g.ubPrev.x == 16 && g.ubPrev.y == 16)
+	{
+		// Group doesn't have a previous sector. Create one first.
+		CalculateGroupRetreatSector(&g);
+	}
+
+	UINT8 direction;
 	{ // Group has a previous sector
 		g.ubNext = g.ubPrev;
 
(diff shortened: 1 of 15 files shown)


src/game/Strategic/Strategic_Movement.cc
```diff
@@ -2720,8 +2720,13 @@ void RetreatGroupToPreviousSector(GROUP& g)
 {
 	AssertMsg(!g.fBetweenSectors, "Can't retreat a group when between sectors!");
 
-	UINT8 direction = 255;
-	if (g.ubPrev.x != 16 || g.ubPrev.y != 16)
+	if (g.ubPrev.x == 16 && g.ubPrev.y == 16)
```
lynxlynxlynx, line 2723: Would be cleaner to use IsValid(), but that allows 16. I find the original check suspect — why exclude this one sector?

src/game/Strategic/Strategic_Movement.cc
```diff
@@ -2720,8 +2720,13 @@ void RetreatGroupToPreviousSector(GROUP& g)
...
+	{
+		// Group doesn't have a previous sector. Create one first.
+		CalculateGroupRetreatSector(&g);
+	}
+
+	UINT8 direction;
 	{ // Group has a previous sector
```
lynxlynxlynx, line 2730: You can remove this level of braces and unindent.

momoko-h: I didn't want to touch more lines than necessary for the sake of git blame, and I liked the fact that these braces limit the scope of delta, although C++17 would allow me to move the initialization into the if() statement instead. I'll change it if you want.

src/game/Strategic/Strategic_Movement.cc
momoko-h: The problem is trying to fit 256 different sector ids in one byte and still having a value to indicate 'invalid'. There is at least one other place, [in AutoResolve](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/ff9cf59cdb0860c90e258ef43ac8a403ef058308/src/game/Strategic/Auto_Resolve.cc#L3889), which uses P16 as the odd one out.

src/game/Strategic/Strategic_Movement.cc
lynxlynxlynx: Can be a separate commit, but I'll also say I definitely don't like declarations in if statements themselves.

src/game/Strategic/Strategic_Movement.cc
lynxlynxlynx: Ok. It's inaccessible anyway, likely in all modded versions as well.

lynxlynxlynx:
Btw, you've consistently been doing good work. Do you want to become an "official" team member? Besides the clout, you get the power to merge other people's changes, assign issue tags and things like that.

momoko-h:
If you don't mind having an anonymous team member I'd gladly accept your invitation 😊. I guess I have some reading to do then; other than my own I have never managed a GitHub repository before.

lynxlynxlynx:
Don't worry about it, there's nothing extra to manage. :) I've sent you the invite.

PR #1615, opened 2022-07-22
Let enemy soldiers climb onto roofs again

momoko-h:
The `GenerateBuildings()` call was deleted by mistake in a previous commit.
 
I wondered why the soldiers weren't following me onto the roof right next to helicopter drop off zone any more.


lynxlynxlynx:
Thanks! It was cruft removal in a3595cb2 from 2020 if anyone is wondering.

PR #1616, opened 2022-07-23
Debug: another attempt at avoiding asserting coverity reports

lynxlynxlynx:
Let's see, two days until the next coverity run.
std::unique_ptr has operator bool, but the compiler didn't find it sufficient, hence the `get()` chunk.

### DIFF
--- a/src/coverity_model.cpp
+++ b/src/coverity_model.cpp
@@ -29,3 +29,8 @@ constexpr void LogMessageST(bool isAssert, LogLevel level, const char* file, Arg
 		__coverity_panic__();
 	}
 }
+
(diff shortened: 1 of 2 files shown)


momoko-h:
Not really surprising that the compiler didn't like Assert(filename). Conversion from a raw pointer to a bool is probably only allowed for backwards compatibility with C.

`#define Assert(a) VerboseAssert(static_cast<bool>(a), "", __FILE__, __LINE__)` should always give you the desired result. Placing `a` in parentheses is advisable anyway.

lynxlynxlynx:
Yeah, good idea, thanks.

PR #1617, opened 2022-07-24
Adressing some of the CIDs in NPC.cc

momoko-h:
CID 81519-81523, 81526, 146376. Also attempts to address the 'untrusted value
as argument' CIDs between 249129 and 249151.

The pSoldier null checks after `FindSoldierByProfileID` are usually present in other places in this file, so at the very least this is now consistent. 

I hope narrowing the types of `Fact` and the constants is enough to convince Coverity that we've validated our loaded data.


lynxlynxlynx:
Hmm, a compiler bug on ARM? Maybe bumping the android runner image would resolve it.

momoko-h:
If I understand the log, the Android build is done with Ubuntu 18.04? The default GCC version there was 7.4, that's probably too old. Can it be updated separately?

lynxlynxlynx:
GCC? Yes:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/.ci/ci-setup.sh (even have a handy function for it)
While for a newer image:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/.github/workflows/github-ci.yml#L29

momoko-h:
Ok, let's see if I did this correctly...

lynxlynxlynx:
Add `linux-install-via-apt-get "gcc-$GCC_VER" "g++-$GCC_VER"` before the setting.

momoko-h:
Wait, this can't work. The log says it's using clang++ to cross-compile, not gcc. How do I find out which version it is currently using and what the alternatives for Ubuntu 18.04 are?

lynxlynxlynx:
it's version 9:
https://github.com/actions/virtual-environments

momoko-h:
I just tried install clang 10 in a virtual machine but for some reason it doesn't appear in the list of alternatives, so it would still use clang 6. Are there any objections to update all runners to Ubuntu 20.04?

momoko-h:
I'll just try it and see what happens. Last thing I do today.

ja2-derek:
We prefer the oldest image (oldest clib) feasible for Linux to support as many platforms as possible (see #1360).

selaux:
The Android build does not use the system clang. It uses the clang version included in the NDK we build the Android build with. NDK verison is currently [21.0.6113669](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/android/app/build.gradle#L9), so we would probably need to upgrade this.

momoko-h:
And now I've [just read](https://github.com/actions/virtual-environments/issues/5930) that NDK r21 will be replaced with r25, so should we just use that one instead?

selaux:
If it's possible lets upgrade to r25, so we are on the latest LTS release. If it's more than just upgrading the version I would not invest the time right now though. Keeping up with Android changes can be exhausting ^^

src/game/Strategic/Facts.h
```diff
@@ -1,9 +1,10 @@
 #ifndef FACTS_H
 #define FACTS_H
+#include <cstdint>
 
-enum Fact
+enum Fact : uint16_t
 {
-	FACT_NONE = -1,
+	FACT_NONE = UINT16_MAX,
```
selaux, line 7: Did you check whether this is persisted anywhere, e.g. in save games or resource files? If so this would be a break reading from old files.

src/game/TacticalAI/NPC.cc
```diff
@@ -1876,38 +1879,43 @@ void ConverseFull(UINT8 const ubNPC, UINT8 const ubMerc, Approach bApproach, UIN
...
 				{
 					{
-						INT8 bInvPos;
-
 						SOLDIERTYPE* const pSoldier = FindSoldierByProfileID(ubNPC);
+						if (pSoldier)
+						{
+							// Look for item....
+							INT8 const bInvPos = FindObj( pSoldier, pQuotePtr->usGiftItem );
+							if (bInvPos == NO_SLOT) throw std::logic_error("NPC.cc: Gift item does not exist in NPC.");
```
selaux, line 1887: Do we really want to throw an exception here? We could just log the error and not gift the item, which would be more robust.

src/game/Strategic/Facts.h
momoko-h: `FACT_NONE` didn't exist in the original code base, it was added by Tron in commit 58839b9b7c633d25d500783cb8a65a32b6858af4 and is currently only used in two places to signify an invalid value. It now has the same value as NO_FACT (65535), which it replaces everywhere so that we are only using values from the Fact enum. It would also be possible to do it the other way around, keep NO_FACT and drop FACT_NONE, but this approach seems more consistent.

src/game/TacticalAI/NPC.cc
momoko-h: It's first and foremost here to tell Coverity that we're really checking our return values, I would not expect to ever see this exception unless somebody breaks something in the future, in which case a game abort might even be preferable to give the problem more urgency.

 ATM it would already assert in a debug build and corrupt memory in release builds if we continue past this point.

momoko-h:
> If it's possible lets upgrade to r25, so we are on the latest LTS release. If it's more than just upgrading the version I would not invest the time right now though. Keeping up with Android changes can be exhausting ^^

v23 is already present, v25 should arrive in the next two days. If it's included in the virtual environment, I just have to set the version in build.gradle and could remove the linux-install-via-android-sdkmanager line from ci-setup.sh, correct?

selaux:
I'd leave the `linux-install-via-android-sdkmanager` in, so we ensure that the version we expect is independent from whatever is installed in the runner. Otherwise every NDK update in the runner images will cause the build to fail.

momoko-h:
I need some help here: why would my changes cause any problem with SDL2? What do I need to do to get this Android build working?

selaux:
TBH I have no idea why a toolchain update would cause this, but I can have a look.

lynxlynxlynx:
This builds now. :) Is the macos image bump a remnant or intentional?

momoko-h:
That's intentional, MacOS 10.15 support [ends this month.](https://github.com/actions/runner-images/issues/5583) It's not really necessary for this PR of course, it just was something I added while I was making the changes to get this PR to build.

lynxlynxlynx:
Ok, thanks!

PR #1619, opened 2022-07-25
Let Auto Resolve free the soldiers created for it

momoko-h:
Make it crystal clear that we're not freeing one of the soldiers from the Menptr
array. Also fix the (harmless) warnings caused by #1571.

CID 275310

About these warnings: the AR soldiers aren't fully initialized but `DeleteSoldier()` doesn't know about that and tries to free the animation cache. That makes `UnLoadAnimationSurface()` unhappy and causes it to print one warning for every AR soldier. Ugly. Now every AnimationCache is always in a valid state, no matter where the outer SOLDIERTYPE was created.

And `ProfileID` is used for Merc Profiles, not soldier ids.

### DIFF
--- a/src/game/Tactical/Animation_Cache.cc
+++ b/src/game/Tactical/Animation_Cache.cc
@@ -6,19 +6,10 @@
 #include "Overhead.h"
 #include <climits>
 
-#define EMPTY_CACHE_ENTRY 65000
 
-
-void AnimationSurfaceCacheType::init(ProfileID const pid)
+void AnimationSurfaceCacheType::init(UINT8 const pid)
 {
 	mPid = pid;
-
-	// Zero entries
-	for ( int cnt = 0; cnt < ANIM_CACHE_SIZE; cnt++ )
-	{
-		usCachedSurfaces[ cnt ] = EMPTY_CACHE_ENTRY;
-		sCacheHits[ cnt ] = 0;
-	}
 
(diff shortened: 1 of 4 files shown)


src/game/Tactical/Animation_Cache.cc
```diff
@@ -8,17 +8,22 @@
 
 #define EMPTY_CACHE_ENTRY 65000
 
-
-void AnimationSurfaceCacheType::init(ProfileID const pid)
+// Ensure cache state is valid even if init is never called
+// (the special soldiers created by Auto Resolve)
+AnimationSurfaceCacheType::AnimationSurfaceCacheType()
```
lynxlynxlynx, line 13: This just screams for further simplification, as the whole ctor can be removed in favour of in-class initializers.

lynxlynxlynx:
Other than that it looks good. 👍🏾

src/game/Tactical/Animation_Cache.cc
momoko-h: Yeah, I don't like C++'s syntax for initializing an entire array to the same value too much, but it's probably still better than the separate constructor.

lynxlynxlynx: For the second member you could've used default initialisation, but yeah, when you don't want zeros, it can get ugly without switching to std::array.

PR #1621, opened 2022-07-26
coverity: enable asserts

lynxlynxlynx:
Continuation of trying to get Assert* false positives to vanish.

In the meanwhile I played with various models, for example trying to manually provide the one function needed from the templating of LogMessageST, toyed with other ideas, but this is the simplest next attempt — just don't ifdef the assert out for coverity, since it's not analyzing a debug build.


momoko-h:
Looks good, if that still doesn't convince coverity that failed asserts are fatal I don't know what would.

PR #1622, opened 2022-07-27
Fix several incorrect gridno comparisons

momoko-h:
`CreatePhysicalObject`:
`MAPROWCOLTOPOS` returns -1, not NOWHERE in case of invalid arguments. Also
removes the constant `oneOverMass` and adds a new function for the vector_3
to gridno computations.

Enable bounds checking in `CheckForCollision`. Checking for collisions in not
visible tiles seems ok, but not on invalid tiles.

I searched for comparisons with WORLD_MAX and these are all locations I could find which were wrong.


src/game/Tactical/LOS.cc
```diff
@@ -4311,15 +4307,10 @@ INT32 CheckForCollision(FLOAT dX, FLOAT dY, FLOAT dZ, FLOAT dDeltaX, FLOAT dDelt
...
 	sY = (INT16)( dY / CELL_Y_SIZE );
 
-	// Check if gridno is in bounds....
-	if ( !GridNoOnVisibleWorldTile( (INT16) (sX + sY * WORLD_COLS) ) )
-	{
-		//return( COLLISION_NONE );
-	}
-
-	if ( sX < 0 || sX > WORLD_COLS || sY < 0 || sY > WORLD_COLS )
+	if (sX < 0 || sX >= WORLD_COLS || sY < 0 || sY >= WORLD_COLS)
```
lynxlynxlynx, line 4310: One of these should be WORLD_ROWS, I guess the last.

lynxlynxlynx:
Nice little cleanup.

src/game/Tactical/LOS.cc
momoko-h: Yes, I didn't even notice that one.

PR #1629, opened 2022-08-01
coverity: properly pass a cmake define [ci skip]

lynxlynxlynx:
Grrr, now I really feel silly.

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -39,7 +39,7 @@ jobs:
           cd ${COV_BUILD_DIR}
           cmake ${CMAKE_BUILD_OPTIONS} ..
         env:
-          CMAKE_BUILD_OPTIONS: "-DCMAKE_BUILD_TYPE=Release -DUSE_SCCACHE=OFF -DENABLE_ASSERTS"
+          CMAKE_BUILD_OPTIONS: "-DCMAKE_BUILD_TYPE=Release -DUSE_SCCACHE=OFF -DENABLE_ASSERTS=1"
 
       - name: Run Coverity Scan Analysis Tool
         run: |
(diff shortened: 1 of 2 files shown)


momoko-h:
Why does that make a difference, we're just testing #ifdef ENABLE_ASSERTS, right?

lynxlynxlynx:
CMake is being pedantic about its own syntax. What I wrote would work on the compiler command-line, but cmake itself wants a specific value. But this won't be enough either, come to think of it.

momoko-h:
Ah OK. I've just read in the [CMake documentation](https://cmake.org/cmake/help/latest/command/add_definitions.html) that add_definitions has been superseded by add_compile_definitions since v3.12. Do we have a minimum supported CMake version older than that?

lynxlynxlynx:
Currently: `cmake_minimum_required (VERSION 3.4)`

momoko-h:
3.4 was released in 2015, and 3.12 in July 2018. GitHub's Ubuntu 18.04 has 3.23.2, so maybe it's time to raise that limit?

lynxlynxlynx:
Do we gain anything? CMake is good about backward compatibility.

lynxlynxlynx:
I don't know if it'd be useful, unless someone wanted to refactor our current android build. But I'm not against a bump, just was curious if it had any tangible benefits.

momoko-h:
According to [this](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html), we would already need at least 3.8 for our current use of C++17 and if you want C++20 you'd need 3.12.

selaux:
I would be fine with updating it to a more recent version (maybe at least 3.8 for proper compatibiltiy with C++17). We even have [a file](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e23678b74af7e7fa2ccaf067eb44993184af0a33/dependencies/lib-smacker/CMakeLists.txt) which gives warnings about the required version (2.8) being too low and compatibility being removed. But it should probably not be part of this PR.

momoko-h:
That did it! 25 of the 27 newly eliminated issues are due to this change.

lynxlynxlynx:
Yep, will submit another PR to remove the now useless model file.

PR #1630, opened 2022-08-03
Do not use abs() before hypot()

momoko-h:
The sign of hypot's arguments has no influence on its result.

So `GetPythDistanceFromPalace()` rounds its result while the others truncate. I doubt that was a conscious decision by the developers back then but I also see no need to change the game logic now.


selaux:
Thanks for all the PRs :slightly_smiling_face:

PR #1637, opened 2022-08-08
Fix seven more Coverity issues

momoko-h:
• Wrong index used into soldier's inventory (CID 249130)
• Buffer overrun in SaveLoadGame unit test (215865)
• harden logic in a workaround for a potential problem in ancient saves (250860)
• Coverity does know that Random(10) can only return 0-9 (81654)
• Safer array access in CheckFact (254706)
• Remove special SKI handling in InitItemDescriptionBox (249140)
• Write after free in HandleNonCombatGroupArrival (209947)

Two real bugs in the game (209947 and 249130), one in an unit test, and another that's unlikely to still be relevant (the fix for saves version 61-65). The other three are "make Coverity happy" changes, but at least one also gets rid of some horrible code in `InitItemDescriptionBox`

### DIFF
--- a/src/game/SaveLoadGame.cc
+++ b/src/game/SaveLoadGame.cc
@@ -1114,17 +1114,19 @@ void LoadSavedGame(const ST::string &saveName)
 		if (s) TacticalRemoveSoldier(*s);
 
 		// add the pilot at a random location!
-		SGPSector sMap;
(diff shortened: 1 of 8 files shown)


lynxlynxlynx:
LGTM!

PR #1639, opened 2022-08-10
Upgrade Android NDK to r25

selaux:
This is a result of checking what is actually happening with the NDK upgrade in https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1617.

Turns out SDL2 does not compile yet on NDK r25 (https://github.com/libsdl-org/SDL/issues/6019) and somehow the Android build hides all the error messages of the configure step for CMake, so this is not noticeable.

I upgraded the gradle build plugin, so I could actually see the error messages and added the commit that fixes the SDL2 build as a patch. Lets see how well it fares in CI, as with the gradle upgrade a different Java version is expected to be used.


momoko-h:
Thank you for this, it would have taken me ages to figure all that out myself.

selaux:
Yep, that is how it goes with Android 😉 . That's why I'd like to keep the Android wrapper as small as possible. Its just annoying to keep up with their changes...

lynxlynxlynx:
So is there anything left to do, now that it builds?

selaux:
No. I'll merge.

PR #1640, opened 2022-08-10
Fix two out of range array accesses

momoko-h:
• `gsTerrainTypeSpeedModifiers` didn't have an entry for NO_TERRAIN
• `LookForAndMayCommentOnSeeingCorpse` can also be called for EPCs with a profile id > 74
• Remove unused array `QuoteExp_TeamSpecific`

I found this while testing my new fix for 1635 with an ASan build. More to come, but I need to investigate another crash I encountered a bit more first.

### DIFF
--- a/src/game/Tactical/QArray.cc
+++ b/src/game/Tactical/QArray.cc
@@ -1,7 +1,7 @@
 #include "QArray.h"
 
 
-BOOLEAN QuoteExp_HeadShotOnly[75] =
+std::array<INT8, 75> const QuoteExp_HeadShotOnly
 {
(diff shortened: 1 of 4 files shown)


momoko-h:
So the AppVeyor build failed because of a missing `#include <array>`. I'd like to propose to make some of the most useful standard library types "first-class citizens": add their includes and `using std::array` etc. statements to Types.h. Would that be acceptable or what are the counter-arguments?

lynxlynxlynx:
Types.h gets included a whole lot, so I wouldn't needlessly bloat the binary / slow the compilation. Also, being more precise has historically been better, as compilers changed their internal header dependencies, split stuff off etc., breaking builds on version bumps.

PR #1641, opened 2022-08-10
Display correct location of militia training in Finances

momoko-h:
ATM all militia training is reported to happen in sector B10.


lynxlynxlynx:
Thanks!

Issue #1592, opened 2022-07-07
Current Android target API set to Oreo (API 26)


Srandista:
It's not so much a bug, rather then question.

Is there any specific reason, why the current target API is set to just an Oreo? That's already a quite old Android version, and it sits there for as long as I can remember with Android releases. It's not that is some big issue or something, but if you would push Stracciatella one day to Play Store, this would need to be changed and push higher, since such an old API is no longer accepted by Google.

As far as I know, target API is independent on minimum API, so it can be changed without needing to push the minimum Android OS requirement higher.

Again, not a bug, I'm just curious about it, since JA2 Stracciatella is one of the "oldest" app in my phone (old as a app targeting such an old API).

lynxlynxlynx:
Yeah, that scoped storage is a PITA, so I also wouldn't bump the API. Is there something still to do here or can we close?

PR #1642, opened 2022-08-11
Upgrade Android dependencies and clean up deprecations

selaux:
After doing the NDK upgrade, I decided to do the same for other android dependencies, while keeping the target sdk the same (because the issues described in https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1592). I fixed all deprecations as well. Functionally there should be no changes.


selaux:
Seems like one depecation remains. But it's better than before...

Issue #1635, opened 2022-08-07
Segfault in DistanceVisible()


guijan:
### What is wrong

The game segfaults

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0ec7a348c6be2a6dd5dd414e261ca17542e6e58d (current master) | 1.12 GoG  |    don't know    |    don't know     |

- how to reproduce the bug and, if possible, a save game

[2022-07-31t06-12-04z-j.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/9276697/2022-07-31t06-12-04z-j.sav.zip)

Load my save, enter the tactical screen. To avoid confusion, the right sector (J9) is already selected, it's the one with Buns inside Skyrider's helicopter.

This crash doesn't seem to happen 100% of the time. 

- the operating system

OpenBSD 7.1 amd64

- backtrace

```gdb
Thread 1 received signal SIGSEGV, Segmentation fault.
0x000009418eb94512 in DistanceVisible (pSoldier=0x9418f803e28 <Menptr+17640>,
    bFacingDir=3 '\003', bSubjectDir=2 '\002', sSubjectGridNo=8594, bLevel=0 '\000')
    at ../src/game/Tactical/OppList.cc:1014
1014            if ( gpWorldLevelData[ pSoldier->sGridNo ].ubExtFlags[ bLevel ] & (MAPELEMENT_EXT_TEARGAS | MAPELEMENT_EXT_MUSTARDGAS) )
(gdb) bt
#0  0x000009418eb94512 in DistanceVisible (pSoldier=0x9418f803e28 <Menptr+17640>,
    bFacingDir=3 '\003', bSubjectDir=2 '\002', sSubjectGridNo=8594, bLevel=0 '\000')
    at ../src/game/Tactical/OppList.cc:1014
#1  0x000009418eb9d510 in ManLooksForMan (pSoldier=0x9418f803e28 <Menptr+17640>,
    pOpponent=0x9418f846f20 <Menptr+292320>, ubCaller=1 '\001')
    at ../src/game/Tactical/OppList.cc:1460
#2  0x000009418eb932c5 in OtherTeamsLookForMan (pOpponent=0x9418f846f20 <Menptr+292320>)
[...]

Issue #1634, opened 2022-08-06
Yet another segfault in DistanceVisible()


guijan:
### What is wrong

The game segfaults.

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0ec7a348c6be2a6dd5dd414e261ca17542e6e58d (current master) | 1.12 GoG  |    don't know    |    don't know     |

- how to reproduce the bug and, if possible, a save game)

[2022-07-31t06-12-09z-j.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/9275291/2022-07-31t06-12-09z-j.sav.zip)

Load my save, enter the tactical screen. 

- the operating system

OpenBSD 7.1 amd64

- backtrace

```gdb
Thread 1 received signal SIGSEGV, Segmentation fault.
0x000009418eb94512 in DistanceVisible (pSoldier=0x9418f803e28 <Menptr+17640>,
    bFacingDir=3 '\003', bSubjectDir=2 '\002', sSubjectGridNo=8594, bLevel=0 '\000')
    at ../src/game/Tactical/OppList.cc:1014
1014            if ( gpWorldLevelData[ pSoldier->sGridNo ].ubExtFlags[ bLevel ] & (MAPELEMENT_EXT_TEARGAS | MAPELEMENT_EXT_MUSTARDGAS) )
(gdb) bt
#0  0x000009418eb94512 in DistanceVisible (pSoldier=0x9418f803e28 <Menptr+17640>,
    bFacingDir=3 '\003', bSubjectDir=2 '\002', sSubjectGridNo=8594, bLevel=0 '\000')
    at ../src/game/Tactical/OppList.cc:1014
#1  0x000009418eb9d510 in ManLooksForMan (pSoldier=0x9418f803e28 <Menptr+17640>,
    pOpponent=0x9418f846f20 <Menptr+292320>, ubCaller=1 '\001')
    at ../src/game/Tactical/OppList.cc:1460
#2  0x000009418eb932c5 in OtherTeamsLookForMan (pOpponent=0x9418f846f20 <Menptr+292320>)
    at ../src/game/Tactical/OppList.cc:1995
#3  0x000009418eb92745 in HandleSight (s=..., sight_flags=SIGHT_LOOK)
    at ../src/game/Tactical/OppList.cc:669
[...]

PR #1643, opened 2022-08-12
Make OppList.cc more robust

momoko-h:
This adds more checks in OppList.cc if a soldier should even be considered in several places, some of which even already had comments that this should be done.

The changes in the other files are necessary to stop us from already trying to update sightings while we are still on the overhead map placing mercs.

I've tested this patches with an ASan build (with a hacky workaround for the Z blitter problem) with @guijan's saves from #1634 and #1635, some of the other saves I have around and played my own campaign for over an hour without problems. No promises that the game is now completely stable because I'm still early in the campaign, but it's definitely improving.

### DIFF
--- a/src/game/Tactical/OppList.cc
+++ b/src/game/Tactical/OppList.cc
@@ -625,6 +625,7 @@ static bool IsSoldierValidForSightings(SOLDIERTYPE const& s)
 	return !(!s.bActive
 	      || !s.bInSector
 	      || s.uiStatusFlags & SOLDIER_DEAD
+	      || s.bLife <= 0 // should be redundant, but who knows?
 	      || (s.bAssignment == VEHICLE && s.iVehicleId == iHelicopterVehicleId));
 }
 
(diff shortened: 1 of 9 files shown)


src/game/Tactical/OppList.cc
```diff
@@ -619,13 +619,19 @@ static void OurTeamRadiosRandomlyAbout(SOLDIERTYPE* about);
...
+static bool IsSoldierValidForSightings(SOLDIERTYPE const& s)
 {
 	extern INT32 iHelicopterVehicleId; 	// I don't want to include a map screen header file for this
-	if (!s.bActive)                     return;
-	if (!s.bInSector)                   return;
-	if (s.uiStatusFlags & SOLDIER_DEAD) return;
-	if (s.bAssignment == VEHICLE && s.iVehicleId == iHelicopterVehicleId) return;
+	return !(!s.bActive
+	      || !s.bInSector
+	      || s.uiStatusFlags & SOLDIER_DEAD
```
selaux, line 627: Should we add a `|| s.bLife <= 0` as well. It should probably be covered by `s.uiStatusFlags & SOLDIER_DEAD`, but some of the conditions below checked it before (although they checked explicitly whether its not 0).

momoko-h: Yes, I don't trust the game to keep these two always in sync either.

PR #1644, opened 2022-08-12
Fix memory leak in tactical placement

momoko-h:
The gMercPlacement array was never freed. Old vanilla bug which also didn't have the corresponding `MemFree` call.

### DIFF
--- a/src/game/TileEngine/Tactical_Placement_GUI.cc
+++ b/src/game/TileEngine/Tactical_Placement_GUI.cc
@@ -43,7 +43,7 @@
 
 #include <string_theory/format>
 #include <string_theory/string>
-
+#include <memory>
 #include <stdexcept>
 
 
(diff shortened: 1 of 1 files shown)


src/game/TileEngine/Tactical_Placement_GUI.cc
```diff
@@ -56,11 +56,11 @@ struct MERCPLACEMENT
 	BOOLEAN				fPlaced;
 };
 
-static MERCPLACEMENT* gMercPlacement = 0;
-static INT32          giPlacements   = 0;
+static std::unique_ptr<MERCPLACEMENT[]> gMercPlacement;
```
selaux, line 59: Could this just be a `std::vector<MERCPLACEMENT>` instead?

momoko-h: Only with some additional work. It's like so many other things in this code base: it could be better but it gets the job done. My current policy about such changes is to do them only when the code is really bad or the improvement is really trivial, ideally by simply removing some lines, like #1630. This one isn't something I'd like to spent my time on while real bugs are waiting.

PR #1645, opened 2022-08-13
Fix OOB array access in UpdatePublic

momoko-h:
Plus misc. cleanup of OppList.cc

The cleanup and the macros hide the bug here a bit, so I'll explain:
The bug here is the `if` block starting in line 4588: the condition ensures that `*pbPublOL` must be either -5 or 6 when we call `UpdatePublic()`. `UpdatePublic()` then subtracts -4, so it's now either -1 or 10 and uses that as an index into a 10x10 array, so both cases are out of bounds.


lynxlynxlynx:
LGTM

PR #1649, opened 2022-08-16
More Coverity fixes

momoko-h:
• Missing null check (14638, 254710)
• Logical dead code (81428, 81439, 81442, 81443)
• Unused value (81671)
• Identical code in different branches (81477)
• Dereference after null check (81474)

Nothing really substantial this time. I used the opportunity to change the bad algorithm to loop over all directions in a random order.


selaux:
Thanks.

PR #1652, opened 2022-08-26
Another round of Coverity fixes

momoko-h:
> Release the wrapper object in an items unit test (CID 219694)
> Out of bounds array accesses (81526, 81541, 249147, 254706)
> 
> Attempt to address the "Untrusted value as argument" CIDs
>     
> Ensure that `usFactMustBeFalse` and `usFactMustBeTrue` are in range before using them as array indices.
 
The check in `EditScreen.cc` may be necessary (I use the editor too rarely to be sure) but other than that these changes shouldn't do anything except reduce the number of incidents.


lynxlynxlynx:
Thanks!

PR #1654, opened 2022-08-31
More Coverity fixes

momoko-h:
>     Do not throw an exception into an empty catch block
>     
>     Also remove a strcpy(), CID 214897, and clean up Mercs.cc a bit (clang-tidy
>     found a dead store).
> 
>     Fix several logical dead code CIDs
>     
>     CID 81429, 81432, 81433, 81437, 81438, 81441
> 
>     Fix two "Improper use of negative value" CIDs
>     
>     Coverity can't know that if pGroup->fVehicle is true s.bAssignment == VEHICLE
>     must also be true.
>     
>     CIDs 81500, 81503
> 
>     Do not deref a null return value in LoadStrategicAI()
>     
>     CID 210845
> 
>     Don't deref a null pointer in SoldierGiveItemFromAnimation()
>     
>     CID 81466
> 
>     Do not deref a null pointer when ChangeStat receives an invalid argument
>     
>     CID 81464
> 
>     Use const references in ranged for loops
>     
>     CIDs 275760, 275761, 275779
> 
>     Rework quest number logic in NPCConsiderQuote()
>     
>     Try to convince Coverity that NPCQuoteInfo::ubQuest is properly validated.
>     Add an assertion in GetHeadMinersMineIndex() to protect against an OOB access.
>     
>     CIDs 81526, 254706 and 13 "Untrusted value as argument" CIDs

If Coverity accepts all these changes (which of course isn't a given) it would reduce the number of incidents by 29. Game play wise these changes should not have any effect, since the changes just remove some dead code or implement the same thing a bit differently.

### DIFF
--- a/src/externalized/mercs/MercProfileInfo.cc
+++ b/src/externalized/mercs/MercProfileInfo.cc
@@ -53,7 +53,7 @@ void MercProfileInfo::validateData(const std::map<uint8_t, const MercProfileInfo
 	{
 		throw std::runtime_error("Too many merc profiles");
 	}
-	for (auto p : models)
+	for (auto const& p : models)
 	{
 		if (p.first != p.second->profileID)
 		{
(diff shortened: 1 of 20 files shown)


lynxlynxlynx:
LGTM, thanks!

Issue #1547, opened 2022-05-23
Communicate better how to choose a proper resolution / scaling method


selaux:
The fact how to choose a proper resolution / scaling method has been asked quite a bit in Discord and other places. I think we need to do a better job at communicating how to choose a good resolution and scaling method for the system that the game is currently running on.

One way to achieve this would be to have some explanatory text for the different options in the launcher.

Also selecting a nice default for each system would be beneficial (e.g. for resolution 1920x1080, choose fullscreen 960x540 with pixel perfect scaling).

Any other suggestions?

lynxlynxlynx:
Better defaults are definitely welcome, but I think we need imagery. Since that requires space, we'd have to move the resolution settings to a subwindow, but I think it would pay off. Something like the KDE display input switcher [comes to mind](https://i.stack.imgur.com/SMAJ6.jpg), but expanded.

Points to convey:
- differences between fullscreen/window modes
- differences between scaling modes?
- main menu is not representative, high res is only for tactical maps

Or maybe rather start from the other end, with a decision tree, "what do you want to achieve?":
- take up as much of my screen
  - no: windowed mode, no scaling
  - yes: upscale pixels?
    - no: fullscreen mode
    - yes: windowed mode, scaling, resolution lower than the desktop's

I think we can disregard anyone still having a CRT or a monitor with a resolution that is a nice factor of the original ja2 one. And we still need a note about black bars / non-tactical views.

MrMaggot:
I think a note mentioning the game's intended resolution was 640x480 and the option to select 960x540 without inputting a custom resolution for those who don't know 16:9 resolutions would be sufficient.

AlmightyBeaver:
I just installed it on my Intel Mac, but it is not possible to change the resolution of the game. You can change it in the launcher, to e.g. fullscreen and 1920x1080 but it will give you a black fullscreen with a tiny (probably 640x480) JA2 game window. 640x480 and fullscreen will give you an scaled to fullscreen JA2 game window, but the resolution is still 640x480. Is it even possible to change the game resolution to you can see more of the map while playing? 
A little explanation would be really appreciated :)

MrMaggot:
The laptop screen and main menu will always be 640x480. I would recommend trying 960x540 or 1280x720 to see more on the battle map while the menus are still legible.

lynxlynxlynx:
[...]

PR #1663, opened 2022-10-06
Tooltips to explain about resolution and scaling modes

ja2-derek:
Cosmetics only updates to the launcher tab to explain about resolution and scaling options (#1547). This does not include any material change to scaling and rendering.

- Calling it the "Internal Resolution" - that it is not the window size
- Moving the resolution up to indicate the scaling is applied over the internal resolution
- Added tooltips trying to explain what they are

### New layout

Look for `o->tooltip` in the diff to see the tooltip texts.

image


src/launcher/StracciatellaLauncher.cc
```diff
@@ -98,75 +99,70 @@ le from the mod with the highest priority is used.");
...
             invalidResolutionLabel->labelcolor((Fl_Color)1);
             invalidResolutionLabel->align(Fl_Align(FL_ALIGN_LEFT|FL_ALIGN_INSIDE));
           } // Fl_Box* invalidResolutionLabel
           o->end();
         } // Fl_Group* o
+        { Fl_Group* o = new Fl_Group(10, 170, 500, 55);
+          o->tooltip("Choose a scaling mode to stretch the game video to fit the screen. Linear and\
+ Near-Perfect modes stretch the video to match any window sizes, but the image\
+ may look blurred. Pixel-Perfect can only scale up by integer (1x, 2x and so o\
+n) but gives sharp images.");
```
selaux, line 148: Not sure how much space there is in the tooltip, but it might make sense to add that you need to enable fullscreen for pixel-perfect scaling and use the "low-resolution" that will be scaled up.

src/launcher/StracciatellaLauncher.cc
```diff
@@ -10,6 +10,7 @@ StracciatellaLauncher::StracciatellaLauncher() {
       { Fl_Group* o = new Fl_Group(0, 50, 520, 300, "@> Play  ");
         o->labelsize(16);
         o->labelcolor((Fl_Color)24);
+        o->hide();
```
selaux, line 13: Just a guess, but I think you need to change to the first tab before saving and exporting to C++ in Fluid. Otherwise the tab currently open in Fluid will be open in the application. Or we just add some code that shows the first tab on startup.

src/launcher/StracciatellaLauncher.cc
ja2-derek: You don't need full screen mode for it. Just resize the window until it's large enough.

We should later make a 2x/3x button in the scaling option, instead of having the code to calculate on window resize. That should be much less confusing

Issue #1661, opened 2022-09-30
Graphical corruption


dijidiji:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
I'm not really sure how to describe this graphical error. It seems to occur randomly. Restarting fixes it but it reoccurs after a while. Some sessions I can play without it cropping up. I'd test it on the last stable but I get a save error on that version which is why I switched to the nightly. Resolution is set to 1280x720 with pixel perfect upscaling. Native res is 2560x1440.
![image](https://user-images.githubusercontent.com/6995669/193290471-2181250f-0e43-4012-b1b3-a5bec35da914.png)

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  Nightly 0.20.0-20220920+fa3273c | JA2 (GOG)  |   no    |    N/A     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)
Specs:
Intel i7-4790K
Nvidia GTX 1060 6GB (driver version 512.15)
24GB DDR4
Running Windows 10 21H2 build 19044.2006

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)
Log:
```
2022-09-30T11:41:59 [INFO] stracciatella::logger: Logging to file "C:\\Users\\Daniel\\AppData\\Local\\Temp\\ja2.log"
2022-09-30T11:42:00 [INFO] stracciatella::vfs: VFS item with priority 25: DirFs { "D:\\Users\\Daniel\\Documents\\JA2\\data" }
2022-09-30T11:42:00 [INFO] stracciatella::vfs: VFS item with priority 24: DirFs { "F:\\Games\\JA2 Stracciatella\\externalized" }
[...]

PR #1665, opened 2022-10-11
Ensure RegisterBackgroundRect properly initializes BACKGROUND_SAVE structs

momoko-h:
`GetFreeBackgroundBuffer` does not guarantee that the returned `BACKGROUND_SAVE` is in a clean state so `RegisterBackgroundRect` must always set all members of this struct.

Without a reliable way to reproduce #1661 this is nothing more than an educated guess. I have definitely seen structs which had `fDisabled` set which can certainly cause graphical corruptions.  Another possible cause might be non-null `pSaveArea` or `pZSaveArea` pointers which would confuse `RestoreBackgroundRects` and `SaveBackgroundRects`. Vanilla had some extra checks which would have hidden this problem.

### DIFF
--- a/src/game/TileEngine/Render_Dirty.cc
+++ b/src/game/TileEngine/Render_Dirty.cc
@@ -48,8 +48,6 @@ static VIDEO_OVERLAY* gVideoOverlays;
 		if (iter##__next = iter->next, iter->fDisabled) continue; else
 
 
-SGPRect gDirtyClipRect;
-
 
 static BOOLEAN gfViewportDirty = FALSE;
 
@@ -92,6 +90,8 @@ void ExecuteBaseDirtyRectQueue(void)
 }
 
 
+// Callers of this function must assume that the returned struct is in a
+// random state and initialise ALL members!
 static BACKGROUND_SAVE* GetFreeBackgroundBuffer(void)
 {
 	auto before{gBackSaves.before_begin()};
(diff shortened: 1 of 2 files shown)


momoko-h:
I'm now fairly confident that this is the cause of the problem. Try setting a break point at [Render_Dirty.cc:142](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/4e06cac7ddfb76283f0ead3aeaa9cb27984a155d/src/game/TileEngine/Render_Dirty.cc#L142) in gdb and set the following commands for this BP:
```
silent
set b->fDisabled=1
continue
```

Moving the cursor around now causes the same corruption seen in the screenshot from #1661.

lynxlynxlynx:
LGTM & looks like the test failures are unrelated and just a temporary hiccup considering the tests passed elsewhere.

> [ RUN      ] ExternalizedData.readAllData
> Error: -11T13:41:38 [ERROR] externalized/DefaultContentManager.cc: Could not read vanilla item strings from binarydata/itemdesc.edt: openGameResForReading: VfsFile_open "binarydata/itemdesc.edt": entity not found

ja2-derek:
Thanks!

dijidiji:
Was about to comment that I've not encountered the issue since building this PR but it's already been merged. Thanks for the speedy fix!

PR #1667, opened 2022-10-15
Options to build with or without magic_enum

### LINKED ISSUE #1599: Allow using system magic_enum
Many systems forbid downloading or embedding dependencies, please provide an option to use a system magic_enum.

ja2-derek:
Now we will have 3 options with `magic_enum`:

1. By default, without any arguments, it will download and build with `magic_enum` v0.7.3
2. Another option is to build with a system-provided `magic_enum`. Supply `-DLOCAL_MAGICENUM_LIB=0 -DMAGICENUM_INCLUDE_DIR=/path/to/magic_enum`. The resulting binary may have `-enumgen` capability, depending on the build environment.
3. To build without `magic_enum` use `-DWITH_MAGICENUM=0`. The build will skip `magic_enum` and the resulting binary will not be able to do `-enumgen`.

`magic_enum` v0.8 seems to have breaking changes, so you would want v0.7.3 no matter how to provide the `magic_enum.hpp`.

This resolves #1599.


lynxlynxlynx:
Thanks!

PR #1669, opened 2022-10-15
Autosave please

lynxlynxlynx:
As no other opinion was provided, I went with my own and just removed the file check.


momoko-h:
Looks fine code wise, if someone complains about these auto saves we could still add a new option to game.json.

PR #1672, opened 2022-10-16
Script extension points on new merc hired

ja2-derek:
This adds new functions for mods to manipulate soldiers via scripts. This allows more complex logic on soldiers (skills, name, stats and appearances of RPC, hired mercs, enemies) which can depend on progress, current squad composition, fact, events, or items owned.

## Summary

Exposing a new function and a data type via Lua:

 - `GetMercProfile()`
 - `SGPSector`

New observable extension points:

 - `OnMercHired`
 - `OnRPCRecruited`
 - `OnSoldierCreated` now covers AutoResolve battles too

## Example use-cases

 - Make enemies soldiers tougher on specific events
 - Dynamically grant RPCs, depending on what skills you need the most (or the least)
 - Update stats or skills of un-hired soldiers to simulate growth

### DIFF
--- a/src/externalized/scripting/FunctionsLibrary.cc
+++ b/src/externalized/scripting/FunctionsLibrary.cc
@@ -8,6 +8,7 @@
 #include "MessageBoxScreen.h"
 #include "Queen_Command.h"
 #include "SaveLoadGameStates.h"
+#include "Soldier_Profile.h"
 #include "StrategicMap.h"
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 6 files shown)


src/externalized/scripting/FunctionsLibrary.cc
```diff
@@ -70,6 +71,11 @@ void PlaceItem(const INT16 sGridNo, OBJECTTYPE* const pObject, const INT8 ubVisi
 	AddItemToPool(sGridNo, pObject, static_cast<Visibility>(ubVisibility), 0, 0, 0);
 }
 
+MERCPROFILESTRUCT* GetMercProfile(const UINT8 ubProfileID)
+{
+	return &(GetProfile(ubProfileID));
```
lynxlynxlynx, line 76: Why is it returning the address, not the pointer itself?

ja2-derek: The C function `GetProfile` returns by reference. Here we convert the reference to pointer, because Sol/Lua can only pass by value. And we want Lua to have the pointer (instead of a copy) so scripts can modify the data.

Issue #1679, opened 2022-10-19
Rapidly right clicking stacked items exits with code -1073741819


dijidiji:
I first ran into this while checking through multiple stacks of mags to see if any were partially depleted.
Easiest way I've found to reproduce is to set up multiple stacked items (mags, first aid kits, doesn't seem to matter) next to each other and spam right click while moving the mouse around them.
Happens on a brand new save, too.

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  Nightlies | 1.12/GOG  |    no    |    yes    |

Has occurred on:
0.20.0-20221019+6539e41
0.20.0-20221017+d95d98d
0.20.0-20221012+4e06cac

Haven't been able to reproduce on:
0.20.0-20220920+fa3273c
0.20.0-20220812+21f5310
0.20.0-20220627+8d5389b
0.19.1-git+6ad1b6f

Specs:
Intel i7-4790K
Nvidia GTX 1060 6GB (driver version 512.15)
24GB DDR4
Running Windows 10 21H2 build 19044.2006

Log:
```
2022-10-19T12:00:54 [INFO] stracciatella::logger: Logging to file "C:\\Users\\Daniel\\AppData\\Local\\Temp\\ja2.log"
2022-10-19T12:00:54 [INFO] stracciatella::vfs: VFS item with priority 25: DirFs { "D:\\Users\\Daniel\\Documents\\JA2\\data" }
2022-10-19T12:00:54 [INFO] stracciatella::vfs: VFS item with priority 24: DirFs { "F:\\Games\\ja2-stracciatella_0.20.0-20221019+6539e41_win-msvc64-142\\externalized" }
2022-10-19T12:00:54 [INFO] stracciatella::vfs: VFS item with priority 23: DirFs { "F:\\Games\\GOG Library\\Jagged Alliance 2\\data" }
2022-10-19T12:00:54 [INFO] stracciatella::vfs: VFS item with priority 22: SlfFs { source: DirFsFile { "laptop.slf" in "F:\\Games\\GOG Library\\Jagged Alliance 2\\data" } }
2022-10-19T12:00:54 [INFO] stracciatella::vfs: VFS item with priority 21: SlfFs { source: DirFsFile { "anims.slf" in "F:\\Games\\GOG Library\\Jagged Alliance 2\\data" } }
[...]

Issue #1666, opened 2022-10-14
deposit/withdraw button just doesn't work


wertyzp:
vanilla game, master branch, 0.20
deposit/withdraw button just doesn't work

PR #1682, opened 2022-10-21
Touch control fixes

selaux:
This should fix:

- #1679: Caused by the mouse region deleted within its callback, leading to a segfault. We now guard against deleted / disabled regions in `MouseCallbackPrimarySecondary` and `ButtonCallbackPrimarySecondary` before calling the callbacks
- #1666: We now always compare callback reason flags using the `&` operator
- Assertion messages on tactical placement screen: The array was not properly zeroed out when resetting merc placements

### DIFF
--- a/src/sgp/Button_System.h
+++ b/src/sgp/Button_System.h
@@ -1,7 +1,6 @@
 // by Kris Morness (originally created by Bret Rowden)
 
-#ifndef BUTTON_SYSTEM_H
-#define BUTTON_SYSTEM_H
+#pragma once
 
 #include "MouseSystem.h"
 
@@ -40,6 +39,13 @@ struct GUI_BUTTON;
 
 // GUI_BUTTON callback function type
 typedef std::function<void(GUI_BUTTON*, UINT32)> GUI_CALLBACK;
+
+GUI_CALLBACK ButtonCallbackPrimarySecondary(
+	GUI_CALLBACK primaryAction,
+	GUI_CALLBACK secondaryAction,
+	GUI_CALLBACK allEvents = nullptr,
+	bool triggerPrimaryOnMouseDown = false
+);
 
 // GUI_BUTTON structure definitions.
 struct GUI_BUTTON
@@ -286,5 +292,3 @@ const ButtonDimensions* GetDimensionsOfButtonPic(const BUTTON_PICS*);
 UINT16 GetGenericButtonFillColor(void);
 
 void ReleaseAnchorMode(void);
-
-#endif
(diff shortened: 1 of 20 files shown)


momoko-h (commented):
I dislike this change. I'd rather remove the assertion in MSYS_DefineRegion, because its making the wrong assumption that the memory block it's initializing was a `MOUSE_REGION` before, so all it is testing is whether the previous owner had set some random bit. It's basically a pre-C++ version of a constructor, and as such must not assume zero initialized memory.

In the original code the `AssertMsg` was inside an `#ifdef MOUSESYSTEM_DEBUGGING` anyway.

src/sgp/Button_System.h
```diff
@@ -41,6 +40,13 @@ struct GUI_BUTTON;
 // GUI_BUTTON callback function type
 typedef std::function<void(GUI_BUTTON*, UINT32)> GUI_CALLBACK;
 
+std::function<void(GUI_BUTTON*, UINT32)> ButtonCallbackPrimarySecondary(
+	std::function<void(GUI_BUTTON*, UINT32)> primaryAction,
+	std::function<void(GUI_BUTTON*, UINT32)> secondaryAction,
+	std::function<void(GUI_BUTTON*, UINT32)> allEvents = nullptr,
+	bool triggerPrimaryOnMouseDown = false
+);
+
```
momoko-h, line 49: primaryAction, secondaryAction and allEvents should probably use GUI_CALLBACK.

src/sgp/MouseSystem.cc
```diff
@@ -974,3 +974,13 @@ void MOUSE_REGION::AllowDisabledRegionFastHelp(bool const allow)
...
 }
+
+std::function<void(MOUSE_REGION*, UINT32)> MouseCallbackPrimarySecondary(
+	std::function<void(MOUSE_REGION*, UINT32)> primaryAction,
+	std::function<void(MOUSE_REGION*, UINT32)> secondaryAction,
+	std::function<void(MOUSE_REGION*, UINT32)> allEvents,
+	bool triggerPrimaryOnMouseDown
+)
+{
+	return CallbackPrimarySecondary<MOUSE_REGION, MSYS_REGION_ENABLED>(primaryAction, secondaryAction, allEvents, triggerPrimaryOnMouseDown);
```
momoko-h, line 985: MOUSE_CALLBACK for the parameter types?

momoko-h:
The `gfItemPopupRegionCallbackEndFix` based workarounds in Interface_Items.cc seem to be no longer needed?

selaux:
Good catch regarding `gfItemPopupRegionCallbackEndFix`. I removed the related code.

Regarding `MSYS_DefineRegion` I don't have a strong oppinion. From my understanding it was less a check for the specific flag, but for a zeroed out memory region. But I agree that this only tests for basically a single bit, so it can probably be removed.

lynxlynxlynx:
The less code the better and since this was ifdefed out before, I see no reason to keep it.

PR #1683, opened 2022-10-21
Remove the MouseRegion::Base hack

momoko-h:
It is only used once in auto resolve, where it can easily be replaced because the mouse region's user data is not used for anything else.


selaux:
Thanks!

PR #1684, opened 2022-10-21
Allow to choose vanilla version in Android launcher

selaux:
Adds Spinner for the game version to the Android launcher. Unfortunately auto-detecting the game version is not possible yet, as it would require interacting with the Stracciatella library via FFI, which requires some work.

image


lynxlynxlynx:
A fine workaround, thanks.

PR #1688, opened 2022-10-28
Cleanup clang compiler warnings

selaux:
This cleans up some compiler warnings I get from clang:

- Variables that are only assigned to, but never used
- Conflicts between function declaration and function definition
- Comparisons of signed values using `< 0`
- Missing cases in switches


lynxlynxlynx:
Thanks!

PR #1696, opened 2022-11-04
Fix segmentation fault in NumberOfPeopleInSquad()

momoko-h:
Groups are not squads. Regressed by patch from #1608.

I found this while testing the 0.20 RC. It's a must-fix IMO.


selaux:
Sounds and looks good to me.

PR #1697, opened 2022-11-04
Fix two touch control issues

selaux:
This should fix two touch control issues, that still make the game quite hard to play on touch devices.

- Using the item in your hand was not possible, now there is a button for it. Now you can actually bandage your mercs before the end of battle
- Interacting with closed doors was not possible, so you could not get into places

I recommend this still going into `0.20.0`, as it only affects touch controls and enhances playability quite a lot.


lynxlynxlynx:
Looks fine on a cursory inspection, thanks.

PR #1698, opened 2022-11-07
Properly restore two Path AI variables upon return from function

momoko-h:
Several functions store the old values of the global Path AI variables `gubNPCAPBudget` and `gubNPCDistLimit` and restore them upon function exit. However, four of these functions have early returns where the values are not restored.

The first commit fixes the problem, the second commit uses the new class where applicable.

I don't believe this is an urgent problem and who knows if this doesn't somehow cause some regression with the fragile path finding code, so it's probably a good idea to delay this until after the 0.20 release.


lynxlynxlynx:
Coming back to this: it looks fine, thanks.

Issue #1680, opened 2022-10-19
0.20 release checklist


lynxlynxlynx:
Copying the template from docs/

## Pre-release development

- [x] Fix or postpone issues in the [milestone](https://github.com/ja2-stracciatella/ja2-stracciatella/milestone/8)
- [x] Triage open pull requests and recent bugs
- [x] Finalise choice of version number
  - [x] Decide if this version should be 1.0 #443 
- [x] Add new contributors to contributors.txt
- [x] Update changes.md (run docs/draft-changelog.sh to get started)
- [x] Update man page if necessary

## Release candidate

- [x] Update Version number to `-rc` and tag
- [x] Mark the tag as a Github prerelease
- [x] Runtime tests of prerelease packages
  - [x] Linux
    - [ ] Ubuntu 20.04 x64
  - [x] Windows
  - [ ] OS X
  - [x] Android
  
Runtime tests:
 * IMP creation
 * AIM and extending contracts
 * taking Drassen airport or more
 * buying from Bobby Ray's
 * visual inspection of laptop screens

If you want to build the packages manually:

    $ make clean
    $ make build-releases # to build all releases
    $ make build-win-release-on-linux # crosscompile Windows release on Linux
    $ make build-release-on-mac # build the mac bundle
    $ make build-debian-package # build the .DEB package

# Release

- [x] Attach built versions to GitHub release
  - [x] Linux, Windows
  - [x] OS X
  - [x] Android
- [x] Document new features on the website
- [x] Make GitHub prerelase the release, update texts
- [x] Announce
  - [x] Write main announcement
  - [x] Website download page (update frontmatter variables as needed)
  - [x] Website news
  - [x] Bear's pit
  - [x] Moddb

  
## Post-release
- [x] Set version to +1 and add the `-dev` suffix
- [x] Close this bug and GitHub milestone
- [x] Create new milestone if none exists yet

selaux:
All of the items of the milestone are resolved. Any other issues that we should have a look at before the release?

lynxlynxlynx:
I think we're ready to go. There are some assertion bugs open, but without reproducers.

selaux:
I tagged `v0.20.0`. Release candidate packages are here for testing: https://storage.googleapis.com/ja2-builds/index.html#releases/v0.20.0/

lynxlynxlynx:
Did the standard test and all seems fine. :)

Two oddities that I noticed:
[...]

PR #1701, opened 2022-11-11
Fix exiting the map to the south on higher resolutions

selaux:
This is a fix for the issue @lynxlynxlynx discovered in #1680. Exiting south was only possible in low resolutions. This should be fixed now. TBH the scrolling code is still quite incomprehensible to me, so I just increased the padding when the world is considered "scrolled all the way down". As the world actively scrolls down anyways when the exit cursor is shown I think this is an ok solution.


lynxlynxlynx:
Looks good enough, thanks.

PR #1707, opened 2022-11-15
Proposal: WIP: Use type-safe ids

selaux:
This introduces a type-safe ID for Items. It's a struct with a single field, so it should serialize the same as the underlying type, making it usable in save games and such.

**Why?**

Currently we always use the underlying type (in this case `UINT16`) or a type alias. This has the disadvantage that the type checker does not check whether what you are using is actually an itemid and not just another `UINT16`. It also encourages things like iterating over item ids using `++` and strict item id bounds such as `MAX_WEAPONS`, which limits moddablility (what if I want to have more weapons in my mod).

**Continuing the refactor**

This would require other entities such as weapons to use the `TypedId<>` template (see `Types.h`) as well and refactoring the places where `.inner` or `ItemId(x)` is used to use iterators.

**Implications**

Most places using `.inner()` and `ItemId(x)` should be considered technical debt and should be replaced, e.g. by iterators.


lynxlynxlynx:
Wouldn't it be cleaner to use an enum type? No need for an extra class, while still providing the extra strictness and arbitrary size.

momoko-h:
Yes, I also don't quite see the advantage of this approach over simply changing `ITEMDEFINE` to an enum class that inherits from UINT16. Since ITEMDEFINE looks fairly ugly you could also change the name, of course. You would have to change every occurrence of `JAR`, etc. to `ItemId::JAR` or add constexprs for them, but at least that's more idiomatic than `JAR.inner()`.

One other thing I noticed: we really should give OBJECTTYPE a constructor that simply calls CreateItem(). `OBJECTTYPE o; CreateItem(HMX, 100, &o);` is so old school C.

selaux:
I disagree. An enum class implies that there is a known number of known variants. I would argue we currently only have the known number and that also only holds as long as we don't allow adding any additional items in mods. What the actual item id points to is known only after loading the `items.json`. My general standpoint is we shouldn't have any `ITEMDEFINE`s in C++ at all, although that is more of a long term goal :wink:.

lynxlynxlynx:
I wouldn't get stuck on the fact that it's an enum, since it'd clearly be used as a type. Just like nobody would find the underlying ugliness of your class approach, unless they looked specifically. ;P
And we don't actually need to care what underlying values are used for dynamic instances. Nor for the several hardcoded cases if we put some annoying work into it. But these two things are the same for both approaches anyway.

src/game/Strategic/Quest_Debug_System.cc
```diff
@@ -560,7 +560,7 @@ void QuestDebugScreenInit()
 
 	gItemListBox.usItemDisplayedOnTopOfList			= 1;
 	gItemListBox.usStartIndex										= 1;
-	gItemListBox.usMaxArrayIndex								= MAXITEMS;
+	gItemListBox.usMaxArrayIndex								= MAXITEMS.inner();
```
momoko-h, line 563: Do you really want MAXITEMS to be an ItemId? size_t looks like a better choice to me.

src/game/Strategic/Quest_Debug_System.cc
```diff
@@ -1731,7 +1731,7 @@ static void DisplaySelectedItem(void)
 
 		SetFontShadow(NO_SHADOW);
 
-		ST::string ItemName = GCM->getItem(gpActiveListBox->sCurSelectedItem)->getShortName();
+		ST::string ItemName = GCM->getItem(ItemId(gpActiveListBox->sCurSelectedItem))->getShortName();
```
momoko-h, line 1734: What do you think about a getItem overload that still takes an UINT16?

src/game/Strategic/Quest_Debug_System.cc
```diff
@@ -2056,9 +2056,9 @@ static void BtnQuestDebugGiveItemToNPCButtonCallback(GUI_BUTTON* btn, UINT32 rea
 {
 	if (reason & MSYS_CALLBACK_REASON_POINTER_UP)
 	{
-		OBJECTTYPE		Object;
+		OBJECTTYPE		Object = {};
```
momoko-h, line 2059: This isn't necessary because it is the first thing CreateItem does, but as I said, I'd prefer a proper constructor for OBJECTTYPE.

src/game/Tactical/LoadSaveObjectType.cc
```diff
@@ -38,7 +38,7 @@ void ExtractObject(DataReader& d, OBJECTTYPE* const o)
 			EXTR_U8(d, o->ubGunAmmoType)
 			EXTR_U8(d, o->ubGunShotsLeft)
 			EXTR_SKIP(d, 1)
-			EXTR_U16(d, o->usGunAmmoItem)
+			o->usGunAmmoItem = d.read<ItemId>();
```
momoko-h, line 41: You're doing this often enough that adding EXTR_ItemId and INJ_ItemId macros are probably worthwhile

magras:
> An enum class implies that there is a known number of known variants.

I know I am just a passing stranger, but that's not true. There are a bunch of counterexamples from the standard:
1. [`std::byte`](https://eel.is/c++draft/cstddef.syn#lib:byte):
   ```
   enum class byte : unsigned char {};
   ```
2. [`std::align_val_t`](https://eel.is/c++draft/new.syn):
   ```
   enum class align_val_t : size_t {};
   ```
3. [`Handle` example](https://eel.is/c++draft/dcl.init.list#example-7):
   ```
   enum class Handle : uint32_t { Invalid = 0 };
   ```
The last one demonstrates that the committee sees scoped enums as opaque type aliases for integral types with associated constants.

selaux:
Interesting. I might have another look at that then.

PR #1709, opened 2022-11-16
Do not use GetSoldier in AdjustToNextAnimationFrame

momoko-h:
`GetSoldier` returns a `shared_ptr`, which isn't cheap. `AdjustToNextAnimationFrame` can be called hundreds of times per second, but this `shared_ptr` is only used in four of the animation frames.

Usually I am not too concerned about performance of a game that was designed to run on a Pentium 133, but this one is really ugly, we are hammering the memory allocator just to set one `SOLDIERTYPE` member in very few cases.

We could also consider removing `GetSoldier` entirely. The other use cases aren't as bad, but the fact remains that none of them actually needs a shared pointer.

### DIFF
--- a/src/game/Tactical/Soldier_Ani.cc
+++ b/src/game/Tactical/Soldier_Ani.cc
@@ -114,8 +114,6 @@ BOOLEAN AdjustToNextAnimationFrame( SOLDIERTYPE *pSoldier )
 	BOOLEAN bWeaponJammed;
 	UINT16 usUIMovementMode;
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Sure, why not. Were you looking at cachegrind output or how did you come across this?

momoko-h:
No, just stepping through this function while trying to debug something. Since this GetSoldier() call was the first instruction I became curious what it does and why we need it.

Do you think we should remove GetSoldier? It wouldn't be that much work.

PR #1711, opened 2022-11-17
Fix double events for mouse input on touch devices

### LINKED ISSUE #1710: LMB doubleclicks on Android version after introducing touch controls

As title states, left mouse button always doubleclicks, making entering some menus or toggling some options troublesome. I use my tablet with mouse and keyboard, playing the game just like on a PC, but this change broke that, and I don't really want to switch to touch controls. 

I assume it's related to new touch controls. I'm currently using 0.20.0-20221009+10b3516, which is the most recent nightly version I found without this issue, and it doesn't seem to have those new controls either. 0.20.0-20221012+4e06cac already has both new controls and the issue. 

If it's not easy to fix, it would be sufficient to at least add an option to disable new controls, maybe.

VIDEO
0.20.0-20221009+10b3516:
https://youtu.be/lNMc3PrFYew
0.20.0-20221012+4e06cac:
https://youtu.be/RLP86lJ3fQs

|  0.20.0-20221012+4e06cac onwards | GOG  |    no    |    yes     |

selaux:
It seems like there is also touch emulation for mouse events not only the other way around.

It's a bit more complex. There is a hint `SDL_HINT_ANDROID_SEPARATE_MOUSE_AND_TOUCH` for SDL2 `2.0.4` to `2.0.10`, to separate touch and mouse on Android. After `2.0.10` we have `SDL_MOUSE_TOUCHID`. I chose just to cover the second part, as we compile with `2.0.20` anyways for Android. The `#ifdefs` are there for older SDL versions on Linux.

Should close #1710.

### DIFF
--- a/src/sgp/Input.cc
+++ b/src/sgp/Input.cc
@@ -274,6 +274,9 @@ void MouseWheelScroll(const SDL_MouseWheelEvent* WheelEv)
 }
 
 void FingerMove(const SDL_TouchFingerEvent* event) {
+	#ifdef SDL_MOUSE_TOUCHID
+	if (event->touchId == SDL_MOUSE_TOUCHID) return;
+	#endif
 	if (event->fingerId != gMainFingerId) return;
 	gfIsUsingTouch = true;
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks

PR #1713, opened 2022-11-18
Clean up SGPStrings and remove many uses of C string functions

lynxlynxlynx:
Gone since the move to string theory.


momoko-h:
Excellent. `strlcpy` is also not used any more, and I think we should disallow any new uses of these horrible C string functions.

lynxlynxlynx:
On it and it turns out there's more to be cleaned ...

lynxlynxlynx:
Once you start looking at the code ... ><

momoko-h:
This day should be marked on the calendar, conversion of the entire vanilla codebase from C string functions completed?! Just one harmless looking `strcpy` left. Brilliant.

src/game/Editor/Sector_Summary.cc
```diff
@@ -2199,8 +2194,7 @@ static void RegenerateSummaryInfoForAllOutdatedMaps(void)
 	for (sMap.y = 1; sMap.y <= 16; sMap.y++) for(sMap.x = 1; sMap.x <= 16; sMap.x++)
 	{
 		INT32 x = sMap.x - 1, y = sMap.y - 1;
-		char str[40];
-		sprintf(str, "%s", sMap.AsShortString().c_str());
+		const ST::string& str = sMap.AsShortString();
```
momoko-h, line 2197: Without the &, please. The compiler may be able to optimize this away, but in theory you're creating a temporary string and a reference here.

src/game/Tactical/PathAI.cc
```diff
@@ -1735,13 +1730,13 @@ INT32 FindBestPath(SOLDIERTYPE* s, INT16 sDestination, INT8 ubLevel, INT16 usMov
 						SLOGD(zTempString );
```
momoko-h, line 1730: This no longer compiles because SLOGD expects a char * as its first argument. When was the last time anyone used PATHAI_SKIPLIST_DEBUG anyway?

src/sgp/Types.h
```diff
@@ -54,11 +54,6 @@ typedef char            CHAR8;
...
 typedef void *					PTR;
 typedef UINT8						BYTE;
-typedef CHAR8						STRING512[512];
-
-#define SGPFILENAME_LEN 100
-typedef CHAR8 SGPFILENAME[SGPFILENAME_LEN];
-
 
 #ifndef TRUE
 #define TRUE 1
```
momoko-h, line 59: While we have to recompile everything because you're touching Types.h, BAD_INDEX is also unused.

src/sgp/VObject.cc
```diff
@@ -369,12 +364,12 @@ static void DumpVObjectInfoIntoFile(const char* filename, BOOLEAN fAppend)
 	UINT32 uiUniqueID = 0;
 	for (SGPVObject const* i = gpVObjectHead; i; i = i->next_)
 	{
-		char const* const Name = i->name_;
-		char const* const Code = i->code_;
+		char ST::string& Name = i->name_;
+		char ST::string& Code = i->code_;
```
momoko-h, line 368: I doubt this compiles, but `SGP_VIDEO_DEBUGGING` was already broken. As far as I am concerned it should be next on the chopping block.

src/game/Tactical/PathAI.cc
lynxlynxlynx: Doesn't hurt to fix it; unlike the video debugging, this might still be useful.

lynxlynxlynx:
Thanks for the detailed review. :)

For full de-cstring-ification, there are still a handful of plain (f)printf and sscanf calls, and about 12x21 uses of const char*. Full removal likely isn't very viable, but we can get much closer still.

src/sgp/VObject.cc
lynxlynxlynx: Always happy to delete code. 👍🏾

momoko-h:
> For full de-cstring-ification, there are still a handful of plain (f)printf and sscanf calls, and about 12x21 uses of const char*. Full removal likely isn't very viable, but we can get much closer still.

I don't believe const char * has to go entirely, it's perfectly acceptable as a way to pass constant strings around. For example, I believe converting all the strings in the _EnglishText.cc, etc, to ST::string was a mistake, that just creates tons of ST::string objects at startup that will never be used.

src/sgp/VObject.h
```diff
@@ -100,20 +96,10 @@ void InitializeVideoObjectManager(void);
...
-#else
-	SGPVObject* AddStandardVideoObjectFromHImage(SGPImage*);
-	SGPVObject* AddStandardVideoObjectFromFile(const ST::string& ImageFile);
-	#define AddVideoObjectFromHImage(a) AddStandardVideoObjectFromHImage(a)
-	#define AddVideoObjectFromFile(a)   AddStandardVideoObjectFromFile(a)
-#endif
+SGPVObject* AddStandardVideoObjectFromHImage(SGPImage*);
+SGPVObject* AddStandardVideoObjectFromFile(const ST::string& ImageFile);
+#define AddVideoObjectFromHImage(a) AddStandardVideoObjectFromHImage(a)
+#define AddVideoObjectFromFile(a)   AddStandardVideoObjectFromFile(a)
```
momoko-h, line 102: These defines are no longer needed, just remove the Standard from the function names.

momoko-h:
Otherwise this is looking great now. :+1:

lynxlynxlynx:
I disagree about the translations not benefiting, but there's room for improvement for sure. I can't find it on the tracker, however there was already some talk of externalizing them as a start (json, gettext). If we're now preloading all the languages, that's one clear thing to forego.

lynxlynxlynx:
So, any objections to merging?

momoko-h:
No objections from me. Do you want to merge as is or squash your commits first?

lynxlynxlynx:
The history was clean (save tiny 2cbac99), so squashing would do more bad than good.

PR #1715, opened 2022-11-20
Double the maximum number of corpses in a sector to 200

### LINKED ISSUE #1664: Segfault on killing any next enemy, even if you leave and re-enter sector
### What is wrong
Details, details, details.

|  0.20.0 commit fa3273c6bcbcbcdf6f73becc3cf8718253d40ac9 | head  |    yes    |    yes     |
Savegame:

https://easyupload.io/xourr9

Kill any next enemy, results in segfault
No backtrace generated

Screenshot from 2022-10-08 16-35-11

momoko-h:
Fixes #1664. Also cleans up Rotting_Corpses and others a little (dead and duplicated code).

I believe this should give us enough leeway for all but the most extreme cases, after all, we never had any reports about 100 corpses not being enough before #1664. Yes, 1.13 has a limit of 500 but they also have fights of a much larger scale like 30 mercs + 30 militia against 80 enemies.

Attached is a save to demonstrate this patch: there are already 198 corpses in this sector, if you kill the three remaining enemies two will leave corpses, the third won't but the game doesn't crash any more.
[2022-11-20t10-27-29z-198-corpses-3-enemies.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/10050146/2022-11-20t10-27-29z-198-corpses-3-enemies.zip)


lynxlynxlynx:
LGTM. With the change to `TurnSoldierIntoCorpse` a few more lines could be removed — just assign it directly to `pfMadeCorpse`.

ja2-derek:
The temp files now will have 200 slots for corpses. Will this create problems when we load the game in a previous version?

momoko-h:
No, `LoadRottingCorpsesFromTempCorpseFile` uses `AddRottingCorpse` and that function just returns without doing anything if it can't get a free slot. The game will simply be in the same fragile state where the next corpse will crash it, same as with the save attached to #1664.

You can test this with a build of the 0.20 release and the save I provided: all that happens here is 98 messages that the game failed to add a corpse, if you run it with the -debug switch.

PR #1718, opened 2022-12-01
Change function signature of (G|S)etClippingRect

momoko-h:
Passing `SGPRect` by value instead of by pointer so we do not have to assert the pointers are not null.

Not that it matter much but this should be about as efficient as the current method or even a tiny bit better because an `SGPRect` fits into a 64-bit register (2 registers on 32-bit systems).


selaux:
Looking good. Thanks.

PR #1722, opened 2022-12-05
Check indices of MSYS user data functions at compile time

momoko-h:
All of the current callers of MSYS_(G|S)etRegionUserData use constants for the index so it is pointless to check it at runtime.

In theory this prevents us from using variables for the indices in the future, but I don't see that as a real problem:
- nothing is stopping you from accessing the user.data array directly
- now that MOUSE_CALLBACK is defined as a std::function, you have a much more powerful way to pass data to your callbacks by using lambdas.


selaux:
LGTM. Thanks.

PR #1726, opened 2022-12-10
Add an overload for AddVideoObjectFromFile

momoko-h:
A shortcut for AddVideoObjectFromFile(GetMLGFilename()), this also saves a pair of ST::string constructor/destructor calls at the calling location.

Fairly trivial stuff, but it reduces code size at the calling sites by combining four function calls into one.


lynxlynxlynx:
Thanks!

PR #1727, opened 2022-12-12
Upgrade GitHub Action builder image

ja2-derek:
The image we use (ubuntu-18.04) is in the process to be deprecated.

- Announcement: https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/
- Details: https://github.com/actions/runner-images/issues/6002

### DIFF
--- a/.ci/ci-build.sh
+++ b/.ci/ci-build.sh
@@ -88,13 +88,6 @@ if [[ "$RUN_TESTS" == "true" ]]; then
   if [[ "$CI_TARGET" == "linux" ]]; then
     ./AppDir/usr/bin/ja2 -unittests
     ./AppDir/usr/bin/ja2-launcher -help
-
-    # Smoke test to check if the binary can run on older distros
-    DISTRO_IMAGE="debian:10"
-    PACKAGES="libfontconfig1 libx11-6 libsdl2-2.0.0 libfltk1.3 libfltk-images1.3"
-    SETUP_COMMAND="apt-get update && apt-get -yq install $PACKAGES"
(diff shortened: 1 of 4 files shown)


ja2-derek:
We may have a problem here. Simply upgrading the version number will lose us the compatibility with older Linux (e.g. Debian 10). The `ubuntu:18.04` image will still be available until April 2023, so we still have some time to decide. 

I can think of a few options:

1. Build inside a docker container (e.g.  `debian:10` or CentOS/Rocky). So our build should remain good as long as the `debian:10` is still on DockerHub. Not sure if this impacts build time.
2. Build with newer ubuntu but linking old libc. We may need to install older version of glibc and all toolchains, and have `cmake` link to the correct version. We already explicitly use older gcc on CI builders.
3. Statically link glibc and bundle it with the executable. This might result in larger release packages (and maybe also marginally longer start time?).
4. Disregard older Linux releases. `ubuntu-21.10` comes with glibc 2.31. Running on Debian 10 gives you ``version `GLIBC_2.29' not found``. We might want to document the libc compatibility somewhere. We can keep our script and build simple.

I would like to keep the compatibility, because I believe some of us plays it on old machines.  Let me know your thoughts.

lynxlynxlynx:
1. overkill.
2. perhaps, not sure if worth the effort.
3. is impossible, there's no way to include the whole glibc.
4. I like this the most. Users on older platforms can build it themselves.

lynxlynxlynx:
LGTM now. :)

PR #1728, opened 2022-12-12
Remove SoldierSP and GetSoldier

momoko-h:
Continuation of #1709: allocating a `shared_ptr` just to call one simple method is really inefficient. These pointers are never shared.


lynxlynxlynx:
Thanks!

PR #1729, opened 2022-12-13
Remove unnecessary ST::string.c_str calls

momoko-h:
• Several are no longer needed since PR #1713
• `JsonObject::AddMember` has a matching overload since PR #1045
• `IoException` always had ST::string as its parameter

This saves some ST string ➟ C string ➟ ST string round trips, which is just a convoluted way to create string copies that no one needs.


lynxlynxlynx:
Nice, thanks!

Issue #1714, opened 2022-11-20
Missing sound of RPK-74 in burst mode


clemenskaa:
Howdy,

I am playing JA 2 (English Steam version) with Stracciatella 0.20.0 on Windows 10 and I noticed a small audio bug:

The RPK-74 has no burst sound (single shot works). The strange thing is, the AK-74 works just fine, even though both guns share the same ammo type. And as far as I am aware, the sound depends on the ammo type (at least the externalized data looks that way).

I played around with the weapons.json and realized that the value "ubShotsPerBurst": 5 makes a difference: 3 and 4 shots per burst work fine, but a soon as I use 5 (the default value for the RPK-74) the sound is no longer played.

I played unmodded JA 2 in the day and cannot remember any missing sounds of this gun, but I am not entirely sure. Also I am unfortunately not able to tell of this was a problem in a previous version of Stracciatella.

Please find attached a save game. Buzz has both - the RPK-74 and the AK-74 - in her inventory.

[2022-11-20t15-01-30z-ak-and-rpk.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/10049995/2022-11-20t15-01-30z-ak-and-rpk.zip)

Thank you and keep up the good work!
Clemens

lynxlynxlynx:
Thanks for the excellent report!

selaux:
Looked into it quickly: It should indeed use the sound from the calibre, but there is no sound for 5 shots for the `AMMO545` calibre in the original data files. Only for 4 shots (`5,45 BURST 4.WAV`). Before externalizing weapon sounds the value was `BURSTTYPE1.WAV`. That is probably the sound we want, so our data files are wrong here.

Danf96:
Tried fixing the bug by adding an extra key for the RPK-74, but the game crashes when spawning the RPK in with `[ERROR] sgp/SGP.cc: Game has been terminated due to an unrecoverable error: Found no default ammo for gun` as well as crashing when right clicking the RPK in your inventory.

PR #1730, opened 2022-12-16
Fix RPK burst sound

Danf96:
RPK was looking for a burst sound file that didn't exist so bursttype1 is now used for the RPK while other 5.45 weapons continue to use the sounds defined in calibres.json. Fixes the problem mentioned in #1714


selaux:
Awesome. Thanks!

PR #1732, opened 2022-12-18
Use more automatic memory manangement in ContentManager

momoko-h:
• Use smart pointers for ContentManager members.
    • deleteElements templates no longer copy the map or vector
      they're deleting first.
    • Store ST::strings in vectors, not pointers to them.
    • Remove useless clear() calls of vectors that are about to be destructed.
    • DefaultContentManager::readJsonDataFile is already public, no need to
      "expose" anything in DefaultContentManagerUT


lynxlynxlynx:
Thanks, nice cleanup.

PR #1734, opened 2022-12-21
Externalize Flo's weapon deals bonus

momoko-h:
Add a new field to mercs-profile-info.json which lets users specify a modifier when trading with arms dealers.

Always hiring Flo just to get more money from Tony is boring, so now others can get a bonus (or a handicap), too. Here is an overview of what's possible:

| Modifier | Buying a 100% dart gun | Selling a 100% MAC-10 |
| :---         |    ---:      |          ---: |
| 100  | 625 | 877 |
| 110 (Flo) | 563  | 964  |
| 10 (Minimum) | 1187 | 88 |
| 180 (Maximum) | 125 | 1578 |

I've never worked with the yaml schema files before, so it'd be good to have a closer look at these changes.

### DIFF
--- a/assets/externalized/mercs-profile-info.json
+++ b/assets/externalized/mercs-profile-info.json
@@ -13,7 +13,12 @@
  *        - "RPC": Recruitable characters - NPCs that can be recruited
  *        - "NPC": Non-playing characters, that can never be controlled by the player
  *        - "NOT_USED": Reserved slots that are not used anywhere
- */        
+ *   - weaponSaleModifier: what prices does this character get when buying or selling weapons?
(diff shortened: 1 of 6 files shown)


lynxlynxlynx:
Why a max of 180%? Seems oddly specific. I'd put it at 300, 400, 500, if anyone ever wants to simulate a high-scarcity economy. Or at least 250 if you want to keep it within a byte.

momoko-h:
The problem with values >= 200 is that you'd be buying everything for free (well, 1$). If you take a look at the table it already seems excessive to  be able to buy something at one fifth of the normal price.

lynxlynxlynx:
I don't see a problem with that, but fine.

PR #1736, opened 2022-12-22
Add two missing usActionData assignments

momoko-h:
DecideActionRed had two code paths where it would return AI_ACTION_CHANGE_STANCE without setting the new stance in usActionData.

I found this rather by accident [here](https://github.com/hb029/JA-2-Stracciatella/commit/00936bcfa680f48fe1e0316b987edc4151892af0). The author claims this can cause an AI deadlock, if that's true then this is a quite important fix and it would have been nice if he had let us know about it.


misanthropos (approved):
Looks good to me.

PR #1737, opened 2022-12-22
enhancement: Quit the Game from Main Screen on CTRL-Q / CTRL-ESC

misanthropos:
Signed-off-by: Richard Fröhning <misanthropos@gmx.de>

### DIFF
--- a/src/game/MainMenuScreen.cc
+++ b/src/game/MainMenuScreen.cc
@@ -261,9 +261,8 @@ static void HandleMainMenuInput(void)
 		{
 			switch (InputEvent.usParam)
 			{
-/*
+				case 'q': if (_KeyDown(CTRL)) gbHandledMainMenu = QUIT; break;					
 				case SDLK_ESCAPE: gbHandledMainMenu = QUIT; break;
-*/
 
 				case 'c':
 					if (_KeyDown(ALT)) gfLoadGameUponEntry = TRUE;
--- a/src/sgp/Input.cc
+++ b/src/sgp/Input.cc
@@ -586,12 +586,9 @@ void SimulateMouseMovement( UINT32 uiNewXPos, UINT32 uiNewYPos )
 
 void DequeueAllKeyBoardEvents(void)
 {
-#if 1 // XXX TODO
-	FIXME
-#else
 	//dequeue all the events waiting in the windows queue
-	MSG KeyMessage;
-	while (PeekMessage(&KeyMessage, ghWindow, WM_KEYFIRST, WM_KEYLAST, PM_REMOVE));
+	//	MSG KeyMessage;
+	//while (PeekMessage(&KeyMessage, ghWindow, WM_KEYFIRST, WM_KEYLAST, PM_REMOVE));
 
 	//Deque all the events waiting in the SGP queue
 	InputAtom InputEvent;
@@ -599,7 +596,6 @@ void DequeueAllKeyBoardEvents(void)
 	{
 		//dont do anything
 	}
-#endif
 }
 
 


ja2-derek:
Looking back in git-blame, it has been disabled since the initial import 18 years ago.

I have a habit of tapping `Esc` repeatedly on game launch, to skip the SirTech logo animation intro. I will for sure quit the game unexpectedly.

lynxlynxlynx:
We should also have a skip_intro_videos settings key, so we don't even try loading them, obviating the need for skipping.

To prevent accidental escaping on entry, we can use a timer and only process esc in the main menu if at least, let's say, 2 seconds have passed since it was loaded.

misanthropos:
> We should also have a skip_intro_videos settings key, so we don't even try loading them, obviating the need for skipping.
> 
https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1738

momoko-h:
How about changing that line to
 ```case SDLK_ESCAPE: if (GetJA2Clock() >= 30000) gbHandledMainMenu = QUIT; break;```

Then the ESC key would be enabled 30 seconds after the game started.

misanthropos:
> How about changing that line to `case SDLK_ESCAPE: if (GetJA2Clock() >= 30000) gbHandledMainMenu = QUIT; break;`
> 
> Then the ESC key would be enabled 30 seconds after the game started.

I changed it to CTRL-ESC as a compromise.

lynxlynxlynx:
Nobody will think of ctrl-esc.

misanthropos:
> Nobody will think of ctrl-esc.

Agreed - I added CTRL-Q in addition.

lynxlynxlynx:
I don't think that's any better; anything hidden behind a modifier just won't come up.

misanthropos:
> I don't think that's any better; anything hidden behind a modifier just won't come up.

To me CTRL-q comes naturally. Another idea could be to empty the input event queue before leaving intro and going back to ESC.

momoko-h (changes requested):
Some bits of nitpicking:

- there is no need to keep these remnants of the game's Windows-only past around
- the function name and what it actually does do not match
- The other use of this function in `LoadWorldFromSGPFile` should probably be removed. If we did not need to drop events there before we won't need it now either.

misanthropos:
> 
>     * there is no need to keep these remnants of the game's Windows-only past around

which remnants?

> 
>     * the function name and what it actually does do not match

can you please be more specific?

> 
>     * The other use of this function in `LoadWorldFromSGPFile` should probably be removed. If we did not need to drop events there before we won't need it now either.

This would be a cleanup for something else?

misanthropos:
damn i just committed to master. If thats is ok I wont revert it.

momoko-h:
> > ```
> > * there is no need to keep these remnants of the game's Windows-only past around
> > ```
> 
> which remnants?

The part you commented out, the MSG definition and the  PeekMessage call. Those are parts of the Windows API.
> > * the function name and what it actually does do not match
> 
> can you please be more specific? So I should rename it to DequeueAllEvents?

Yes.
> 
> > ```
> > * The other use of this function in `LoadWorldFromSGPFile` should probably be removed. If we did not need to drop events there before we won't need it now either.
> > ```
> 
> This would be a cleanup for something else?

A new PR just to remove one line?

misanthropos:
Does it hurt to leave it?

momoko-h:
Loading a game is so fast nowadays that there won't be many events to dequeue anyway. And if someone knows beforehand that they want to select a certain merc or squad and presses the key for it while the loading screen is still up, why shouldn't they be able to do that? It has worked until now. 

I just dislike it in general when software ignores my input.

misanthropos:
Since it seemed to have had no side effects until reactivating DequeueAllEvents I don´t object on removing that from LoadWorldFromSGPFile - But it was the original intention to not do that (press keys before game is ready).

PR #1738, opened 2022-12-22
[TEST] reduce intro splash duration

misanthropos:
picking 8fcc33db1d4d836c2cf2b70e993decc10be5d398 from maxrd2

### DIFF
--- a/src/game/MainMenuScreen.cc
+++ b/src/game/MainMenuScreen.cc
@@ -84,7 +84,7 @@ static void RestoreButtonBackGrounds(void);
 
 ScreenID MainMenuScreenHandle(void)
 {
-	if (guiSplashStartTime + 4000 > GetJA2Clock())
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
LGTM!

src/game/MainMenuScreen.cc
```diff
@@ -84,7 +84,7 @@ static void RestoreButtonBackGrounds(void);
 
 ScreenID MainMenuScreenHandle(void)
 {
-	if (guiSplashStartTime + 4000 > GetJA2Clock())
+	if (guiSplashStartTime + INTRO_SPLASH_DURATION > GetJA2Clock())
```
momoko-h, line 87: Isn't this redundant now? `guiSplashStartTime` is set before `splashDisplayedMoment` (it is set in `InitJA2SplashScreen` almost immediately after `InitializeJA2Clock`, so it should always be 0 or maybe 1 on slower machines).

PR #1742, opened 2022-12-27
Address problems when compiling with GCC 13

momoko-h:
• string_theory uses uint64_t but does not include stdint.h
• dangling references to rapidjson arrays
• incorrect initialization order in MercProfileInfo
• warning about possible OOB array access in switchHeadGear (not really
  possible, but this code would not be safe if we'd use multithreading).

### DIFF
--- a/src/externalized/AmmoTypeModel.cc
+++ b/src/externalized/AmmoTypeModel.cc
@@ -1,11 +1,11 @@
 #include "AmmoTypeModel.h"
 #include "JsonObject.h"
+#include <stdint.h>
 #include <stdexcept>
-#include <string_theory/format>
 #include <utility>
+#include <string_theory/format>
 
-AmmoTypeModel::AmmoTypeModel(uint16_t index_,
-				ST::string internalName_)
+AmmoTypeModel::AmmoTypeModel(uint16_t index_, ST::string && internalName_)
 	:index(index_), internalName(std::move(internalName_))
 {
 }
@@ -21,8 +21,7 @@ void AmmoTypeModel::serializeTo(JsonObject &obj) const
 AmmoTypeModel* AmmoTypeModel::deserialize(JsonObjectReader &obj)
 {
 	int index = obj.GetInt("index");
-	ST::string internalName = obj.GetString("internalName");
-	return new AmmoTypeModel(index, internalName);
+	return new AmmoTypeModel(index, obj.GetString("internalName"));
 }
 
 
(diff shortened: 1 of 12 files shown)


src/externalized/AmmoTypeModel.cc
```diff
@@ -1,11 +1,11 @@
 #include "AmmoTypeModel.h"
 #include "JsonObject.h"
+#include <stdint.h>
```
lynxlynxlynx, line 3: Why not cstdint?

src/externalized/AmmoTypeModel.cc
```diff
@@ -21,8 +21,7 @@ void AmmoTypeModel::serializeTo(JsonObject &obj) const
 AmmoTypeModel* AmmoTypeModel::deserialize(JsonObjectReader &obj)
 {
 	int index = obj.GetInt("index");
-	ST::string internalName = obj.GetString("internalName");
-	return new AmmoTypeModel(index, internalName);
+	return new AmmoTypeModel(index, obj.GetString("internalName"));
```
lynxlynxlynx, line 24: There's no need to explicitly use std::move here?

src/externalized/AmmoTypeModel.cc
momoko-h: Because the C++ standard only requires cstdint to put these names in the std namespace; putting them in the global namespace  (where string_theory needs them) is optional.

src/externalized/AmmoTypeModel.cc
momoko-h: I thought so too, but [Compiler Explorer](https://godbolt.org/z/539Yccqn3) proved me wrong. 84 lines with the move, 106 without it. C++ move semantics are complicated, that's for sure.

momoko-h: Oh, I see, I misunderstood your question. The std::move has to be in the member initialization list. Compiler Explorer does not show any difference between the constructor calls F(c) and F(std::move(c)).

lynxlynxlynx: Ah right, it already is.

lynxlynxlynx:
Thanks!

Danf96:
Not sure why, but I was still having issue compiling with GCC 13 unless using the precompiled headers.

momoko-h:
That's probably fallout from issue #1884 then. I forgot to test this without PCHs and the GitHub runners use older compiler versions where the standard headers weren't cleaned up yet.

PR #1744, opened 2022-12-28
Restore the "Interrupt for ..." message from vanilla

momoko-h:
The function call for this message was probably unintentionally removed in a previous commit. Right now, we display the list only when there are more than six guys interrupting, so almost never.

[(This commit, if you're curious.)](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/706477d84f95ab26095f7c022c488c9f7142de1b#diff-683714cce60d534b535dfdc2addbcb887374d604f05f72ccfa6a9f325e37af05L568)


lynxlynxlynx:
Thanks!

Issue #45, opened 2013-04-06
Crashing in tactical Combat mode while scrolling in fullscreen (Mac)


lynxlynxlynx:
Originally reported by: **Anonymous**

---

While in fullscreen it is impossible to scroll through the tactical combat mode by either arrow keys or mouse movement. No problems in window mode.

ja2 always crashes with the following output in shell and the logging output back to the end:

```
#!shell
Game version: ENGLISH
Configuration file:            '/Users/matze/.ja2/ja2.ini'
Root game resources directory: '/Users/matze/ja2/JaggedAlliance2'
Data directory:                '/Users/matze/ja2/JaggedAlliance2/Data'
Tilecache directory:           '/Users/matze/ja2/JaggedAlliance2/Data/Tilecache'
------------------------------------------------------------------------------
2013-04-06 12:13:43.492 ja2[13004:707] Loading Maximizer into bundle: (null)
2013-04-06 12:14:11.682 ja2[13004:707] Maximizer: Unsupported window created of class: SDL_QuartzWindow

```

```
#!log

Process:         ja2 [12970]
Path:            /Users/USER/Downloads/*/ja2
Identifier:      ja2
Version:         ??? (???)
Code Type:       X86 (Native)
Parent Process:  bash [12968]

Date/Time:       2013-04-06 12:11:58.076 +0200
OS Version:      Mac OS X 10.7.5 (11G63b)
Report Version:  9

Interval Since Last Report:          3312749 sec
Crashes Since Last Report:           91
Per-App Crashes Since Last Report:   3
Anonymous UUID:                      8866E1E1-F76F-4E55-B22E-67C956A266D0

Crashed Thread:  1  Dispatch queue: com.apple.libdispatch-manager

Exception Type:  EXC_CRASH (SIGBUS)
Exception Codes: 0x0000000000000000, 0x0000000000000000

Application Specific Information:
objc[12970]: garbage collection is OFF

Thread 0:: Dispatch queue: com.apple.main-thread
0   libsystem_c.dylib               0x904f8a97 memmove$VARIANT$sse42 + 218
1   ja2                             0x00347a21 Blt16BPPTo16BPP(unsigned short*, unsigned int, unsigned short*, unsigned int, int, int, int, int, unsigned int, unsigned int) + 177
2   ja2                             0x002e1181 RestoreShiftedVideoOverlays(short, short) + 929
[...]

PR #1748, opened 2023-01-03
Remove SGPVSurfaceAuto

momoko-h:
The only difference between SGPVSurfaceAuto and SGPVSurface is that the former frees the underlying SDL_Surface upon destruction while the latter does not. There is only one surface where we do not want this automatic free to happen: the screen buffer.

This behaviour was added back in 2013 to fix issue #45: crashes on MacOS. So why do we not need this special behaviour any more? Back then the wrapped surface for the screen buffer was obtained by a call to the SDL 1.2 function SDL_SetVideoMode, and that surface must not be freed by the application, similar how the surface obtained by a call to SDL_GetWindowSurface in SDL 2.0 must not be freed.

Nowadays Stracciatella uses CreateRGBSurface for the screen buffer, so the need for SGPVSurfaceAuto simply is not there any more. SGPVSurface can always free its surface.

### DIFF
--- a/src/game/GameScreen.cc
+++ b/src/game/GameScreen.cc
@@ -106,7 +106,8 @@ static void BlitMFont(VIDEO_OVERLAY* const ovr)
 void MainGameScreenInit(void)
 {
 	// all blit functions expect z-buffer pitch to match framebuffer pitch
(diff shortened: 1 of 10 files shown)


lynxlynxlynx:
Looks fine, thanks!

PR #1750, opened 2023-01-06
Some housekeeping

momoko-h:
Avoid unnecessary construction of SGPSector objects
    • Town_Militia: sector is the result of pSoldier->sSector.AsByte()
    * Strategic_AI: pGroup->ubSector is already a SGPSector

    • Remove two unused function in Video.cc
    • Reseed the existing mt19937 engine instead of creating a new one
    • Move one variable declaration into the include guard in Random.h
    • Remove an unnecessary assignment in SoundMan (CID 215563)
    • Do not unlock a unique_lock that is about to go out of scope

Just some assorted stuff I came across.


selaux:
Thanks!

PR #1753, opened 2023-01-12
Remove a check for profile ID 53 in HandleSuppressionFire

momoko-h:
It is true that drunk Larry does not have a quote for being under fire, however the dialogue system does not expect every merc to have a full set of quotes and handles this situation just fine. All this check does now is prevent modders from adding the missing quote.


lynxlynxlynx:
Thanks!

Issue #1751, opened 2023-01-08
VerboseAssert is too costly


momoko-h:
Some number first. These were produced by starting the program, loading a save which goes directly into tactical, let the game sit idle for 3 minutes, then ALT + F4.

Current situation:
> total heap usage: 3,171,924 allocs, 3,168,351 frees, 331,719,674 bytes allocated

With Assert and AssertMsg neutered (`void(0)`)
> total heap usage: 1,256,880 allocs, 1,253,272 frees, 238,810,283 bytes allocated

Or in graph form, for these I started a new game and played until I left the rebel basement with Ira:

Currently:
![graph_current](https://user-images.githubusercontent.com/25463312/211189494-74f48185-a3a8-4873-867d-a23e88aabf7d.png)

Patched:
![graph_patched](https://user-images.githubusercontent.com/25463312/211189537-19f90f66-e803-4099-a371-fcc4e08b8b01.png)

The runtime cost of asserts should be as close to zero as possible. VerboseAssert currently does not meet this criteria: two of its parameters are strings, so each time we come across an Assert or AssertMsg we must construct two ST::strings, sometimes even through ST::format. The second string is the filename, which is always too long for short string optimization, so we must always allocate these.

The very least we should do is change the type of the second string (the filename) to `char const *`, but we should also consider to nuke AssertMsg:

1. For our users, it doesn't really make a difference if it just says 'assertion failed, file X line Y' or if there's also some nerd speak about 'BlaBlaBla out of range'.
2. For us, it's debugging time anyway, and a break point will give us all that information and much more.
3. We should also check each instance of AssertMsg if it is really the right tool, or if SLOGE or an exception wouldn't be better.

PR #1754, opened 2023-01-12
Reduce the cost of Assert and AssertMsg

momoko-h:
• Do not construct strings for the filename prematurely • Do not create a new instance of an empty string every time an Assert
  line is reached, use just one global instance instead.
• Use Assert instead of AssertMsg for null pointer checks, the message
  does not add anything useful.

I repeated my test from #1751 with this patch and this saved ~1.9 million allocations in four minutes, or ~7800 allocations every second.

### DIFF
--- a/src/game/Utils/Slider.cc
+++ b/src/game/Utils/Slider.cc
@@ -71,7 +71,7 @@ void InitSlider(void)
 
 void ShutDownSlider(void)
 {
-	AssertMsg(guiSliderBoxImage != NULL, "Trying to ShutDown the Slider System when it was never inited");
+	Assert(guiSliderBoxImage); // Trying to ShutDown the Slider System when it was never inited?
 
 	//Do a cehck to see if there are still active nodes
 	for (SLIDER* i = pSliderHead; i != NULL;)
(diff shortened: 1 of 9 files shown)


src/game/Utils/Slider.cc
```diff
@@ -96,7 +96,7 @@ static void SelectedSliderMovementCallBack(MOUSE_REGION* r, UINT32 reason);
 
 SLIDER* AddSlider(UINT8 ubStyle, UINT16 usCursor, UINT16 usPosX, UINT16 usPosY, UINT16 usWidth, UINT16 usNumberOfIncrements, INT8 sPriority, SLIDER_CHANGE_CALLBACK SliderChangeCallback)
 {
-	AssertMsg(guiSliderBoxImage != NULL, "Trying to Add a Slider Bar when the Slider System was never inited");
+	if (!guiSliderBoxImage) InitSlider();
```
lynxlynxlynx, line 99: This could still fail, but then we'd have other problems anyway. Not sure if coverity would complain.

lynxlynxlynx:
LGTM!

PR #1755, opened 2023-01-15
Do not update the entire screen texture every frame

momoko-h:
Before this commit, RefreshScreen() used SDL_UpdateTexture to blit the entire screen buffer surface to the screen texture every frame. Since this function has to convert the screen texture format to the actual output format of our window, this is extremely costly. The game spent well above 50% of its time under this function.

With this commit, only the union of all screen buffer regions that actually got modified are updated. In the most extreme cases (Load Screen and others) this union can be smaller than 1% of the screen area.

Blit_RGB565_32 is by far the heaviest function in the entire program, it usually accounts for more than 60% of all CPU usage on my system. Here is a graph showing how this very simple PR affects this function:
UpdateTexturePerfData

The performance data for the Laptop screen with this PR is not missing, the number is simply too small to show in this graph: 2,532 IR per call, compared to 2,923,232 IR per call with the current master branch.

### DIFF
--- a/src/sgp/Video.cc
+++ b/src/sgp/Video.cc
@@ -544,6 +544,11 @@ void RefreshScreen(void)
 
 	const BOOLEAN scrolling = (gsScrollXIncrement != 0 || gsScrollYIncrement != 0);
 
+	// This variable will hold the union of all modified regions.
+	struct rect : SDL_Rect {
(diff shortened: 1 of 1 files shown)


selaux:
Awesome improvement! LGTM. Thank you.

PR #1759, opened 2023-01-18
Prevent mercs from swapping places when that make one of them unhittable

momoko-h:
When a prone merc M1 swapped places with a not prone merc and M1's new grid and position didn't allow him to be prone there, he became unhittable.

This is not what is causing the enemies to become unhittable because the AI does not exchange places, but the added warning in `SetSoldierGridNo` should make it easier to figure out who's calling that function without proper checks.


lynxlynxlynx:
Thanks for both!

PR #1761, opened 2023-01-19
Remove SDL_RenderClear

momoko-h:
From the SDL 2 documentation:

SDL_RenderClear
Clear the current rendering target with the drawing color.

This means our use of this function is completely useless because our next next step is to fill the rendering target with the screen texture.

It's not a huge deal because `SDL_RenderPresent` is smart enough to figure out that it can just drop this request, so this removes just the little bit of overhead required to add and remove something from SDL's rendering queue.


selaux:
Thanks.

PR #1762, opened 2023-01-19
Upgrade Rust toolchain to 1.59 and Rust edition to 2021

selaux:
As discussed in Discord, the latest version of Rust currently installable on Debian Stable (the oldest system we kind of support) is `1.59` with the [rustc-mozilla](https://packages.debian.org/buster/rustc-mozilla) package. So we should be safe to upgrade to that version, which this PR does.

- Upgrades to Rust 1.59
- Fixes all Clippy lints and formatting issues
- Changes the used [Rust edition](https://doc.rust-lang.org/edition-guide/editions/index.html) to `2021` (which resulted in no changes, to code)


selaux:
I had to extract the rust formatting and linting checks into a different github actions job as there were incompatibilities with `sccache` and the new rust version. I think its better to do it in a separate job anyways, as the feedback cycle is way lower compared to running them at the end of the build step.

selaux:
No complaints, so I will merge.

PR #1768, opened 2023-01-22
Fix visual artifacts when resizing the window or toggling fullscreen

momoko-h:
I misunderstood the SDL documentation, `RenderCopy` only stretches the texture to fill the viewport, while `RenderClear` clears the entire window area. Fortunately it is still not necessary to clear everything every frame.


selaux:
Thanks.

PR #1769, opened 2023-01-22
Proposal: Transition away from SGP::Buffer

momoko-h:
`SGP::Buffer` is essentially just a pre C++11 version of unique_ptr's array variant. It is less functional and certainly far less known than its modern successor.

This PR demonstrates a practical benefit of this approach: `SGPImages`  are short-lived (often unnamed and temporary) objects that are only created as a step in the creation of other things ike `SGPVObject`. It is therefore safe to move data away from `SGPImage` instead of copying it.


selaux:
Sounds good to me.

momoko-h:
Are there any objections against merging this? I think it's a nice little cleanup, less and faster code.

PR #1770, opened 2023-01-27
Undo damage from PR #1759

momoko-h:
Not setting `sGridNo` causes problems when NPCs leave a sector, for example when Fatima moves from A9 to A10.

Make the new warning appear less often, we only need it figure out the problem with the enemy AI.

I should have known better than to blindly trust code comments, this causes several error messages and OOB accesses in current builds.


lynxlynxlynx:
Tricky, thanks!

PR #1771, opened 2023-01-30
Give CheckForGunJam and OKFireWeapon a proper return type

momoko-h:
These function used a BOOLEAN to express four different possible results. This commit also uses the new return type for `FireWeapon`, even though nobody is currently using the returned value. This adds a fifth possible result to the enum.

More code comments that contradict the actual code, lovely.

### DIFF
--- a/src/game/Tactical/Soldier_Ani.cc
+++ b/src/game/Tactical/Soldier_Ani.cc
@@ -657,8 +657,8 @@ BOOLEAN AdjustToNextAnimationFrame( SOLDIERTYPE *pSoldier )
 					else if (pSoldier->bDoBurst == 1)
 					{
 						// CHECK FOR GUN JAM
-						if (auto const weaponJammed = CheckForGunJam(pSoldier);
-						    weaponJammed == FireWeaponResult::JAMMED)
+						auto const weaponJammed = CheckForGunJam(pSoldier);
+						if (weaponJammed == FireWeaponResult::JAMMED)
 						{
 							fStop = TRUE;
 							// stop shooting!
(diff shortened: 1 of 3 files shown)


src/game/Tactical/Soldier_Ani.cc
```diff
@@ -659,8 +657,8 @@ BOOLEAN AdjustToNextAnimationFrame( SOLDIERTYPE *pSoldier )
 					else if (pSoldier->bDoBurst == 1)
 					{
 						// CHECK FOR GUN JAM
-						bWeaponJammed = CheckForGunJam( pSoldier );
-						if ( bWeaponJammed == TRUE )
+						if (auto const weaponJammed = CheckForGunJam(pSoldier);
+						    weaponJammed == FireWeaponResult::JAMMED)
```
lynxlynxlynx, line 661: Bljak is this ugly, please leave the first statement out of the test.

momoko-h: Really? I kind of like it because it limits the scope of the variable to exactly the if statement.

lynxlynxlynx: It's worse for readability and bug prone. So much in fact there are several standards warning against it (copying from Sonar):

- MISRA C:2004, 13.1 - Assignment operators shall not be used in expressions that yield a Boolean value
- MISRA C++:2008, 6-2-1 - Assignment operators shall not be used in sub-expressions
- MISRA C:2012, 13.4 - The result of an assignment operator should not be used
- [MITRE, CWE-481](https://cwe.mitre.org/data/definitions/481) - Assigning instead of Comparing
- [CERT, EXP45-C.](https://wiki.sei.cmu.edu/confluence/x/ZNYxBQ) - Do not perform assignments in selection statements
- [CERT, EXP51-J.](https://wiki.sei.cmu.edu/confluence/x/ITZGBQ) - Do not perform assignments in conditional expressions

lynxlynxlynx: Thanks.

PR #1772, opened 2023-01-30
Update rust dependencies to most recent versions

selaux:
NB: Seems that the hash for string theory has changed, so I changed it in the PR as well.


Exarkuniv:
glad someone came across this also, i thought i was going crazy.

selaux:
Seems like the change will be reverted for now: https://github.blog/changelog/2023-01-30-git-archive-checksums-may-change/

PR #1773, opened 2023-01-31
Replace rapidjson with rust based implementation

selaux:
This replaces the `rapidjson` dependency with a wrapper around the `serde-json` rust parser and `jsonschema` module.

**Rationale**: `rapidjson` does not get any releases anymore and it seems like there is no intention to do so. This causes issues with compiler warnings and such all the time. Additionally there might be bugs, which linger around in rapidjson unnoticed. Additionally with [jsonschema-rs](https://github.com/Stranger6667/jsonschema-rs) we can validate using newer json schema drafts. It also unifies json parsing, as some things (`ja2.json`) were parsed with serde already. 

Performance wise it is worse than rapidjson tbh, as all json values are owned, so e.g. when accessing an object property it will be cloned. I have not tested memory usage yet, but probably it's worse as well. On my machine this boils down to:

- 8ms for loading all JSON with rapidjson
- 16ms for loading all JSON with serde

jsonschema-rs also validates the schema itself using json schema, which is how I found an issue in our json schemas as well (see `translation-tables.schema.yaml`).


momoko-h:
A somewhat easier alternative would be to do what several Linux distributions are already doing, ignore "official" rapidjson releases and just pick a snapshot from their master branch. But I don't mind moving away from rapidjson altogether, it's not even state of the art in C++ any more, where [simdjson](https://github.com/simdjson/simdjson) seems to be the premier way to handle json nowadays.

Could you show a comparison what valgrind reports for number of allocations and memory used until the program reaches the splash screen? I assume memory usage would only be increased during startup, which isn't that important because the game will need to allocate much more later anyway.

lynxlynxlynx:
So I take it serde supports comments?

selaux:
Not serde itself, but we already have code for that to support it in `ja2.json`: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/stracciatella/src/json.rs

src/externalized/json/Json.cc
```diff
@@ -0,0 +1,207 @@
...
+std::vector<JsonValue> JsonValue::toVec() const {
+	RustPointer<RJsonArray> array(RJsonValue_toArray(m_value.get()));
+	throwRustError(!array);
+	auto length = RJsonArray_length(array.get());
+	std::vector<JsonValue> vec;
+	vec.reserve(length);
+	for (size_t i = 0; i < length; i++) {
+		RustPointer<RJsonValue> val(RJsonArray_get(array.get(), i));
+		throwRustError(!val);
+		vec.push_back(val.release());
```
momoko-h, line 18: Like in `JsonObject::keys()`, this should be `val.get()` or you'd leak the `RJsonValue` if `push_back` throws. This could also use `emplace_back`.

src/externalized/json/Json.h
```diff
@@ -0,0 +1,99 @@
...
+
+#include <string_theory/string>
+#include <vector>
+
+class JsonObject;
+
+class JsonValue {
+	public:
+		// Takes ownership of RJsonValue
+		JsonValue(RJsonValue* value) {
```
momoko-h, line 13: You should not just write a comment about it but express it in code: `JsonValue(std::unique_ptr<RJsonvalue> && value)`.

src/externalized/json/Json.h
```diff
@@ -0,0 +1,99 @@
...
+			m_value.reset(value);
+		}
+		JsonValue(int value) {
+			m_value.reset(RJsonValue_fromInt(value));
```
momoko-h, line 17: I would prefer it if the constructors used the member initialization list to set m_value. The optimizer may be able to handle this but it's better not to rely on it too much.

src/externalized/tactical/MapItemReplacementModel.cc
```diff
@@ -1,27 +1,28 @@
...
-	}
-	if (val.IsInt())
-	{
-		return items->getItem(val.GetInt());
-	}
+	try {
+		auto str = val.toString();
+		return items->getItemByName(str);
+	} catch (...) {}
+		try {
```
momoko-h, line 9: Either the indentation is weird here or this does not do what you meant to do.

src/sgp/SGP.cc
```diff
@@ -506,15 +436,15 @@ int main(int argc, char* argv[])
...
-	} catch (...) {
-		try {
-			TerminationHandler();
-		} catch (...) {
-			// If you ever see return code 27, try to set a breakpoint here
-			return 27;
-		}
-		return EXIT_FAILURE;
-	}
+	// } catch (...) {
```
momoko-h, line 439: Is this a remnant from the development phase or do you really want to check this in?

src/externalized/json/Json.h
```diff
@@ -0,0 +1,99 @@
...
+		JsonValue(unsigned int value) {
+			m_value.reset(RJsonValue_fromUInt(value));
+		}
+		JsonValue(double value) {
+			m_value.reset(RJsonValue_fromDouble(value));
+		}
+		JsonValue(bool value) {
+			m_value.reset(RJsonValue_fromBool(value));
+		}
+		JsonValue(const ST::string value) {
```
momoko-h, line 28: Missing a `&` here.

momoko-h (commented):
Looks very nice overall. :+1:

src/externalized/DefaultContentManager.cc
```diff
@@ -579,32 +555,28 @@ bool DefaultContentManager::loadMusicModeList(const MusicMode mode, rapidjson::V
...
-	if(!document->IsObject()) {
-		SLOGE("music.json has wrong structure");
-		return false;
-	}
+	auto obj = document.toObject();
 
-	auto obj = document->GetObject();
 	SLOGD("Loading main_menu music");
-	loadMusicModeList(MUSIC_MAIN_MENU, obj["main_menu"]);
+	loadMusicModeList(MUSIC_MAIN_MENU, obj.GetValue("main_menu"));
```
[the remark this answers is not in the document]
selaux: Yes, I had considered that. JsonObject replaces both the `JsonObjectReader` as well as some `rapidjson::Value` occurences. `.GetValue()` is from the JsonObjectReader. I think we could support both for the moment. I will have a look.

src/externalized/CalibreModel.h
```diff
@@ -1,12 +1,14 @@
...
+#include "Json.h"
+
 #include <map>
 #include <stdexcept>
 #include <stdint.h>
 #include <string_theory/string>
 
 class JsonObject;
-class JsonObjectReader;
+class JsonObject;
```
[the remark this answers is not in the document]
selaux: Probably a search and replace error. It's actually in there three times now, as its defined in the `Json.h` header already. Will remove.

src/externalized/RustInterface.cc
```diff
@@ -0,0 +1,8 @@
+#include "RustInterface.h"
+
+void throwRustError(bool condition) {
+	if (condition) {
+		RustPointer<char> err{ getRustError() };
```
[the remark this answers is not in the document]
selaux: Yes. It's a `std::unique_ptr` with a custom destructor afaik.

src/externalized/json/Json.cc
selaux: Good catch!

src/externalized/tactical/MapItemReplacementModel.cc
selaux: Good catch. This is one of two places I still need to refactor, as they should not use `try {} catch {}`. I was just too lazy to define `.isXXX` methods on `JsonValue`.

src/game/SaveLoadGameStates.cc
```diff
@@ -71,114 +61,100 @@ void SavedGameStates::Set(const ST::string& key, const char* s)
 }
 
 #define RETURN_PRIMITIVE(v) \
-	if (v.IsString()) return ST::string(v.GetString()); \
-	else if (v.IsBool()) return v.GetBool(); \
-	else if (v.IsInt()) return v.GetInt(); \
-	else if (v.IsFloat()) return v.GetFloat();
+	try { return v.toString(); } catch (...) {} \
```
selaux, line 64: Don't use `try {} catch {}` here, but define `JsonValue::isXXX` methods.

src/game/SaveLoadGameStates.cc
```diff
@@ -71,114 +61,100 @@ void SavedGameStates::Set(const ST::string& key, const char* s)
...
-	auto readVal = [](const rapidjson::Value& v) -> PRIMITIVE_VALUE {
+	auto readVal = [](const JsonValue& v) -> PRIMITIVE_VALUE {
 		RETURN_PRIMITIVE(v) throw std::runtime_error("not a supported type");
 	};
 
 	RETURN_PRIMITIVE(v)
 
-	if (v.IsArray())
-	{
+	try {
```
selaux, line 82: Same here.

src/game/SaveLoadGameStates.cc
```diff
@@ -71,114 +61,100 @@ void SavedGameStates::Set(const ST::string& key, const char* s)
...
+		for (auto& el : v.toVec())
 		{
 			vec.push_back(readVal(el));
 		}
 		return vec;
-	}
-	if (v.IsObject())
-	{
+	} catch (...) {}
+	try {
```
selaux, line 90: And here

src/sgp/SGP.cc
selaux: Yep, this should not be checked in. We really should have a way to define this via CMake. Otherwise most errors will end up here and not in the debugger, or maybe I am missing something?

selaux:
So from the overall memory consumption it looks quite similar to the rapidjson variant:

**serde**

image

**rapidjson**

image

I will have a look at the amount of allocations/deallocations tomorrow, I still have to figure out how to get those numbers.

The failing tests are trivial to fix. Mostly it's different orders of properties in serialized JSON.

momoko-h:
Judging by these images the memory consumption looks really good, it's even a reduction of several 100 kb. I doubt the number of allocations is worth worrying about then, so please don't spend your time measuring them on my behalf (unless you're curious yourself, or course).

src/externalized/json/Json.cc
selaux: I just looked into this. And I think we should leave it as is, otherwise we would need to clone the RJsonValue again (which would be a second allocation just in this loop). My rationale behind this is that the `push_back` / `emplace_back` should only throw when reallocating in this case, but we already reserve the length of the vector beforehand, so reallocation should never be necessary.

selaux:
Rework is done now. I addressed most comments, for the others I added some explanation why it's not necessary imho.

selaux:
Incidentially this should also fix crashes on Android when using the ARMv7 architecture, probably caused by this: https://github.com/Tencent/rapidjson/issues/388 and reported on Discord.

lynxlynxlynx:
👍🏾 A cursory look found nothing from me. I'm just a bit surprised that it still requires so much scaffolding.

momoko-h:
I had another look and I didn't find anything that seemed wrong. Let's check it in and see how it performs in action.

PR #1775, opened 2023-02-01
Fix assertion failure in GetSelectedAssignSoldier

momoko-h:
DetermineWhichAssignmentMenusCanBeShown must not call this function before checking some flags, otherwise mousing over a vehicle triggers a stream of assertions.

This commit includes another fix for a problem found by clang-tidy: being a POW was supposed to be stressful and impede healing, but this never worked due to a missing 'else'.


lynxlynxlynx:
Nice catch. Also another reason to use more compact styles in the future, but that's bikeshedding already.

PR #1776, opened 2023-02-02
Change return type of four AI-related functions to bool

momoko-h:
These functions were using int, INT16 or INT8 to return TRUE or FALSE. Also remove declarations of four nonexistent functions.

If you're wondering why I care, I tried building with `-Wconversion`. Ouch. I believe the developers were just throwing a dice when they had to decide which type to use.


lynxlynxlynx:
Thanks!

PR #1782, opened 2023-02-06
Implement tab to view logs in Android launcher

selaux:
Just here for the builds again :wink:.

This adds a tab to the Android launcher for easier bug reporting / debugging.

### DIFF
--- a/android/app/src/main/java/io/github/ja2stracciatella/ConfigurationModel.kt
+++ b/android/app/src/main/java/io/github/ja2stracciatella/ConfigurationModel.kt
@@ -53,7 +53,8 @@ class Resolution(
 
 
 object ResolutionSerializer : KSerializer<Resolution> {
-    override val descriptor: SerialDescriptor = PrimitiveSerialDescriptor("Resolution", PrimitiveKind.STRING)
+    override val descriptor: SerialDescriptor =
+        PrimitiveSerialDescriptor("Resolution", PrimitiveKind.STRING)
 
     override fun serialize(encoder: Encoder, value: Resolution) {
         val width = value.width.toString()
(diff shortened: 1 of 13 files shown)


src/sgp/Video.cc
```diff
@@ -144,9 +144,17 @@ void InitializeVideoManager(const VideoScaleQuality quality)
...
+	if (g_game_window == NULL) {
+		throw std::runtime_error(ST::format("failed to create game window: {}", SDL_GetError()).c_str());
+	}
 
 	GameRenderer = SDL_CreateRenderer(g_game_window, -1, 0);
-	SDL_RenderSetLogicalSize(GameRenderer, SCREEN_WIDTH, SCREEN_HEIGHT);
+	if (GameRenderer == NULL) {
+		throw std::runtime_error(ST::format("failed to create renderer: {}", SDL_GetError()).c_str());
+	}
+	if (SDL_RenderSetLogicalSize(GameRenderer, SCREEN_WIDTH, SCREEN_HEIGHT) < 0) {
```
momoko-h, line 155: Ugh, yes, the correct "error handling" in this function is really completely broken.

I really wish the SDL documentation wasn't so terse. What do we gain from this SDL_RenderSetLogicalSize() call? Doesn't it make the calculations in [SimulateMouseMovement](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/e32ca529836a75fc8ee1b0ac0e9a0bafdd079fd2/src/sgp/Input.cc#L556) obsolete or even plain wrong?

selaux: Yeah. I will extract this into a separate PR, as we probably need check if the error is handled properly for all `SDL_` functions.

selaux:
This should now be done. It looks like this:

image

src/sgp/Video.cc
selaux: I've removed this part from the PR, as it should focus on the logs part.

PR #1783, opened 2023-02-09
Improve scrolling

momoko-h:
As its last step, ScrollJA2Background() currently creates a temporary SDL texture from the screen buffer and copies this texture to the renderer. This is pointless because after the function returns, RefreshScreen() then updates the permanent screen texture from the screen buffer and copies it to the renderer, over the texture created by ScrollJA2Background().

This commit removes this last step, resulting in faster scrolling and at least 900 K less memory used (in 640x480).

Additionally this commit addresses all -Wconversion warnings in Video.cc.


selaux:
Thanks!

PR #1792, opened 2023-02-13
Don't crash when removing a dead merc from the roster

### LINKED ISSUE #1788: Attempt to remove dead merc from list results in CTD
|  ja2-stracciatella_0.20.0-git+439a6d3_win-msvc64 | JA2 Gold 1.12 English CD  |    no   |    ?    |

Tried to remove my dead merc from list to make space for other but it crashes instantly after clicking "remove merc" button. Unfortunately cant reproduce this as it seemed to happen when i was trying to add mercs using ctrl+g so there was full 18 merc squad and one dead in it. Dont know what was real cause
Desktop_2023_02_13_08_39_58_465

momoko-h:
Use a more reliable way to find the selected merc to remove. Fixes #1788.

This also changes the logic in `CreateDestroyMouseRegionsForRemoveMenu` a bit: `GetNumberOfLinesOfTextInBox(ghRemoveMercAssignBox)`  always returns 2 here, no other value would make sense, there are only two mouse regions available for this.

### DIFF
--- a/src/game/Strategic/Assignments.cc
+++ b/src/game/Strategic/Assignments.cc
@@ -79,7 +79,6 @@
 #include <iterator>
 #include <string_theory/format>
 #include <string_theory/string>
-struct PopUpBox;
 
 
 // various reason an assignment can be aborted before completion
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
LGTM, thanks!

PR #1798, opened 2023-02-25
HandleTownLoyaltyForNPCRecruitment: no loyalty to increase if we're n…

lynxlynxlynx:
…ot in a town

fixes Return to San Amaro assertion when handing the note to the doc in the second area.


momoko-h:
So, bTownId uses the 'b' prefix which is supposed to indicate an INT8, but is declared as an UINT8. Why not, `GetTownIdForSector`'s return type is also UINT8, even though it just returns `StrategicMapElement::bNameId,` which is an INT8... I'm crying.

As for this PR, i'd appreciate it if you made the comparison more explicit: `bTownId != BLANK_SECTOR`

momoko-h:
Thanks!

PR #1799, opened 2023-02-26
Fix: IsThereASoldierInThisSector uses wrong SGPSector to index method

momoko-h:
Fallout from the introduction of `SGPSector`: `fSectorsWithSoldiers` used the strategic index (0-323) as its array index, not the sector index (0-255). Changing this array to a bitset and adding a dedicated function to convert a `SGPSector` to a position should help to avoid this mistake in the future.

This function was only used once, but in the really important function `FindStratPath`, which probably caused the enemy troops to use incorrect routes.


lynxlynxlynx:
Nice, thanks!

Issue #99, opened 2013-09-24
Game speed is generally slow


lynxlynxlynx:
Originally reported by: **Anonymous**

---

As the title says, the game speed is slow in general, fo example the mercs moving slowly, screen scrolling, pressing the laptop button, etc. Most particularilly it can be noticed from the slow bullet flight speed; in vanilla JA2 it's lighting fast but in stracciatella it is very slow, someone else had this issue too. This slowness of gameplay appears to only be on gennady's new stracciatella versions since the game speed is the same in vanilla JA2 in Tron's original stracciatella releases, perhaps poorly optimized code? SDL issue maybe, have you updated it or some other component from Tron's original? If so perhaps my very old computer just doesn't handle the newer versions of them very well. Still a thing to look into IMO.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/99

derSteFfi:
soo... the link is down, ja-galaxy moved in the meantime. i think this should be the new one:
http://thepit.ja-galaxy-forum.com/index.php?t=msg&goto=102405&&srch=Cool+Patch

The patch done there is trivial, added sleep cycles after each game loop, so no, i don't think this helps. The author even references ja2-stracciatella in the end as a 'cool' (temperaturewise) port. :)

There is a second idea presented in that thread that patches the game to consume a specific amount of time for one game cycle, but i think we already do this and it is controllable via MS_PER_GAME_CYCLE (see #148 ), right?

lynxlynxlynx:
It was externalised, so it's even easier to test. Some side-by-side timing is required to see if 25 is the best possible default. All three bugs linked above test various aspects of that.

kavol:
well, the user actions speeds seem to be okay, laptop opening roughly the same as I remember (if I remember anything at all after all those years :-)), but I have a great speed issue with enemy turns - it takes ages to resolve enemy movement ... if I see the enemy, it walks/shoots the same as the player but sometimes it looks just like it got stuck, taking e.g. 5 secs between shots from the same position, and if I don't see the enemy, sometimes it takes even over a minute before the progress bar gets to zero and sides switch; I don't observe such delays during the civilians' turn (except for Kingpin's "civilians" in San Mona, obviously)
this is pretty annoying, and I just don't understand that when my machine is Core i7-3520M, by orders of magnitude faster than what I had when playing the original ...

lynxlynxlynx:
[...]

PR #1806, opened 2023-03-12
Keep JA2Clock in lockstep with SDL's clock

oldlaptop:
As of #1173, the game's built-in clock is incremented monotonically by an `SDL_AddTimer()` callback called every `ms_per_time_slice`, as part of the effort to address the issues surrounding #99. Unfortunately this has a significant problem: `SDL_AddTimer()` can't necessarily be counted on to have granularity beyond 10ms or so on all platforms, and on some (OpenBSD at the very least) the time often varies upwards by 10ms (one step!) in practice, with the practical effect that the game continues to run painfully slowly. One way to solve this would be to instead use the timer callback to synchronize the JA2Clock with SDL's internal clock via SDL_GetTicks64().

This (unfinished) patch is a quick and dirty attempt at doing just that; empirically, the game is *far* more playable on SDL platforms that won't reliably call a timer callback every 10ms with this patch. Unfortunately, it also breaks the clock's configurability: it always advances at the same overall rate as SDL's internal clock. (`ms_per_time_slice` only affects the frequency at which the game's clock is synchronized with SDL's, and is therefore more or less useless.) One way to address this might be to add a new setting that applies a multiplier (or perhaps a bitshift?) to the new `elapsed` value in `TimeProc()`, but this would still break `ms_per_time_slice`'s existing semantics: to double the rate, for example, one would need to set the multiplier to 2, instead of dividing `ms_per_time_slice` by 2.

Assuming there's interest in this idea, I'll work on it further to make the appropriate configuration changes, etc.

### DIFF
--- a/src/game/Utils/Timer_Control.cc
+++ b/src/game/Utils/Timer_Control.cc
@@ -17,6 +17,8 @@
 UINT32	guiClockTimer = UINT32_MAX;
 UINT32	guiTimerDiag  =  0;
 
+Uint64 last_ticks = 0;
+
 UINT32 guiBaseJA2Clock = 0;
 
 static BOOLEAN gfPauseClock = FALSE;
@@ -71,23 +73,30 @@ extern UINT32 guiFlashCursorBaseTime;
 extern UINT32 guiPotCharPathBaseTime;
 
 
+#define UPDATECOUNTER( c, n )		( ( giTimerCounters[ c ] - n ) < 0 ) ?  ( giTimerCounters[ c ] = 0 ) : ( giTimerCounters[ c ] -= n )
+#define UPDATETIMECOUNTER( c, n )		( ( c - n ) < 0 ) ?  ( c = 0 ) : ( c -= n )
+
 static UINT32 TimeProc(UINT32 const interval, void*)
 {
+	Uint64 new_last_ticks = SDL_GetTicks64();
+	Sint64 elapsed = new_last_ticks - last_ticks;
+	last_ticks = new_last_ticks;
+
 	if (!gfPauseClock)
 	{
-		guiBaseJA2Clock += BASETIMESLICE;
+		guiBaseJA2Clock += elapsed;
 
 		for (UINT32 i = 0; i != NUMTIMERS; i++)
 		{
-			UPDATECOUNTER(i);
+			UPDATECOUNTER(i, elapsed);
 		}
 
 		// Update some specialized countdown timers...
-		UPDATETIMECOUNTER(giTimerTeamTurnUpdate);
+		UPDATETIMECOUNTER(giTimerTeamTurnUpdate, elapsed);
 
 		if (gpCustomizableTimerCallback)
 		{
-			UPDATETIMECOUNTER(giTimerCustomizable);
+			UPDATETIMECOUNTER(giTimerCustomizable, elapsed);
 		}
 
 #ifndef BOUNDS_CHECKER
@@ -96,8 +105,8 @@ static UINT32 TimeProc(UINT32 const interval, void*)
 			// IN Mapscreen, loop through player's team
 			FOR_EACH_IN_TEAM(s, OUR_TEAM)
 			{
-				UPDATETIMECOUNTER(s->PortraitFlashCounter);
-				UPDATETIMECOUNTER(s->PanelAnimateCounter);
+				UPDATETIMECOUNTER(s->PortraitFlashCounter, elapsed);
+				UPDATETIMECOUNTER(s->PanelAnimateCounter, elapsed);
 			}
 		}
 		else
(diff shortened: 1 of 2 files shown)


oldlaptop:
One thing I see in CI logs is that `SDL_GetTicks64()` is relatively new (https://github.com/libsdl-org/SDL/commit/99c9727dc0bbc9be96f9e023d8b6e7e0fe120b11) - `SDL_GetTicks()` is a perfectly reasonable substitute for basic testing, and correctly handling the overflow case is another "if anyone's interested" thing.

ja2-derek:
Relevant docs on SDL: https://wiki.libsdl.org/SDL2/SDL_AddTimer

> Timing may be inexact due to OS scheduling. Be sure to note the current time with [SDL_GetTicks](https://wiki.libsdl.org/SDL2/SDL_GetTicks)() or [SDL_GetPerformanceCounter](https://wiki.libsdl.org/SDL2/SDL_GetPerformanceCounter)() in case your callback needs to adjust for variances.

oldlaptop:
> Relevant docs on SDL: https://wiki.libsdl.org/SDL2/SDL_AddTimer
> 
> > Timing may be inexact due to OS scheduling. Be sure to note the current time with [SDL_GetTicks](https://wiki.libsdl.org/SDL2/SDL_GetTicks)() or [SDL_GetPerformanceCounter](https://wiki.libsdl.org/SDL2/SDL_GetPerformanceCounter)() in case your callback needs to adjust for variances.

Precisely so. Those 'variances' aren't accounted for now, JA2Clock is advanced a fixed amount every time the timer callback fires, no matter how much time has actually elapsed.

> And [SDL_GetTicks64](https://wiki.libsdl.org/SDL2/SDL_GetTicks64):
> 
> > This function is available since SDL 2.0.18.

And I take it the Linux CI box has some older SDL version than that, as does Debian stable. I'm guessing you'll prefer `SDL_GetTicks()` for that reason?

lynxlynxlynx:
Or we use `<chrono>`, eg.:
```
using tick_t = unsigned long; // milliseconds
inline tick_t GetMilliseconds()
{
    using namespace std::chrono;
    return duration_cast<tick_t>(steady_clock::now().time_since_epoch()).count();
}
```

momoko-h:
@oldlaptop, could you give [this commit](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/40c757e11b6ecc4884755c17ccdab2f8ea3dc44e) a trial run to see if this improves the situation with OpenBSD? I didn't test it much, so some weird behaviour wouldn't be unexpected.

oldlaptop:
> Is there a reasonable explanation why OpenBSD (or SDL on OpenBSD, whoever the culprit may be) is so bad at this?

I don't get the impression "this" (use SDL_Timer to increment a monotonic clock at a set frequency) is something very many other SDL users do. Presumably the SDL developers haven't gone to great lengths to tune this interface so it's more useful for something it isn't supposed to do in the first place.

> @oldlaptop, could you give [this commit](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/40c757e11b6ecc4884755c17ccdab2f8ea3dc44e) a trial run to see if this improves the situation with OpenBSD? I didn't test it much, so some weird behaviour wouldn't be unexpected.

I'll try that, yes. What would you do about the "speed up all the animations" use case for `ms_per_time_slice` under this approach? Perhaps externalize some or all of the TimerIntervals? (I don't care about it at all, for my part, but I assume the ability to control the game clock's speed is or was useful to someone.)

momoko-h:
If this approach works, then it would be quite easy to add a user configurable multiplier for all intervals, for example users could specify that all intervals should be multiplied by 0.8, so that an interval that is specified as 400ms would result in an 320ms delay.

Or, if it turns out that just some things are too slow, we could also tweak specific intervals. There is no reason to take the values Sir-Tech used 25 years as gospel; it's quite possible that the only design decision that went into them was "250 works for me with our current timer code under Windows 98, let's use that".

momoko-h:
[This commit](https://github.com/momoko-h/ja2-stracciatella/commit/f164c82ee56ea97710ba9685a7f9454428a54aa6) adds the new configuration option. You can either set game_durations_multiplier in game.json or g_durations_multiplier if you're running the game under a debugger. Values > 1 slow down the game,  values < 1 speed it up.

oldlaptop:
I've run https://github.com/momoko-h/ja2-stracciatella/commit/40c757e11b6ecc4884755c17ccdab2f8ea3dc44e through one long-ish battle (Tixa ground level, so a reasonably good variety of animations, including tear gas) and it seems about as playable as the quick-and-dirty patch in this PR. I'll see about https://github.com/momoko-h/ja2-stracciatella/commit/01bf5be3fd831fb216c0021687dc43d8c4046728 soon-ish.

> Or, if it turns out that just some things are too slow, we could also tweak specific intervals. There is no reason to take the values Sir-Tech used 25 years as gospel; it's quite possible that the only design decision that went into them was "250 works for me with our current timer code under Windows 98, let's use that".

I think it's *very* likely Stracciatella's existing timers aren't quite right... :-)

oldlaptop:
I've not had that much time to fool with this, but https://github.com/momoko-h/ja2-stracciatella/commit/01bf5be3fd831fb216c0021687dc43d8c4046728 still seems fine to me. I'll assume that if either of these branches should get merged, it's that one.

lynxlynxlynx:
@momoko-h please submit these as a PR, so the work doesn't get lost.

momoko-h:
> @momoko-h please submit these as a PR, so the work doesn't get lost.

I'm currently experimenting with a new version that removes the SDL timer completely, but since that's a fairly massive change I'd like to test it some more before I ask for review. But if you're curious, then it's [here](https://github.com/momoko-h/ja2-stracciatella/tree/RefactorTimerControl).

PR #1807, opened 2023-03-15
Proposal: Revamp calling of the game loop

momoko-h:
Change how often GameLoop() gets called
    
    Before this commit, the duration between two GameLoop calls was
    configurable (ms_per_game_cycle, default 25ms). This is problematic
    because many of the predefined timecout counters use multiples of 5ms,
    and the delay between two soldier animations is even calculated with
    a resolution of 1ms.
    
    With this commit the game now tries to call GameLoop at a rate of
    144Hz (every ~7ms), corresponding to the refresh rate of modern
    monitors. To satisfy the needs of our users who do not need 144 FPS
    and would prefer a lower CPU usage, GameLoop()'s heaviest subfunction
    RefreshScreen() now has its own timeout.

This patch should make our timings be more in line with those originally intended by the JA2 developers. For example, `ExecuteOverhead` was supposed to run every 5ms, but with our current default settings it is limited to once every 25ms.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -31,11 +31,11 @@
   //  vanilla setting: true
   "draw_item_shadow": true,
 
-  //  Number of milliseconds for one game cycle.
-  //  25 ms gives approx. 40 cycles per second (and 40 frames per second, since the screen is updated on every cycle).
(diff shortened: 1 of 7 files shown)


lynxlynxlynx:
Looks fine. 👍🏾

Issue #1808, opened 2023-03-18
Blood type possible inconsistency


sevenfm:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/42628dc4dba47991dda3f8501372ac4785bafbb2/src/game/TileEngine/Smell.cc#L365

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/42628dc4dba47991dda3f8501372ac4785bafbb2/src/game/TileEngine/Smell.cc#L385-L386

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/42628dc4dba47991dda3f8501372ac4785bafbb2/src/game/TileEngine/Smell.cc#L463

The code says that "blood floor type stored in smell byte!", but in one instance it tries to read it from ubBloodInfo, which may be incorrect.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|   |   |    yes    |         |

PR #1809, opened 2023-03-19
Fix three blood / smell bugs

momoko-h:
• Wrong calculation of blood amount in `ExpAffect()`
• Wrong MAP_ELEMENT member checked for blood type (#1808)
• Wrong variable used for dropping new blood on roof


lynxlynxlynx:
LGTM, thanks!

PR #1818, opened 2023-04-19
Refactor timer control

momoko-h:
As discussed in #1806, using an SDL timer to keep track is very imprecise on some platforms. This set of commits removes this timer and instead queries the steady_clock whenever necessary.

I didn't notice anything odd while testing this, but I also never had much problem on Linux with the old code, so I have to rely on oldlaptop's word that this improves the situation on OpenBSD.


src/game/Utils/Timer_Control.cc
```diff
@@ -73,52 +73,17 @@ extern UINT32 guiPotCharPathBaseTime;
...
-				UPDATETIMECOUNTER(s->NextTileCounter);
-				UPDATETIMECOUNTER(s->PanelAnimateCounter);
-			}
-		}
-#endif
+		auto const timeDiff = now - lastUpdate;
+		guiBaseJA2Clock += std::chrono::duration_cast<std::chrono::milliseconds>(timeDiff).count();
 	}
 
+	lastUpdate = now;
```
lynxlynxlynx, line 78: Why not keep this inside the above block? Now when the player unpauses the timers will update as if there was no pause in the interim.

rust/stracciatella/src/schemas/yaml/game.schema.yaml
```diff
@@ -48,15 +48,18 @@ properties:
       power consumption, higher values can make the game appear to be more responsive.
 
       Stracciatella default value: `40`
-  ms_per_time_slice:
```
lynxlynxlynx, line 51: Will this break upgraded installs, requiring user intervention? I don't remember when all we check against the schema, but if it's also at runtime, it'd be better to leave this entry in and mark it as deprecated instead.

src/game/Utils/Timer_Control.cc
momoko-h: I'll rewrite this a little to make the logic how JA2Clock is handled while time is paused easier to follow. The problems with paused time also shouldn't be overemphasized; it happens quite rarely: while either a message box, the item pickup menu or the door open menu is open. This means it's not really an issue for timing parts of the UI, even though such parts could and probably should better use the GetClock() function defined in Timer.h

rust/stracciatella/src/schemas/yaml/game.schema.yaml
momoko-h: I didn't know the answer, so I just tried it: the validation checks whether all the properties marked as required are present and whether properties have the correct type within the allowed range. Additional properties are ignored, so having an old and unused value for ms_per_time_slice won't cause an error when the game starts.

src/game/Utils/Timer_Control.cc
lynxlynxlynx: Most of the interval constants are for the gui, so that's good to hear. What I was worried about is quests, bleeding, daytime, that sort of thing. Consider you've got the open door pop-up and leave to make some tea. Can someone die due to that?

momoko-h: > Most of the interval constants are for the gui, so that's good to hear. What I was worried about is quests, bleeding, daytime, that sort of thing. Consider you've got the open door pop-up and leave to make some tea. Can someone die due to that?

No, that's one of those things pausing time actually prevents. If you've a wounded merc who would usually bleed every couple of seconds in realtime, he'll not bleed while those menus are open, but instead resume bleeding a couple of seconds after the menu are closed. This behavior is not changed by this commit. 

Quests do not use the JA2Clock() or `TIMECOUNTERs` but the the "game clock" function in GameClock.h, which has it's own functions to interrupt or pause time, so they're not affected either.

momoko-h:
I've been using this some more without any issues. At this point I believe it would be helpful to merge it so it gets more exposure and hopefully some feedback from people who were affected by the poor accuracy of the SDL timer.

selaux:
Had a look through the code and tested it a bit myself. Looks good :+1:

PR #1819, opened 2023-04-20
Address several problems in Sound_Control.cc

momoko-h:
• `getSoundSample()`: unused function that didn't do anything useful.
• `PlayJA2StreamingSample()`: exactly the same as `PlayJA2Sample`
• `PlayJA2Sample`: could call SoundPlay with a NULL filename
• `PlaySoldierJA2Sample`: wrong return value in case of an error
• simpler implementation of `SoundRange()`


selaux:
Thanks.

PR #1820, opened 2023-04-26
Fix return value of CalcPercentBetter

momoko-h:
As the name suggests, this function is supposed to calculate how much better a new cover position is compared to the old one, so returning NOWHERE would indicate the new cover is 25,600% better.

In theory this could change the behavior of the AI slightly so that enemies choose a different gridno to take cover, in practice I didn't any case where this function actually returned 25600.


selaux:
LGTM. Thanks!

PR #1821, opened 2023-04-28
Add a new function CoinToss() to Random

momoko-h:
This addresses several cases where a probability of 50% was intended but the result was either a bit or completely off. Notable change in gameplay: critical hits to the torso will no longer automatically cause a loss of dexterity.

### DIFF
--- a/src/game/Laptop/Mercs.cc
+++ b/src/game/Laptop/Mercs.cc
@@ -1339,10 +1339,9 @@ static BOOLEAN ShouldSpeckStartTalkingDueToActionOnSubPage(void)
 		HandlePlayerHiringMerc(listing[gubCurMercIndex]);
 
 		//get speck to say the thank you
-		if( Random( 100 ) > 50 )
-			StartSpeckTalking( SPECK_QUOTE_GENERIC_THANKS_FOR_HIRING_MERCS_1 );
-		else
-			StartSpeckTalking( SPECK_QUOTE_GENERIC_THANKS_FOR_HIRING_MERCS_2 );
+		StartSpeckTalking(CoinToss()
+			? SPECK_QUOTE_GENERIC_THANKS_FOR_HIRING_MERCS_1
+			: SPECK_QUOTE_GENERIC_THANKS_FOR_HIRING_MERCS_2);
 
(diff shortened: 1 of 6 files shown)


lynxlynxlynx:
Thanks!

misanthropos:
Looking at line 2817 in src/game/Tactical/Soldier_Ani.cc does not seem to be a coin tossed but something much rarer: 1/50th. Now I would expect a comment by characters too often. It looks fine otherwise.

momoko-h:
I know that this changes the distribution of the possible reactions to killing an enemy (and it's not a civilian, a bloodcat or a head shot and what else the function is checking for), which is controlled by line 2817 in combination with line 2803. Before this change, the chances for the different reactions were:

33% "Killed an enemy" quote
65.6% Laugh
1.34% "Cool!"

which seemed odd to me, and is one of those cases I meant when I wrote that the result was completely off in the commit message.

With this change the distribution is now:

33% "Killed an enemy" quote
33.5% Laugh
33.5% "Cool!"

which I believe makes more sense.

lynxlynxlynx:
I think both of us with @momoko-h took that as a bug in the original. It's was even rarer before, unless you're fighting bloodcats.

misanthropos:
I was not aware that it was just 1.34% "Cool!". I had the impression it would be said if a nearly impossible shot hit and maybe killed the enemy.
To my taste a distribution like the former 
33% "Killed an enemy" quote
65.6% Laugh
1.34% "Cool!"
seems more natural than distributing them equally. A cool shot should only be expressed if it was actually cool :-).

By the way, I have not seen an exploding head for quite some time. Does it still happen?

lynxlynxlynx:
The code doesn't even consider whether it was a cool shot or not.

momoko-h:
Well, if you want to figure out and implement a heuristic to determine whether a kill was particularly cool I won't object. But as long as it's purely random I'd prefer to hear the maniacal laughter a bit less often.
Head explosions definitely still happen, I see them quite often when I start a new game and fight the enemies without helmets in Omerta.

misanthropos:
As I said I was under the impression a "cool" shot would trigger another member to make a comment. If it was random before then I won`t object to the new approach.

PR #1825, opened 2023-05-14
Upgrade string_theory to 3.7

selaux:


momoko-h:
Looks good. A couple of the `clear()` aren't strictly necessary but I prefer the way you wrote it because it makes the intention clear and for symmetry.

PR #1826, opened 2023-05-18
CMakeLists.txt maintenance

momoko-h:
• remove unused define WITH_SOUND_DEBUG
• fix WITH_ASAN option
• only add compiler warnings flags when compiling, not when linking
• disable some MSVC warnings here instead of in Platform.h

I could've used generator expressions for the warnings but I find the if-else-endif construct much easier to read and understand.


momoko-h:
The new warnings in the AppVeyor log are not caused by this but the ST 3.7 update.

PR #1829, opened 2023-05-23
Update CMakeLists

### LINKED ISSUE #1828: Should we bump the cmake_minimum_required value?
So [it looks like](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8274) CMake 3.27, which is due out in two weeks will deprecate versions < 3.5. Our main CMakeLists.txt currently uses 3.4, so we either have to increase that or live with the new warning.

I'd suggest we bump this to 3.20. That version was released two years ago and I don't think it's too much too ask that anyone who'd like to build the game themselves uses tools that are not completely outdated.

The github runners and AppVeyor are all at 3.26, but I'm not sure what the Android build uses.

momoko-h:
• Set minimum required version to 3.18.1 to avoid deprecation warnings
    • Set maximum policy version in the getter CMakeList.txt.in files
      to 3.25 to avoid policy CMP0135 warnings
    • Move the main ja2 target up, so we can add properties to
       it instead of setting them globally for all targets.

closes #1828.


momoko-h:
Is there a way to trigger an AppVeyor check without pushing a commit?

lynxlynxlynx:
In it's GUI you can rerun jobs, but that's it.

CMakeLists.txt
```diff
@@ -111,8 +111,6 @@ if(MSVC)
 		CMAKE_C_FLAGS_MINSIZEREL
 		CMAKE_C_FLAGS_RELWITHDEBINFO
 	)
-		# force /MD runtime (future note: cmake 3.15 has CMAKE_MSVC_RUNTIME_LIBRARY)
```
selaux, line 114: Is this replaced by anything, or is it just deemed unnecessary?

momoko-h: It is replaced by [policy CMP0091](https://cmake.org/cmake/help/latest/policy/CMP0091.html#policy:CMP0091) which is automatically enabled by setting the minimum required version to 3.18. If we don't set MSVC_RUNTIME_LIBRARY CMake defaults to MultiThreaded$<$<CONFIG:Debug>:Debug>DLL ( /MD or /MDd for debug builds).

PR #1830, opened 2023-05-23
Remove BUTTONSYSTEM_DEBUGGING

momoko-h:
It is pointless and doesn't even build any longer since GUI_CALLBACK was changed to std::function.

This was causing the Appveyor build failure in #1829.


selaux:
Thanks.

PR #1832, opened 2023-05-27
Strategic-level modding support (Event, Quests and Facts)

ja2-derek:
Adding several more event hooks and Lua functions for mods to handle Strategic level events.

- Allowing mods to process, or ignore StrategicEvents - `OnStrategicEvent`
- Allowing mods to trigger  StrategicEvents - `AddEveryDayStrategicEvent`, `AddStrategicEvent`
- Allowing mods to override the current value of game progress (which affects weapons availability, enemy levels) - `OnCalcPlayerProgress`
- Allowing mods to process quests - `OnCheckQuests`, `OnQuestEnded`
- Allowing mods to process facts - `SetFactTrue`, `SetFactFalse`, `CheckFact`
- Mods can read the current game clock - `GetWorldTotalSeconds`, `GetWorldTotalMin`, `GetWorldDay`
---

Example: Disabling Meanwhile scenes from mod

```lua
require('enums.lua')
RegisterListener("OnStrategicEvent", "skip_meanwhile")

function skip_meanwhile(event, processed)
    if event.ubEventKind == StrategicEventKind.EVENT_MEANWHILE then
        log.debug('intercepted and cancelled MEANWHILE cutscene')
        processed.val = TRUE
    end
end
```

### DIFF
--- a/src/externalized/scripting/FunctionsLibrary.h
+++ b/src/externalized/scripting/FunctionsLibrary.h
@@ -233,7 +233,7 @@ OBJECTTYPE* CreateMoney(const UINT32 amt);
  */
 void PlaceItem(const INT16 sGridNo, OBJECTTYPE* const pObject, const INT8 bVisibility);
 /**
- * Adds a recurring events that happens at the same time every day.
+ * Adds a recurring event that happens at the same time every day.
  * @param ubCallbackID strategic event ID
  * @param uiStartMin the time (minutes of day) that the event
  * @param uiParam a parameter that will be passed to the event handler
--- a/src/externalized/scripting/FunctionsLibrary.h
+++ b/src/externalized/scripting/FunctionsLibrary.h
@@ -91,6 +91,12 @@ extern Observable<INT16, INT16, INT8, INT16, STRUCTURE*, UINT8, BOOLEAN> OnStruc
 extern Observable<STRATEGICEVENT*, BOOLEAN_S*> OnStrategicEvent;
 
 /**
+ * Allows to override the player progress calculation.
+ * @param the progress percentage calculated by the base game. This can be adjusted or overridden.
+ */
+extern Observable<UINT8_S*> OnCalcPlayerProgress;
+
+/**
  * When the game about to be saved. This is the place to persist mod game states.
  * @ingroup observables
  */
--- a/src/externalized/scripting/FunctionsLibrary.h
+++ b/src/externalized/scripting/FunctionsLibrary.h
@@ -247,6 +247,9 @@ void AddEveryDayStrategicEvent_(UINT8 ubCallbackID, UINT32 uiStartMin, UINT32 ui
  * @param uiParam a parameter that will be passed to the event handler
  */
 void AddStrategicEvent_(UINT8 ubCallbackID, UINT32 uiMinStamp, UINT32);
+UINT32 GetWorldTotalMin();
+UINT32 GetWorldTotalSeconds();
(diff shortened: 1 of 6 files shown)


src/externalized/scripting/FunctionsLibrary.h
```diff
@@ -200,6 +232,28 @@ OBJECTTYPE* CreateMoney(const UINT32 amt);
  * @ingroup funclib-items
  */
 void PlaceItem(const INT16 sGridNo, OBJECTTYPE* const pObject, const INT8 bVisibility);
+/**
+ * Adds a recurring events that happens at the same time every day.
```
lynxlynxlynx, line 236:
```suggestion
 * Adds a recurring event that happens at the same time every day.
```

src/externalized/scripting/FunctionsLibrary.h
```diff
@@ -200,6 +232,28 @@ OBJECTTYPE* CreateMoney(const UINT32 amt);
...
+ * @param ubCallbackID strategic event ID
+ * @param uiStartMin the time (minutes of day) that the event
+ * @param uiParam a parameter that will be passed to the event handler
+ */
+void AddEveryDayStrategicEvent_(UINT8 ubCallbackID, UINT32 uiStartMin, UINT32 uiParam);
+
+/**
+ * Adds an one-off strategic event
+ * @param ubCallbackID strategic event ID
+ * @param uiMinStamp earliest time (in world-seconds) that this event will be processed
```
lynxlynxlynx, line 246: Do lua users have a way to know the current world time? IOW is this realistically usable?

src/externalized/scripting/FunctionsLibrary.h
```diff
@@ -200,6 +232,28 @@ OBJECTTYPE* CreateMoney(const UINT32 amt);
...
+ * @param uiParam a parameter that will be passed to the event handler
+ */
+void AddStrategicEvent_(UINT8 ubCallbackID, UINT32 uiMinStamp, UINT32);
+void StartQuest_(UINT8 ubQuestID, std::string sectorID);
+void EndQuest_(UINT8 ubQuest, std::string sectorID);
```
lynxlynxlynx, line 251: These two should also use const references for `sectorID`.

lynxlynxlynx:
Very nice, getting closer to be able to externalize more of the story.

ja2-derek:
Examples of scheduling a Strategic Event:

```lua
function bring_merc_site_back_online()
    local event_param = 0  -- unused for this event
    local event_time = GetWorldTotalSeconds() + 3600
    AddStrategicEvent(StrategicEventKind.EVENT_MERC_SITE_BACK_ONLINE, event_time, event_param);
end
```

Mods may also hard-coded the time if it's relatively to game/campaign start.

src/externalized/scripting/FunctionsLibrary.h
ja2-derek: Added a commit to expose also the `GetWorldTotalSeconds` method to return game world time.

src/externalized/scripting/FunctionsLibrary.h
ja2-derek: This is a function to be invoked from Lua, and the args will go through serialization. Pass-by-reference does not work. We can only pass by pointer or pass by value.

This arg is not to be modified, and it is not supposed to be called frequently. So it is easier passed by value, as it doesn't have potential problems of null pointers.

PR #1833, opened 2023-05-29
WIP: Externalize translations

momoko-h:
Part 1: add code to export the translations to JSON.

Since this code is supposed to run exactly once to produce the JSON files before it is removed again this is just a quick and dirty implementation. Only Git will remember it ever existed...

I am somewhat concerned about the legal status of these translations. Aren't these strings copyrighted material that we aren't allowed to ship? Would be a shame to implement loading these files only to find out the work was all for naught.


ja2-derek:
I am not a lawyer, but those strings were part of the original source code. So shouldn't they be under the same license?

lynxlynxlynx:
I can't see an issue either.

selaux:
I also think it's fine. Everything is still distributed under the same license.

lynxlynxlynx:
Please also run a formatter on the json files, so translators will have an easier time editing and reviewing.

PR #1834, opened 2023-06-07
New implementation of the FPS display

momoko-h:
Since #1807 the FPS display was rather useless: Game Loop always displayed '1000', but even before that it was not in great shape: it had tearing while scrolling and Game Loop usually just printed what your ms_per_time_slice setting was. 
From a code perspective it had to be supported by each screen individually, so it was only available on the editor, game, map and laptop screen and all of these just different key combinations to toggle it  (and Laptop didn't even have a key combo). Also, hiding it behind the cheats toggle was pointless, IMO.

This new version should not only print more accurate numbers, it is also available everywhere under the control-f key combination.


lynxlynxlynx:
Thanks!

PR #1836, opened 2023-06-12
Add an option to use precompiled headers

momoko-h:
Add a CMake option to use precompiled standard library headers. This cuts down compilation times quite a bit when not using ccache or sccache.

I obviously didn't try every possible combination of headers , compilers and configurations, but to give you some idea of the savings for a fresh build without cache:
Compiler, Config | no PCH | PCH |
---------------- | ------ | --- |
GCC 13, Debug | 1143.46user |883.24user |
Clang 16, RelWithDebInfo | 1638.52user | 1249.86user

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,7 @@ option(BUILD_LAUNCHER "Build the ja2 launcher application" ON)
 option(WITH_EDITOR_SLF "Include the latest free editor.slf" OFF)
 option(WITH_RUST_BINARIES "Include rust binaries in build" ON)
 option(WITH_ASAN "Build with address sanitizer" OFF)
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Sure, why not, it's also just for some system headers, so less likely to cause obscure bugs (I think most reports of pch being problematic are where projects precompile all headers).

momoko-h:
Yes, there are plenty of dire warnings out there about situations where precompiled headers can be problematic, but many of them are really old and it's far from clear which ones were resolved in modern compilers and which still apply to our project. That's why I choose the safest option, system headers only and default off. After all, anyone who's more adventurous can simply add more headers themselves (SDL.h looks like a promising candidate, for example).

PR #1837, opened 2023-06-15
Remove three unnecessary functions handling video refresh

momoko-h:
ExecuteBaseDirtyRectQueue and EndFrameBufferRender just set two flags that the entire screen should be refreshed or that the frame buffer is dirty, something RefreshScreen() can figure out itself by looking at the state of three variables it needs to check anyway.
These changes turned made AddBaseDirtyRect just a synonym for InvalidateRegionEx, so it could be removed, too.


lynxlynxlynx:
LGTM!

PR #1838, opened 2023-06-23
Compute gsRoadMacroStartIndex array at compile time

momoko-h:
The function InitializeRoadMacros simply computes an array of constants from another array of constant values, so there is no need to do this at run time and include the code for it in the executable.


lynxlynxlynx:
Sure, why not.

PR #1839, opened 2023-07-03
Clean up guiSavedGameVersion related code

momoko-h:
In some places guiSavedGameVersion, which was itself a constant, was compared to another constant, so the result of this comparison was always predetermined, leading to some dead code that could never be executed: the functions LoadEnemySoldiersFromTempFile and JA2EncryptedFileWrite.


lynxlynxlynx:
Nice find, thanks!

PR #1841, opened 2023-07-05
Doc cleanup

AnthonyGiorgio:
I fixed a variety of spelling, grammar, and formatting issues in some of the Markdown files.  I also resolved a bunch of warnings from the Markdown linter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,17 +29,17 @@ To get in touch with the developers you can use the [#ja2-stracciatella Discord 
 
 ### Without the optional launcher
 
-3. Start the game the first time.  It will create the configuration file %USERPROFILE%\Documents\JA2\ja2.json on Windows or ~/.ja2/ja2.json on Unix-like systems.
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Thanks, I'll leave some comments inline.

README.md
```diff
@@ -9,7 +9,7 @@ This is continuation of venerable JA2-Stracciatella project.
 
 The goal of the project is to make Jagged Alliance 2 available on a wide range of platforms, improve its stability, fix bugs and provide a stable platform for mod development.  At the moment the goal is mostly to fix bugs.
 
-Official Homepage: http://ja2-stracciatella.github.io/
+[Official Homepage](http://ja2-stracciatella.github.io)
```
lynxlynxlynx, line 12: This is not good for two reasons: the URL is not visible any more and the line stands out. The intent is to lay it out, so if the linter is complaining about the bare URL, just add the (reduntant) URL syntax around it with itself as the text.

AnthonyGiorgio: I've reverted the change.

lynxlynxlynx:
Thanks!

PR #1845, opened 2023-07-16
Fix GetNumberOfVisibleWorldItemsFromSectorStructureForSector

momoko-h:
This function returned wrong results while loading the game, which lead to misleading warning messages like "SynchronizeItemTempFile() Reported 0, should be 1".

`SynchronizeItemTempFileVisbleItemsToSectorInfoVisbleItems`  simply wants to compare the number of
WORLDITEMs in the temp file with the `uiNumberOfWorldItemsInTempFileThatCanBeSeenByPlayer` in the 
`SECTORINFO` or `UNDERGROUND_SECTORINFO` structs, but `GetNumberOfVisibleWorldItemsFromSectorStructureForSector`
does a little more than just return that number.


lynxlynxlynx:
Nice!

PR #1843, opened 2023-07-10
Fix division by 0 in CalcBestStab()

momoko-h:
MinAPsToAttack() can return 1 for extreme values of ubShotsPer4Turns in weapons.json, so always subtracting 1 AP for the cost of turning even if the soldier does not need to turn for the attack gave 0 for ubRawAPCost and a division by 0 followed when computing the hit rate.

CalcTentacleAttack() had the same problem.

Removed declaration of non-existant function GetBPsTouseJar().

Fixes #1840 (the crashing part, I didn't observe the unexpected item class warning while testing this).

PR #1846, opened 2023-07-17
Address two new coverity defects

momoko-h:
Make it crystal clear even to static analyzers that ubRawAPCost can never be 0. CIDs 322125, 322126.

This is almost funny, now that I've fixed the situations where this might actually happen in #1843, Coverity finally detects that MinAPsToAttack might return 1 and 1-1=0. Why it didn't find that when we were still unconditionally subtracting 1 is a mystery.


lynxlynxlynx:
Yep, thanks!

Issue #1850, opened 2023-07-22
Can't change merc's duty assignment to "repair". 


qqlkan:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

Hello there!

From the beginning (Omerta or elsewhere) no one can use repair kit/tool kit to repair stuff. After clicking on repair button nothing happens, 'items' and 'cancel' buttons are inactive. I have to click on the ground to quit it. This bug occurs with every merc. It's a serious bug affecting the entire gameplay.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
| Stracciatella v0.20.0-git+439a6d3 |JA2 1.12 GOG, JA2 GOLD and JA2 GOG + Wildfire maps STEAM   |    no    |    ?   |

Enabled plugins don't matter
Windows 11 x64, version 22H2

momoko-h:
Oh, you mean trying to assign a merc to repair in the tactical screen. Yes, that triggers an assertion: 
`sgp/Debug.h: Assertion failed in /home/devel/ja2-stracciatella/src/game/Strategic/Assignments.cc, line 6854:`

You can still assign mercs to repair from the map screen so it's not game breaking, but certainly annoying.

PR #1855, opened 2023-07-26
Fix assigning mercs to Repair in tactical

momoko-h:
RepairMenuBtnCallback got the merc who should do repairs from a global variable, which could get clobbered between the time the popup box was created and the time the callback function was executed.

Now the ID of the merc is stored in the user data of the popup boxes' mouse regions during creation and retrieved from there by the callback function. Issue #1850.


selaux:
Thanks. Less global state is always good...

Issue #1856, opened 2023-07-27
Game reliably crashes in this particular situation.


MoerasGrizzly:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### Game reliably crashes when MD dies in this savegame.

Whilst playing I've noticed a couple crashes that seem to occur when MD dies. I've now gotten a savegame where this situation happens a 100% of the time. 

MD bleeds out at the start of the next turn in the provided savegames.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  both stable and latest nightly |  JA2 whatever the current GOG version is |    unknown |    unknown |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)
```2023-07-27T22:40:01.3356242Z [INFO] stracciatella::logger: Logging to file "C:\\Users\\Grizzly\\AppData\\Local\\Temp\\ja2.log"
2023-07-27T22:40:01.6774923Z [INFO] stracciatella::vfs: VFS item with priority 24: DirFs { "C:\\Games\\JA2 Stracciatella\\externalized" }
2023-07-27T22:40:01.6777669Z [INFO] stracciatella::vfs: VFS item with priority 23: DirFs { "C:\\gog games\\Jagged Alliance 2\\data" }
2023-07-27T22:40:01.6779593Z [INFO] stracciatella::vfs: VFS item with priority 22: SlfFs { source: DirFsFile { "mercedt.slf" in "C:\\gog games\\Jagged Alliance 2\\data" } }
[...]

PR #1857, opened 2023-07-28
Fix crash when insured merc dies without an attacker

momoko-h:
Issue #1856.

### DIFF
--- a/src/game/Strategic/Strategic_Merc_Handler.cc
+++ b/src/game/Strategic/Strategic_Merc_Handler.cc
@@ -99,7 +99,8 @@ void StrategicHandlePlayerTeamMercDeath(SOLDIERTYPE& s)
 				{ // Killed within an hour of being insured
 					p.ubSuspiciousDeath = VERY_SUSPICIOUS_DEATH;
 				}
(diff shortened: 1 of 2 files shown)


momoko-h:
The AppVeyor build failed again because it could not download sccache. Maybe we should default USE_SCCACHE to off, at least on Windows?

lynxlynxlynx:
Thanks.

As for sccache: no, we should just bump the version, since this is one of our slowest builds. Apparently 0.2.13 got removed:
https://community.chocolatey.org/packages/sccache#versionhistory
Sure, it may still break now and then, but I think we can live with that, the benefits outweigh it.

momoko-h:
Hmm, if I understand ci-setup.sh and ci-functions.sh correctly, we are only selecting version 0.2.13 for the unix-like builds, while windows-install-via-chocolatey leaves the version unspecified, so bumping the version would only benefit the Linux builds. The log of the last successful AppVeyor build confirms this:
```
Cache location                  Local disk: "C:\\Users\\appveyor\\AppData\\Local\\Mozilla\\sccache\\cache"
Version (client)                0.5.4
Cache size                          500 MiB

```

PR #1858, opened 2023-07-28
AppVeyor: Cache sccache binary

momoko-h:
[skip actions]

After reading https://www.appveyor.com/docs/build-cache/#caching-chocolatey-packages I think we could make those builds faster and more reliable by caching the sccache binary itself, too.

### DIFF
--- a/.ci/ci-functions.sh
+++ b/.ci/ci-functions.sh
@@ -10,7 +10,7 @@ unix-install-rustup () {
 }
 
 linux-install-sccache () {
-    local SCCACHE_VERSION="0.2.13"
+    local SCCACHE_VERSION="v0.5.4"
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Looks fine; we'll see a proper test once something else gets pushed.

momoko-h:
Yes, that worked:
```
[00:01:04] Warnings:
[00:01:04]  - sccache - sccache v0.5.4 already installed.
[00:01:04]  Use --force to reinstall, specify a version to install, or try upgrade.
```

But sccache added a 'v' to its version tag in the meantime.

lynxlynxlynx:
Not sure if you wanted to do anything else, but LGTM — feel free to merge.

PR #1862, opened 2023-07-30
Use the old way to load squads if save is of an older version

### LINKED ISSUE #1861: Crash when loading original 1.02 savegames
[//]: # (If you have more than one item to report, please open separate issues for each one!)

### Invalid memory accesses during loading of original 1.02 savegames

During the loading process of an old savegame from the original JA2 German 1.02 version, the ja2-stracciatella crashes.

|  ja2-stracciatella_0.20.0-git+439a6d3_win-msvc64-142 | eg. JA2 1.02 German  |   ?    |    yes, e.g. with 0.17     |

I debugged the issue, and at least one reason (not sure yet if there are more further along) seems to be commit 24dc17101 "* Update the Squad save game data format".

Inside `LoadSquadInfoFromSavedGameFile()` variable `ubSquadID` has a very suspicious, high value, which leads to out-of-bounds access in `slotPos`.
I suspect the issue is the way the commit introduced the `ubFormatVersion` distinction between the old and new save format. Inside `SaveSquadInfoToSavedGameFile` `SQUAD_INFO_FORMAT_VERSION` (==1) is written for the new format, but before this area was just skipped via `INJ_SKIP`, which I presume does not necessarily mean zero-value bytes, but can be any random garbage value.

What makes the crash more likely is that in `LoadSquadInfoFromSavedGameFile()` the check `if (ubFormatVersion == 0)` is performed. So uninitialized data will in nearly all cases lead to the code erroneously  assuming the new-style savegame format.

It seems to me encoding a squad info version number inside the save data area for every squad is not a good idea. Isn't there a more global version number? Alternatively maybe the check could be hardened by introducing an integrity check via a hash value, although that might be a bit over the top.

ja2-derek:
Fixes #1861

### DIFF
--- a/src/game/Tactical/Squads.cc
+++ b/src/game/Tactical/Squads.cc
@@ -688,7 +688,7 @@ void SaveSquadInfoToSavedGameFile(HWFILE const f)
 }
 
 
-void LoadSquadInfoFromSavedGameFile(HWFILE const f)
+void LoadSquadInfoFromSavedGameFile(HWFILE const f, UINT32 const uiSaveGameVersion)
 {
 	// Load in the squad info
 	BYTE data[SQUAD_INFO_NUM_RECORDS * 12] = { 0 };
@@ -715,7 +715,7 @@ void LoadSquadInfoFromSavedGameFile(HWFILE const f)
 
 		if (id < 0) continue; // skip empty slot
 
-		if (ubFormatVersion == 0) ubSquadID = i / 6; // use the implicit squadID
+		if (uiSaveGameVersion < 102 || ubFormatVersion == 0) ubSquadID = i / 6; // fixed squads of six, if we are loading a save from old versions
 
 		if (slotPos[ubSquadID + extraSquads] == gamepolicy(squad_size)) extraSquads++; // squad already full, split to a new one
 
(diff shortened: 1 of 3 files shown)


momoko-h:
I had a look at GameVersion.cc and at the time you #1411 was checked in, we were still at version 101:

https://github.com/ja2-derek/ja2-stracciatella/blob/24dc17101d7a4521979cc79546bdfc1ad4d91984/src/game/GameVersion.cc#L23

Fortunately, you had bumped the version from 100 to 101 just two weeks earlier in #1282, so the risk that there are version 101 save files out there with the old squad format is extremely slim.

src/game/Tactical/Squads.cc
```diff
@@ -715,7 +715,7 @@ void LoadSquadInfoFromSavedGameFile(HWFILE const f)
 
 		if (id < 0) continue; // skip empty slot
 
-		if (ubFormatVersion == 0) ubSquadID = i / 6; // use the implicit squadID
+		if (uiSaveGameVersion < 102 || ubFormatVersion == 0) ubSquadID = i / 6; // fixed squads of six, if we are loading a save from old versions
```
momoko-h, line 718: So I think this should be uiSaveGameVersion < 101 || (uiSaveGameVersion == 101 && ubFormatVersion != SQUAD_INFO_FORMAT_VERSION)

momoko-h:
LGTM

PR #1865, opened 2023-08-02
Force a cursor mode switch when the current targt soldier dies

### LINKED ISSUE #1778: Crosshair doesn't change back to movement after hovering an enemy
### What is wrong

Hovering over a dying enemy in real-time mode switches the cursor to a crosshair, but moving the mouse cursor away doesn't change it back to movement - unlike the turn-based mode where it immediately changes back to the movement cursor.

|  0.20.0, 0.21.0-dev | JA2 GOG English  |    no    |    yes, regression from 0.19.0 (broken by 3352e5d71d7db7b64eeb875c9aa74ae4fc21f559)     |

**Reproduction**

1. get an enemy into a Dying state
2. manage two turns without either them bleeding out or another enemy showing up
3. now in real-time, hover the mouse over the enemy

Here's a save to try it right away
[2023-02-03t22-24-08z-potentialdyingsave.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/10609663/2023-02-03t22-24-08z-potentialdyingsave.zip)

**Env**
* Linux (Fedora 37)
* using the x86_64 AppImages

momoko-h:
The cursor could get stuck in a no longer valid mode if the current gUIFullTarget died, for example from bleeding out. Fixes #1778.

Given the byzantine labyrinth of functions and global variables that are involved in handling the tactical cursor my confidence that this is the best way to fix this is fairly low. I believe the worst thing that could happen is that the cursor now changes to move mode when someone dies where this currently does not happen and might not be desirable, but right now I can't think of such a situation.

### DIFF
--- a/src/game/Tactical/Soldier_Create.cc
+++ b/src/game/Tactical/Soldier_Create.cc
@@ -964,7 +964,12 @@ static void InitSoldierStruct(SOLDIERTYPE& s)
 void InternalTacticalRemoveSoldier(SOLDIERTYPE& s, BOOLEAN const fRemoveVehicle)
 {
 	if (GetSelectedMan() == &s) SetSelectedMan(0);
-	if (gUIFullTarget    == &s) gUIFullTarget   = 0;
+	if (gUIFullTarget    == &s)
+	{
+		gUIFullTarget = nullptr;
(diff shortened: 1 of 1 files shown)


Nindaleth:
Thanks for taking care of this!

PR #1867, opened 2023-08-04
Correct documentation of minimum required SDL2 version

momoko-h:
I unintentionally raised this when I used SDL_CreateRGBSurfaceWithFormat in 67654996626ea5ceb20abe5236f4d625e6063b42. So far nobody has complained and I don't believe requiring your SDL version to be less than seven years old is too much too ask.


lynxlynxlynx:
Please add the minimum to the main CMakeLists.txt too, so it will fail nicely. IIRC find_package can take a version; if not, it's trivial to do manually.

lynxlynxlynx:
Thanks!

Issue #1871, opened 2023-08-09
Can reliably reproduce crash with error code -1073741819


MoerasGrizzly:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
I've noticed the game crashing a bit when giving movement orders. Was able to create a savegame where this crash always happens.
Crash occurs on both stable and a recent nightly.
Crash seems to be tied to mercs falling asleep whilst walking.

Here's the savefile. Just enable time compression and the crash should occur automatically.
[2023-08-09t16-17-23z-crash.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/12304766/2023-08-09t16-17-23z-crash.zip)

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  release&git | eg. JA2 GOG  |    unknown    |    unknown     |

```2023-08-09T16:20:31 [INFO] stracciatella::logger: Logging to file "C:\\Users\\Grizzly\\AppData\\Local\\Temp\\ja2.log"
2023-08-09T16:20:31 [INFO] stracciatella::vfs: VFS item with priority 25: DirFs { "C:\\Users\\Grizzly\\Documents\\JA2\\data" }
2023-08-09T16:20:31 [INFO] stracciatella::vfs: VFS item with priority 24: DirFs { "C:\\Games\\ja2-stracciatella_0.20.0-git+439a6d3_win-msvc64-142\\externalized" }
2023-08-09T16:20:31 [INFO] stracciatella::vfs: VFS item with priority 23: DirFs { "C:\\Games\\Jagged Alliance 2\\data" }
2023-08-09T16:20:31 [INFO] stracciatella::vfs: VFS item with priority 22: SlfFs { source: DirFsFile { "intro.slf" in "C:\\Games\\Jagged Alliance 2\\data" } }
2023-08-09T16:20:31 [INFO] stracciatella::vfs: VFS item with priority 21: SlfFs { source: DirFsFile { "bigitems.slf" in "C:\\Games\\Jagged Alliance 2\\data" } }
[...]

PR #1872, opened 2023-08-10
Do not dereference a null pointer in RebuildWayPointsForGroupPath

momoko-h:
Also change a couple of assertions to SLOGE messages. This avoids the crash with the save game provided in issue #1871.

The strategic pathfinding problem is still there, that's why I'd leave 1871 open, it has a nice testcase. This is what it looks like with this commit:

```
2023-08-10T09:14:16.024072348Z [ERROR] game/Strategic/Strategic_Movement.cc: Invalid DIAGONAL waypoint being added for groupID 30. AM-0
2023-08-10T09:14:16.024268076Z [ERROR] game/Strategic/Strategic_Movement.cc: Invalid DIAGONAL waypoint being added for groupID 30. AM-0
2023-08-10T09:14:16.024369941Z [ERROR] game/Strategic/Strategic_Movement.cc: Invalid DIAGONAL waypoint being added for groupID 30. AM-0
2023-08-10T09:14:16.024461411Z [ERROR] game/Strategic/Strategic_Pathing.cc: Path exists, but no waypoints were added!  AM-0
```

### DIFF
--- a/src/game/Strategic/Strategic_Pathing.cc
+++ b/src/game/Strategic/Strategic_Pathing.cc
@@ -794,7 +794,6 @@ void RebuildWayPointsForGroupPath(PathSt* const pHeadOfPath, GROUP& g)
 	INT32 iOldDelta = 0;
 	BOOLEAN fFirstNode = TRUE;
 	PathSt* pNode = pHeadOfPath;
-	WAYPOINT *wp = NULL;
 
 	//KRIS!  Added this because it was possible to plot a new course to the same destination, and the
 	//       group would add new arrival events without removing the existing one(s).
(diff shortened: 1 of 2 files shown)


src/game/Strategic/Strategic_Pathing.cc
```diff
@@ -868,10 +867,18 @@ void RebuildWayPointsForGroupPath(PathSt* const pHeadOfPath, GROUP& g)
...
 
+		if (wp == nullptr)
+		{
+			SLOGE("Path exists, but no waypoints were added!  AM-0" );
+		}
+		else if (pNode->uiSectorId != static_cast<UINT32>(wp->sSector.AsStrategicIndex()))
+		{
+			SLOGE("Last waypoint differs from final path sector!  AM-0");
+		}
+	}
```
lynxlynxlynx, line 881: Why the extra block?

lynxlynxlynx: It makes the code uglier though. Also, if anyone tried to add code that would use wp without ensuring it's not null, coverity would file a complaint anyway.

momoko-h: Uglier by two lines and one indentation level? Also, I prefer to not have it when bad code doesn't even compile instead of having to wait until Monday for Coverity to tell me about it (that is, if it finds the problem at all and it doesn't get buried in the other 374 defects).

PR #1876, opened 2023-08-18
Fix bug in two FindGridNoFromSweetSpot* functions

momoko-h:
The method these functions used to filter out gridnos without a path to the sweet spot was broken: uiRange was set to 999 and then compared to uiLowestRange. However, since uiLowestRange initial value is 999999, this invalid gridno could be considered as the best, and if there are no gridnos with a valid path in the search range, it could be returned as the result.

I believe this bug is not very likely to happen, but it could potentially the source (or one of the sources) of the path finding warnings we get occasionally.


lynxlynxlynx:
Nice find!

PR #1878, opened 2023-08-20
Fix heavy weapons chance to hit

magras:


magras:
There is a think to consider: the game might have been balanced with this bug in effect.

lynxlynxlynx:
The change to effective level looks ok, but I don't agree with the 10x factor. It just seems too favourable. That alone would enable a 100 % chance to hit before other modifiers are applied for a maxed out merc. You can get +15/+30 from traits, +5 from morale, +10 from the same target on top, before considering aiming bonuses (+10 per point by default).

magras:
@lynxlynxlynx, I'm still on a fence with this change, but is it possible to get aim bonus for mortar and similar weapons? I see aim handling in the code, but I don't think that it's possible in the game. LAW is the only exception and it is actually use the correct formula already: https://github.com/ja2-stracciatella/ja2-stracciatella/blob/455a8b71097a1c72852f05bab4485e18d5375dfd/src/game/Tactical/Weapons.cc#L2060

lynxlynxlynx:
True, no aim bonuses for the mortar.

momoko-h:
So this change would increase iChance for a level 8 merc with an average of 80 in effective marksmanship, dexterity and wisdom from 62 to 80.  That seems quite high; I don't believe the game developers intended mortars to be very accurate. 
It would also greatly diminish the value of the heavy weapons trait: gaining 3 levels has the same effect as being a heavy weapons expert? Compare that to the branch just above: more experience as a merc doesn't turn you into a MLB pitcher.

magras:
> It would also greatly diminish the value of the heavy weapons trait: gaining 3 levels has the same effect as being a heavy weapons expert? Compare that to the branch just above: more experience as a merc doesn't turn you into a MLB pitcher.

With the correct formula heavy weapons expert is equivalent to 12 levels. Level component is 1/4 of `iChance`.

The comparison with ordinal throwing weapons is strange because it's physical action, while aiming a mortar ["need brains & know-how"](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/455a8b71097a1c72852f05bab4485e18d5375dfd/src/game/Tactical/Weapons.cc#L3657).

> So this change would increase iChance for a level 8 merc with an average of 80 in effective marksmanship, dexterity and wisdom from 62 to 80. That seems quite high; I don't believe the game developers intended mortars to be very accurate.

If there were no mistakes in other arguments, I would withdraw the proposed change without question because I agree that potential increase of 22.5% CTH and practical increase of 15-20% CTH is significant, although I don't think it is game breaking.

Can I get the final confirmation that my corrections about absence of aiming for mortar like weapons and comparison to heavy weapons skill did not change your mind?

momoko-h:
Right, the effective level is inside the parentheses that get divided by 4. I suck at reading long lines...

I'm fine with just giving this a trial run to see how it plays out.  If we don't like the change in game balance we can still tone it down later by lowering the factor from 10.

Issue #539, opened 2017-06-05
Right ALT key doesn't work the same as left ALT key


JAsmine-ja2:
Because it was mentioned here: #533

In both vanilla and Stracciatella, the right ALT key (ALT Gr) doesn't work the same as the left ALT key. Instead, it works like the CTRL key.

So I can bring up the hand cursor with the left CTRL key, the right CTRL key AND the right ALT key.
I can make my mercs strafe or walk backwards with only the left ALT key.

Is this intended? Is the right ALT key really supposed to work like the left ALT key?

Vanilla 1.12 English
Stracciatella 0.15.1

Windows 8.1

Tercus:
Found a solution. After searching around for this error, I found [this forum thread](http://forums.libsdl.org/viewtopic.php?t=8590), which explains it a bit, but doesn't solve it. ALT GR is (or can be) emulated by CTRL+ALT. Some keyboards have a right ALT button, which is *not* the same as the ALT GR button. Checking for RALT thus isn't always enough.

But, after finding [this table](http://www.jose.it-berater.org/sdl/core/sdl_keysym.htm), I tried adding SDLK_MODE and that worked perfectly:
```
		case SDLK_LALT:
		case SDLK_RALT:
		case SDLK_MODE:
			gfKeyState.set(ALT);
			break;
```

Manuel-K:
A question to both of you: What keyboard layout do you use?

A second question for @Tercus: Can you please post the output of `localectl` or something similar?

Manuel-K:
I'll try to upload my simple SDL2 key test application that displays the name of the pressed key later.

JAsmine-ja2:
I use a german keyboard layout.

Manuel-K:
I've attached the test program. sdl2-key-test-src.zip contains the source code with a makefile that works on Linux. sdl2-key-test-win.zip contains an exe that I created using the mxe cross compilation environment and the source code with a modified makefile.

If you press a key while the window is focused, it should display messages on the terminal (on Windows it should open a terminal automatically). On my Windows 8.1 system AltGr is recognized as Left Ctrl and Right Alt. On all Linux systems that I tested AltGr is recognized as Right Alt.

[sdl2-key-test-src.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/1054841/sdl2-key-test-src.zip)
[...]

PR #1879, opened 2023-08-27
Improve handling of modifier keys

momoko-h:
• Add the fix for #539 suggested by Tercus
• Use the special treatment for some keys added in #1853 also
  when Num Lock, Caps Lock or Scroll Lock are active


lynxlynxlynx:
Thanks!

PR #1882, opened 2023-09-03
Fix crash when opening email box after save and load

### LINKED ISSUE #1591: Random crashes when opening emails on the in-game laptop

### Random crashes when opening emails on the in-game laptop

Not many details unfortunately, I cannot cleanly re-produce this. It seems to be creep up from time to time. Seems to occur after a fight and when new emails arrive.

```
/Users/runner/work/ja2-stracciatella/ja2-stracciatella/ci-build/lib-string_theory/include/string_theory/st_utf_conv.h:263: String data buffer is too large
[1]    67228 abort      ./ja2
```

|  v0.19.1 | JA2 Gold English  |   ?    |    ?    |

Operating system: macOS 12.3.1 (21E258), Apple M1

magras:
`CurrentMail` was not zeroed during load, so it pointed to the deleted object. Results can vary, but one of them is an assert firing in the string theory library:

```
string_theory/st_utf_conv.h:263: String data buffer is too large
```

fixes #1591

### DIFF
--- a/src/game/Laptop/EMail.cc
+++ b/src/game/Laptop/EMail.cc
@@ -279,6 +279,9 @@ void GameInitEmail()
 	pEmailList=NULL;
 	pPageList=NULL;
 
+	CurrentMail = NULL;
+	PreviousMail = NULL;
+
 	iLastPage=-1;
 
 	iCurrentPage=0;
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Nice, thanks!

Issue #1885, opened 2023-09-05
Crash to Desktop Involving Helicopter and Missing Merc


sabertooth3:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### CTD using Helicopter

The helicopter shows 1 merc (Biff) on board, however he's assigned to a squad on the ground in Drassen airport. When I try to move Biff to a different sector or go to ground view where the helicopter is, the game crashes to desktop. OS is Win11 desktop. Log is attached.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  Stratcciatella | JA2 base, DL from Amazon  |    Unknown   |    yes     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)
[ja2.log](https://github.com/ja2-stracciatella/ja2-stracciatella/files/12525647/ja2.log)

lynxlynxlynx:
Do you have a save? Can you get into the same situation from an earlier one?

You shouldn't be able to have a merc both be assigned to a squad and in a heli, so a crash is a special kind of luck.

sabertooth3:
[2023-08-21t18-04-35z-test-novice-run.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/12586921/2023-08-21t18-04-35z-test-novice-run.zip)

Yes, here's the save file that was generated from the error. I loaded an autosave from just a few game hours earlier that had the same error as well.

momoko-h:
This doesn't crash in a nightly build, but that's probably only because I changed the asserts to error messages in #1872. It would be nice to know how Biff got into this messy situation, but in the meantime I think we should add a sanity check for this situation when we load a save file.

PR #1893, opened 2023-09-18
Detect wrong soldier state when loading a VEHICLETYPE

momoko-h:
It is somehow possible for a merc to be in a vehicle's passenger list while having an assignment other than VEHICLE. This commit detects this situation and tries to bring the game state back to normal.

Obviously this doesn't fix the underlying issue that it shouldn't even possible to get into the state described in #1885, but at least now players have the chance to continue their game without having to dismiss the merc in question.


lynxlynxlynx:
LGTM, thanks!

Issue #1735, opened 2022-12-22
Strange input handling on follow-up shots.


Danf96:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
Game requires you to double click for every subsequent shot after an initial shot. Here's a save attached for quick testing with Gunther but it should occur in any game.
https://drive.google.com/file/d/1K1IBbDEK_Yl2ttSr-Uk6ca_cix-SHJ4U/view?usp=sharing

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  git+8eafd1a| JA2 Gold  |    no    |    yes   |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

Danf96:
Posting some video footage so the problem is clearer.


https://github.com/ja2-stracciatella/ja2-stracciatella/assets/8466937/9422d0e2-427f-4909-9b61-c0a35e450bd3



Note how the cursor will immediately show the crosshair with the circle showing aim mode, then switch back to a cursor can't be fired with, and the player must wait for it to reappear. On the second turn, I clicked right after shooting and the aiming cursor did not appear (which allows you to fire) did not reappear until I clicked again.



https://github.com/ja2-stracciatella/ja2-stracciatella/assets/8466937/b72d9ef5-b28f-45dd-a7c2-1a8bc63d3cb8



Here is how the original behavior was in JA2 1.12. If you watch the stracciatella clip afterwards the issue will become more apparent that the new input code is not behaving the same way as the original for PC players. I didn't record footage but stracciatella before the touch controls update behaves the same way as 1.12.

Danf96:
[...]

PR #1900, opened 2023-09-28
Cursor hides on action, reselects targets.

Danf96:
Reintroduces pre-touch screen behavior of the cursor disappearing during an action in turn-based mode as well as no longer requiring a double-click or moving the cursor off a target for follow up shots as the cursor is re-examined after an action.  Solves issue #1735


lynxlynxlynx:
Would this better fit into SetUIMouseCursor?

Danf96:
I agree it's a better fit there, I wasn't aware it was called only once at the bottom of HandleTacticalUI. I've made the changes accordingly.

PR #1901, opened 2023-09-29
Workaround for the Z-Blitter address sanitizer problem

momoko-h:
The Z-Blitter functions disagree with AddZStripInfoToVObject about the format of the zchange array in a ZStripInfo, which leads to constant complaints (with valgrind) or crashes (in ASAN builds) when blitting some multitile VObjects. Since nobody seems to fully understand what this code is supposed to do, this commit changes the zchange array to a fixed size of 16 bytes. Each byte is supposed to represent the z change of 20 pixels, so this is enough for multitile VObjects up to 320 pixels wide. AFAIK the biggest asset in the original game data is the queen monster, which requires a zchange array of 10 bytes, so 16 bytes should be plenty.


lynxlynxlynx:
Just from the bounds checks it seems it could be much higher, but let's give it a spin in the wild.

Issue #1764, opened 2023-01-20
CAWS some burst fire weapon sounds are missing/not playing.


ja2forever:
JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  ja2-stracciatella_0.20.0-git+439a6d3_win-msvc64  | JA2 Gold 1.12 English CD  |    no    |    dunno     |

CAWS have missing buckshot(blue mag) and slug(red mag) ammo burst fire sounds - checked on default 3-shot bursts, no mods:
-BS 3 or more shots left burst - missing
-BS 2 shots left burst - missing
-BS 1 shot left burst - missing
-SLG 3 or more shots left burst - OK
-SLG 2 shots left burst - OK
-SLG 1 shot left burst - missing

Screenshot with errors:

![JA2_CAWS_SOUND](https://user-images.githubusercontent.com/122088263/213758337-3573d50c-31a2-407d-b6a9-ddd2dd4d5204.png)

arnodick:
I've noticed this as well, and it seems like the value for how many shots are fired in the burst is randomized when trying to find the correct wav. (burst 9.wav, burst 27.wav etc.) Could it be retrieving a datetime value instead of the shot count?
Also, even in vanilla JA2 I believe there wasn't actually any sound file for the caws burst, and so it defaulted to bursttype1.wav, which I believe doesn't happen in stracciatella any more. Instead there is no default wav and stracciatella just doesn't play sound.

momoko-h:
No, this isn't random. If you look closely you'll notice that these numbers are multiples of 9, that's because the game treats one unit of buckshot ammunition as 9 separate bullets, with 1⁄4 damage per bullet. The problem might be solvable just like #1730: find the CAWS entry in `weapons.json` and add a `burstSound` entry (I didn't test this hypothesis).

selaux:
I think with this new information even the fix for #1730 might be flawed, because the correct sound to play also depends on the number of bullets remaining in the gun. So even the RPK now probably always plays the "full" burst in all cases, even when only a single bullet remains in the gun.

arnodick:
Interesting. For me it kind of stick to multiples of 9, but I've gotten other values sometimes.

The CAWS has shotgun burst %d.wav as its burstSound. Every .wav file except shotgunburst1.wav is in the sounds.slf file.

[...]

PR #1902, opened 2023-09-29
Rework burst sound logic

momoko-h:
This should fix the problem of the missing sound for CAWS buckshot ammo and the RPK74 using the generic sound even when the 5.45 burst sound is available (when we have four or less shots left). Issues #1730 and #1764.

This could definitely use more testing, I'd appreciate help here. This is what I found with two known problematic weapons:

|CAWS, CAWS buckshot magazine||
|---|---|
|>=3 bullets left| bursttype1.wav|
|2 bullets left| bursttype1.wav|
|1 bullet left| bursttype1.wav|

|CAWS, CAWS magazine||
|---|---|
|>=3 bullets left| shotgun burst 3.wav|
|2 bullets left| shotgun burst 2.wav|
|1 bullet left| shotgun single shot.wav|

|RPK74||
|---|---|
|>=5 bullets left:|bursttype1.wav|
|4 bullets left| 5,45 burst 4.wav|
|3 bullets left| 5,45 burst 3.wav|
|2 bullets left| 5,45 burst 2.wav|
|1 bullet left| 5,45 burst 1.wav|

Does that  seem reasonable?

Save game for testing, so you don't have to set things up yourself: 
[2023-09-29t11-29-13z-burst-sound-test.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/12765552/2023-09-29t11-29-13z-burst-sound-test.zip)

### DIFF
--- a/assets/externalized/weapons.json
+++ b/assets/externalized/weapons.json
@@ -1726,7 +1726,6 @@
         "usRange": 500,
         "ubAttackVolume": 82,
         "ubHitVolume": 8,
-        "burstSound": "sounds/weapons/bursttype1.wav",
         "attachment_SniperScope": true,
         "attachment_LaserScope": true,
         "attachment_Bipod": true,
--- a/src/game/Tactical/Weapons.cc
+++ b/src/game/Tactical/Weapons.cc
@@ -2,6 +2,7 @@
 #include "Font_Control.h"
 #include "Handle_Items.h"
 #include "Overhead_Types.h"
+#include "SGPStrings.h"
 #include "Soldier_Control.h"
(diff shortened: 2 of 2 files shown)


Danf96:
Tested it and didn't see any issues with any burst-fire weapons. One thing I did notice (but isn't a result of this patch as it occurs in older builds and 1.12) is that in real-time mode if you click too quickly for a follow-up burst for a weapon like the G3, the two round burst sound will play instead of the three round burst but I don't think that's in the scope of this commit.

selaux:
Looks good to me as well.

Issue #1864, opened 2023-08-01
[Windows] Change Stracciatella config dir location to AppData


ChristianKaeser:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### Proposal for a better Stracciatella configuration directory location on Windows

On Windows, Stracciatella's configuration directory defaults to `dirs::document_dir()` (via `find_stracciatella_home()`).
This seems suboptimal to me.

For one, most users probably never/super rarely need to interact with this directory. Stracciatella's launcher application can be used to modify the settings which are stored there.
Furthermore, the save game directory can be customized anyway, and then the "SavedGames" subdirectory there seems to be unused (any potentially confusing).
So the folder is simply cluttering the all-important Documents location, without an option to avoid this while using Stracciatella.

While many games traditionally seem to have selected the user's Documents directory for their save and configuration data, Windows plattform recommendations rather seem to prefer %APPDATA% as the place for application configuration data.
This is also reflected in the utilized Rust "dirs" crate, which provides `config_dir()`, which points to just that.

But the most important reason for me are Windows' security defaults, which nowadays give special protection to the user's Documents and similar directories - but not to %APPDATA%: https://learn.microsoft.com/en-us/microsoft-365/security/defender-endpoint/controlled-folders

This means for me there is a security notification popup the first time JA2 Stracciatella is used, and the Launcher/Game simply exits/is not working without any indication as to what happened. This seems to be due to the processes trying to write in the config location, which Windows does not allow without the user having to give an admin-rights-protected exception with multiple clicks: https://support.microsoft.com/en-us/windows/allow-an-app-to-access-controlled-folders-b5b6627a-b008-2ca2-7931-7e51e912b034

I see this as a big hurdle for many casual users, which probably will simply give up on Stracciatella right there and then.

Of course a change would mean that a migration fallback code path would have to be introduced, so existing installations continue to work.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  ja2-stracciatella_0.20.0-git+439a6d3_win-msvc64-142 | all  |    -    |    -     |

[...]

PR #1903, opened 2023-10-06
Change configuration directory on Windows to AppData

selaux:
The new dir should be: `C:\Users\Alice\AppData\Roaming\JA2`, which is not influenced by MS Defender on Windows.

Closes https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1864.

I need someone to test this though, as I dont have a Windows system available to me.


lynxlynxlynx:
I removed it from the milestone, since the bug is already in it and this way the changelog entries won't be duplicated.

PR #1907, opened 2023-10-13
Upgrade string_theory to 3.8

selaux:
Closes the overflow bug reported by @momoko-h  https://github.com/zrax/string_theory/issues/29.

Also closes the stdint include issues.


momoko-h:
Three annoying problems solved with just two changed lines, nice.

PR #1908, opened 2023-10-15
Load the big map of Arulco just once

momoko-h:
b_map.pcx was loaded twice and copied to a surface twice during startup, once just to compute the green and red alternative palettes for the map. The palettes are now generated later, from the surface that is actually used during the game.

Not really a big deal, modern CPUs are just too fast to care about loading a ~450k PCX file.


lynxlynxlynx:
LGTM, thanks!

PR #1911, opened 2023-10-16
Clarify Lua 5.3 version dependency

timp87:
The project needs exactly lua 5.3 to build.
Add EXACT to lib-lia's CMakeLists.txt.


timp87:
Hmm, you made me doubt. I look at my build log and not sure if there is a problem actually.
Let me recheck everything and get back.

ja2-derek:
The current release of sol3 ([3.3.0](https://github.com/ThePhD/sol2/releases/tag/v3.3.0)) mentions:

> ... and fixes for Lua 5.4.

Maybe we should try upgrading sol3 too.

timp87:
I'm so sorry for the noise, it's builds ok with lua54

Issue #1693, opened 2022-10-31
Warn if 1.13 is installed


lynxlynxlynx:
We get clueless users often enough that it would be cool if the launcher detected a 1.13 install and warned about it. I'm sure we can create a list of let's say 3-5 marker files that are likely to be installed in any configuration and test for them.

Perhaps our `assets/externalized/resource_packs` mechanism can be reused?

ja2-derek:
Found a repo for the 1.13 game dir: https://github.com/kitty624/JA2-v1.13-7609-GameDir

Some obvious files to look for:
- Ja2.ini
- Data-1.13/
- Data/Ja2Set.dat.xml
- Data/Ja2_Options.INI

hypertexthero:
Is 1.13 support planned or actively being worked on by someone?

lynxlynxlynx:
No. To quote the website:

> Stracciatella walks a fine line between preserving the original gameplay, adding ease-of-life improvements and providing new inobtrusive optional features.

hypertexthero:
Understood, thank you!

lynxlynxlynx:
From what I can tell our ResourcePack stuff in guess.rs relies on archives and their hashes. For this we can't use hashes or sizes, since the contents will be arbitrary. A simple existence check is enough.

PR #1913, opened 2023-10-17
Warn about 1.13 modified game files in launcher

selaux:
Now a warning is shown after three user interactions

- Selecting a game directory
- Autoguessing version
- Starting the game

whenever one of the following conditions is true:

- Game directory does not contain `Data` dir
- Game directory contains a 1.13 data file

The user has the choice to continue anyways or cancel the action.

As discussed in #1693.

image

### DIFF
--- a/android/app/src/main/java/io/github/ja2stracciatella/GameDir.kt
+++ b/android/app/src/main/java/io/github/ja2stracciatella/GameDir.kt
@@ -1,0 +1,73 @@
+package io.github.ja2stracciatella
+
+import android.app.AlertDialog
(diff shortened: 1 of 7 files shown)


selaux:
True, I will add some logic to the Android launcher as well.

selaux:
The logic has now been added to the Android launcher as well. Unfortunately we cannot access the Rust library from Kotlin yet, so it's custom logic.

I also added an error log when the game is started with a detected 1.13 installation.

lynxlynxlynx:
Thanks!

PR #738, opened 2018-10-29
WIP: Proper video scaling, sdl blitting, 32bpp resource support, etc.

maxrd2:
Have done an experiment with scaling all resources to screen size immediately as they are loaded to improve size and look of the screen graphics.

Here is the progress of the planned changes:

- [x] Base resolution in-game is increased from 640x480 to whatever is setup in json file, screen is setup in 32bpp
- [x] All image resources when loaded are immediately decoded into 32bpp RGBA and scaled/interpolated proportionally to configured resolution
- [x] Bitmap fonts are properly rendered
- [x] Intro, Main menu, Options, Credits screens are scaled and working/looking properly
- [x] Strategic interface screen is scaled and working/looking properly
- [x] Tactical interface screen is scaled and working/looking properly
- [ ] Laptop screen is scaled and working/looking properly
- [x] All blitting functions are reworked to use 32bpp
- [ ] Ability to dump loaded and scaled resources into png files into some path (so some artist could edit them and make them look better on higher res)
- [ ] Ability to override loading of official resources (if png file is found in override path it's loaded instead of official res)
- [ ] Add support for freetype fonts.
- [ ] Replace all/most blit code with SDL function calls and load resources as SDL textures

Am not sure if SDL 2D allows for z buffers which are used for blits on tactical map.

Anyone has any thoughts on this? Is this something you guys wanted to implement at some point?

PR #1923, opened 2023-10-27
RFC: Stop hiding the SDL surface

momoko-h:
As you know, a SGPVSurface is just a thin wrapper around a SDL surface. SGPVSurface tries to hide this fact by making the pointer to the SDL surface a private field. This approach probably made sense when this project started and made the transition from DirectDraw to SDL; this is pure speculation on my part, but I may have been done to keep the option open to use something else other than SDL.

IMO nowadays this design just hurts us. We're just denying ourselves easy access to tons of functionality SDL has to offer; instead we're stuck with terrible APIs from vanilla like LineDraw.

This PR doesn't really do all that much:
1. Allow getting the SDL surface pointer non-const
2. Create a SDL software renderer for the surface
3. Add four convenience methods for this renderer

This small change already lays the foundation to move Line.cc to the trashcan. You'll notice that a nice side effect of this change: it removes our dependency on the pixel format of the surface, something that hurts us so much in #738. We would get 32-bit line drawing for free.

I'd welcome input on this idea. Am I missing something here, some downside I'm not seeing? 

IMO the upcoming release would be a great opportunity to make this change. I don't foresee any huge problems, but it is invasive; new bugs could certainly happen.

### DIFF
--- a/src/sgp/VSurface.h
+++ b/src/sgp/VSurface.h
@@ -20,8 +20,10 @@ inline SGPVSurface* g_mouse_buffer;
 struct SDLDeleter
 {
 	void operator()(SDL_Surface *p) noexcept { SDL_FreeSurface(p); }
+	void operator()(SDL_Renderer *p) noexcept { SDL_DestroyRenderer(p); }
 };
 using SurfaceUniquePtr = std::unique_ptr<SDL_Surface, SDLDeleter>;
+using RendererUniquePtr = std::unique_ptr<SDL_Renderer, SDLDeleter>;
 
 
 /** Utility wrapper around SDL_Surface. */
@@ -30,7 +32,7 @@ class SGPVSurface
 	public:
 		SGPVSurface(SDL_Surface*);
 		SGPVSurface(UINT16 w, UINT16 h, UINT8 bpp);
-		virtual ~SGPVSurface();
+		~SGPVSurface();
 
 		UINT16 Width()  const { return surface_->w; }
 		UINT16 Height() const { return surface_->h; }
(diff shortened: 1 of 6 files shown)


src/sgp/VSurface.h
```diff
@@ -50,8 +52,45 @@ class SGPVSurface
...
 		void ShadowRectUsingLowPercentTable(INT32 x1, INT32 y1, INT32 x2, INT32 y2);
 
-		/* Allow read access to the underlying SDL_Surface */
-		SDL_Surface const& GetSDLSurface() const noexcept { return *surface_; }
+		/* Allow access to the underlying SDL_Surface */
+		SDL_Surface & GetSDLSurface() const noexcept { return *surface_; }
+
+		/* Return a software renderer for this surface. The renderer is created
+		 * on demand the first time this method gets called and will be
+		 * destroyed together with the surface. */
```
lynxlynxlynx, line 60: I don't know how long-lived our SGPVSurfaces are or if SDL_CreateSoftwareRenderer is an expensive call (doesn't look like it from SDL source). Is it something we should worry about?

momoko-h: 99% of all uses of the software renderer will probably be on the frame buffer, which is created once at startup and lives until the program exits. I came to the same conclusion after looking at the SDL sources, a creating a software renderer seems to involve nothing more than filling out a couple of fields based on the surface's value. I just didn't want to create a renderer for every surface because most will never need one.

But now that we seem to have managed to motivate @maxrd2 to continue his work on #738, we may want to put this on hold. If I create even more merge conflicts for him he'll probably hit me. :laughing:

momoko-h:
I don't intend to pursue this further, it's doubtful the benefits outweigh the risk of modifying not pretty but working code.

PR #1927, opened 2023-10-29
Reduce CPU usage of the main menu screen

momoko-h:
Due to some leftovers from really ancient debug code which used the main menu screen to test the display of foreign language fonts, the main menu screen redrew its buttons every game loop. The result was that the game required a highly inflated number of CPU cycles to render a static image.

Some callgrind numbers (IR per call, inclusive)

| Function    | Unpatched |Patched|
| -------- | -------: |-----:|
| GameLoop  | 661460    |32112|
| MainMenuScreenHandle | 246454     |1449|

Looks like [this junk](https://github.com/dariusk/ja2/blob/876ccf5dfdad7e6821c5b26d6d783132ea1ab7a2/ja2/Build/MainMenuScreen.c#L661) wasted plenty of energy over the last 25 years...

### DIFF
--- a/src/game/MainMenuScreen.cc
+++ b/src/game/MainMenuScreen.cc
@@ -32,7 +32,6 @@
 #include <string_theory/format>
 
 
-//#define TESTFOREIGNFONTS
 
 // MENU ITEMS
 enum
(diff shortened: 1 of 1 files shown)


src/game/MainMenuScreen.cc
```diff
@@ -338,7 +330,6 @@ static void RenderMainMenu(void)
 {
 	BltVideoObject(FRAME_BUFFER, guiMainMenuBackGroundImage, 0, STD_SCREEN_X,       STD_SCREEN_Y     );
 	BltVideoObject(FRAME_BUFFER, guiJa2LogoImage,            0, STD_SCREEN_X + 188, STD_SCREEN_Y + 15);
-	BltVideoSurface(guiSAVEBUFFER, FRAME_BUFFER, 0, 0, NULL);
```
lynxlynxlynx, line 350: Was this supposed to be in the previous commit?

momoko-h: Yes, I just had overlooked it. Copying the frame buffer to the save buffer is only necessary when we must use `RestoreExternBackgroundRect()` to restore parts of the frame buffer later. The main screen however is so simple (just two images and five quick buttons) that there isn't anything in the frame buffer that needs to be saved.

PR #1930, opened 2023-11-05
Limit the damage to items done by alt+o to almost nothing

momoko-h:
This should only do minimal damage to items: worst case is 3 points to bloodcat pelts.


lynxlynxlynx:
Thanks!

PR #1950, opened 2024-01-06
add LOCAL_SMACKER_LIB ON/OFF option

misanthropos:
use const for several variables to please compiler.
However: when using system libsmacker the colors of the text flicks is wrong/changing and the console spits out something like:
libsmacker::palette_render(s,p,size) - ERROR: overflow, 0x40 attempt to copy 5 entries from 5 to 6


dependencies/lib-smacker/CMakeLists.txt
```diff
@@ -1,3 +1,14 @@
+option(LOCAL_SMACKER_LIB "Use bundled libsmacker instead of searching the system" ON)
+if (NOT LOCAL_SMACKER_LIB)
+    message(STATUS "Using system libsmacker")
+    find_library(LIBSMACKER_found libsmacker.so)
+	if (NOT LIBSMACKER_found)
+	       message(FATAL_ERROR "libsmacker not found")
+    endif()
+    return()
```
lynxlynxlynx, line 8: this can go and the above is misindented.

lynxlynxlynx:
Why are the const changes needed? It doesn't look like any API is touched by those vars. Not that it hurts anything (for the same reason), just curious.

And is this viable at all? We made at least two changes to the original sources. In a sense your trouble is not surprising then.

misanthropos:
The const changes are necessary for libsmacker-1.1.1_p35 and later. JA2S compilation will fail otherwise.
I am pretty sure the changes we made do not affect decoding the stream. But I remember there were artifacts in some closing scenes at the time I was adding cinematics to ja2 - but I belive that went away. Something  must have changed to libsmacker in the meantime but I can not say what or how to address it. I have mailed @greg-kennedy. Maybe he can give some insight to that matter.

Kangie:
oh I pretty much duplicated this in #2354 - didn't hit those issues but I believe you - I'll see what I can work out. It looks like the best solution might be to bump the vendored version and fix the flickering issue - looks like there have been some changes to the way palettes are handled.

Issue #1951, opened 2024-01-10
In map editor when you click RPM on the civilian, it sometimes crashes


Inveriss:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong

Hello 

I noticed that this is happening especially on civilians, but on the others also.

Regards

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.20.0 | JA2 Gold 1.12 ENG  |    NO    |    YES     |

```
[//]: # I have no pattern on this. It just sometimes happens. The bug log is always literally that long as below.

2024-01-08T20:08:56 [INFO] game\GameLoop.cc: Version Label: Stracciatella v0.20.0-git
2024-01-08T20:08:56 [INFO] game\GameLoop.cc: Version #:     Build 04.12.02
2024-01-08T20:08:57 [INFO] game\Init.cc: Beginning JA2 using -editor commandline argument...
2024-01-08T20:08:57 [INFO] game\GameScreen.cc: Aborting normal game mode and entering editor mode...
2024-01-08T20:08:57 [INFO] game\Editor\EditScreen.cc: Entering editor mode...
2024-01-08T20:09:06 [WARN] game\TileEngine\Structure.cc: Possible bad structure 11
2024-01-08T20:09:06 [WARN] game\TileEngine\Structure.cc: Possible bad structure 11
2024-01-08T20:11:11 [ERROR] sgp\Debug.h: Assertion failed in E:\master1\src\game\Utils\Text_Input.cc, line 1022:

2024-01-08T20:11:11 [ERROR] sgp\Font.cc: Invalid character given U+0000
2024-01-08T20:11:11 [ERROR] sgp\Debug.h: Assertion failed in E:\master1\src\game\Utils\Text_Input.cc, line 1022:

2024-01-08T20:11:11 [ERROR] sgp\Font.cc: Invalid character given U+0000
2024-01-08T20:11:11 [ERROR] sgp\Debug.h: Assertion failed in E:\master1\src\game\Utils\Text_Input.cc, line 1022:

[...]

PR #1952, opened 2024-01-13
Fix crash in HandleTextInput()

momoko-h:
This function still tried to handle input even when there was no active input field, for example after calling DisableAllTextFields(), which could lead to a access violation when trying to dereference a NULL gpActive pointer.

STR if you want to see the crash for yourself:
1. Open the editor
2. Load sector A9
3. Click on the Mercs tab
4. Find Pacos or Fatima and right click on them
5. Click the "Merc Att" button to see their stats in disabled text fields
6. Press the right arrow or the End key


lynxlynxlynx:
Thanks! I guess it may resolve #1951 as well.

PR #1969, opened 2024-04-05
Update getrandom and tempfile with cargo update

### LINKED ISSUE #1968: ja2-0.21.0 fails to be built when linking against musl libc
As maintainer of gentoos ja2-stracciatella I have been informed about this:
https://bugs.gentoo.org/928627

This line:
```
/usr/lib/gcc/x86_64-gentoo-linux-musl/13/../../../../x86_64-gentoo-linux-musl/bin/ld: /var/tmp/portage/games-strategy/ja2-stracciatella-0.21.0/work/cargo_home/gentoo/getrandom-0.2.6/src/util_libc.rs:117:(.text._ZN9getrandom9getrandom17h056e21e14da38d0fE+0x231): undefined reference to `open64'
```

looks to me it is related to a cargo package. Is there a way we could fix that?

[musl](http://musl.libc.org/) libc is a replacement for glibc.

selaux:
Should close #1968

The Android built failure should be looked at separately.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -408,9 +408,9 @@ dependencies = [
 
 [[package]]
 name = "getrandom"
-version = "0.2.6"
+version = "0.2.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks!

PR #1974, opened 2024-04-11
Allow total weight of objects to exceed 25.5 kilograms

momoko-h:
The member OBJECTTYPE::ubWeight is kept for savegame compatibility but no longer used; instead the object's current weight is calculated when needed.

Converting from hectograms to kilograms and pounds gave me a headache so the new function Weight() returns the object's weight in grams (currently the return value is always a multiple of 100).


lynxlynxlynx:
LGTM, thanks!

Issue #1980, opened 2024-04-17
If you dismiss from your squad NPC like Skyrider, John and Mary or runaway kid Joey, when you are escorting them, then the game freezes


Inveriss:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong

Hello 

If you decide to dismiss mentioned NPCs from your squad while escorting them, then the entire game will be blocked (time compression, entering sector, laptop will stop working). I could only exit the game by ALT+X.

I checked other NPCs which required escorting and situation was the same. In vanilla everything after dismising those NPC worked fine.

Thanks

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  Stable 0.21.0 | JA2 Gold 1.12 ENG  |   NO    |    YES     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

momoko-h:
How do you dismiss them? I used the "Unescort" option from the Assign menu in strategic view on John and it worked just fine.

Inveriss:
I always was using alternative way to get rid of them from my squad and was clicking on "CONTRACT" button where you can renew your contract with other mercs. I was clicking dismiss there.

momoko-h:
Ah, ok, that even works with the Maria you can get from the Alt+y cheat, so it's easy to reproduce. I think the best option is to simply disable the Dismiss option in the contract menu.

PR #1982, opened 2024-04-21
Disable contract menu for EPCs

momoko-h:
This menu does not have useful options for EPCS; the Dismiss option could even cause serious problems (issue #1980).

Also addresses two clangd complaints about unnecessary copies and removes some duplicated code with a function call in DrawCharacterInfo().


lynxlynxlynx:
LGTM, thanks!

PR #1984, opened 2024-04-28
Fix bug returning to the game screen from the load screen

### LINKED ISSUE #1972: When you enter load game screen while clock is spinning, then you can't return to the game
[//]: # (If you have more than one item to report, please open separate issues for each one!)

### What is wrong

Hello

When we are fighting with enemies and it's their turn, we are waiting for our turn and the clock is spinning. When in such situation we open load game screen, we won't be able return to the game. We will have to load any game or quit the game. Instead of returning to the game, the option screen is appearing. 

Thanks

|  0.21.0 stable | JA2 Gold 1.12 ENG  |    NO    |    YES     |

momoko-h:
Pressing Ctrl+L to go to the Load screen during the AI's turn is a Straciatella feature not present in Vanilla, however cancelling the load screen did not work, it took you to the Options screen instead.

Fixes #1972.

### DIFF
--- a/src/game/Tactical/Turn_Based_Input.cc
+++ b/src/game/Tactical/Turn_Based_Input.cc
@@ -2162,7 +2162,7 @@ void GetKeyboardInput(UIEventKind* const puiNewEvent)
 		{
 			gfSaveGame              = FALSE;
 			gfCameDirectlyFromGame  = TRUE;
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Nice and simple, thanks!

PR #1985, opened 2024-04-29
Clean up MessageBoxScreen a bit

momoko-h:
• Pass the message box by const reference. A missing &
  caused us to do unnecessary string copies in ~70 places.
• Remove the unused default message box style. This style
  could only be reached by doing something really silly
  like DoMessageBox((MessageBoxStyleID)9, ...).
• MakeButton is now a lambda so we do not have to always
  pass the same constant arguments to it
• Combined the six seperate callback handlers that only
  differed by their return value into one common handler

If you are wondering what the deleted default style looked like, here it is: 
Screenshot_20240428_211238

### DIFF
--- a/src/game/MessageBoxScreen.cc
+++ b/src/game/MessageBoxScreen.cc
@@ -1,3 +1,4 @@
+#include "Debug.h"
 #include "Directories.h"
 #include "Font.h"
 #include "Handle_UI.h"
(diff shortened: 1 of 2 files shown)


momoko-h:
clangd warned me about the extra string copies and then I had a look at the entire source file and found it wanting. :-1:

Issue #1986, opened 2024-04-29
Increas the displayed character limit, so that the full lines in AIM and MERC Bios is useable.


Moddi:
Recently I rewrote Scope's AIM bio. I wanted to exchange the upper and the lower part of her text, so that the apology about the mistake does not take up her actual new bio. While doing so I realized, that the character limit of the actual displayed text in both parts, the actual bio and the additional info, is smaller then what text could actually be displayed on the screen in a single line. E.g. the additional info has two lines of space available, but the actually displayed character limit is roughly only 1.5 lines. Same is true for the main bio. Here, also the text stops being displayed, despite there being half a line or even more empty.

There is probably an integer number somewhere using "normal" bit numbers (32, 64, 108, etc.) and not using the number of the character limit of the actual line.

momoko-h:
The actual limits are 400 characters for the main description and 160 characters for the additional information., hardly enough to fill the available display area.

The root of the problem are simply the awful .edt files. EDT can hardly even be called a file format, it is just a blob of obfuscated UCS-2 code points (which are often even incorrectly encoded and have to be fixed up) and it is up to the user of the files to make some sense out of that blob. Since EDT has absolutely no metadata you cannot simply decree that some strings should be longer or shorter than others.

Ideally we would have some mechanism in place that would allow users to override all .edt files with JSON files, not just mercbios.edt and aimbios.edt. That would not only solve this problem; users wouldn't have to use specialized Windows-only tools to edit these files.

selaux:
Yep. I think there is already an approach that does that for some dialogue files. Not sure if it simply first searches for `.json` and then falls back to `.edt`.

ja2-derek:
We have this "test-json-dialog" mod: https://github.com/ja2-stracciatella/ja2-stracciatella/tree/master/assets/mods/test-json-dialogs/

PR #1988, opened 2024-05-07
Improve loading strings from EDT files

momoko-h:
This commit adds the following:
1. A new interface IEDT. This very simple interface has just one method to get a single string from an EDT file.
2. Class ClassicEDT is a helper class for files in the original .edt format. It does not do much, it simply opens the file and takes care of offset calculations.
3. Class JsonEDT is only available for mods via the ModPackContentManager. It uses a classic .edt file as a fallback option and allows to override individual strings with entries taken from a .json file.
4. Class EDTFile is the intended replacement for the current loadEncryptedString calls. It contains the details needed to interpret the layout of the supported .edt files.

As Derek mentioned, the quote files in the mercedt folder were already done and selaux added support for itemdesc.edt in #1575. This PR adds aimbios.edt, mercbios.edt and help.edt and should be flexible enough for the remaining files in binarydata.

This seems to work but I'm marking it as a draft because I'd like to get some feedback about the general design and more testing first. Don't forget to enable the test-json-dialogs mod to see some changes.


momoko-h:
Here is a screenshot to demonstrate that this addresses the problem mentioned in #1986. It's possible to fill the entire available area with text now.

Screenshot_20240507_083338

lynxlynxlynx:
Nice work! I only noticed a blog/blob typo, but have no comments on the architecture.

Is there any reason we wouldn't use the same implementation for the existing users, as a second step to this?

momoko-h:
VanillaItems also uses a schema so that would require a bit more work and that's time that is probably better spent adding more .edt files.

The DefaultContentManager version of loadDialogQuoteFromFile could easily be converted but the ModPack version expects the .edt.json file to contain a complete set of all quotes in that file so that it can cache these strings so that would require a redesign that probably wouldn't have much benefit.

PR #1990, opened 2024-05-07
Put the content of the primary hand in the secondary (not a pocket) when assigning doctor/repair tasks 

ShishkinP:
Currently when a task is assigned, equipped one-handed guns are put in a pocket when swapped with a toolkit or medkit, which then makes it a bit of a hassle to open every soldier's inventory and equip them back. The proposed enhancement puts items in the secondary hand so that we can quickly swap them by right clicking on team panel in tactical mode.


lynxlynxlynx:
This may change the order of things getting repaired, since the handheld item goes first AFAIK, but I don't think that matters.

ShishkinP:
>I don't think that matters

I agree. The order of repair is a much more rare concern than lots of redundant clicks this fix removes. The order still can be changed manually if the player really cares.

lynxlynxlynx:
The code LGTM, I'd just like to see another pair of eyes on the overall issue.

lynxlynxlynx:
Thanks!

PR #1992, opened 2024-05-20
Make our tools less unhappy about LogMessageST

momoko-h:
Perfect forwarding of the variadic arguments should reduce the number of Coverity's COPY_INSTEAD_OF_MOVE defects.

Added [[maybe_unused]] to tell the Compiler that it is okay that isAssert is not used when ENABLE_ASSERTS is not defined.

Since these changes require a complete recompilation of almost all C++ code I used the opportunity to mark some functions in Types.h as constexpr.


selaux:
Thanks!

PR #1993, opened 2024-05-21
Implement an easier way to call RenderTiles()

momoko-h:
Instead of passing the same three start and end coordinates pairs every time we now have a new class to hold these value; RenderTiles is a method of this class and can access its former arguments without any changes.

A little helper function allows us to pass the RenderLayerIDs directly inside the function call, no need to fill in an array first and count the IDs.


lynxlynxlynx:
thanks!

Issue #83, opened 2013-08-31
walking diagonally through large doors - impossible?


lynxlynxlynx:
Originally reported by: **JAsmine-ja2 (Bitbucket: [JAsmine-ja2](http://bitbucket.org/JAsmine-ja2), GitHub: [JAsmine-ja2](http://github.com/JAsmine-ja2))**

---

This is a minor issue, that was already present in vanilla JA2.
![issue with doors](http://img.xrmb2.net/images/679331.jpeg)
It is not possible to walk through large doors diagonally.
The merc will always turn two times instead, if you try to do it.
You can see the movement path in the screenie.

This is true for all doors and both angles.

---
- Bitbucket: https://bitbucket.org/gennady/ja2-stracciatella/issue/83

derSteFfi:
because i just came across it: this is rejected in Build/Tactical/PathAI.cc, currently around line 1075:
```C
				else if ( IS_TRAVELCOST_DOOR( nextCost ) )
				{

					// don't let anyone path diagonally through doors!
					if (iCnt & 1)
					{
						goto NEXTDIR;
					}
```

lynxlynxlynx:
I guess that is in for normal doors, where it makes sense. For larger we should skip the check at least in the internally angled diagonals. However, I have a nagging feeling that everything is cell based here and there is no concept of a wide door. If that's true, we have to be careful not to break single doors that are side by side.

RinneT:
I deleted the check @derSteFfi posted just to see what happens.
Walking through the double wide door from the screenshot above works correctly, so does walking through a single door.
The same behaviour as shown above does happen on two side by side doors as in cambria hospital.
It seems the two single doors are handled differently.

Essentially the code is useless.

I attached two savegames for testing.
[Issue83_Doors.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/614754/Issue83_Doors.zip)

JAsmine-ja2:
Do we have any opinions on this issue? Can we "fix" it without making a mess of something?

PR #1995, opened 2024-05-22
Fix double doors' diagonal travel and inside-oriented versions

ShishkinP:
This one should be linked with #83

PROBLEM
1. Diagonal travel through the middle of a double door (ddoor) is disabled, because the vanilla tile engine can only handle a travel permission for a tile depending on only 1 tile (or rather the state of a structure on that tile).
2. It doesn't matter much for the sliding door/pitdoor, because structures on both of its tiles are always in the same state (open, closed or destroyed). They still disallowed it to probably make movement consistent with standard ddoors so as not to confuse the player.
3. Of the full set of ddoors only the outside-oriented half is fully-fledged. The inside-oriented only appear in Meduna tilesets with signs of being hastily copy-pasted and given up on. They behave like this:
inside-left-ddoor-bug2
The right door from the gif even has the wrong sound for closing: metal instead of plywood.
And the left-oriented are as bad and not even placed anywhere in vanilla maps.

SOLUTION
1. Fix incorrect flags for the inside ddoors at the stage of the tile db creation
2. Add new ddoor defines pointing at 2 tiles with doors instead of 1
3. Using these defines assign correct travelcost permissions at the worldbuilding stage
4. Add new logic for using them in pathfinding

TESTING

All fixed 8 doors have been tested for:
1. diagonal travel with both doors in different states (including the perceived ones)
2. destruction of adjacent walls
3. destruction of left or right part of a ddoor
4. item and room reveal through (half-)closed ddoors
5. propagation of light and smoke through (half-)closed and (half-)open ddoors
6. NPC travel (turn-based militia battles and real time)
7. attacking with a knife someone behind a closed door

Modded map (queen's palace) with all kinds of ddoors around the lobby and removed decor structures near them: https://disk.yandex.ru/d/ccPyI1d-ryPRkw
Savefile for this mod with trained militia and a soldier with TNTs: https://disk.yandex.ru/d/tIS9PGaRpMZDTg


src/game/Tactical/PathAI.cc
```diff
@@ -1063,13 +1067,53 @@ INT32 FindBestPath(SOLDIERTYPE* s, INT16 sDestination, INT8 ubLevel, INT16 usMov
...
 
 					// don't let anyone path diagonally through doors!
-					if (ubCnt & 1)
-					{
-						goto NEXTDIR;
-					}
+					//if (ubCnt & 1)
+					//{
+					//	goto NEXTDIR;
+					//}
```
lynxlynxlynx, line 1073: Please just delete these five lines instead.

src/game/Tactical/PathAI.cc
```diff
@@ -2529,6 +2648,54 @@ UINT8 InternalDoorTravelCost(const SOLDIERTYPE* pSoldier, INT32 iGridNo, UINT8 u
 
 		switch( ubMovementCost )
 		{
+			case TRAVELCOST_DOORS_CLOSED_W_SW:
+				fDoorIsObstacleIfClosed = TRUE;
+				iDoorGridNo = iGridNo + DirIncrementer[WEST];
+				iDoorGridNo2 = iGridNo + DirIncrementer[SOUTHWEST];
+				ubReplacementCost = TRAVELCOST_DOOR;
+				break;
```
lynxlynxlynx, line 2656: This is quite repetitive. Can the whole switch be moved to a separate function that then this and FindBestPath could use (and we get to maintain only one copy)?

src/game/TileEngine/TileDef.cc
```diff
@@ -116,7 +117,35 @@ void CreateTileDatabase()
 			}
 
 			SetSpecificDatabaseValues(cnt1, gTileDatabaseSize, TileElement, TileSurf->bRaisedObjectType);
-
+			// fix incorrect double door flags. in vanilla it only affects DOOR3 in PALACE! tileset and DOOR1 in QUEEN'S TROPICAL
```
lynxlynxlynx, line 120: Ideally this whould be in a separate commit.

ShishkinP: Got it. I'll keep that in mind in the future.

lynxlynxlynx:
Thanks!

PR #1996, opened 2024-05-23
Fix disabled yellow highlight of those in the selected sector

ShishkinP:
A one-character typo in #1479 disabled yellow highlighting of the soldiers in a selected sector.
mercsInSectorHighlight


lynxlynxlynx:
Thanks!

lynxlynxlynx:
It wouldn't help in this particular case though.

PR #1998, opened 2024-05-25
Upgrade all GitHub action tasks

selaux:
I have not seen any breaking changes for our usecases, so lets see if upgrading everything works.

Edit: The issue with the Android SDK is quite weird. It seems to use the wrong java version, although all pointers point to a different one. Will continue debugging this soon.


lynxlynxlynx:
Thanks. The mac  uploaderror should be transient, while the android build now fails on some rust stuff instead.

PR #1999, opened 2024-05-25
Fix tile path-blocking on roofs due to nearby explosion

ShishkinP:
PROBLEM (in vanilla too)

Structure-destroying explosions make nearby roof tiles obstacle-blocked if occupied by soldiers at the moment of the explosion.
Steps to reproduce:
1. Load this savefile: https://disk.yandex.ru/d/NhSrPal-hll_zw
2. End turn
3. The enemy soldier will attack Gus with a LAW, destroying some ground level structures in the process:
LAWattack
(LAW missiles going through roofs unimpeded is an unrelated bug)
4. Within a certain radius the roof tiles occupied at the moment of the explosion by soldiers will be blocked for travel:
scullyBlocked
gusBlocked
clownBlocked
5. The same story will be with destruction by TNT, for example.

SOLUTION

Add a check excluding anything flagged STRUCTURE_PERSON (which includes monsters and bloodcats, by the way) from explosion-induced travel cost recompilation.

TESTING

People structures are supposed to block travel into a tile by WhoIsThere2 test anyway, not by assigning travel costs to the world, so I don't expect any complications from this.

### DIFF
--- a/src/game/TileEngine/WorldDef.cc
+++ b/src/game/TileEngine/WorldDef.cc
@@ -987,7 +987,7 @@ static void CompileTileMovementCosts(UINT16 usGridNo)
 			}
 			else
 			{
-				if (!(pStructure->fFlags & STRUCTURE_PASSABLE || pStructure->fFlags & STRUCTURE_NORMAL_ROOF))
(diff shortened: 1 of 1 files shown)


momoko-h:
The movement cost values are actually part of the debug info you get when you press Y. This info confirms that the costs for the grids below the guys after the explosion are 253 without this PR and 10 with so this LGTM.

PR #2001, opened 2024-05-26
Fix hang while displaying FOV debug info

momoko-h:
In debug builds (when _DEBUG is defined) you can press Num Lock to display a soldier's FOV when he is moving. The problem is that he JA2 clock is no longer updated on a seperate thread so you can't busy loop on GetJA2Clock() to freeze the program for a constant duration, it would just enter an infinite loop. Use SDL_Delay() instead.


lynxlynxlynx:
Thanks!

PR #2002, opened 2024-05-28
Attempt to work around crashes with the MSVC build

momoko-h:
Apparently MSVC's STL does not like how we use initializer_list so see if we can replace them with vectors to get a working build.


ShishkinP:
No crashes so far. Thank you!

momoko-h:
It's not pretty but better than a constantly crashing and useless program.

selaux:
Thanks.

PR #2006, opened 2024-06-06
Fix jump location validation

ShishkinP:
PROBLEM

1. Jumping over the base tile of a soldier is too tricky. First you have to trigger blocked path cursor over their legs, then move it to the jump location in the pattern shown in this pic:
jump1
2. You can't even jump over someone facing west, because their "extended" structure is flagged as passable, so there is no blocked path cursor to take advantage of:
jump2
3. There is no block test for jumper's source tile, only for destination, which allows you to phase through walls and doors:
jump3

SOLUTION

Add a check that takes into account selection cursor.
Add a test for jumper's source tile block.


lynxlynxlynx:
LGTM, thanks!

PR #2017, opened 2024-06-23
Fix several compiler warnings

momoko-h:
• Default signedness of char is platform-dependent, which leads to compiler warnings while compiling libsmacker
• Assignment in TileDef should be a comparison
• PAGE_SIZE macro is already defined on some platforms
• Unused macros warnings


lynxlynxlynx:
Thanks. I'll go look at the v4 failures now, so the builds will be fine again.

PR #2022, opened 2024-06-25
github-ci.yml: upload-artifact@v4 requires unique names

lynxlynxlynx:
https://github.com/actions/download-artifact/blob/main/docs/MIGRATION.md

This is basically just for PR builds, if one wants to look at the artifacts directly. We upload to our infra in a previous step already.


selaux:
Thanks. I will have a look at the Android stuff soonish.

PR #2024, opened 2024-06-30
Refactor SearchForClosestMapEdgepoint functions

momoko-h:
Functions SearchForClosestPrimaryMapEdgepoint and
SearchForClosestSecondaryMapEdgepoint were almost identical except for a switch statement to select the vector to use at the start of these functions. This commit moves this switch to a seperate function, which makes one copy of the search functions redundant.


lynxlynxlynx:
LGTM, thanks!

PR #2026, opened 2024-07-05
Fix Android build by upgrading minimum Android SDK to 21

selaux:
We introduced a dependency on some functions which only become available with the Android SDK 21 (equivalent to Android 5). As Android 4 (the previous equivalent minimum version) is already > 10 years old I think this requirement should be fine (covers 99.5% of all running Android devices according to https://apilevels.com/).


lynxlynxlynx:
Cool, thanks, we're back to building everywhere. :)

PR #2029, opened 2024-07-10
Externalize NPC script records

ShishkinP:
There doesn't seem to be a task related to NPC scripting in #665. Unless it's "Dialogues" but script records define any scripted actions not just dialogues. Also, this has nothing to do with quote strings and sound files for dialogues.

Some notes:
- ~~The records are split into NPC and RPC parts purely to avoid one gigantic file.~~
- The global script record array is now filled with ALL NPCs on game startup. The old way of populating it on sector load would lead to noticeable stutter, because of reiterated schema validation and traversal, if there are more than 1 profiled NPC in the sector.
- Queen and Elliot have a separate file for meanwhile "cutscenes" due to 50 records per NPC limitation so they have to be extracted on each meanwhile load instead of game startup


src/externalized/tacticalai/NPCQuoteInfo.cc
```diff
@@ -0,0 +1,141 @@
...
+	auto jsonRecords = reader["records"].toVec();
+	int recordIndex = 0;
+	auto buf = std::make_unique<NPCQuoteInfo[]>(NUM_NPC_QUOTE_RECORDS);
+	for (int i = 0; i < jsonRecords.size(); i++)
+	{
+		auto jsonRec = jsonRecords[i].toObject();
+		recordIndex = jsonRec.GetUInt("index");
+		NPCQuoteInfo* rec = &buf[recordIndex];
+		rec->ubIdentifier = recordIndex;
+		if (jsonRec.getOptionalBool("alreadySaid")) rec->fFlags |= QUOTE_FLAG_SAID;
```
lynxlynxlynx, line 55: This should be fine, but is relying on default zero-initialization of `fFlags`. To remove any future doubt, I suggest you use normal assignment.

momoko-h:
You're probably using Visual Studio, that's why your build succeeded? You'll have to add a CMakeLists.txt for your new tacticalai directory although I wouldn't mind seeing your NPCQuoteInfo.{cc|h} files in externalized/content instead then you'll have add the new source file to the existing cmake file.

lynxlynxlynx:
Very nice! Looks like there are some linking issues though.

Btw, that issue is not supposed to be a definitive list. For users we just point them to the files ATM:
https://ja2-stracciatella.github.io/mods/#other-externalized-settings

src/externalized/DefaultContentManager.cc
```diff
@@ -1333,6 +1354,41 @@ const VehicleModel* DefaultContentManager::getVehicle(uint8_t const vehicleID) c
...
+	ST::string filename;
+	if (profile.isPlayerMerc() || (profile.isRPC() && profile.isRecruited()))
+	{
+		filename = "script-records-recruited.json";
+		auto json = readJsonDataFile(filename);
+		return NPCQuoteInfo::deserialize(json.toVec()[0], this, this);
+	}
+	else
+	{
+		return (NPCQuoteInfo*)m_scriptRecords[profileId];
```
momoko-h, line 1369: This cast is needed because m_scriptRecords is a vector of NPCQuoteInfo const *, right? Ideally all function's returning NPQuoteInfo * should be checked if they should not return a pointer to const instead but if you don't want to do this now please use a const_cast here.

momoko-h:
Did you use a script to convert the .npc files to .json? If yes, it could be useful to have this script in the repository to make life a little bit easier for modders.

ShishkinP:
> Did you use a script to convert the .npc files to .json?

No, I inserted my extraction code into the codebase and made it use its LoadQuoteFile(). Plus, instead of proper json serialization it's just a bunch of fprintf's. Long story short: it's unpublishable in its current state.

selaux:
First of all: Nice :). Could we make the enums in json a string instead of a number. So the schema in general would be (if all values are known, we wouldn't even need the integer):

```yaml
oneOf:
  - type: string
    title: Quest enum
    enum:
      - QUEST_CREATURES
      - ...
  - type: integer
     title: Quest index
     minimum: 0
     maximum: 49
```

I think this makes the json more readable and the external interface for mods less dependent on specific integer values. Which might become important when mods are able to add new values.

ShishkinP:
> Could we make the enums in json a string instead of a number

Yes, it's doable.

Another concern is the json file size. Suppose a modder wants to add script records for only a couple NPCs. Now they have to include one of those big json files in their mod's data dir . Instead, I want to:
1. Merge back together those script-records-NPCs.json and script-records-RPCs.json. (as they are functionally the same this split only adds to confusion)
2. Add another empty file called something like script-records-modextra.json where they can put their small amounts of records instead. It will add to or overwrite the records in that one big file.

The other solution, which I actively don't like, is to make around 100 files, separate for each NPC. Let me know if I'm missing something or if it's even a problem.

selaux:
Yeah. I think we have this issue in multiple places. E.g. when someone just wants to change a few items, they need to add the full item list. That has been bugging me for some time as well. Ideally there would be a generic way for most of the JSON data to do this. One thing I thought of in the past would be to also allow this patch-like structure:

```js
{
  // Overwrite records for Miguel (or create new one if Miguel wouldn't have existed in the original dataset)
  "MIGUEL": {
     "records": [ ... ]
  },
  // Delete records for Carlos
  "CARLOS": null,
}
```

This would allow us to do a layered loading approach, loading files from all enabled mods:

- If the input is an array overwrite everything
- If the input is an object as described above, apply the patch

This might work for most of the cases, but not all. E.g. the meanwhile json in this PR has two keys (profileID and meanwhileID).

ShishkinP:
This is it. Quests are now referred to by name instead of index. Also, I had to do some rewriting due to pointer double deletion error and modifiable data that must not be modifiable.

The failed checks complain about content manager's unit test failure to write to or read a temp file. No sure if it's relevant to these last commits.

> One thing I thought of in the past would be to also allow this patch-like structure

Do you think an actual [RFC6902](https://datatracker.ietf.org/doc/html/rfc6902) patching system would be an overkill? I found a couple of games that use it. There are supporting libraries. Is it worth looking into?

lynxlynxlynx:
Nothing new to add after checking the new code, but I will echo momoko-h's preference for placing the sources in externalized/content instead.

momoko-h:
This LGTM. This code would really benefit from a type definition for `std::array<NPCQuoteInfo, NUM_NPC_QUOTE_RECORDS>` but that was already the case before this PR and is material for a potential cleanup PR in the future. For now I think the best plan is to merge the PR and see how this code performs in action. 

Very nice work!

PR #2030, opened 2024-07-11
Fix moddability of Walter's bribe amount

momoko-h:
The amount specified in tactical-npc-action-params.json had no effect because the code used the wrong NpcActionParamsModel method (getGridNo instead of getAmount).

I moved the code to handle these two cases into a new function because HandleNPCDoAction() is the worst function in the sense of code complexity measurements.


lynxlynxlynx:
Nice catch!

PR #2031, opened 2024-07-12
Small refactoring around ArmsDealerID

selaux:
This is a small refactoring around arms dealer ids. My hope is that something like https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1899 can be achieved with way smaller, more easily mergeable PRs. Basically taking baby steps in the right direction.


lynxlynxlynx:
LGTM, thanks!

src/externalized/DefaultContentManager.cc
```diff
@@ -181,8 +181,13 @@ const DealerInventory* DefaultContentManager::getBobbyRayUsedInventory() const
...
 
-const DealerModel* DefaultContentManager::getDealer(uint8_t dealerID) const
+const DealerModel* DefaultContentManager::getDealer(ArmsDealerID dealerID) const
 {
+	auto dealerIndex = static_cast<std::size_t>(dealerID);
+	if (dealerIndex >= m_dealers.size()) {
+		ST::string err = ST::format("failed to get dealer: invalid dealer ID: {}", dealerID);
+		throw std::runtime_error(err.to_std_string());
+	}
 	return m_dealers[dealerID];
```
momoko-h, line 191: I wouldn't expect to see this exception too often (hopefully never) so the additional error info shouldn't be needed and the range checking could be handled by vector::at.

src/externalized/DefaultContentManager.cc
```diff
@@ -854,7 +859,7 @@ const std::map<uint16_t, uint16_t>& DefaultContentManager::getMapItemReplacement
 	return m_mapItemReplacements;
 }
 
-const DealerInventory* DefaultContentManager::getDealerInventory(int dealerId) const
+const DealerInventory* DefaultContentManager::getDealerInventory(ArmsDealerID dealerId) const
```
momoko-h, line 862: If getDealer range checks its argument I think this method should also do.

selaux: Good point. I will include this in a follow up PR.

PR #2032, opened 2024-07-14
Refactor iteration over items

selaux:
This changes the way items are iterated over in most of the places. `GCM` has a new method, which returns a new class called `ItemRange`, which is a range over the items vector.

It can be used where we need iteration over items instead of the loop over the item ids.

I've used this in most places where we do iterations over item ids, except for two cases:

- When the thing we are iterating over is not really the items vector but something different (e.g. the arms dealer inventory)
- When the iteration logic is not a straight forward iteration over all items.

NB: This would be a place where we actually would benefit from the C++20, as it supports ranges and views over the items vector.

### DIFF
--- a/src/externalized/ItemRange.h
+++ b/src/externalized/ItemRange.h
@@ -1,0 +1,21 @@
+#pragma once
+
+#include <vector>
+#include "ItemModel.h"
+
+class ItemRange {
+	using Iterator = std::vector<ItemModel const *>::const_iterator;
+
+	public:
+		ItemRange(Iterator begin, Iterator end) noexcept : m_begin(begin), m_end(end) {};
+		Iterator begin() const noexcept {
+			return m_begin;
+		};
+		Iterator end() const noexcept {
+			return m_end;
+		};
+
+	private:
+		Iterator m_begin;
+		Iterator m_end;
+};
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -804,6 +804,11 @@ bool DefaultContentManager::loadAllDealersAndInventory()
 	return true;
 }
 
+ItemRange DefaultContentManager::getItems() const
+{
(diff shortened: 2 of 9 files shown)


momoko-h:
I like the basic idea very much but the implementation could be more user friendly. Adding a simple helper type (let's call it UsableItems) which only has two methods `begin()` and `end()` would enable the use of ranged for loops.

Rough sketch:
```
struct UsableItems
{
	std::vector<ItemModel const *>::const_iterator begin();
	std::vector<ItemModel const *>::const_iterator end();
};
```
Add a new ContentManager method
```
UsableItems ContentManager::getUsable();
```
Then use it like this:
```
for (auto * item : GCM->getUsable()) {}
```

selaux:
@momoko-h True, good idea. I've adapted the PR accordingly. I've still called the method `getItems` and the class `ItemRange`, as it returns all items (index 0 is not really an item, just a hint that there is none).

lynxlynxlynx:
Much better, yeah. (const) references or pointers in the ranged loops would be negligibly faster, but whatever, it's not a hot path.

selaux:
In my understanding the result of the `auto` is a `const ItemModel*` already as its a `const_iterator`. At least that is what my IDE is showing me:

image

lynxlynxlynx:
Oh, good, no copy then.

src/externalized/ItemRange.h
```diff
@@ -0,0 +1,21 @@
...
+			return m_begin;
+		};
+		Iterator end() {
+			return m_end;
+		};
+
+	private:
+		Iterator m_begin;
+		Iterator m_end;
+};
```
momoko-h, line 21: The methods of this class could all be marked `const noexcept`  which might help the compiler optimize them.

momoko-h:
Other than that it all LGTM!

src/externalized/ItemRange.h
selaux: Fixed.

PR #2033, opened 2024-07-16
Fix warnings when running the unit tests

momoko-h:
During unit testing the file itemdesc.edt from the original game resources is not available which caused several irritiating warning messages in DefaultContentManager:: loadGameData(). This method is now virtual and overridden by the unit test content manager. The new loadGameData method simply uses an empty VanillaItemStrings instance because the unit tests do not need the actual item descriptions.


selaux:
Thanks.

PR #2035, opened 2024-07-18
Remove some more references to MAXITEMS

selaux:
Remove some more references to `MAXITEMS`.

### DIFF
--- a/src/game/Editor/EditorItems.cc
+++ b/src/game/Editor/EditorItems.cc
@@ -184,8 +184,7 @@ static void DrawItemCentered(const ItemModel * item, SGPVSurface* const vs, INT3
 
 void InitEditorItemsInfo(ToolbarMode const uiItemType)
 {
-	INT16 i, x, y;
-	UINT16 usCounter;
+	INT16 x, y;
 	ST::string pStr;
 	BOOLEAN fTypeMatch;
 	INT32 iEquipCount = 0;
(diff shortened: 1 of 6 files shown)


src/game/Editor/EditorItems.cc
```diff
@@ -324,26 +325,30 @@ void InitEditorItemsInfo(ToolbarMode const uiItemType)
...
-				else
-					usCounter = ACTION_ITEM;
+				itemIt = std::find_if(items.begin(), items.end(), [](const ItemModel* item) -> bool {
+					return item->getItemIndex() == ACTION_ITEM;
+				});
+				if( i < PRESSURE_ACTION_ID && !(i % 2) ) {
+					itemIt = std::find_if(items.begin(), items.end(), [](const ItemModel* item) -> bool {
+						return item->getItemIndex() == SWITCH;
+					});
+				}
```
lynxlynxlynx, line 349: Shouldn't these searches start at itemIt? Consider having more than one matching item in the list, wouldn't always the first one be returned?

momoko-h: Do you have plans to no longer store the ItemModel pointers in a vector ordered by the item index in the future? If yes, then we'll probably need a generalized find function anyway. If not, then you can simply set the iterator to a specific value with std::next, for example  `itemIt = std::next(GCM->getItems().begin(), SWITCH - 1);` (-1 uses the documented fact that begin() starts after the NOTHING item).

selaux: @lynxlynxlynx This part of the code is quite convoluted. As far as I understood the original code, when in `TBAR_MODE_ITEM_TRIGGERS`, it will alternate between the `ACTION_ITEM` and `SWITCH` depending on `i % 2`. So the searches should always return either the iterator pointing to `ACTION_ITEM` or to `SWITCH`. Then at the bottom of the function it will render different labels for the two items, depending on `i`. I've extracted the find operations to the beginning of the function to make this a bit clearer.

@momoko-h I'd still expect them to be a vector ordered by `itemIndex`, but I would not expect all `itemIndexes` to exist though. E.g. when a mod decides to remove a certain item. At least I don't want the code to depend on the fact.

lynxlynxlynx: Yeah, this way also avoids repeated lookups.

lynxlynxlynx:
Thanks!

PR #2036, opened 2024-07-19
Remove the limitation that all weapons must have an index <= 70

momoko-h:
The only code reason for this limitation was the size of the STRATEGIC_STATUS::fAlreadyDropped array which used one BOOLEAN per weapon.

This commit utilizes the formerly unused passing bytes at the end of STRATEGIC_STATUS and uses one bit per weapon for a new theoretical limit for the weapon's index of 629. The total number of items is still limited to MAXITEMS (351), however.

Considering that weapons are by far the most popular target for modders this PR should at least offer them a bit more leeway, especially by making the range 328-350 available.


src/game/Strategic/Strategic_Status.h
```diff
@@ -99,7 +99,10 @@ struct STRATEGIC_STATUS
 	UINT8			ubNumNewSectorsVisitedToday;
 	UINT8			ubNumberOfDaysOfInactivity;
 
-	INT8 bPadding[70]; // XXX HACK000B
+	// For all weapons with an item index >= MAX_WEAPONS this array stores
+	// their "already dropped" status as a single bit. The size of this array
+	// must be exactly 70 because the struct size must be 192 bytes.
+	std::array<uint8_t, 70> additionalWeaponsAlreadyDropped;
```
lynxlynxlynx, line 106: How come you didn't choose std::bitset? It could simplify SetWeaponAlreadyDropped and WasWeaponAlreadyDropped.

selaux:
Could we do this using the `SavedGameStates` instead? It's basically a JSON appended to the end of the save. This would have several advantages:

- It would also work for other cases, e.g. dealer inventories, where no unused padding is available
- It would also avoid the new limit, basically making the limit the max of `uint16_t`
- We wouldn't need a new save game version, we could just check if the key exists in `SavedGameStates`

Disadvantage is that we probably need to serialize the `fAlreadyDropped` to JSON somehow, which shouldn't be that hard. We could simply do an array of already encountered weapons, ideally using the `internalName`. At least that is how I planned to do it for the dealer inventories.

src/game/Strategic/Strategic_Status.h
momoko-h: Because the implementation of bitset is up to the compiler authors. I cannot guarantee that a bitset of 560 bits has a size of exactly 70 bytes. Usually the size gets aligned to the CPU's natural word size; GCC on x64 for example has sizeof(std::bitset<560>) == 72.

Also, a save produced by an executable build with MSVC might not be compatible with a GCC or a Clang executable.

momoko-h:
> Could we do this using the `SavedGameStates` instead? It's basically a JSON appended to the end of the save.

Sure, let's do that.

src/game/Strategic/Strategic_Status.h
lynxlynxlynx: Ouch, that's quite bad for a standard. One could cmake around it, but bleargh and why bother.

lynxlynxlynx:
Looks good to me. 👍🏾  @selaux did you want to take another look?

assets/externalized/weapons.json
```diff
@@ -3016,5 +3016,58 @@
             "type": "GUNS",
             "subIndex": 1
         }
+    },
+    {
```
selaux, line 3020: Could we set `bNotBuyable` and `bDefaultUndroppable` for this item?

selaux:
Back from vacation. LGTM. One minor comment regarding the newly added item.

momoko-h:
So typing git rebase master before git push wasn't the best idea. I'm not sure trying to fix the history here would look correct, so maybe it would be better to close this PR and open a new one?

lynxlynxlynx:
As you wish. 
I think rerebasing on an old HEAD of master would work and maybe the previous one is visible in the reflog. But the simplest should be `git rebase -i HEAD~35` to just reorder your commits to the end, as if they're all new.

PR #2041, opened 2024-07-23
Reduce the number of unnecessary vector copies

momoko-h:
Several ContentManager methods return a reference to a vector. Simply calling this methods with` auto v = GCM-> methodName() `creates a new copy of the vector in v but in most of this cases we simply want to read from the vector.


lynxlynxlynx:
Thanks!

PR #2042, opened 2024-07-28
Fix NPC removal failure due to untimely sector unload

### LINKED ISSUE #750: Doreen Harrows still in Drassen C13 after quest completion
### Doreen Harrows is still in sector C13 even though I already convinced her to leave

On day 1, I peacefully resolved the quest number 20 "Doreen is cruel to children" by convincing her to leave.
On day 4, I found that she is still in sector C13 Drassen.
Both, the history log and the Quest Debug Screen show the quest as done.
[SaveGame11.sav.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/2705315/SaveGame11.sav.zip)

|  ja2-stracciatella_0.16.1_amd64.deb | JA2 Gold English (GOG)  |    ~no~ yes    |    yes    |

- savegame attached
- operating system is an up to date Debian 9.6 (Stretch)

ShishkinP:
Closes #750.

NPCs that are removed from the game by first walking to the edge of the map and then triggering NPC_ACTION_REMOVE_NPC 
can fail to do so if the map is unloaded during the walk.
Affected NPCs are those with script records like this:
```javascript
{
   //...
   "goToGridno": 8449,
   "actionData": {
      "code": 264 // NPC_ACTION_REMOVE_NPC
   }
}
``` 
Namely: Doreen, Maria and Angel, Sergeant, Kyle, Carmen


src/game/TacticalAI/NPC.cc
```diff
@@ -1876,6 +1876,12 @@ void ConverseFull(UINT8 const ubNPC, UINT8 const ubMerc, Approach bApproach, UIN
 						}
 						else
 						{
+							// This check prevents NPCs from forever lingering if the player unloads the sector
+							// before an NPC chosen for removal reaches their destination (issue #750).
+							if (pNPC && pQuotePtr->sActionData == NPC_ACTION_REMOVE_NPC)
+							{
+								gMercProfiles[pNPC->ubProfile].sSector = SGPSector();
```
momoko-h, line 1883: Nice, I like bug fixes that are a net negative in code lines. I'm wondering why you need to check if pNPC is not null, though. ubNPC should always be the same as pNPC->ubProfile and gMercProfiles[ubNPC] is available under the variable name `p`.

Or as an alternative, wouldn't moving the line 					`gMercProfiles[ ubTargetNPC ].sSector = SGPSector();` in the NPC_ACTION_REMOVE_NPC case in `HandleNPCDoAction()` out of the `if ( pSoldier )` block also work?

ShishkinP: > wondering why you need to check if pNPC is not null

Just to be on the safe side. With `p` the check is not needed of course.

> gMercProfiles[ubNPC] is available under the variable name `p`

All the examples of using `SGPSector()` in the codebase are with `gMercProfiles[id]` so I didn't put much thought into it.

> wouldn't moving the line `gMercProfiles[ ubTargetNPC ].sSector = SGPSector();` in the NPC_ACTION_REMOVE_NPC case in `HandleNPCDoAction()` out of the `if ( pSoldier )` block also work?

Whatever happens in the NPC_ACTION_REMOVE_NPC case is irrelevant to this issue because it happens only after the destination is reached. And it's never reached if the map is unloaded during the travel. Spawning on the destination tile on sector load is not considered reaching.
And we can't do this action beforehand because it'll make the NPC disappear right in front the player's eyes because of `TacticalRemoveSoldier(*pSoldier)`.

momoko-h: Ok then, thanks for the explanation.

PR #2044, opened 2024-07-29
Auto resolve cleanup

momoko-h:
- Remove three unused functions in Auto_Resolve

    
    Previously auto resolve had several cheat keys but these
    all got removed as part of the JA2BETAVERSION removal.
    Since then these function could only be reached through
    a debugger (I didn't test whether they'd still work).

-   Clean up RemoveAutoResolveInterface()
    
    The removed debug code was the only place that called
    this function with a false argument for delete_for_good.

-   Allocate SOLDIERCELL arrays together with AUTORESOLVE_STRUCT
    
    These arrays and the AR struct now have the same lifetime so
    it makes no sense to allocate them seperately.

-   Remove INVULNERABILITY

-   Remove CalculateSoldierCells's parameter
    
    Was now always false; this also made RefreshMerc obsolete.

-   Remove iActualMercFaces
    
    This variable was also only used when the AR screen still
    had cheat keys and it was possible to add more mercs to
    the team.

### DIFF
--- a/src/game/Strategic/Auto_Resolve.cc
+++ b/src/game/Strategic/Auto_Resolve.cc
@@ -2167,75 +2167,6 @@ static void CalculateAutoResolveInfo(void)
 }
 
 
-static void CreateTempPlayerMerc(void);
-
-
-// Debug utilities
-static void ResetAutoResolveInterface(void)
-{
-	guiPreRandomIndex = gpAR->uiPreRandomIndex;
-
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
LGTM! `fUnlimitedAmmo` can get the same treatment, axe.

lynxlynxlynx:
Thanks!

PR #2046, opened 2024-07-31
Fix various script records and deduplicate EPC fact checking code

### LINKED ISSUE #1370: Joey can be escorted again after delivering him to Martha
### What is wrong
Joey was in abandoned mines, threatened him to follow me, got back to Cambria. The quest ended properly. The problem is, I could still "recruit" him afterwards and get him to follow me (from his room), which probably shouldn't be happening. When recruited like this (for a second time) and coming close to Martha, the following error is logged:
```2021-05-13T08:52:49 [ WARN] game\TacticalAI\NPC.cc: trigger of 90, record 9 cannot proceed, possible error```
Attaching a qucksave, just talking to him with DSeyka and choosing either "Threaten" or "Recruit" should trigger this.
[QuickSave.zip](https://github.com/ja2-stracciatella/ja2-stracciatella/files/6471377/QuickSave.zip)
Oh, and I'm using latest nightly (Stracciatella v0.19.0-20210507+2ba89cb) w/ English GOG version of JA2.

ShishkinP:
Closes #1370

Fixed issues:

- Typo in Mike's item-giving quote: EXTENDEDEAR (210) instead of intended MONEY (219)
- After the quest completion Joey's quotes assume he's still in San Mona and he can be re-recruited again
- Joey and Skyrider: if unescorted before their quest completion they are stuck on the same quote complaining about being abandoned even after their quests are done
- Needlessly hardcoded and redundant code for checking if an NPC is being escorted
- `NumMercsNear()` works in the way opposite of its intended use. Went unnoticed because it's used only in Martha's records and doesn't seriously break anything.

Added some comments to script-records-recruited.json to make its purpose less confusing.


lynxlynxlynx:
Good stuff, thanks!

PR #2048, opened 2024-07-31
Fix NPCs getting stuck on diagonal travel through closed sliding door

ShishkinP:
Introduced in #1995

Reproducible by allowing Seargeant to leave. The last door is the sliding door on which he freezes if it's closed.


lynxlynxlynx:
Thanks!

PR #2051, opened 2024-08-07
Lift the MAX_WEAPONS restriction; weapons no longer must have an index < 70

momoko-h:
Well, that was slightly embarrassing. See #2036.


lynxlynxlynx:
Thanks!

PR #2052, opened 2024-08-08
Add JSON Patch functionality for modding

ShishkinP:
Big JSON files make it hard to filter out and debug changes introduced by a mod. The JSON Patch format solves this problem by allowing partial modifications with these operations: "add", "remove", "replace", "move", "copy", "test" (described in [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902/#section-4)).

The patch can be easily created by using [online tools](https://chbrown.github.io/rfc6902/) or JSON editor's extensions that generate patches based on a diff between the vanilla and the modded JSON docs.

The way I see it in the future:
- The currently used JSON files in mods will be reserved for "total conversions". If multiple mods have `music.json` for example only one of them substitutes the vanilla one.
- On the other hand, multiple `music.patch.json`'s will be compounding (this PR doesn't make it possible - only one `music.patch.json` is applied. If I'm not mistaken the first mod in `"mods": ["mod1", "mod2", ... ]` that has this file always wins)


lynxlynxlynx:
Fancy, I like it. It's the first time we need all resource matches, not just the first one found, so it will really need some more VFS work to make patch stacking possible.
@selaux any thoughts?

selaux:
Yep. I like it as well, finally some composability for mods :). As @lynxlynxlynx said it might make sense to create functionality that allows to read all patch files in all VFS layers. But that can also come in the future.

One thing I am not 100% sure about is that JSON patch is the exact format we want, as it makes it hard to work with arrays nicely as it is using JSON pointer as the path.  With all of the array based data we have it might be hard to do some things, especially if entities are removed and multiple mods are in play.

Example: We have two mods: `mod1` and `mod2`, which are developed independent of each other

- `mod1` removes an item at item index `123`
- `mod2` edits the item at item index `124` -> This will lead to issues as all of the indexes have been shifted by the removal and a different item will be edited

What would be an option here is to use the JSON patch standard with a subtle difference: Use JSON path instead of JSON pointer as the path, which allows for selectors in arrays. E.g. for `weapons.json`:

```json
[
  {
    "op": "remove",
    "path": "$[?(@.internalName == 'CROWBAR')]"
  }
]
```

As with this PR we can only read a single patch file anyways we can still merge and think about how to handle this in the future.

lynxlynxlynx:
I just realized this will also make user configs more resilient. Instead of adding a whole game.json, one will be able to modify just what they need, avoiding problems with missing new keys and explicitly failing with removed ones.

ShishkinP:
> hard to work with arrays nicely as it is using JSON pointer as the path

JSON Path of course would be preferable from this POV. The partial workaround is to enforce the "test" operation before modifying arrays. This way `mod2` will abort its patch if the test for the value at a certain index fails.

Another idea: recompile all patches into one - BUT! - now with index-shifting operations put at the end of the patch.

The reason I don't want to abandon JSON Patch is I just did a bit of googling and couldn't find anything as easy as JSON Patch tools. Or do we assume modders must have a certain skill level and put more effort (if they want their mod to be compatible with other mods)?

selaux:
That is a good question how much technical knowledge we can expect from modders. Do you think this could also be solved with more documentation on the website? Or we push users to use the [stracciatella-toolset](https://github.com/ja2-stracciatella/stracciatella-toolset). For that it would need to be more usable though.

ShishkinP:
I completely forgot about the toolset. That and documentation can indeed make it a non-problem. I'll have to think this through.

selaux:
I'll merge now anyways, we still can adapt it in the future, as long as we haven't released a new version.

lynxlynxlynx:
@ShishkinP would you like commit access? We all work through PRs, so it's not much of a change, but it would mean you could merge other peoples work.

ShishkinP:
@lynxlynxlynx Yes, it wouldn't hurt to have it, I guess

lynxlynxlynx:
Done, welcome to the team. :)

PR #2053, opened 2024-08-10
Externalize explosives

selaux:
Just a foreword: This still needs some testing, but I think its ready for preliminary review, as it's quite big again unfortunately.

This externalizes the the following:

- `EXPLOSIVETYPE const Explosive[]` -> `explosives.json`: Explosive items
- `ExplosionInfo explosion_info[]` -> `explosion-animations.json`: Animation data for explosives

Similar to `weapons.json`, explosives will not part of `items.json` anymore, as they need some extra data which was previously defined in `Explosive[]`. If a current mod has defined explosives in `items.json` they now will be ignored with a warning.

Next up would be Armours as they have a similar logic to explosions.


rust/stracciatella/src/schemas/yaml/items.schema.yaml
```diff
@@ -1,142 +1,29 @@
...
-          title: Small
-          description: Small graphics. Used e.g. in strategic & tactical inventory
-          $ref: types/graphic.schema.yaml
-        big:
-          title: Big
-          description: Big graphics. Used e.g. in item detail view
-          $ref: types/graphic.schema.yaml
+        usItemClass:
+          title: Item class
+          $ref: types/uint32.schema.yaml
```
lynxlynxlynx, line 19: Doesn't need to happen now, but this could also be changed to a fixed list to be easier to work with.

rust/stracciatella/src/schemas/yaml/explosives.schema.yaml
```diff
@@ -0,0 +1,71 @@
+$schema: http://json-schema.org/schema#
+type: array
+title: Explosives
+description: |
+  **Limitations**: Some explosives are still referenced in source code, so they should not be changed.
```
lynxlynxlynx, line 5: The itemIndex or something else?

rust/stracciatella/src/schemas/yaml/types/item-cursor.schema.yaml
```diff
@@ -0,0 +1,28 @@
...
+  - KEY
+  - SAW
+  - WIRECUT
+  - REMOTE
+  - BOMB
+  - REPAIR
+  - TRAJECTORY
+  - JAR
+  - TINCAN
+  - REFUEL
```
lynxlynxlynx, line 28: This can now be used for `ubCursor` in `items.json` as well, right?

lynxlynxlynx:
LGTM, nothing jumped out at me in the code itself.

rust/stracciatella/src/schemas/yaml/types/item-cursor.schema.yaml
selaux: Yes. I just didn't want to break existing mods that modify the `items.json`.

rust/stracciatella/src/schemas/yaml/items.schema.yaml
selaux: Yes. I just didn't want to break existing mods that modify the `items.json`.

rust/stracciatella/src/schemas/yaml/explosives.schema.yaml
selaux: Yeah, the item indexes are used for some things, e.g. explosive weapons.

lynxlynxlynx: It would be good to explain this a bit more. As a modder I wouldn't be sure what I can and cannot modify. Are some of the explosives off-limits, are just some fields problematic, that sort of thing.

rust/stracciatella/src/schemas/yaml/items.schema.yaml
lynxlynxlynx: It will always be a problem with already released schemas. I guess we could solve that with anyOf and then eventual deprecation.

src/externalized/ItemModel.h
```diff
@@ -118,5 +120,5 @@ struct ItemModel
 	uint16_t   fFlags;
 
 	void serializeFlags(JsonObject &obj) const;
-	uint32_t deserializeFlags(JsonObject &obj) const;
+	static uint32_t deserializeFlags(JsonObject &obj);
```
momoko-h, line 123: While you're fixing this oversight, please also change the return type to `uint16_t` (same as `ItemModels::fFlags`) and make obj a const reference.

selaux:
So while looking into how the explosives are referenced in source code, I've noticed that there is a lot of other things that could be externalized. It feels like every explosive is referenced somewhere. Ideally I don't want to do this in this PR, as it's already quite big. Would it be fine if I fix the comments in this PR and then go on in a separate PR to make the changes more clear. Things I've found that could be externalized (not exhaustive):

- Which explosives cause Blast / Stun / Smoke / Light effects
- Which explosives can be loaded into e.g. the grenade launchers
- Which explosives can be equipped with a 
- Action Items (I still need to figure out how they work)

lynxlynxlynx:
Yeah, sure.

src/externalized/ItemModel.h
selaux: Fixed, thanks.

rust/stracciatella/src/schemas/yaml/explosives.schema.yaml
selaux: Right now I just stated that this should not be adapted. I will update this in the future.

selaux:
Updated the PR to current master and adressed the comments. Good to go from my side.

momoko-h:
Looks good. As you said, anything else can be done in new PRs.

PR #2054, opened 2024-08-10
Fix running out of disk space for Android github runner

selaux:
Tries to fix the disk space issues for the Android build on GitHub actions by clearing some space at the beginning of the pipeline.


momoko-h:
Thank you for taking care of that annoyance!

PR #2056, opened 2024-08-15
Fix sign-compare warning in NPCQuoteInfo

momoko-h:
Also, it is not necessary to pass a MercSystem and a ItemSystem pointer to deserialize() because
ContentManager inherits both interfaces.

The code is working as is but this is currently the only warning on the Linux runner; it's nice to keep that one free of warning because it makes it easy to notice any new ones.


lynxlynxlynx:
LGTM, thanks!

PR #2058, opened 2024-08-17
Clean up CalibreModel

momoko-h:
This struct neither inherits from anything nor is it inherited anywhere so its functions don't need to be virtual.

Let the constructor do all the work of creating a new instance from JSON instead of moving all data from deserialize().


lynxlynxlynx:
Yep, thanks!

PR #2059, opened 2024-08-17
Fix mercs refusing to join/renew contract

ShishkinP:
`GetFirstBuddyOnTeam()` returns merc ID instead of expected index for the array of friends. As a result, mercs who are joining or renewing the contract ignore friends on the team if there are also enemies. The bug is not present in vanilla.

Steps to reproduce (using Grizzly who is friends with Bull and enemies with Dr. Q): 
1. Hire Dr. Q and make sure he's in Omerta
2. Attempt to hire Grizzly. He will decline and mention Dr. Q as the reason.
3. Hire Bull and make sure he's in Omerta
4. Attempt to hire Grizzly. He will decline and mention Dr. Q as the reason. The correct behavior: he must agree and mention Bull as the reason

Additionally: code that's unreachable because `hated` by design cannot be more then 2. Doesn't actually affect anything except 
code readability.

### DIFF
--- a/src/game/Laptop/AIMMembers.cc
+++ b/src/game/Laptop/AIMMembers.cc
@@ -90,7 +90,6 @@
 #define SIZE_MERC_ADDITIONAL_INFO		160
 
 #define MERC_ANNOYED_WONT_CONTACT_TIME_MINUTES	6 * 60
-#define NUMBER_HATED_MERCS_ONTEAM		3
 
 
 #define STATS_X					IMAGE_OFFSET_X + 121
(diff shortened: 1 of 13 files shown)


momoko-h:
So this was broken since Tron added this function almost 16 years ago in commit 6e6bed89d2b32fdd4cc73a5a6c11079d79ba0f55. Of course it didn't help that the function has no comment to describe what it does, its name is misleading and its return type INT8 doesn't give any hint either. I would appreciate it if you could add a comment to the function declaration and change its return type. We don't have anything like gsl::index, so std::ptrdiff_t is probably the best match.

Another nice find!

ShishkinP:
Might as well go all the way through and add enums because these -1's, 0's, 1's, 2's are all over the codebase.

lynxlynxlynx:
LGTM, thanks!

PR #2060, opened 2024-08-18
Externalize explosive calibres

selaux:
Similar to calibres, this extracts explosive calibres (meaning the association between launchers/cannons and their ammo). This was added to `explosives.json` and `weapons.json`. For missing values it will print a warning and fall back to vanilla values.

Tested it by adding a new launcher weapon with a different calibre:
[Short video of test](https://github.com/user-attachments/assets/5f77f65b-1d5d-4764-b0d9-e798cad803e6)


rust/stracciatella/src/schemas/yaml/explosives.schema.yaml
```diff
@@ -16,6 +16,10 @@ items:
           title: Animation ID
           description: ID of the explosion animation used for this explosive
           $ref: types/id.schema.yaml
+        calibre:
+          title: Calibre
+          description: The calibre of the launcher weapon that must be used to launch this explosive.
+          $ref: types/id.schema.yaml
```
lynxlynxlynx, line 22: Shouldn't this use the new schema?

selaux: As it's just a reference to the new schema, this is just a string. The calibres are their own file.

selaux:
Had to fix some forward declarations (mixed up class and struct) for MSVC to compile.

lynxlynxlynx:
Nice 👍🏾

selaux:
NB: I am aware that I introduced some new coverity issues with this and the previous PR (copy instead of move). I will fix them once I have finished everything around explosives.

PR #2061, opened 2024-08-23
Address HACK000E comments in DisplayTeamStats

momoko-h:
No more uninitialized variables in this function.


lynxlynxlynx:
Thanks!

PR #2062, opened 2024-08-25
Fix crash in CalcBestThrow

momoko-h:
Crash happened when the AI considered throwing a grenade. Grenades do not count as weapons, so the weapon pointer was NULL.

This commit also tones down the gigantic block of variable declarations at the top of the function a bit by moving down several variables to the location of their initialization.


lynxlynxlynx:
Thanks!

PR #2063, opened 2024-08-26
Extract smoke effects

selaux:
Still needs some testing, but should be about done. This externalizes `SmokeEffectModel` from code, which represents the effects of smoke (e.g. Smoke, Teargas).

### DIFF
--- a/src/game/Tactical/Overhead.cc
+++ b/src/game/Tactical/Overhead.cc
@@ -90,6 +90,7 @@
 #include "End_Game.h"
 #include "Strategic_Status.h"
 #include "PreBattle_Interface.h"
+#include "SmokeEffects.h"
 
 #include "ContentManager.h"
 #include "GameInstance.h"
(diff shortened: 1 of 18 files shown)


src/game/Tactical/Overhead.cc
```diff
@@ -1523,37 +1524,26 @@ BOOLEAN HandleGotoNewGridNo(SOLDIERTYPE* pSoldier, BOOLEAN* pfKeepMoving, BOOLEA
 					bPosOfMask = NO_SLOT;
 				}
 
-				const ExplosiveModel* pExplosive = nullptr;
-				if (!AM_A_ROBOT(pSoldier))
+				const SmokeEffectModel* smokeEffect = nullptr;
+				auto smokeEffectID = GetSmokeEffectOnTile(pSoldier->sGridNo, pSoldier->bLevel);
```
lynxlynxlynx, line 1528: Is there any risk of applying consequences several times in a turn this way? Checking the tile for presence vs `fHitByGasFlags`, which seems to be a marker specifically for such cases.

src/game/TileEngine/SmokeEffects.cc
```diff
@@ -196,7 +196,7 @@ void NewSmokeEffect(const INT16 sGridNo, const UINT16 usItem, const INT8 bLevel,
 	pSmoke->ubRadius    = ubStartRadius;
 	pSmoke->bAge				= 0;
 	pSmoke->fAllocated  = TRUE;
-	pSmoke->bType				= bSmokeEffectType;
+	pSmoke->bType				= static_cast<int8_t>(smokeEffectID);
```
momoko-h, line 199: Looks like SMOKEEFFET::bType is not actually used for anything so you could just remove it.

selaux: True. But it should be there for savegame compatibility. I think with the next PR i might be able to read it from the explosive item though and can remove it from the struct.

src/game/Tactical/Overhead.cc
selaux: True I've added `IsSoldierAlreadyAffectedBySmokeEffect` to reflect the correct behavior.

selaux:
Did some testing and as far as I can see everything looks good.

src/game/TileEngine/SmokeEffects.h
```diff
@@ -34,15 +25,22 @@ struct SMOKEEFFECT
...
 
 
-// Returns NO_SMOKE_EFFECT if none there...
-SmokeEffectKind GetSmokeEffectOnTile(INT16 sGridNo, INT8 bLevel);
+// Returns the smoke effect id for a specific tile on the current map or SmokeEffectID::NOTHING if no smoke effect is present.
+SmokeEffectID GetSmokeEffectOnTile(INT16 sGridNo, INT8 bLevel);
+// Returns the smoke effect id for the world flags of a tile or SmokeEffectID::NOTHING if no smoke effect is present.
+SmokeEffectID FromWorldFlagsToSmokeType(UINT8 ubWorldFlags);
+
+struct SOLDIERTYPE;
```
momoko-h, line 33: No need for this line because JA2Types.h gets included. SOLDIERTYPE* is already used above in struct SMOKEEFFECT.

selaux: Fixed.

momoko-h:
Very nice, thanks!

PR #2065, opened 2024-08-26
Address TODO0009: possible endless loop

momoko-h:
-   It's no longer possible to get stuck in an endless loop in the very unlikely case where we find a LEVELNODE with an usIndex oF NO_TILE.
    
-   Refactored TypeRangeExists... functions.  Changed three "find roof or floor" functions to use   TypeRangeExists...

-   Replace eight levelnode remove functions with a function template

-    Change the return type of TypeRangeExistsInLandLayer so it matches the other two TypeRangeExists... functions.


selaux:
Thanks!.

PR #2066, opened 2024-08-28
Merge three arrays that control Render()

momoko-h:
Merging the three arrays RenderFX, RenderFXStartIndex and g_render_fx_layer_flags has two advantages:

1. It indicates the close relationship of these arrays and how they control the program flow in Render().
2. This makes it more obvious that copying values from RenderFX and RenderFXStartIndex to temporary arrays at the start of Render() was not necessary. We can just look up these values in the constant table when needed.

Seeing that this data is constant also made it easier to see that we do not need to initialize various variables for every levelnode because they will stay constant while Render() processes an entire render layer.

### DIFF
--- a/src/game/TileEngine/RenderWorld.cc
+++ b/src/game/TileEngine/RenderWorld.cc
@@ -4,6 +4,7 @@
 #include "English.h"
 #include "Font.h"
 #include "Font_Control.h"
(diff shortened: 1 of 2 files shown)


selaux:
LGTM. Thanks!

PR #2067, opened 2024-09-01
Add basic support for CMake unity builds

momoko-h:
-  Address small problems that disturb a CMake unity build
    
    This commit contains only some hopefully not controversial
    changes that are needed for an unity build. On its own it
    is not sufficient, either more substantial code changes or
    the definition of unity groups are needed.

- Specify unity groups for SGP and editor.

To try a unity build you'll have to set the CMake environment variable CMAKE_UNITY_BUILD:
 `cmake -DCMAKE_UNITY_BUILD:bool=true` 

Unity builds are quite useful when you have to recompile large parts of the project, for example if you touch `Types.h`. Not recommended when you only change one or two source files.

[Doumentation for UNITY_BUILD](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html)


momoko-h:
Some numbers to give you an idea how much build time this saves:

|    | Normal       | Unity           |
| ---- | ---------------- | ---------------- |
|Usr | 22.87 mins | 19.89 mins |
|Sys | 2.37 mins   | 2.02 mins   |

(Clean build, GCC 13.3.1, Debug configuration). Not too bad considering that several subdirectories aren't covered at all yet.

selaux:
Never heard of that before, but as it is opt-in it does not hurt. There is a merge conflict though.

momoko-h:
It was only added in CMake 3.16, so it's relatively new; with all the features that get constantly added to CMake it's easy to miss. 

The basic idea behind it is really old, however. C and C++ compilers prefer one huge source file over many smaller files that all include the same headers, so creating an intermediate source file that just includes several source files is quite effective.

PR #2070, opened 2024-09-04
Externalize profiles for named characters (MERCPROFILESTRUCTs)

ShishkinP:
Added:
1. mercs-profiles.json - stats, inventory and other stuff
2. mercs-relations.json - opinions and friend/enemy relations


assets/externalized/mercs-profiles.json
```diff
@@ -0,0 +1,12326 @@
+[
+    {
+        "profile": "BARRY",        // Required. Must correspond to the internalName field in mercs-profile-info.json
```
ja2-derek, line 3: Should we merge `mercs-profile-info.json` into `mercs-profiles.json`? 

There is no point to have two separate JSONs and hence the need to be consistent. Both map to same data classes, and the `mercs-profile-info` is just a single data field.

ShishkinP: I'm not sure about the mod compatibility. Are there even enough of them to care? Or are they all primarily weapons-focused?

ShishkinP: I guess if we almost didn't have any char externalization why would anyone bother modding it? I'll merge it then.

src/externalized/DefaultContentManager.cc
```diff
@@ -1163,6 +1164,25 @@ bool DefaultContentManager::loadMercsData()
 	}
 	MERCListingModel::validateData(m_MERCListings);
 
+	std::vector<MERCPROFILESTRUCT*> temp_mercStructs;
+	temp_mercStructs.resize(NUM_PROFILES);
```
momoko-h, line 1168: You can combine these two lines: `std::vector<MERCPROFILESTRUCT*> temp_mercStructs(NUM_PROFILES);
`

src/externalized/mercs/MercProfile.cc
```diff
@@ -90,4 +91,409 @@ MERCPROFILESTRUCT &MercProfile::getStruct() const
...
-}
\ No newline at end of file
+}
+
+void MercProfile::deserializeStructRelations(MERCPROFILESTRUCT* prof, const JsonObject& json, const ContentManager* contentManager)
+{
+	for (auto& element : json["relations"].toVec()) {
+		JsonObject r = element.toObject();
+		uint8_t idx = contentManager->getMercProfileInfoByName(r.GetString("target"))->profileID;
+		if (idx >= NUM_RECRUTABLE) {
```
momoko-h, line 101: Spelling: NUM_RECRUITABLE

src/externalized/mercs/MercProfile.cc
```diff
@@ -90,4 +91,409 @@ MERCPROFILESTRUCT &MercProfile::getStruct() const
...
+		else if (r.getOptionalBool("eventualEnemy")) {
+			prof->bLearnToHate = idx;
+			prof->bLearnToHateTime = r.getOptionalUInt("resistanceToMakingEnemy");
+		}
+	}
+}
+
+MERCPROFILESTRUCT* MercProfile::deserializeStruct(const JsonObject& json, const ContentManager* contentManager)
+{
+	MERCPROFILESTRUCT* prof = new MERCPROFILESTRUCT();
```
momoko-h, line 132: Please use unique_ptr here and return that from this function. It's preferable to make ownership of objects more obvious.

src/externalized/mercs/MercProfile.cc
```diff
@@ -90,4 +91,409 @@ MERCPROFILESTRUCT &MercProfile::getStruct() const
...
+	}
+
+	return prof;
+}
+
+JsonValue MercProfile::serializeStruct(const ContentManager* contentManager) const
+{
+	MERCPROFILESTRUCT* prof = this->m_profile;
+	JsonObject obj;
+	obj.set("000profile", this->getInfo().internalName);
```
momoko-h, line 281: Why "000profile", etc. instead of just "profile"?

src/game/Tactical/Soldier_Control.h
```diff
@@ -248,6 +248,8 @@ enum InvSlotPos
 	NUM_INV_SLOTS,
 };
 
+extern UINT8 gubItemDroppableFlag[NUM_INV_SLOTS];
```
momoko-h, line 251: This invalidates the comment above this array's definition in Soldier_Create.cc. And why did you place this declaration in this header? Seems a bit odd here.

src/game/Tactical/Soldier_Profile.cc
```diff
@@ -120,12 +120,12 @@ static void StartSomeMercsOnAssignment(void);
...
 {
-	{ AutoSGPFile f(GCM->openGameResForReading(BINARYDATADIR "/prof.dat"));
-		LoadRawMercProfiles(f, NUM_PROFILES, gMercProfiles, getDataFilesEncodingCorrector());
+	{
+		GCM->resetMercProfileStructs();
 		for (UINT32 i = 0; i != NUM_PROFILES; ++i)
 		{
 			MERCPROFILESTRUCT& p = gMercProfiles[i];
-
+
```
momoko-h, line 128: Please revert this.

src/game/Tactical/Soldier_Profile_Type.h
```diff
@@ -180,177 +181,166 @@ struct MERCPROFILESTRUCT
...
+	BOOLEAN fUseProfileInsertionInfo; // Set to various flags, ( contained in TacticalSave.h )
+	INT8 bTown;
+	INT8 bTownAttachment;
 	UINT8 ubRoomRangeStart[2];
-	UINT16 inv[19];
+	UINT8 ubRoomRangeEnd[2];
 
-	UINT16 usStatChangeChances[ 12 ]; // used strictly for balancing, never shown!
-	UINT16 usStatChangeSuccesses[ 12 ]; // used strictly for balancing, never shown!
+	INT8 bBuddy[5]{ -1, -1, -1, -1, -1 }; // Only indices 0, 1, 2 are used. Contain id's for friend1, friend2 and eventual friend respectively
```
momoko-h, line 281: I know this is already wrong in the current state but these are profile ID's so the type of this array and several other members of this struct should be ProfileID or at least UINT8. I'm not sure if we would run into problems if a mod added recruitable characters with profile IDs > 127 otherwise.

assets/externalized/mercs-profiles.json
```diff
@@ -0,0 +1,12326 @@
...
+                "slot": "BIGPOCK1",
+                "item": "CLIP38_6",
+                "quantity": 5
+            },
+            {
+                "slot": "BIGPOCK2",
+                "item": "TNT"
+            },
+            {
+                "slot": "BIGPOCK3",
```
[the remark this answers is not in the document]
ShishkinP: Turns out they get rearranged automatically: `HireMerc()` => `CopyProfileItems()` => `AutoPlaceObject()`. Along with auto-attaching DETONATOR to TNT and other stuff. They were probably added this way to display inventory on the MERC site.

src/externalized/mercs/MercProfile.cc
ShishkinP: This is a workaround to place fields logically grouped instead of alphabetically. I forgot to add a comment explaining this.

src/game/Tactical/Soldier_Control.h
ShishkinP: Since `gubItemDroppableFlag[]` defines which inv slots can be marked undroppable I placed it under the InvSlotPos enum and at the same time from here it makes it visible to MercProfile.cc.

src/game/Tactical/Soldier_Profile_Type.h
ShishkinP: We would certainly run into problems:
1. Currently only IDs < 75 (`NUM_RECRUITABLE` I had to add) can be hired. For example, in this same struct it's why `bMercOpinion[]`'s size is 75.
2. We need "-1" in these members because it's used throughout the codebase to denote a buddy or enemy slot is vacant. "0" would be BARRY's ID so we can't use it.
3. Besides, this structure is saved into a save file which is another headache.

So, I think changing the type is not realistic within this PR. It should be a separate undertaking.

ShishkinP: I forgot about BuddySlot and HatedSlot enums I added in #2059. I will use them as the type.

assets/externalized/mercs-profiles.json
ja2-derek: How about a simple document to log all mod-breaking changes? That would help people to change customized game settings or mods.

assets/externalized/mercs-profiles.json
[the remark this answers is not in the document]
momoko-h: I see. It might be surprising for modders when items end up in other slots than they specified but we could wait and see if someone complains.

ShishkinP:
Added: CHANGELOG.md
Seems to be all for now. Follow-up PRs have already been planned

src/game/Tactical/Soldier_Profile.cc
```diff
@@ -120,8 +120,8 @@ static void StartSomeMercsOnAssignment(void);
 
 void LoadMercProfiles()
 {
-	{ AutoSGPFile f(GCM->openGameResForReading(BINARYDATADIR "/prof.dat"));
-		LoadRawMercProfiles(f, NUM_PROFILES, gMercProfiles, getDataFilesEncodingCorrector());
```
momoko-h, line 124: Nice, looks like this was the only user of `getDataFilesEncodingCorrector` which means not only this function but the entire `IEncodingCorrector` interface can go. You don't have to do that in this PR but if it's not one of your planned followup PRs please let me know so I don't forget it.

src/externalized/DefaultContentManager.cc
```diff
@@ -1459,6 +1471,15 @@ const std::vector<const MercProfile*>& DefaultContentManager::listMercProfiles()
 	return m_mercProfiles;
 }
 
+void DefaultContentManager::resetMercProfileStructs() const
+{
+	for (size_t i = 0; i < NUM_PROFILES; i++) {
+		if (m_mercStructs[i] != nullptr) {
```
momoko-h, line 1477: Can any of the mercStructs pointers really be null? If yes, you should add an else branch to at least set gMercProfiles[i] to a default value.

src/externalized/internals/enums.h
```diff
@@ -1,15 +1,53 @@
...
+#include "MercProfileInfo.h"
+#include "Overhead_Types.h"
+#include "Soldier_Control.h"
+#include "Soldier_Profile_Type.h"
 
 namespace Internals
 {
 	/** Return approach name or NULL if not found. */
 	const char* getApproachName(enum Approach approach);
+	const Approach getApproachEnumFromString(const ST::string& s);
```
momoko-h, line 15: Please have a look at the Github runner logs for this PR. Marking the return value as `const`  doesn't do anything useful here but adds several new warnings (ignored-qualifiers) to the build.

src/game/Tactical/Soldier_Profile.cc
ShishkinP: No, It's not in my plans

src/externalized/DefaultContentManager.cc
ShishkinP: Yes, they can. I just put a safeguard against it in `loadMercsData()`

src/externalized/internals/enums.h
ShishkinP: Got it. My next PR is a cleanup one targeting enums.h among other things.

momoko-h:
LGTM now. I'll hold off merging for now to give the others a chance to chime in.

PR #2071, opened 2024-09-09
Implement TilesetTileIndexModel::serialize

momoko-h:
Before this function just always returned an empty string.


lynxlynxlynx:
LGTM, thanks!

PR #2073, opened 2024-09-12
Fix Realtime mode spread burst.

Danf96:
Removed a gpItemPointer check that I suspect used to not be correctly set to null as well as a check in Weapons.cc that caused pSoldier->sSpreadLocations[5] to never be used, leading to the last shot of a minimi burst to go to the start of the aimed/spread burst rather than the end.


momoko-h:
Looks good. Thanks!

PR #2074, opened 2024-09-14
Remove IEncodingCorrector and related code

momoko-h:
This interface is no longer needed after issue #2070.

Also remove the unused member function DataWriter:writeUTF32 and move some trivial functions to the LoadSaveData header.


lynxlynxlynx:
Thanks!

PR #2076, opened 2024-09-15
Change the way we initialize WeaponModel instances

momoko-h:
Instead of passign the strings for name, short name and description around from constructor to another we now initialize the WeaponModel member variables after we have constructed the object.

This should cut down the enormous amount of complaints Coverity (COPY_INSTEAD_OF_MOVE) has about WeaponModels.cc.

This file is currently responsible for the most CIDs by far with 103. This solution seems better to me than adding lots of `std::move` everywhere.

### DIFF
--- a/src/externalized/WeaponModels.cc
+++ b/src/externalized/WeaponModels.cc
@@ -25,7 +25,7 @@ enum
 	GUN_SHOTGUN
 };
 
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Thanks!

PR #2086, opened 2024-09-25
Factor out common code from two functions in Video.cc

momoko-h:
`InvalidateRegion` and `AddRegionEx` were the same function, just operating on different arrays. The now merely call a common helper function with references to these arrays.

Also removes some duplicated debug messages that are already present in SGP.cc


lynxlynxlynx:
LGTM

PR #2087, opened 2024-09-28
Add a constexpr overload for DirectionInc

momoko-h:
There are about 60 places where DirectionInc is called with one of the WorldDirections enumerators which always yields the same result. This result can already be computed at compile time, resulting in code that is a tiny bit smaller and faster.


lynxlynxlynx:
Thanks!

PR #2088, opened 2024-09-30
Bug fix: size of SaveStates in the save game header

momoko-h:
The size of ste SaveStates was written to the file header at the start of SaveGame but since #2051 was checked in, saving the strategic status added more data to those SaveStates. The result of this bug was that SaveGameInfo could no longer determine the mods activated for a save, resulting in wrong tool tips on the load screen and wrong warnings about a mismatch in the activated mods.

### DIFF
--- a/src/game/SaveLoadGame.cc
+++ b/src/game/SaveLoadGame.cc
@@ -80,6 +80,7 @@
 #include "Quests.h"
 #include "Random.h"
 #include "RenderWorld.h"
+#include "SGPFile.h"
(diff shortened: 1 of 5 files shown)


lynxlynxlynx:
Looks ok, but haven't done any testing.

momoko-h:
Seems the failure of the Mac runner was just temporary. Is there anything specific you'd like to see tested?

lynxlynxlynx:
No, was just noting the extent of my review.

PR #2089, opened 2024-10-04
Mark include directories of dependencies as SYSTEM

momoko-h:
This informs the compiler not to generate warnings for code in this header. This change was brought about by clang's complaints about miniaudio.h.


lynxlynxlynx:
Thanks!

PR #2092, opened 2024-10-06
Fix ReduceStringLength()

### LINKED ISSUE #1491: Tiny graphic issue in strategic map sector name display
The Drassen Airport is the only sector with this problem, since all other appear to have shorter names.
image
Could be a vanilla thing — bug or design choice. but it's worthy to check how we handle ellipsis. I couldn't easily find it, perhaps it's even in the data, but we do have a function for this in AttemptToAddSubstring.

EDIT: similarly in the sector inventory:
image

momoko-h:
Because the width variable was never updated this function didn't really do anything. Fixes #1491.
Screenshot_20241006_114634


selaux:
LGTM. Thanks.

PR #2093, opened 2024-10-10
Remove four unused variables in Creature_Spreading

momoko-h:
Just initialized to 0 but never modified or read anywhere.


lynxlynxlynx:
Best sorts of stats. :) +0 −12

PR #2094, opened 2024-10-10
Bring back support for modded prof.dat binaries

ShishkinP:
Was removed in #2070. Now it's back. ~If `prof.dat` is present in one of `mods\modname\data\BinaryData` dirs and the mod is enabled, the binary is used instead of the json data.~ 

Next in line is the restoration of modded .NPC files loading (removed in #2029).


momoko-h:
Hm, don't you need to bring back the workaround for the incorrectly encoded Russian prof.dat files I removed in #2074 in some form? I'll admit I know nothing about how popular these version were back then or if somebody's still using them.

ShishkinP:
You are right. It's worth restoring too.

src/externalized/DefaultContentManager.cc
```diff
@@ -1185,18 +1193,33 @@ bool DefaultContentManager::loadMercsData()
...
+		AutoSGPFile f(openGameResForReading(BINARYDATADIR "/prof.dat"));
+		bool const isCorrectlyEncoded = !(isRussianVersion() || isRussianGoldVersion());
+		for (int i = 0; i != NUM_PROFILES; ++i) {
+			BYTE data[MERC_PROFILE_SIZE];
+			JA2EncryptedFileRead(f, data, sizeof(data));
+			std::unique_ptr<MERCPROFILESTRUCT> prof = std::make_unique<MERCPROFILESTRUCT>();
+			UINT32 checksum;
+			ExtractMercProfile(data, *prof, false, &checksum, isCorrectlyEncoded);
+			// not checking the checksum
+			m_mercStructs.push_back(std::make_unique<const MERCPROFILESTRUCT>(*prof));
```
momoko-h, line 1222: No need to make a copy here, `m_mercStructs.push_back(std::move(prof));` should suffice.

src/externalized/DefaultContentManager.cc
```diff
@@ -1185,18 +1193,33 @@ bool DefaultContentManager::loadMercsData()
...
+			}
+		}
+	}
+	else {
+		AutoSGPFile f(openGameResForReading(BINARYDATADIR "/prof.dat"));
+		bool const isCorrectlyEncoded = !(isRussianVersion() || isRussianGoldVersion());
+		for (int i = 0; i != NUM_PROFILES; ++i) {
+			BYTE data[MERC_PROFILE_SIZE];
+			JA2EncryptedFileRead(f, data, sizeof(data));
+			std::unique_ptr<MERCPROFILESTRUCT> prof = std::make_unique<MERCPROFILESTRUCT>();
```
momoko-h, line 1218: I'd use auto here to avoid the repetition.

src/sgp/LoadSaveData.h
```diff
@@ -120,9 +121,10 @@ class DataReader
 	   encoded strings.	*/
 	auto readString(size_t const numChars, bool const stracLinuxFormat)
 	{
-		return std::invoke(
-			stracLinuxFormat ? &DataReader::readUTF32 : &DataReader::readUTF16,
-			this, numChars, ST_DEFAULT_VALIDATION);
+		if (stracLinuxFormat) {
+			return std::invoke(&DataReader::readUTF32, this, numChars, ST_DEFAULT_VALIDATION);
+		}
+		return std::invoke(&DataReader::readUTF16, this, numChars, true, ST_DEFAULT_VALIDATION);
```
momoko-h, line 127: You can just call readUTF32 and readUTF16 directly; I used invoke() here only because the C++ syntax to make a call through a pointer to a member function is so ugly that I can never remember it.

src/externalized/DefaultContentManager.cc
```diff
@@ -1185,18 +1193,33 @@ bool DefaultContentManager::loadMercsData()
...
-	json = readJsonDataFileWithSchema("mercs-relations.json");
-	for (auto& element : json.toVec()) {
-		JsonObject reader = element.toObject();
-		const MercProfileInfo* inf = this->getMercProfileInfoByName(reader.GetString("profile"));
-		MercProfile::deserializeStructRelations(temp_mercStructs[inf->profileID].get(), reader, this);
-	}
-	for (auto& element : temp_mercStructs) {
-		if (element != nullptr) {
-			m_mercStructs.push_back( std::make_unique<const MERCPROFILESTRUCT>( *element ) );
+	if (!doesGameResBelongToMod(BINARYDATADIR "/prof.dat")) {
```
selaux, line 1196: Not sure if this is the best approach to do this. We could also always load `prof.dat` and only use it if no JSON data is provided. Similar to what is done with `VanillaItemStrings`, where the data is passed to the individual JSON deserialization methods of the items.

ShishkinP:
@selaux Would something like this work?
```c++
// If there are no json fields beside the 3 mandatory ones (profileID, internalName, type)
if (charProperties.keys().size() == 3) {
	// use only the prof.dat data
	temp_mercStructs[profileID] = std::move(binary_mercStructs[profileID]);
}
else {
	temp_mercStructs[profileID] = MercProfile::deserializeStruct(binary_mercStructs[profileID].get(), charProperties, this);
}
``` 
And the deserialization method using the passed binary data (only with displayed strings) :
```c++
prof->zName = r.has("fullName") ? r.GetString("fullName") : binaryProf->zName;
prof->zNickname = r.has("nickname") ? r.GetString("nickname") : binaryProf->zNickname;
```

selaux:
I am a bit confused, but maybe that is because I don't fully understand what is stored in `prof.dat` and `MercProfile` exactly. Won't using this if approach imply that as soon as I set e.g. `agility` for a MERC in JSON, it will use all of the default values for the other stats of the MERC? Why not have this logic for all properties that have related information in `prof.dat`. E.g.

```
prof->bLifeMax = stats.getOptionalUInt("health", binaryProf->bLifeMax);
```

This way we wouldn't need the `if` at all and could just always merge the binary and JSON representations, with properties defined in JSON being preferred over `prof.dat`. I would also imply that for vanilla data we would just need the minimal data in the JSON file, as most of the data is in `prof.dat` anyways.

ShishkinP:
OK, bin-to-json merging it is. Also `VanillaItemStrings` is renamed to `BinaryData` and now contains `prof.dat` profiles as well.

src/externalized/mercs/MercProfile.cc
```diff
@@ -178,36 +207,52 @@ std::unique_ptr<MERCPROFILESTRUCT> MercProfile::deserializeStruct(const JsonValu
...
+	ST::string jEvolution = stats.getOptionalString("evolution");
+	if (jEvolution.empty()) {
+		prof->bEvolution = binaryProf->bEvolution;
+	}
+	else if (jEvolution == "NONE") {
+		prof->bEvolution = CharacterEvolution::NO_EVOLUTION;
+	}
+	else if (jEvolution == "REVERSED") {
+		prof->bEvolution = CharacterEvolution::DEVOLVE;
+	}
```
momoko-h, line 223: Probably worth at least a comment here that normal evolution does not need to be handled here because *prof was default initialized and NORMAL_EVOLUTION's numerical value is 0.

src/externalized/mercs/MercProfile.cc
```diff
@@ -178,36 +207,52 @@ std::unique_ptr<MERCPROFILESTRUCT> MercProfile::deserializeStruct(const JsonValu
...
+	prof->bSkillTrait2 = stringForEnum.empty() ? binaryProf->bSkillTrait2 : Internals::getSkillTraitEnumFromString(stringForEnum);
+
+	stringForEnum = r.getOptionalString("attitude");
+	prof->bAttitude = stringForEnum.empty() ? binaryProf->bAttitude : Internals::getAttitudeEnumFromString(stringForEnum);
 
 	ST::string jSector = r.getOptionalString("sector");
-	if (!jSector.empty()) {
+	if (jSector.empty()) {
+		prof->sSector = binaryProf->sSector;
+		prof->sSector.z = binaryProf->sSector.z;
```
momoko-h, line 253: Why? sSector.z should already be set by the line above.

src/externalized/strings/ItemStrings.cc
```diff
@@ -1,58 +1,72 @@
...
 #define SIZE_SHORT_ITEM_NAME  80
 #define SIZE_ITEM_INFO       240
 
-VanillaItemStrings::VanillaItemStrings() {}
+BinaryData::BinaryData() {}
 
 // Needed to be able to return a reference to an empty string which is not local
 const ST::string EMPTY_STRING = "";
+// This is to avoid failures due to prof.dat unavailability during unit-testing
+static const MERCPROFILESTRUCT* DEFAULT_PROFILE = new MERCPROFILESTRUCT();
```
momoko-h, line 17: There's no need to heap allocate this and I'd prefer if you defined this in getProfile, the only place where it is used. That would also explain the the mProfiles.size() check in that function which ATM is a bit obscure.

src/externalized/strings/ItemStrings.cc
```diff
@@ -1,58 +1,72 @@
...
 	if (found == items.end()) {
 		return EMPTY_STRING;
 	}
 	return (*found).second.description;
 };
 
-ST::string VanillaItemStrings::filename() {
+const MERCPROFILESTRUCT* BinaryData::getProfile(ProfileID profileIndex) const
+{
+	if (profiles.size() == 0) return DEFAULT_PROFILE;
```
momoko-h, line 45: `if (profiles.empty()) ...`

momoko-h (commented):
Other than those small nits this is LGTM.

PR #2095, opened 2024-10-12
Modernize random number generation for creature spreading

momoko-h:
DetermineCreatureTownComposition and PrepareCreaturesForBattle are almost text book examples for discrete_distribution use cases.

### DIFF
--- a/src/game/Strategic/Creature_Spreading.cc
+++ b/src/game/Strategic/Creature_Spreading.cc
@@ -1028,6 +1028,7 @@ bool GetWarpOutOfMineCodes(SGPSector& sector, INT16* const insertion_grid_no)
 
 #ifdef WITH_UNITTESTS
 #include "gtest/gtest.h"
+#include <numeric>
 
 TEST(CreatureSpreading, CreatureWeightsAddUpTo100Percent)
 {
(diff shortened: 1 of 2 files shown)


src/game/Strategic/Creature_Spreading.cc
```diff
@@ -709,38 +711,56 @@ BOOLEAN MineClearOfMonsters( UINT8 ubMineIndex )
 	return TRUE;
 }
 
+namespace {
+using Weights = std::array<UINT8, 6>;
+constexpr std::array CreatureWeightsByHabitat
```
selaux, line 716: Could this be extracted into a JSON file, e.g. `strategic-map-creature-habitats.json`? As far as I can see, the only things that are needed to make this fully governed by json is the flag whether the queen is in the habitat and the max population of a habitat. E.g.

```json
{
  "name": "QUEEN_LAIR",
  "hasQueen": true,
  "maxPopulation": 32,
  "populationWeights": {
    "larvae": 20,
    "infant":  40,
    ...
  }
}
```

momoko-h: Sure, the table I added could be extended and converted to JSON. I'm wondering though if there is really any demand for it? I got the impression that Sci-Fi mode isn't really popular with modders and I doubt gamers are waiting for the opportunity to customize the monster populations.

My suggestion is to take this PR as is and if somebody wants additional functionality they can create a new issue with an enhancement request.

selaux: Fine for me.

selaux:
LGTM.

PR #2096, opened 2024-10-13
Remove unused code in Strategic_AI

momoko-h:
Two functions that were defined but never called and one function declaration without a matching definition.


momoko-h:
Since this is pure dead code removal I am going to merge it without further review. Please let me know if you have any objections.

PR #2097, opened 2024-10-13
Externalize all explosive effects

selaux:
This is the next step in externalizing explosives: Externalizing all the effects. Now in `explosives.json` we can choose which effects are caused by an explosive:

- `blastEffect`: Effect damaging to health, structures and items on the ground
- `stunEffect`: Effect damaging to stamina
- `smokeEffect`: Smoke, teargas, etc.
- `lightEffect`: Whether the explosive generates a light effect on the ground

Previously those effects were exclusive, except for the blast and stun effects. Now they can be mixed arbitrarily.

What was also extracted was the fact whether an explosive is triggered by pressure (e.g. mines).

The PR is reviewable, but as most of the time I still need to do some testing, especially around mines whether everything is working correctly.

As a preview, here is a grenade launcher grenade that has all of the effects listed above:

[Screencast from 2024-10-13 14-46-34.webm](https://github.com/user-attachments/assets/834389c9-36d5-48ba-b33f-26cad08dc46f)


src/externalized/ExplosiveModel.h
```diff
@@ -23,41 +51,49 @@ struct ExplosiveModel : ItemModel
...
-		uint8_t stunDamage,
-		uint8_t radius,
 		uint8_t noise,
 		uint8_t volatility,
-		uint8_t type,
+		bool isPressureTriggered,
+		const ExplosiveBlastEffect *blastEffect,
+		const ExplosiveStunEffect *stunEffect,
+		const ExplosiveSmokeEffect *smokeEffect,
+		const ExplosiveLightEffect *lightEffect,
```
momoko-h, line 60: What do you think about using std::optional for these four member variables? Just an idea, not terribly important. It would save a bit of manual memory management.

momoko-h (approved):
I didn't see anything problematic, so if you're confident this works please merge it.

lynxlynxlynx:
LGTM too.

selaux:
Did some testing. Looks good so I will merge.

PR #2101, opened 2024-10-19
github-ci: macos-12 is getting sunset, switch to latest

lynxlynxlynx:
Trivial stuff, one other cases is already using this runner.

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -22,7 +22,7 @@ jobs:
           target: linux-mingw64
 
         - name: Mac
-          os: macos-12
+          os: macos-latest
           target: mac
 
         - name: Android


momoko-h:
This would also change the architecture from Intel to ARM, so the artifacts produced by this Runner would no longer work on older Macs, right? Do we care about that? Of course, sooner or later the last Intel-compatible version of MacOS will no longer be supported by Github anyway.

Also noteworthy: since this required a full rebuilt sccache's hit ratio was 0% but the build still needed only 6 minutes to complete. Do we even need it?

lynxlynxlynx:
Yes, we could switch to -13 while it's still available, but -14 is already M1+. I have no idea which users would be upset more, since we here from apple fans so rarely, so I went for the pragmatic choice.

As for sccache, it probably still saves some upstream resources.

momoko-h:
That's true, I also have no idea whether Mac users would prefer support for older machines or native builds for their more recent systems. Maybe someone else could chime in with some insights.

momoko-h:
Alternative idea: how about having two MacOS runners, one MacOS 13 for Intel, one -latest for ARM? To make this a bit more useful, one of them could exercise some CMake options we're currently not testing, for example WITH_MAGICENUM=OFF or WITH_MAEMO=ON.

lynxlynxlynx:
Apparently recent macos versions don't even report their version correctly, so there's no usable market share data.
A compromise would be to remain on latest, but also build releases for x86.

selaux:
Question is whether anyone in the team can even test macOS x86 anymore. I can test at least the ARM version again, as I have a machine available, but I don't have a x86 mac anymore. With the current x86 build, I can at least test through the macOS compatibility layer. Not sure if it would make sense to push out a build that has not, at least, been started up once to check whether it works properly.

NB: For the universal build, we would also need to check whether all of the libraries (including the rust stuff) could be built universally.

lynxlynxlynx:
I agree and from previous experience, making fat builds is not easy. Even MSP hosts separate builds. Anyway, any objection to merging?

momoko-h:
No, let's go ahead. We'll see if we get some reaction from Intel Mac users.

PR #2105, opened 2024-10-23
Fix memory leak in Strategic_AI

momoko-h:
Two arrays were leaked if you started a new game and then quit that game and started another new one.

Looking a bit closer the array gubPatrolReinforcementsDenied was not actually used for anything.

### DIFF
--- a/src/game/Strategic/Strategic_AI.cc
+++ b/src/game/Strategic/Strategic_AI.cc
@@ -35,6 +35,7 @@
 #include "StrategicAIPolicy.h"
 #include "StrategicMap.h"
 #include "Town_Militia.h"
+#include <memory>
 #include <vector>
 
 #define SAI_VERSION		29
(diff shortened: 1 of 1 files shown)


src/game/Strategic/Strategic_AI.cc
```diff
@@ -137,8 +138,7 @@ extern UINT8 gubNumGroupsArrivedSimultaneously;
 //group.  When the queen wants to send forces to attack a town that is defended, the initial number of forces that
 //she would send would be considered too weak.  So, instead, she will send that force to the sector's adjacent sector,
 //and stage, while
-UINT8 *gubGarrisonReinforcementsDenied = NULL;
-UINT8 *gubPatrolReinforcementsDenied = NULL;
+std::unique_ptr<UINT8 []> gubGarrisonReinforcementsDenied;
```
selaux, line 141: Couln't we just use a `std::vector` instead?

momoko-h: Sure, but we wouldn't use any of vector's features, most notably its flexibility regarding its size. This array always has as many elements as are defined in army-garrison-groups.json. In theory we could even get rid of the allocation in LoadStragicAI and just reuse the already allocated memory but that seems one micro-optimization step too far to me.

selaux:
LGTM.

PR #2106, opened 2024-10-24
Fix some copy instead of move coverity issues

selaux:
This fixes some of the newly introduced `COPY_INSTEAD_OF_MOVE` issues from coverity. Not sure if this is the best way to do it, but I haven't found a foolproof way to do this. At least the rvalue references introduced in this PR forces some of the `std::move` calls, e.g. when calling the constructor or forwarding to the parent class constructor. It's still prossible to omit it in other locations though, e.g. when setting the class members.


momoko-h:
Thanks for doing this. Personally I don't believe we should spend too much time trying to figure out the perfect way to implement this just to satisfy Coverity. It's a tool that's intendend to help us find real problems, not to be our slave driver.

Could you please also fix the new warning in ExplosiveModel.cc line 119 (assigning field to itself [-Wself-assign-field])? It seems the type member variable is unused.

selaux:
Removed the type and the related enum, which is now unused.

momoko-h:
Thanks!

PR #2107, opened 2024-10-25
Fix possible null pointer dereference when ignoring smoke effects

selaux:
Not sure if that would ever happened, as a robot cannot have a gas mask equipped, but it was detected by Coverity as a possible null pointer dereference if the first condition was evaluated as `true`.

### DIFF
--- a/src/game/Tactical/Overhead.cc
+++ b/src/game/Tactical/Overhead.cc
@@ -1524,18 +1524,14 @@ BOOLEAN HandleGotoNewGridNo(SOLDIERTYPE* pSoldier, BOOLEAN* pfKeepMoving, BOOLEA
 					bPosOfMask = NO_SLOT;
 				}
 
-				const SmokeEffectModel* smokeEffect = nullptr;
 				auto smokeEffectID = GetSmokeEffectOnTile(pSoldier->sGridNo, pSoldier->bLevel);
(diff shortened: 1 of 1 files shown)


momoko-h:
(Just for reference, this is CID 417260).

Hm, isn't the real problem here that we are inside an if block that can only be executed when ubExtFlags has ANY_SMOKE_EFFECT set, so that the SmokeEffectID for this tile cannot be NOTHING? I believe Coverity would still complain that bPosOfMask could be != NO_SLOT but smokeEffect could still be NULL in your new line 1535.

I think a better fix would be this:
```

				auto smokeEffectID = GetSmokeEffectOnTile(pSoldier->sGridNo, pSoldier->bLevel);
				auto * smokeEffect = GCM->getSmokeEffect(smokeEffectID);
				Assert(smokeEffect != nullptr);

				// check the cases where we dont take damage
				if ((bPosOfMask != NO_SLOT && !smokeEffect->getIgnoresGasMask()) ||
				    (AM_A_ROBOT(pSoldier) && !smokeEffect->getAffectsRobot())) {
					smokeEffect = nullptr;
				}

```

selaux:
Good catch. Thank you :slightly_smiling_face:. I've adapted the implementation now.

momoko-h:
Thank you!

PR #2109, opened 2024-10-26
Layered patches and scripts

selaux:
This implements a layered approach to load mods for two cases:

### JSON patches

**Previously**: The JSON file of the VFS layer with the highest priority was combined with the patch with the highest priority. This only allows a single mod to apply its patch to a specific JSON file and ignores mod order (a patch from a lower prio mod will always be applied, if it is the only one).

**Now** : The JSON file of the VFS layer with the highest priority is combined with the patches from the layers with the same or higher priorities. This allows patches of multiple mods to be applied to a single JSON file. Ignoring the lower layers is assuming that mods that have full json files can be considered "total conversions". If the lower patches should still be applied, the user can move the patching mod above the "total conversion" in the mod order.

### Lua Scripts

For lua scripts the situation is much easier. Instead of only executing the one with the highest priority, we now execute them all from lowest to highest priority. This allows mods to register their scripting extension, even if a higher prio mod that also has a scripting extension exists.


lynxlynxlynx:
Is priority determined by mod order, the order `Vfs_readLayers` produces (likely alphabetical?), something else?
I'll write some docs for the site, but it's also important to know for the launcher.

selaux:
The order is determined by the inverse order `Vfs_readLayers` produces. And that is the same as mod order from bottom to top (plus other layers).

E.g. with the following mods enabled:

image

This is the VFS layers priorities (excluding the SLF files on the lowest layer):

image

and that order will be returned by `Vfs_readLayers` from top to bottom. For patches and scripts we iterate it from bottom to top, so it will start looking into `/home/stefan/.ja2/data_gold/Data` and end with the highest prio mod dir `/home/stefan/.ja2/Mods/omerta-shooting-range/data`.

selaux:
Thinking about it, maybe it makes sense to change the priority of the `/home/stefan/.ja2/data` layer, as that one probably should overwrite any mod contents.

lynxlynxlynx:
Writing that, wouldn't it be more intuitive if the last item in the mod list had the highest priority (and we appended to the bottom in the launcher)? If we made it more visible, eg. by also writing ordinals besides the names, it would become obvious the list is sorted in descending order now.

selaux:
For me the intuitive way would be as it is now. Because it conveys a sense of priority (high priority -> top of the list). Writing ordinals beside the names imho would not change anything, because then it is just a numeric list...

lynxlynxlynx:
There's no indication of priority now, besides the log, that's the core issue.
I read top down, but also, in general, when installing two packages, I expect the second one to have precedence if there's any overlap. Package managers also list and install dependencies first. In this case also reasoning about mod/patch order would be simpler, since the order would be the same (FIFO), not reversed.

selaux:
So my intuiton might be wong here ;). I've checked other games and most do top -> first, bottom -> last loading.

- Factorio: Automated based on mod dependencies and naming [ref](https://forums.factorio.com/viewtopic.php?t=8554)
- Stellaris: Top are loaded first, bottom is loaded last [ref](https://forum.paradoxplaza.com/forum/threads/how-does-the-new-launcher-work-with-mod-load-order-exactly.1292310/)
- Skyrim: Top are loaded first, bottom is loaded last [ref](https://skyrimseblog.wordpress.com/load-order-structure/)
- Rimworld: Top are loaded first, bottom is loaded last [ref](https://www.reddit.com/r/RimWorld/comments/11xmo5b/is_there_any_order_that_mods_should_be_loaded_in/)

My suggestion would be the following: Because it's probably not a good idea change the mod order in JSON, we can simply change how it is displayed in the launcher and add a comment about how the loading works as well. But I'd prefer to do that in a separate PR.

selaux:
Or we also change it in the JSON file, which might be a breaking change for users that have mods enabled that change the same files.

lynxlynxlynx:
Wait, I don't see priority stored anywhere in mods; do you mean just the mods key in ja2.json?

selaux:
Yes, the order of the `mods` array in `ja2.json`.

PR #2110, opened 2024-10-29
More copy instead of move fixes

selaux:
This attempts to fix all of the occurences inside a `deserialize` method.


lynxlynxlynx:
Thanks!

PR #2111, opened 2024-11-01
Minor cleanup of Physics.cc

momoko-h:
• Removed several unused preprocessor macros
• Removed second parameter of SimulateObject (always DELTA_T) • Removed always false variable fDampingActive
• Changed return type of two functions that always returned
  TRUE to void
• Removed unused last parameter of PhysicsHandleCollisions • Removed overriden second parameter of CalculateSoldierMaxForce • Fixed two incorrect comparisons of structure IDs


selaux:
Thanks :+1:

PR #2112, opened 2024-11-03
Enable CMake policy CMP0156

momoko-h:
This policy de-duplicates libraries when the linker is called and was added in CMake 3.29.

https://cmake.org/cmake/help/latest/policy/CMP0156.html

The point of this PR is to find out if the long link times of the Mingw runner are caused by the huge number of link libraries added by the Rust build step.


momoko-h:
Even with this change linking JA2.exe still takes more than three minutes so I am closing this PR as useless.

PR #1706, opened 2022-11-15
Remove some unused files and one function

momoko-h:
The Quantize code is unused, topic IDs were part of the old debug system and Multi_Language_Graphic_Utils.h just included GameRes.h. TrashMapTile() is not used anywhere.

PR #2114, opened 2024-11-10
Revive the radar map utility

momoko-h:
This commit brings back the utility screen to create radar maps for all map files found in the user's home directory that was removed in #1300.

Also reverts part of #1706.

I chose to make this utility available the same way the 1.13 editor does it because I assumed modders are likely to be familiar with it: by adding a new button on the editor's options tab. After clicking that button the utility does its job and then exits the program.

The code in MapUtility.* and Quantize.* is mostly in the same state it was before it got removed, with just the modifications necessary to get the list of map file names.

In a very brief test this seemed to work, it produced readable radar map files for all maps in my home directory which looked ok at a glance. Nevertheless it would be someone who is more familar with modding maps could give this a more extensive test run.


lynxlynxlynx:
Never tried modding maps, so can't be of use there.

momoko-h:
Me neither, I only start the editor to investigate bug reports. Still it looks like a glaring omission that our tools cannot produce radar maps and modders have to use other editors instead.

Since this PR cannot have a negative impact on the game itself I believe it should be safe to merge this so we can get some feedback from interested people.

lynxlynxlynx:
Agreed, thanks.

Issue #1822, opened 2023-05-05
Some prone enemies force the game to freeze and become untouchable


Ilyab99:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

### What is wrong
prone enemies in sectors are unable to be interacted with force the game into a time clock freeze 

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |

V0.20.0-git+439a6d3(build  (04.12.02) |   JA2 Gold | no | yes 

[//]: # (Please also include:)

save files https://ufile.io/ihr85yt1 all of them but quicksave.sav. Run into the building in alma and try to interact with the prone soldier

OS: Windows 10 enterprise IOT LTSC 21H2
//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)
[ja2.log](https://github.com/ja2-stracciatella/ja2-stracciatella/files/11406066/ja2.log)

PR #2116, opened 2024-11-12
Fix prone positioning

ShishkinP:
Some of the problems with the prone stance:

1. You can go prone when closely facing a structure that is **whole-tile** blocking, but not when facing north. And, if facing away, not when facing south or east:
from-into-tree
 (reminder: the game world north is the direction Dynamo is looking in on the first pic)
2. Crawling along the south (and only south) side of such structures looks like this:
crawling
crawling2
Crawling along all other sides moves the char instantaneously - no stance change needed.
3. None of the above happens with walls and doors. Instead, they have a clipping problem that results in soldiers affecting or being affected by things illogically, like obstructing paths or taking damage with one part of their structure phased through a wall:
wallClips
The only reason it's not very game-breaking is because that part is invisible (no matter if it's clipped or not), so enemies can't choose to attack it.

All of that comes from structure DB descriptions that don't make sense:
structures

The fix for all 3 problems is trivial: flag all "extended" tiles as passable, and you can drop prone anywhere independent of what's in the neighboring tiles.
Another gameplay effect so far: chars lying in the first 3 (out of 8) directions have one less tile to take shockwave and gas damage with.

Probably solves #1822 (the save file is unavailable but if I remember correctly the prone soldier was facing south while phased into a box with his non-passable tile)

### DIFF
--- a/src/game/Tactical/Animation_Data.cc
+++ b/src/game/Tactical/Animation_Data.cc
@@ -550,6 +550,27 @@ void InitAnimationSystem()
 			if (GCM->doesGameResExists(Filename))
 			{
 				STRUCTURE_FILE_REF* pStructureFileRef = LoadStructureFile(Filename);
+
+				// fix non-base prone tiles by making them all passable (#2115)
+				if ((cnt1 >= REGMALE && cnt1 <= REGFEMALE) && cnt2 == P_STRUCT)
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Very thorough, thanks!

PR #2121, opened 2024-11-16
Fix missing buttons in MapScreen

### LINKED ISSUE #2119: When exiting the game with keyboard shortcut from the strategic screen, some graphics are missing
### What is wrong

When you're on the strategic screen and you pres Alt+X fro exiting the game, message for confirmation will appear. But you see, that with that message in foreground (before you confirm and exit the game), some graphics in the background (like sector coordinates and buttons bellow the map) are now missing. The vanilla behaves correctly.

|  latest nightly | JA2 1.06 EN  |    no    |    ?     |

JA2_Exit_1
JA2_Exit_2

momoko-h:
This removes the secret handshake between the message box screen and the map screen where the former asked the latter to redraw itself without the cursor interaction over the Arulco map before rendering itself. Some code change in the past (I did not investigate which) broke this brittle concept and made some buttons and the map coordinates disappear.

Now the map screen redraws itself without the mentioned interaction one last time before handing control over to the message box screen.

The function `HandleTacticalUILoseCursorFromOtherScreen` which does something similar for the game screen is also a bit suspect but unless someone can point to an actual problem with it I'd rather not touch it.

Fixes #2119.

### DIFF
--- a/src/game/MessageBoxScreen.cc
+++ b/src/game/MessageBoxScreen.cc
@@ -134,7 +134,6 @@ void DoMessageBox(MessageBoxStyleID ubStyle, const ST::string& str, ScreenID uiE
 
 	if (fInMapMode)
 	{
-		gfStartedFromMapScreen = TRUE;
 		fMapPanelDirty         = TRUE;
 	}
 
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Yeah, looks fine, thanks!

PR #2126, opened 2024-11-21
Fix crash in StartInterrupt()

momoko-h:
In some rare situations, this function can get called when none of the members of one of involved teams can really act (either dying or out of breath) which means that its AI list would be empty. StartInterrupt() didn't handle this and just tried to use the first entry of the AI list and then crashed when it tried to deref a NULL pSoldier.


lynxlynxlynx:
Nice investigation, thanks!

PR #2127, opened 2024-11-22
Fix out of bounds read in SameBuilding

momoko-h:
Found by valgrind. If a merc and his team have no idea about an opponent's location yet, `ClosestKnownOpponent` calls `SameBuilding` with sGridNo == NOWHERE, which uses it unchecked an as an index to the `gubBuildingInfo` array.

The closely related function `ClosestReachableDisturbance` already checks for this so it is not affected.


lynxlynxlynx:
Thanks!

PR #2130, opened 2024-12-03
Remove unused function LoadSoldierInitListLinks

momoko-h:


lynxlynxlynx:
Thanks, was part of the old save variant 0e32670.

PR #2131, opened 2024-12-10
Simplify RecompileLocalMovementCosts

momoko-h:
Reuse `RecompileLocalMovementCostsFromRadius` to implement this function; they are basically identical except that the former uses a fixed 'radius' of 4.


lynxlynxlynx:
Indeed, nice find.

PR #2134, opened 2024-12-15
Remove SetProgressBarTextDisplayFlag

momoko-h:
This function was only used in code inside JA2BETAVERSION blocks, apparently to give the developers a way to observe the loading progress. Not only is that code gone now, on modern computers those messages would disappear far too fast to be usable.

This also makes two flags and their related code redundant.


lynxlynxlynx:
Thanks!

PR #2135, opened 2024-12-16
Remove gfForceMusicToTense

momoko-h:
Looking back at the original sources it seems this variable never did anything at all.


momoko-h:
I should clarify that the variable never did anything outside the JA2 demo.

PR #2136, opened 2024-12-20
Properly center the history table header columns

momoko-h:
The three column headers were all supposed to be centered but the 'Event' header was not since the the vanilla days.

The imminent cause of this bug is almost certainly a case of copy & paste and then forgetting to adjust the arguments but I believe the real underlying problem is the text API which makes this often needed task far more cumbersome to use than necessary:

The first method is to use FindFontCenterCoordinates or FindFontRightCoordinates followed by MPrint. This method requires passing several arguments twice.

The second method is the DrawToScreen function. This is just one call but it doesn't respect the currenly active font and colors and instead requires the caller to always pass in this information.

For this reason this commit adds a new MPrint overload that takes an additional argument to specify the alignment.

This is how it looks currently:
screenshot_20241220_174831

With this commit:
screenshot_20241220_181159

### DIFF
--- a/src/game/Laptop/History.cc
+++ b/src/game/Laptop/History.cc
@@ -387,7 +387,9 @@ static void DisplayHistoryListHeaders(void)
 
 	// event header
 	x += RECORD_LOCATION_WIDTH;
-	MPrint(x, RECORD_HEADER_Y, pHistoryHeaders[4], CenterAlign(292));
+	// 471 is the width in pixels of one row (the width of guiSHADELINE index 0).
+	constexpr int RECORD_EVENT_WIDTH{ 471 - RECORD_DATE_WIDTH - RECORD_LOCATION_WIDTH };
+	MPrint(x, RECORD_HEADER_Y, pHistoryHeaders[4], CenterAlign(RECORD_EVENT_WIDTH));
 
 	// reset shadow
 	SetFontShadow(DEFAULT_SHADOW);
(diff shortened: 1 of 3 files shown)


src/game/Laptop/History.cc
```diff
@@ -377,20 +377,18 @@ static void DisplayHistoryListHeaders(void)
...
-	MPrint(usX, RECORD_HEADER_Y, pHistoryHeaders[3]);
+	// Location header
+	x += RECORD_DATE_WIDTH;
+	MPrint(x, RECORD_HEADER_Y, pHistoryHeaders[3], CenterAlign(RECORD_LOCATION_WIDTH));
 
 	// event header
-	FindFontCenterCoordinates(RECORD_DATE_X + RECORD_DATE_WIDTH + RECORD_LOCATION_WIDTH + 5,0,RECORD_LOCATION_WIDTH,0, pHistoryHeaders[ 3 ], HISTORY_TEXT_FONT,&usX, &usY);
-	MPrint(usX, RECORD_HEADER_Y, pHistoryHeaders[4]);
+	x += RECORD_LOCATION_WIDTH;
+	MPrint(x, RECORD_HEADER_Y, pHistoryHeaders[4], CenterAlign(292));
```
lynxlynxlynx, line 390: Where is this 292 from? Would be better if it was symbolic

PR #2137, opened 2024-12-21
use std::swap instead of a hand-rolled version

lynxlynxlynx:
Another case of stdlib allowing for some minor cleanup.


momoko-h:
Nice, less crud in `Types.h `can't be wrong.

PR #2138, opened 2024-12-22
Use the new alignment functions in a couple of places

momoko-h:
Achieves the same result with less boilerplate code.

The code in `Email.cc` was particularly weird and passed in nonsensical values for the height. This only worked because the computed y value was just discarded.

### DIFF
--- a/src/game/Laptop/EMail.cc
+++ b/src/game/Laptop/EMail.cc
@@ -1538,27 +1538,22 @@ static void DisplayEmailMessageSubjectDateFromLines(Email* pMail, INT32 iViewerY
 	SetFontAttributes(MESSAGE_FONT, FONT_BLACK, NO_SHADOW);
 
 	// all headers, but not info are right justified
-	INT16 usX;
-	INT16 usY;
+	RightAlign const alignment{ MESSAGE_HEADER_WIDTH };
 
 	// print from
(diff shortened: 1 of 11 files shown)


src/game/Laptop/EMail.cc
```diff
@@ -1538,27 +1538,22 @@ static void DisplayEmailMessageSubjectDateFromLines(Email* pMail, INT32 iViewerY
 	SetFontAttributes(MESSAGE_FONT, FONT_BLACK, NO_SHADOW);
 
 	// all headers, but not info are right justified
-	INT16 usX;
-	INT16 usY;
+	RightAlign const alignment{ MESSAGE_HEADER_WIDTH };
```
lynxlynxlynx, line 1541: This is quite confusing even before the change — now all the content is left aligned.

momoko-h: I'm not quite sure what you mean? This function only draws the text above the actual email content (the 'From:', 'Day:' and 'Subject:' strings) and those are right aligned with or without this change.

lynxlynxlynx: I don't see anything right-aligned:
image

momoko-h: The strings From:, Day: and Subject: are right aligned. The content of the fields ('Enrico' and '1 are supposed to be left aligned, at least based on the original code. The Subject content ('Inquiry') has this left margin in vanilla too.

It might look better if the Day field was right aligned and the Subject content didn't have that margin but I didn't want touch the vanilla look with this PR and mix code cleanup with a change in behavior.

lynxlynxlynx: Ahh, I see, thanks.

momoko-h:
Ok, this is all I want to do here for now. I believe it works as expected, text alignment is the same as before.

lynxlynxlynx:
Thanks!

PR #2140, opened 2024-12-23
Use SDL to simplify AddVideoSurfaceFromFile

momoko-h:
In its original form `CopyImageToBuffer` supported copying compressed data to a surface. It lost that ability long ago apparently because it was not used and now this function is just an elaborate way to copy raw data from a `SGPImage` to a `SGPVSurface`.

`SDL_ConvertPixels` can do this copy just as well which allows us to remove `CopyImageToBuffer` and its helper functions.

The loading screens and the map of Arulco on the map screen are among the users of this function if you want to verify that this approach works.

### DIFF
--- a/src/sgp/HImage.cc
+++ b/src/sgp/HImage.cc
@@ -6,10 +6,8 @@
 #include "ImpTGA.h"
 #include "PCX.h"
 #include "STCI.h"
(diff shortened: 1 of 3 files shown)


lynxlynxlynx:
Nice!

PR #2141, opened 2024-12-26
Use C++ keywords instead of FLOAT, DOUBLE, TRUE and FALSE

momoko-h:
There is really no good reason to still use these eyesores. The C++ standard mandates that in integer conversions true must be converted to 1 and false must be converted to 0.

The end of the year seems like a good time to throw out some old junk. The only file that's a bit more interesting in this PR is Types.h where the typedefs and definitions were removed.


lynxlynxlynx:
Practically this would make merging any part of #738 even harder, so I wouldn't go for it just yet. Also, if we go this route, why not remove the other typedefs as well?

momoko-h:
TRUE and FALSE are macros which make them extra undesirable in my opinion. FLOAT and DOUBLE are special because they can be replaced straight up with keywords with the same number of characters, which means the code formatting is not changed. 

The other types do not have that distinction, they rely on other typedefs themselves and that means there are several options: for example, should `INT16` be replaced with `int16_t`, `std::int16_t` or SDL's `Sint16`?

But ok, if you still hold out hope for the 32-bit patch, then let's drop this, it's not really important.

lynxlynxlynx:
I didn't voice it publicly, but I do commit to at some point extract at least the resolution stuff out (the work on making positional defines dynamic).

PR #2142, opened 2024-12-27
Remove IGameDataLoader.h

momoko-h:
Apparently class IGameDataLoader was supposed to be an interface definition but because its only function is not declared virtual this is really a class without any data members and a declaration for a function that does not exist anywhere.

The actually existing and used readJsonDataFile is a non virtual member function of DefaultContentManager.


lynxlynxlynx:
Good find, thanks!

PR #2144, opened 2025-01-02
Add a new BltVideoObject that loads vobjects on demand

momoko-h:
This new overload has the object's file name as a parameter instead of a SGPVObject *. Unlike the already existing function BltVideoObjectOnce which discards the object after use, the new function uses a very trivial cache so that subsequent blits do not have to reload the object.

Advantages of this approach:
• Less boilerplate code. Several setup functions disappear completely and the number of global variables is reduced (the new global names are all constants).
• Reduced code size
• Less cases where objects are loaded but never used before they are deleted again.


lynxlynxlynx:
Nice!

PR #2145, opened 2025-01-04
Fix possible OOB crash in SetRecalculateWireFrameFlagRadius

momoko-h:
This can be triggered by the Night Ops map for H14, which has buggy room information data.

It's a bit more code than strictly necessary but I really dislike the MAPROWCOLTOPOS macro with its use of  0xFFFF to indicate a problem with its input arguments.


lynxlynxlynx:
Thanks!

Issue #2146, opened 2025-01-05
Two blitters are probably buggy?


momoko-h:
The core of `Blt8BPPDataTo16BPPBufferTransShadowZ`  is this block:
```
					if (px == 254)
					{
						if (*(UINT16*)ZPtr < usZValue)
						{
							*(UINT16*)DestPtr = ShadeTable[2 * (*(UINT16*)DestPtr)];
						}
					}
					else
					{
						if (*(UINT16*)ZPtr <= usZValue)
						{
							*(UINT16*)DestPtr = p16BPPPalette[2 * px];
						}
					}
```

and Blt8BPPDataTo16BPPBufferTransShadow has this:
```
					if (px == 254)
					{
						*(UINT16*)DestPtr = ShadeTable[2 * (*(UINT16*)DestPtr)];
					}
					else
					{
						*(UINT16*)DestPtr = p16BPPPalette[2 * px];
					}
```

What's odd about this is the multiplication by 2 of  `*DestPtr` and px. `*DestPtr` is an UINT16 that can have the full unsigned 16 bit range, so this is from 0-65535. ShadeTable, however, is an array of 65536 UINT16, so this can't be right.
Similar for px: it's an UINT8 that can be anything but 254 here but p16BPPalette points to an array of 256 UINT16.

The problem with testing my theory that this wrong is that I just can't find a way to actually trigger these blitters. I don't know when the conditions for them in RenderTiles are all met.

Another piece of evidence the clipped versions of these blitters (same function names just with `Clip` at the end) should behave exactly the same other than the clipping part. These blitters do not have the '2 *' part.

Since these blitters seem to be very rarely used I would propose to remove their current implementation and just replace them with calls to the clipped versions and passing an unconstrained clipping rectangle.

momoko-h:
So this seems to be a mistranslation of the original x86 code.

Vanilla had this: 
```
		mov		ax, [edi]
		mov		ax, ShadeTable[eax*2]
		mov		[edi], ax
```

(edi is C++'s DestPtr). Here, multiplying eax by 2 is needed to translate the index to an offset (the 2 is basically sizeof(UINT16) in this context). Translated to C this really is just `*DestPtr = ShadeTable[*DestPtr] ` if DestPtr had been defined as an UINT16* as was done in other blitters but not these.

momoko-h:
It seems `Blt8BPPDataTo16BPPBufferTransShadowZNBClip` also has a bug:

[...]

PR #2152, opened 2025-01-14
Fix four blitters

momoko-h:
• Blt8BPPDataTo16BPPBufferTransShadowZNBClip was using the wrong comparison operator (<= instead of <)
• Blt8BPPDataTo16BPPBufferTransShadowZClip was accessing the 16-bit Z Buffer as if it was and array of 8-bit values.
    
Both are not implemented as instantiations of a shared function template to ensure they don't diverge again.

• Blt8BPPDataTo16BPPBufferTransShadow and Blt8BPPDataTo16BPPBufferTransShadowZ multiplied their indices into arrays by 2 as the result of a mistranslation of the original x86 code.

I am not sure how I feel about this PR. Code inspection suggests these changes should be correct, but as I wrote in #2146, I can't to trigger these blitters except Blt8BPPDataTo16BPPBufferTransShadowZNBClip, which is the least problematic of the four. Here, all that can happen that one pixel might have the wrong color occasionally.

So this PR is probably not that important but OTOH it will definitely get in the way of #738 again.

### DIFF
--- a/src/sgp/VObject_Blitters.cc
+++ b/src/sgp/VObject_Blitters.cc
@@ -779,14 +779,14 @@ void Blt8BPPDataTo16BPPBufferTransShadow(UINT16* pBuffer, UINT32 uiDestPitchBYTE
 				do
 				{
 					UINT8 px = *SrcPtr++;
-
+					auto DestPtr16{ reinterpret_cast<UINT16 *>(DestPtr) };
 					if (px == 254)
 					{
-						*(UINT16*)DestPtr = ShadeTable[2 * (*(UINT16*)DestPtr)];
+						*DestPtr16 = ShadeTable[*DestPtr16];
 					}
 					else
 					{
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Hmm, clever. :) This looks fine to me. There's an extra "not" in the first commit message, but that's not important, the code is clear.

momoko-h:
> Hmm, clever. :) This looks fine to me. There's an extra "not" in the first commit message, but that's not important, the code is clear.

You mean the template trick? I learned it from stuff like` std::minus` and friends, the syntax seems a bit clunky but that's normal for C++.

So you don't mind that messing with the blitter code will make all the work maxrd2 put into the 32 bit blitters harder to keep up to date and mergeable?

lynxlynxlynx:
Yeah, it's small enough and all in the same file, unlike the other PR.

PR #2153, opened 2025-01-20
Fix mixed up variables in Blt8BPPDataTo16BPPBufferTransShadowZClip

momoko-h:
This blitter used the variable px in places where it should have used PxCount. Another bug was that the src pointer must always be incremented, no matter if we write to the Z buffer or not.


momoko-h:
Wait a moment, `Blt8BPPDataTo16BPPBufferTransShadowZClip` has the same problem (not that surprising since one of them was clearly copied and pasted.

These are CIDs 210544 and 485743.

lynxlynxlynx:
Thanks!

PR #2156, opened 2025-04-21
github: ubuntu-20.04 was removed, bump used runner

lynxlynxlynx:
Let's see if it all still builds. The image got removed last week, triggering a scheduled coverity runner failure.


selaux:
LGTM. Thanks :)

PR #2157, opened 2025-04-26
Upgrade dependencies for CMake 4

selaux:
I currently cannot build the project with my version of CMake (4.0.0) anymore. It results in the following error:

```
CMake Error at _bin/lib-gtest/src/CMakeLists.txt:4 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

The same happens for `string_theory`.

This PR upgrades `gtest` and `string_theory` to their most recent versions, which fixes the issue.

### DIFF
--- a/.github/workflows/github-ci.yml
+++ b/.github/workflows/github-ci.yml
@@ -18,7 +18,7 @@ jobs:
           target: linux
 
         - name: Linux mingw64
-          os: ubuntu-22.04
+          os: ubuntu-24.04
           target: linux-mingw64
 
         - name: Mac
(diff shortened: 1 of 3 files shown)


selaux:
NB: The runner for the Android build could probably be updated to 24.04 as well. But that resulted in errors and I don't have time to debug right now.

lynxlynxlynx:
That's fine, thanks.

PR #2160, opened 2025-04-27
Bring back support for modded prof.dat binaries

ShishkinP:
Resubmitting...

Restores the ability to use prof.dat by mods (removed in #2070).
JSON profiles and binary are now merged.


lynxlynxlynx:
Looks good. Momoko's last comments from previous review still hold (except the second), but it's nothing critical.

ShishkinP:
Missed the `profiles.size() == 0` thing. As for the heap-allocated `DEFAULT_PROFILE` I'm just not sure as to how to best make it automatic.

lynxlynxlynx:
`empty()` was suggested, since it's faster (doesn't need to iterate the container) and some static analyzers warn about this case. Wouldn't this work?
```c++
	if (profiles.empty()) {
		static const MERCPROFILESTRUCT DEFAULT_PROFILE {};
		return &DEFAULT_PROFILE;
	}
```

ShishkinP:
It does indeed work. Appreciate your help.

lynxlynxlynx:
Thanks!

PR #2161, opened 2025-04-29
Remove mod-breaking changes

ShishkinP:
1. AssociatedTownId property in strategic-mines.json is back but with a deprecation warning
2. With all the changes from the initial concept there is no point in renaming mercs-profile-info.json

Unrelated:
- Drive-by micro-optimization: weaponSaleModifier is already validated by JSON schema
- Git: Ignore the solution state dir for Visual Studio


lynxlynxlynx:
Cool, thanks!

Issue #2099, opened 2024-10-16
Is it possible to make the Github runners do less work?


momoko-h:
Do we have to install Google cloud on the Ubuntu runners? The [documentation](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) lists Google Cloud CLI 495 as preinstalled on the Ubuntu 20.04 and 22.04 runners but we're always downloading and installing the latest version anyway.

Is this really necessary or could we save the ~30 seconds setup time?

Another thing to investigate: do we have to download and compile FLTK every time? This takes more than two minutes.

momoko-h:
Another oddity in the mingw64 log:

> 2024-10-15T19:12:27.1449795Z Building CXX object CMakeFiles/ja2.dir/src/sgp/string_unittest.cc.obj
> 2024-10-15T19:12:28.9200046Z Linking CXX executable ja2.exe
> 2024-10-15T19:15:51.5779576Z Built target ja2

Three minutes just for the linker? Is the mingw linker really this slow or could this be caused by the bizarre numbers of libraries listed at the end of the Rust build step?

lynxlynxlynx:
We currently don't cache anything from dependencies/, but it should be trivial to add to the list in the existing actions/cache action and AppVeyor's cache section.

PR #2164, opened 2025-06-01
Make some Github runners do less work

lynxlynxlynx:
#2099
saves:
- ~15s by not building tar any more on linux
- ~60s by not building fltk on mac
- ~1s by not building gtest on mac

Also fixes too new python on the mac runner and CI_TARGET being ignored.


lynxlynxlynx:
It's odd that I'm getting python failures, since the image was updated two months ago already, but at least now it's clear some things were broken even before.

selaux:
Do we link fltk statically into the launcher? Otherwise the resulting binary will work only on systems that have fltk via homebrew installed.

lynxlynxlynx:
Hmm, good point, we should do that or force the local build for release builds.

lynxlynxlynx:
Cool, that finally worked, at least from the point of view of the runner:

> + ./ja2-launcher -help
> 2025-07-27T07:55:58.603732Z [INFO] stracciatella::logger: Logging to file "/var/folders/y6/nj790rtn62lfktb1sh__79hc0000gn/T/ja2-launcher.log"
> Usage: ja2 [options]
> ...

PR #2166, opened 2025-06-07
Enable diagonal interaction with wall-oriented structures

ShishkinP:
Changes:
- Direct interaction: Both tiles of the garage door are interactable. No need to be positioned directly in front of its base tile.
- Diagonal interaction (enabled by the game.json setting):
Only open/close interactions are allowed diagonally (lockpicking and others are denied).
A char must already be at the adjacent tile to open/close diagonally. If the path to the door is plotted from further tiles, only direct interaction will be picked.
The numbers are APs depending on the "door handle reachability", X - interaction denied:
action-points-for-opening-small
action-points-for-closing-small


src/game/Tactical/Interface.cc
```diff
@@ -1389,14 +1390,20 @@ static void MakeButtonDoor(UINT idx, UINT gfx, INT16 x, INT16 y, INT16 ap, INT16
...
+	ST::string diagWarning = "";
+	if (gOpenDoorMenu.pSoldier->bDesiredDirection & 1 && idx != OPEN_DOOR_ICON && idx != CANCEL_ICON)
+	{
+		diagWarning = *(GCM->getNewString(NS_DIAGONALITY_WARNING));
+		DisableButton(btn);
+	}
 	if (ap == 0 || !(gTacticalStatus.uiFlags & INCOMBAT))
 	{
-		btn->SetFastHelpText(help);
+		btn->SetFastHelpText(help+diagWarning);
```
lynxlynxlynx, line 1401: Is it not missing a space between the two? Same a few lines lower.

lynxlynxlynx:
Other than that it looks good to me, thanks! Perhaps also add another line to game.json to tell the player that enemies don't use this functionality.

src/game/Tactical/Interface.cc
ShishkinP: `diagWarning` always has a new line character: `"\nCan't be done diagonally"`

PR #2081, opened 2024-09-21
Add internalName property to TownModel and other fixes

ShishkinP:
(sorry, I messed up my git repo and had to resubmit)

Various refactors, fixes, deduplications in the externalization code.

PR #2172, opened 2025-07-16
Add backward compatibility for the old strategic-map-towns.json

ShishkinP:
Fixes the problem introduced by #2081.
I think this solves the last known release-blocking issue.
@selaux, thank you for the tip-off.


lynxlynxlynx:
Thanks!

Issue #1623, opened 2022-07-28
ja2-resource-pack doesn't comply with RelWithDebInfo


guijan:
I just tried compiling ja2-stracciatella with -DCMAKE_BUILD_TYPE=RelWithDebInfo and noticed that the other executables are compiled on release mode and with debug symbols as expected, but `ja2-resource-pack` comes out as a release build with no debug symbols.

The issue seems to start here:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/75cce40b9f10c757dd877ecb2e3e34157c59d90b/dependencies/lib-stracciatella/CMakeLists.txt#L67-L69
I can't help because I don't know this Rust stuff.

Noticed while creating a debug package.

lynxlynxlynx:
I suspect it doesn't make sense to package debug symbols for this anyway.

selaux:
The Rust version we use only has support to always include debug symbols in release builds, newer versions (from 1.57 on) can have [custom profiles](https://doc.rust-lang.org/cargo/reference/profiles.html#custom-profiles). We could define a new profile `release-with-debug` that builds a release with debug info. Shouldn't be too hard...

lynxlynxlynx:
Since we're now at rust 1.63, custom profiles are supported out of the box, so this can move forward without special considerations.

Issue #2173, opened 2025-07-23
Depend on a minimum Rust feature level, not a specific version


guijan:
While working on https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1623, I noticed that the CMake scripts are installing Rust 1.63 instead of using the Rust 1.88 I already have installed.

The build scripts should depend on the features provided by the minimum supported Rust version, not on the minimum supported Rust version specifically. As things currently are, compiling ja2 installs an uneccessary Rust toolchain when the user already has one intalled, and it may cause issues on platforms without prebuilt Rust toolchains (all [tier 3](https://doc.rust-lang.org/beta/rustc/platform-support.html#tier-3) platforms) which may nevertheless have it available.

guijan:
Rust doesn't have a stable ABI, but it doesn't matter because Rust statically links everything. Dynamic linking is also supported, but not generally used because Rust's heavy use of generics (generic code always has to be compiled into the program, it can't go in a library) and ABI instability undo any savings. It does support exporting symbols with a C ABI, and the ABI stability of C is platform-specific and breaks/works independently of the Rust version or features used.

Rust uses editions for backwards compatibility: https://doc.rust-lang.org/edition-guide/editions/index.html
The short story is that the only way to guarantee that a Rust program compiles with a certain Rust compiler version is to compile with that version because no breaking changes are made to Rust editions, but additions are made, and these additions won't work on older compilers.

lynxlynxlynx:
We don't install it, except on the continuous integration machines, which is done in `.ci/ci-setup.sh`. You can also just grep for rust-toolchain to see where we use the version.

guijan:
I have this full compilation log: https://paste.opensuse.org/pastes/87a1c889d4a9
With the important part being this:
```console
info: syncing channel updates for '1.63.0-x86_64-unknown-linux-musl'
info: latest update on 2022-08-11, rust version 1.63.0 (4b91a6ea7 2022-08-08)
info: downloading component 'cargo'
info: downloading component 'clippy'
info: downloading component 'rust-docs'
info: downloading component 'rust-std'
info: downloading component 'rustc'
info: downloading component 'rustfmt'
info: installing component 'cargo'
info: installing component 'clippy'
info: installing component 'rust-docs'
info: installing component 'rust-std'
info: installing component 'rustc'
info: installing component 'rustfmt'
[...]

PR #2174, opened 2025-07-23
fix RelWithDebInfo not working with Rust

guijan:
Fixes https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1623

This is only a draft because ja2 doesn't currently compile on my system and I can't test it, something to do with missing libc symbols, and possibly caused by using an old rust toolchain https://github.com/ja2-stracciatella/ja2-stracciatella/issues/2173


lynxlynxlynx:
Cool, thanks. It's interesting to see that the optimized builds are larger than debug ones. I guess some unrolling happens.

PR #2177, opened 2025-07-24
rename rust-toolchain to avoid unnecessary toolchain installs

guijan:
Fixes https://github.com/ja2-stracciatella/ja2-stracciatella/issues/2173


lynxlynxlynx:
LGTM, but this is something @selaux should also comment on.
Is there no way to specify the edition instead? You made quite a case for it already.

guijan:
Specifying the Rust edition ensures forwards compatibility, but not backwards compatibility. It can be done in the Cargo.toml files: https://doc.rust-lang.org/cargo/reference/manifest.html#the-edition-field

Specifying the Rust version is also possible and good for making it explicit, but doesn't ensure anything. 
It's also already done for us by FindRustc and FindCargo:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/b2273d3df48818c113f35cafc44c2f849ffe0e49/cmake/FindRustc.cmake#L32
https://cmake.org/cmake/help/v4.1/module/FindPackageHandleStandardArgs.html (CTRL-F VERSION_VAR)

guijan:
Oh man, sorry, I had to update the PR because one line had the filename spelled `rust-min-version` instead of `min-rust-version`.

lynxlynxlynx:
Works fine for me too.

selaux:
Thanks.

PR #2179, opened 2025-07-24
comply with the manpage directory on Unix systems

guijan:
ja2 was installing the man page to the wrong directory on FreeBSD, NetBSD, and possibly others.

Only tested on Gentoo.

Ref: https://github.com/freebsd/freebsd-ports/blob/a3ecc2dea2351000594d65bdb53ea1a286164a7f/games/jaggedalliance2/files/patch-CMakeLists.txt
Ref: https://www.netbsd.org/docs/pkgsrc/files.html


lynxlynxlynx:
Thanks!

PR #2185, opened 2025-07-26
Fix crashing on destruction of a corner wall

ShishkinP:
Steps to reproduce:
Detonate a bomb on this tile right at the beginning of the game in Omerta:
Image
Crash will happen with the message:
`[ERROR] sgp\SGP.cc: Game has been terminated due to an unrecoverable error: Failed to add node to world`
It will occur with almost any corner consisting of walls placed in this way:
Image

The reason is slightly wider walls used to place adjacent walls like this:
Untitled6
They take up two tiles thus being able to "overlap" with other walls. Since they are purely for visual effect we can easily make a special exception for them.

The bug was introduced by [this](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/119a1d172852ca503e1dda10435e4430af73a134) ancient commit.


lynxlynxlynx:
Nice find!

PR #2189, opened 2025-07-27
Fix exception handling in main menu

ShishkinP:
Steps to reproduce:
1. In `ja2.json` substitute `game_dir` value with UB dir:
```javascript
"game_dir": "C:\\Program Files (x86)\\Steam\\steamapps\\common\\Jagged Alliance 2 Gold Unfinished Business",
``` 
2. The game will crash without giving a useful error text because it tries to create an error save file and fails since there cannot be game world loaded while in main menu.

This fix makes it bypass the error save file creation (if in main menu) and correctly output:
`[ERROR] sgp\SGP.cc: Game has been terminated due to an unrecoverable error: openGameResForReading: Vfs_open "loadscreens/ja2logo.sti": entity not found`
before the crash.

### DIFF
--- a/src/game/GameLoop.cc
+++ b/src/game/GameLoop.cc
@@ -224,27 +224,30 @@ try
 catch (std::exception const& e)
 {
 	guiPreviousOptionScreen = guiCurrentScreen;
-	char const* what;
+	char const* what = "file";
 	ST::string success = "failed";
 	char const* attach = "";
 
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Thanks, looks fine. In the future we could further improve the startup by detecting ja2ub like we do 1.13.

PR #2209, opened 2025-09-05
InternalAutoPlaceObject: fixed comparison against the wrong type

### LINKED ISSUE #2207: Better IMP item placement
`GiveItemsToPC` uses `MakeProfileInvItemAnySlot` for the items that you get. Similarly for mercs (though that may be in the data itself), so you can end up with helmets and armor in your hands.

Image

lynxlynxlynx:
fixes #2207

regression from 1ae407c066


selaux:
Thanks. I will double check the armour PR for similar issues...

lynxlynxlynx:
Didn't notice anything else for armor class.

PR #2223, opened 2025-09-15
Fix some cmake_minimum_required calls for CMake 4

selaux:
Seems like the nightly build is still broken for MSVC and macOS. I guess this will also be true for release builds. E.g.:

- https://github.com/ja2-stracciatella/ja2-stracciatella/actions/runs/17720881873/job/50352833800#step:10:330
- https://ci.appveyor.com/project/ja2-stracciatella/ja2-stracciatella/builds/52742733/job/cpua9b4387dms5ga#L1126

Hopefully this is the last fix for CMake 4.

### DIFF
--- a/cmake/JSONParser.cmake
+++ b/cmake/JSONParser.cmake
@@ -1,7 +1,7 @@
 # Copied from https://github.com/sbellus/json-cmake
 # Licence: MIT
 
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Thanks!

PR #2225, opened 2025-09-18
Replace merging with overwriting for inventory

ShishkinP:


lynxlynxlynx:
I guess it could be if one of the slots wasn't provided, since then they could overwrite each other, but I wouldn't label it as mod breaking.

PR #2226, opened 2025-09-19
Fix schema validation for mercs-profile-info.json

ShishkinP:
This concludes all I wanted to update before the release.


lynxlynxlynx:
Thanks!

PR #2229, opened 2025-09-21
Call RenderClear on every frame again

momoko-h:
Required by some video drivers to avoid ugly artifacts in the border area.


momoko-h:
> Doesn't it obviate the whole [6908edd](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/6908edd50918925037cfa7477cb9fd75d19e359a)?

Pretty much, yes. And while handling SDL window events could maybe be useful for something in the future at the moment `HandleWindowEvent()` could be removed as well if you'd prefer.

lynxlynxlynx:
No preference, was just curious about the justification.

PR #2232, opened 2025-09-25
MinGW/MSYS2: deprecate mingwstd_threads package. Fix ja2-launcher.

askondro:
MinGW/MSYS2: deprecate mingw-std-threads is no longer needed and prints annoying compiler messages.
ja2-launcher could not resolve the HANDLE symbol and that need the windows.h inclusion.


lynxlynxlynx:
Thanks!

PR #2233, opened 2025-09-27
Remove the lib-mingw-std-threads dependency directory.

askondro:
Continue to deprecate mingw-std-threads.


lynxlynxlynx:
Ah, good catch, thanks.

PR #2234, opened 2025-09-28
Some schema fixes

selaux:
These are some fixes for the schema, that come out of updating the stracciatella toolset. Those are quite minor fixes, so I think it is still good to go into this release.


lynxlynxlynx:
Thanks!

PR #2237, opened 2025-09-29
Upgrade minimum rust version to 1.85.1

selaux:
Upgrade rust to 1.85.1

Note that this is a patch version above what is available on Debian stable, this is because of a bugfix (https://github.com/rust-lang/rust/pull/137528) that fixes rename for the latest Windows 10 version. Newer versions are available in backports though (1.87.0).

Upgrade rust edition to 2024


lynxlynxlynx:
LGTM

PR #2240, opened 2025-10-05
Establish a macOS arm build

selaux:
Lets try if we can get around Rosetta on macOS and have a native ARM build. For now it is a separate build. In the future we could strive to build a universal package, but that brings its own challenges.


rklasen:
soooo... whatever happened to this one? Right now, the game doesn't build on a ARM64 Mac after a fresh clone cause fltk 1.3 is required, but 1.4 usually comes from homebrew. Also, I kinda would like to avoid changing the users system (and stuff from homebrew) when the code from the repo should adapt to the system. not the other way around. also, there is a lua package shadowing issue.

Other that that, macos with arm64 builds just fine (after these fixes), the _Intel_ one doesn't (didn't) however. I realize Intel Macs aren't officially supported, but I really dont see a reason why they couldn't be. The failure on the github runner was just a path issue. And for context, the last 27'' iMac ever made is an Intel one, and it's just six years old. I've used computers for much longer than that, and I do know poeople who use this iMac as their daily driver.

I've attempted to re-create this PR on my own branch with some fixes here: https://github.com/rklasen/ja2-stracciatella/pull/13, but I'm hesitant to open a new PR that just copies most form here plus some extra. @selaux , I see you're one of the main contributors, and apparently use use an arm mac as well. what can I do to help you?

_Edit_: If we build both x86 and arm64 macos binaries, we would probably also fix #2436. We should name the macos images with the appropriate suffixes (`_arm64` and `_x86_64`) then, _if_ this project were to support intel macs again. But I now see that the Intel github runner is kinda wonky, it only succeeds every third try or so...

lynxlynxlynx:
We bundle all the dependencies, so you can use those if you don't want to install what we need system-wide. They're all simple cmake flags / ccmake toggles.

By now it would be pointless to continue building for intel macs, since we see no demand. If it arises, doing it one time at release would be enough, the same way misanthropos provides builds for armhf.

selaux:
Lets see how this PR fares now. The issue was that I wanted to build for Intel and ARM macOS. Somehow the build broke for the Intel architecture when updating the runner image. Not sure why anymore. But there were a lot of changes since then anyways, so maybe it just works ^^

NB: I dont really play on macOS, I just have an ARM one for work

selaux:
Regarding #2436: Its probably a different issue. We already build for Intel macOS and from what I understand that person was using Intel macOS, just an old version.

selaux:
Had to adapt the static linking logic a little bit to cater for two different install prefixes, now it seems to work.

rklasen:
Good, then this MR can be merged? @lynxlynxlynx , I think you and I mean separate things. There are still two issues for local apple/arm64 builds, but the CI is immune to both of them by pure chance. The fixes are small, I'll gladly open new PRs for them, or we can discuss them here, but I cant commit or push to this branch.

The first one is with fltk, when both 1.4 and 1.3 are installed locally, the build will fail. Since `brew install fltk`installs 1.4, not 1.3, the headers are wrong for this repos. But since it is the naive (and most of the time correct) way to install those deps via homebrew, I think this repo should include a safeguard for that. The fix is simple.

The other one pertains to a lua shadowing issue. I can open a single PR for both of them, then the build will also succeed on a fresh clone on M series apple laptops.

selaux:
I think so. I'll just merge.

Regarding fltk: I think it would be fine if the project is upgraded to fltk 1.4. As far as I can see it can be installed on every system we use / support anyways.

PR #2241, opened 2025-10-06
Use "git apply" instead of "patch" command to apply custom sol2 patch

rebb:
The sol2 patch was not applied on MSVC using the "patch" command.

Using "git apply" instead makes it work as intended.

### DIFF
--- a/dependencies/lib-sol2/getter/CMakeLists.txt.in
+++ b/dependencies/lib-sol2/getter/CMakeLists.txt.in
@@ -22,5 +22,5 @@ externalproject_add(get-sol2
         BUILD_COMMAND       ""
         INSTALL_COMMAND     ""
         TEST_COMMAND        ""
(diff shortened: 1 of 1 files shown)


rebb:
Since it fails for other platforms, closing for now to make it only apply for MSVC.

rebb:
Since it fails for other platforms, closing for now to make it work for MSVC only.

rebb:
> Patch.exe comes with git but you still have to manually add `C:\Program Files\Git\usr\bin\` to the Path system environment variable for it to be found.

Oh, i wasn't aware of that. In that case feel free to disregard the PR.

I just noticed the compile failing when it worked fine a while back, and using "git apply" ended up helping 😅

ShishkinP:
>feel free to disregard the PR

I would merge it. It can save people some time figuring that out.

PR #2247, opened 2025-10-16
CLI start mode options should always override ja2.json

ja2-derek:
Minor problem - if JA2-Launcher / `ja2.json` is set to start in fullscreen, `ja2 -window` is not respected.

This is because we have 2 conflicting options - run in fullscreen and run in windows. CLI should be overriding in any cases.


lynxlynxlynx:
LGTM, thanks. It's not obvious why we need to track the two separately though.

PR #1577, opened 2022-06-20
Keep hand item and weapon mode in sync

momoko-h:
Ensure that SOLDIERTYPE::bWeaponMode and bDoBurst are sensible after
changing the soldier's hand item. Resolves #196.

This also partly fixes #1184:

- You can't crash the game any more by switching from a weapon with burst mode to another item and then throwing something.
- The problem that a throw attempt can turn into a burst if burst mode is enabled still exists.

PR #2252, opened 2025-10-24
Fix attached launcher mode on non-burstable weapons

ja2-derek:
Fixes an [issue](https://github.com/ja2-stracciatella/ja2-stracciatella/pull/1577#discussion_r2444782388), that in the master and nightlies, it is impossible to switch to launcher mode on a non-burstable weapon.

In `ChangeWeaponMode`, you start with `WM_NORMAL` mode, then add 1 you get `WM_BURST`, but gets reverted because the weapon is not burstable.

This pull request partially reverts bc588c04115eb76d4bb8ad928179014977fd6311. I like the clever one-liner we had, but it does not address non-burstable weapons. The original case-switch, even though verbose, is easier to follow through different cases. Feel free to propose something better than this.


momoko-h (approved):
This is fine, thanks. Better verbose and correct than terse and wrong.

PR #2253, opened 2025-10-24
Add the Informative Tooltips feature

ShishkinP:
- Fixed: Trap level increases the chance of a successful untrap check instead of reducing it as would be proper.
- ~Patched: String Theory puts the plus sign in front of zero if positive integers are signed. This will get rid of it.~
- Added: New control characters for green and red fonts.
- Added: Informative hover box text listing all the modifiers for the door menu skill checks.
Untitled
Untitled2


lynxlynxlynx:
This will  be great for debugging, nice work.
Not very keen on the ST patch from a maintenance POW, but hopefully that won't be an issue.

And TIL dexterity doesn't matter for lockpicking, it's mechanics.

PR #2254, opened 2025-10-25
Enum for soldiers' side alignment

ja2-derek:
There are a lot of usage of SOLDIERTYPE.bSide that looks like `s.bSide == OUR_TEAM`, which is semantically incorrect.

This pull request includes some refactoring:
- Add new Enum `Side` for the values of soldiers' side alignment (for `bSide`, as opposed to `bTeam`)
- Replaced occurences of `bSide == OUR_TEAM` or `bSide == 0` with enum
- Refactoring on MakeCivHostile to remove unused arg
- Regenreated Enums for Lua scripting

### DIFF
--- a/src/game/Tactical/LoadSaveSoldierType_unittest.cc
+++ b/src/game/Tactical/LoadSaveSoldierType_unittest.cc
@@ -12,6 +12,10 @@
 #include <cstring>
 #include <memory>
 
+#ifndef NO_MAGICENUM_LIB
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Yeah, making it an enum class and using it as the type would prevent similar mistakes in the future as well-

lynxlynxlynx:
LGTM, thanks!

PR #2259, opened 2025-10-28
Fix obscured item highlights

rebb:
# Fixes

- Obscured items render properly now when "make items glow" is enabled
- Merc light no longer "wipes away" obscured items, an issue that also exists in vanilla JA2 ( see image )

ja2_obscured_item_glitchx

### DIFF
--- a/src/game/TileEngine/RenderWorld.cc
+++ b/src/game/TileEngine/RenderWorld.cc
@@ -1906,6 +1906,7 @@ static void RenderDynamicWorld(void)
 
 	SaveBackgroundRects();
 
+	// Fixes obscured item highlight issues. Omitting RENDER_STATIC_ONROOF for now as it seems redundant ?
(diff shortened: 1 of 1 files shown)


rebb:
> Nice find!

Thanks :) I had most of my recent stuff sitting in my fork for a while, just getting around to making the PRs now 😅

rebb:
> It looks like in `RenderStaticWorldRect` and `RenderStaticWorld` a very similar call was added but in those functions it also includes the RENDER_STATIC_ONROOF layer. Is it intentional that you didn't include that one here, too?

I tried to make the problem go away with the least change possible, and items on the roof don't show the issue.
In that sense it's intentional.

Maybe adding roof items brings it more in line with the other functions, but i'm not sure it does anything.

lynxlynxlynx:
Thanks!

PR #2260, opened 2025-10-30
Fix mainmenu button artifacts

rebb:
Main-menu buttons no longer show visual artifacts when clicked.

This was a Stracciatella issue, it does not exist in Vanilla.


lynxlynxlynx:
It was visible on preferences, thanks!

PR #2267, opened 2025-11-23
Add tooltips for effective skill values

ShishkinP:
image
image
image

### DIFF
--- a/src/sgp/MouseSystem.cc
+++ b/src/sgp/MouseSystem.cc
@@ -812,7 +812,15 @@ static void DisplayFastHelp(MOUSE_REGION* const r)
 	if (x <  0)                x = 0;
 	if (x >= SCREEN_WIDTH - w) x = SCREEN_WIDTH - w - 4;
 
-	INT32 y = r->RegionTopLeftY - h * 3 / 4;
+	int32_t y = r->RegionTopLeftY;
+	if ((r->RegionBottomRightY - r->RegionTopLeftY) < h)
+	{ // case for regions too small for their tall and obstructive tooltips: savegames, effective attributes and others
+		y += -h - 1;
+	}
+	else
+	{
+		y += -h * 3 / 4;
+	}
 	if (y <  0)                 y = 0;
 	if (y >= SCREEN_HEIGHT - h) y = SCREEN_HEIGHT - h - 15;
 


src/externalized/content/NewStrings.cc
```diff
@@ -611,3 +615,107 @@ const ST::string GetModifiersForLockBlowUp(SOLDIERTYPE* const s)
 
 	return result;
 }
+
+const ST::string GetModifiersForEffectiveAttributes(SOLDIERTYPE* const s, Attributes const attrName)
+{
+	ST::string result{}, comment{}, possession, location;
```
momoko-h: You don't need the {} after the definition of a ST::string, like any proper C++ object it has a default constructor that properly initializes itself.

src/game/Tactical/Interface_Panels.cc
```diff
@@ -947,6 +950,31 @@ void InitializeSMPanel()
 	InitInvSlotInterface(g_ui.m_invSlotPositionTac, &g_ui.m_invCamoRegion, SMInvMoveCallback, smInvClickCallback, SMInvMoveCamoCallback, SMInvClickCamoCallback);
 	InitKeyRingInterface(KeyRingItemPanelButtonCallback);
 
+	if (gamepolicy(informative_popups)) {
+		//Define regions for the informative popups displaying effective attribute values
+		MSYS_DefineRegion(&gSM_SELMERCStatsRegion[ATTR_AGILITY], dx + SM_AGI_X - 10, dy + SM_AGI_Y, dx + SM_AGI_X + SM_STATS_WIDTH, dy + SM_AGI_Y + 10, MSYS_PRIORITY_HIGH, MSYS_NO_CURSOR, SelectedMercPopupMoveCallback, MSYS_NO_CALLBACK);
+		MSYS_SetRegionUserData(&gSM_SELMERCStatsRegion[ATTR_AGILITY], 0, ATTR_AGILITY);
+		MSYS_DefineRegion(&gSM_SELMERCStatsRegion[ATTR_DEXTERITY], dx + SM_DEX_X - 10, dy + SM_DEX_Y, dx + SM_DEX_X + SM_STATS_WIDTH, dy + SM_DEX_Y + 10, MSYS_PRIORITY_HIGH, MSYS_NO_CURSOR, SelectedMercPopupMoveCallback, MSYS_NO_CALLBACK);
+		MSYS_SetRegionUserData(&gSM_SELMERCStatsRegion[ATTR_DEXTERITY], 0, ATTR_DEXTERITY);
```
momoko-h: If the userdata of region[x] is always x itself you don't really need it because you can just compute it inside the callback handler: `auto attr = static_cast<Attributes>(pRegion - gSM_SELMERCStatsRegion)`;

ShishkinP:
Additionally:
- Popups got renamed to tooltips.
- Big tooltips are now displayed without blocking the view of their region
image

lynxlynxlynx:
Nice, thanks!

PR #2275, opened 2025-12-06
Add informative tooltips for the dialogue menu

ShishkinP:
Relentless tooltipization continues.
image

image

With cheats enabled:
image


ShishkinP:
>Is the friendly approach mod deliberately not colored?

Yes. Plain friendly and direct approaches mean they haven't yet been tried, so their modifiers are not applied. Too unintuitive?

>unrelated stuff

Fast Draw info is in the threatening approach tooltip because it can be modified by committing this approach. Maybe it's better to move it to the NPC portrait tooltip. I don't know yet.

lynxlynxlynx:
Alright, thanks!

PR #2278, opened 2025-12-07
Some MouseSystem cleanup

momoko-h:
This function can easily be replaced with std::count.

Nothing really important, just because I stumbled over this while looking at 2277.


lynxlynxlynx:
LGTM, thanks!

PR #2282, opened 2025-12-09
Fix merc's tactical state disabling their answering machine message on the AIM site

ShishkinP:
While an AIM merc is bleeding with less than 15 HP, gassed, unconscious or held as a prisoner their pre-recoded message is not replayed during the video conferencing session.


lynxlynxlynx:
Nice find, thanks!

PR #2283, opened 2025-12-11
Translatable strings for items

selaux:
Currently in the JSON data, it is possible to set `shortName`, `name` and `description` properties for items. This is fine for non-translated mods, but we probably want to give the option to translate strings via JSON. This adds this possibility for items via `TranslatableString`. In JSON the translated string can be one of the following:

- A simple string, which has the same behavior as the current implementation (no translation)
- An object `{ "type": "json", "prefix": "strings/my-items", index: 0 }` to refer to JSON translation files
- An object `{ "type": "binary", "file": "mymercs.edt", "offset": 100, "length": 80 }` to refer to a binary file

All of the implementation is in the `TranslatableString` namespace / file. Currently it is only used for items. I will add this to other locations currently using a translatable string in subsequent PRs so they do not get too big.

### DIFF
--- a/src/externalized/DefaultContentManager.cc
+++ b/src/externalized/DefaultContentManager.cc
@@ -917,7 +917,7 @@ void DefaultContentManager::loadStringRes(const ST::string& name, std::vector<ST
 /** Load the game data and the item descriptions from the original game resources. */
 bool DefaultContentManager::loadGameData()
 {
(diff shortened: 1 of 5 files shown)


selaux:
I don't think there is a mod yet, which uses this feature in any way. We would need a mod that edits items.

lynxlynxlynx:
Ah, you're right, they don't actually change items.

PR #2285, opened 2025-12-13
Fixed a bug that clears map screen messages in the new game at laptop exit

mmamontx:
Before the fix, the bug can be reproduced by doing the following steps:

1. Start the new game.
2. Leave the laptop to the map screen - observe the map screen messages in the bottom (there should be history log notifications at least).
3. Go back to the laptop and hire a merc.
4. Close the laptop and notice that all the messages are now missing - including the one about hiring the new merc.

The messages are cleared by FreeGlobalMessageList(), which is called by InitNewGame(), which in turn is triggered inside LeaveLapTopScreen() by gfAtLeastOneMercWasHired (set on hiring).

This fix is pretty straightforward - it utilizes a variable set in LeaveLapTopScreen().

### DIFF
--- a/src/game/Laptop/Laptop.cc
+++ b/src/game/Laptop/Laptop.cc
@@ -1353,7 +1353,6 @@ static void LeaveLapTopScreen(void)
 			{
 				LaptopSaveInfo.gfNewGameLaptop = FALSE;
 				fExitingLaptopFlag = TRUE;
-				InitNewGame();
 				gfDontStartTransitionFromLaptop = TRUE;
 				return;
 			}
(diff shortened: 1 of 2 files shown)


momoko-h:
Yes, I traced the call flow from the moment you click on start new game to the moment you first get to the map screen. It's a bit bizarre and could certainly be simplified.

Have you tested that simply removing the second InitNewGame call also works when you quit out of a running campaing and then start a new game, not just immediately after starting the program? If that's the case let's go with this solution.

mmamontx:
Removed the call and tested the following scenarios:

1. Start the new game, hire mercs, heli drop -> save (see below), exit to the main menu, start the new game, test using the original steps.
2. Load the save above, exit to the main menu, start the game, test.
3. Scenario 2, but using another save, which is created in the very begging - before hiring mercs (gfAtLeastOneMercWasHired is not yet set).

In all these scenarios the original bug wasn't reproduced, nor did I notice any other issues. Updated the changes.

PR #2287, opened 2025-12-14
Change the return type of LoadGenericButtonIcon

momoko-h:
This function now returns a pointer to a video object directly instead of an index into an array of pointers. This simplifies the code and removes the restriction of max. 40 icons.


lynxlynxlynx:
Yep, thanks!

PR #2289, opened 2025-12-14
Fix raised gun animation disabling sidestepping and backward walking

ShishkinP:
It's currently impossible to make a char walk backwards or sidestep unless their gun is lowered.


ShishkinP:
Never even crossed my mind ☹

lynxlynxlynx:
Don't beat yourself over it. :)

PR #2292, opened 2025-12-16
TranslatableString for EDT overrides

selaux:
This adds the TranslatableString to the EDT overrides function. It also adds most of the existing EDTs to the EDTFile class. For each of the files there is an example override in the `test-json-dialogs` mod. I left out the ones that are better represented as their own model:

- Emails
- Quests
- Files

Other remaining files inside of `BINARYDATADIR` are not string values, but something different, e.g. tilesets.


selaux:
Allright. I've removed the check.

lynxlynxlynx:
LGTM to me too, thanks!

PR #2294, opened 2025-12-17
Introduce containers for externalized models

selaux:
This is an attempt to reduce the responsibilities that `ContentManager` currently has with respect to indexing externalized entities. Right now `DefaultContentManager` takes care of allocating, deleting and indexing every entity. But it is not very consistent in doing that, e.g. `getItem` throws when the item does not exist while `getWeapon` returns a `nullptr` if the weapon does not exist, but the item does. This PR introduces the `Containers` namespace, which currently has:

- `Containers::Entity`: A base class for an entity that can be indexed by an integral value
- `Containers::Indexed`: A container for `Containers::Entity` that takes care of indexing the entities
- `Containers::NamedEntity`: A base class for an entity that can be indexed by integral value or string value (index and name)
- `Containers::Named`: A container for `Containers::NamedEntity` that takes care of indexing the entities

The containers work based on `std::unique_ptr`, so they return pointers, as did the `ContentManager` before. This is necessary because at some point we will need to index items, which can be one of multiple classes. The containers have optional and non optional accessors, which throw if the entity does not exist. They also automatically check for duplicate ids and names upon adding an item to the container and decouples the item id from the index in the array, so we dont need to store empty entities. The interface of using it looks like this:

```
GCM->ammoTypes()->byId(MY_AMMO_TYPE_ID);
GCM->ammoTypes()->optionalById(MY_AMMO_TYPE_ID);
GCM->ammoTypes()->byName("MY_AMMO_TYPE_NAME");
GCM->ammoTypes()->optionalByName("MY_AMMO_TYPE_NAME");
```

We could also change this do do it in a similar way as it is done currently for items with `nothrow_t`. What do you think? Does it make sense to do this? Can you think of better naming?

### DIFF
--- a/src/externalized/WeaponModels.cc
+++ b/src/externalized/WeaponModels.cc
@@ -665,12 +665,12 @@ bool WeaponModel::shootsExplosiveCalibre() const {
 
 bool WeaponModel::matches(const CalibreModel *calibre) const
 {
-	return this->calibre->index == calibre->index;
+	return this->calibre && this->calibre->index == calibre->index;
 }
 
 bool WeaponModel::matches(const MagazineModel *mag) const
 {
-	return (this->calibre->index == mag->calibre->index)
+	return this->calibre && (this->calibre->index == mag->calibre->index)
 		&& (ubMagSize == mag->capacity);
 }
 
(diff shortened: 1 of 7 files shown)


selaux:
I implemented your suggestion and also added some more slight changes:

- Using `size_t` as the correct index type
- Added tests
- Added iterator and `size`, `begin` and `end` functions to the container directly instead of the `all()` method. This way you get the item pointer directly and not a reference to the unique_ptr. It looks like this: `for (auto item : *GCM->items())`
- More throroughly go through the code and add some checks for weapon->calibre now potentially being null

lynxlynxlynx:
I checked the diffs as they flew in — still looks good.

PR #2300, opened 2025-12-27
Externalize the rest of the starting IMP gear granting

lynxlynxlynx:
I was considering just using lua, but that would be harder to change for users.
All the checks but one map nicely to the >= relation and with some clever reordering this is enough to handle either-or and interval cases. That one exception is the wisdom check for helmets, which is now 1% off, which I opted to ignore.
There's a bunch of preparatory commits and a few cleanup ones at the end.

### DIFF
--- a/src/game/Laptop/IMP_Confirm.cc
+++ b/src/game/Laptop/IMP_Confirm.cc
@@ -283,12 +283,12 @@ static void GiveItemsToPC(UINT8 ubProfileId)
 		if (!set.Evaluate(p)) continue;
 		for (const ItemModel* item : set.items) {
 			UINT32 uiPos;
-			if (set.slot == InvSlotPos::NUM_INV_SLOTS) {
+			if (set.slot) {
+				uiPos = *set.slot;
+			} else {
 				INT32 const iSlot = FirstFreeBigEnoughPocket(p, item->getItemIndex());
 				if (iSlot == -1) continue;
 				uiPos = iSlot;
-			} else {
-				uiPos = set.slot;
 			}
 			p.inv[uiPos] = item->getItemIndex();
 			p.bInvStatus[uiPos] = 100;
--- a/src/externalized/policy/IMPPolicy.h
+++ b/src/externalized/policy/IMPPolicy.h
@@ -70,7 +70,7 @@ public:
 struct IMPStartingItemSet
 {
 	std::vector<IMPItemCondition> conditions;
-	uint8_t slot = InvSlotPos::NUM_INV_SLOTS;
+	std::optional<uint8_t> slot;
 	std::vector<const ItemModel *> items;
 
 	bool Evaluate(const MERCPROFILESTRUCT& p) const


rust/stracciatella/src/schemas/yaml/imp.schema.yaml
```diff
@@ -16,23 +16,32 @@ properties:
...
-      $ref: types/id.schema.yaml
+      type: object
+      title: Item condition set
+      description: Data for when, where and what item to create.
+      properties:
+        items:
+          title: List of the items to grant
+          type: array
+          items:
+            type: string
```
selaux: This should probably be `$ref: types/id.schema.yaml`

rust/stracciatella/src/schemas/yaml/imp.schema.yaml
```diff
@@ -16,23 +16,32 @@ properties:
...
+          title: Granting conditions
+          description: Optional pairs of greater-or-equal conditions to evaluate whether the I.M.P. should get the items.
+          type: array
+          items:
+            anyOf:
+              - $ref: types/uint8.schema.yaml
+              - type: string
+    required:
+    - items
+    - conditions
```
selaux: Could we make conditions optional here, so we dont have to use an empty array for non-conditional items?

src/game/Laptop/IMP_Confirm.cc
```diff
@@ -420,9 +269,32 @@ static INT32 FirstFreeBigEnoughPocket(MERCPROFILESTRUCT const& p, UINT16 const u
...
+	// NOTE: Any guns should probably be from those available in regular gun set
 
+	MERCPROFILESTRUCT& p = GetProfile(ubProfileId);
 
-	return(-1);
+	for (const IMPStartingItemSet& set : GCM->getIMPPolicy()->getInventory()) {
+		if (!set.Evaluate(p)) continue;
+		for (const ItemModel* item : set.items) {
+			UINT32 uiPos;
+			if (set.slot == InvSlotPos::NUM_INV_SLOTS) {
```
selaux: Should we make `set.slot` a `std::optional`, so we dont have to use `NUM_INV_SLOTS` as the none value?

rust/stracciatella/src/schemas/yaml/imp.schema.yaml
lynxlynxlynx: That's what I started with, but it didn't work. It's the reason I added extra debugging, but then didn't have the energy to track down the root cause.

src/externalized/internals/enums.cc
```diff
@@ -771,3 +771,19 @@ Quests Internals::getQuestEnumFromString(const ST::string& s)
 	else if (s == "KILL_DEIDRANNA") return Quests::QUEST_KILL_DEIDRANNA;
 	else throw DataError(ST::format("unknown quest name value: {}", s));
 }
+
+Attributes Internals::getAttributeEnumFromString(const ST::string& s)
+{
+	if (s == "") return Attributes::NUM_ATTRIBUTES;
```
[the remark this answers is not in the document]
lynxlynxlynx: It's a natural error value and the only caller checks for it. I could instead add an explicit error value to the enum after NUM_ATTRIBUTES, so it doesn't break other things.

PR #2310, opened 2026-01-10
Fix explosion propagation through walls

ShishkinP:
Missile and bomb explosions don't do damage through walls in certain directions.
In this example the TNT detonation will only affect Igor: 
image
In this - nobody will be damaged:
image
In both situations everybody will be wounded if the bomb and mercs swap sides.

### DIFF
--- a/src/game/TileEngine/Explosion_Control.cc
+++ b/src/game/TileEngine/Explosion_Control.cc
@@ -1323,15 +1323,7 @@ void SpreadEffect(const INT16 sGridNo, const UINT8 ubRadius, const UINT16 usItem
 
 		uiTempRange = sRange;
 
-		INT32 cnt;
-		if (ubDir & 1)
-		{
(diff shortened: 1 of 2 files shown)


ShishkinP:
> can S and SE really be skipped?

Yes, should've excluded EAST too. All the possible directions when the base tile of a wall is on the side opposite to the source:
image

New possible issue found: with this fix the explosion hole on the building is now consistently 5 tiles wide. In vanilla it's 3 or 4 (depending on the wall orientation). Probably has to do with the fact that walls don't absorb damage unlike whole tile structures.

ShishkinP:
Turned out more difficult than I had anticipated. Not only is the tile positioning asymmetric, so is the blast wave spread (in respect to the world direction axes). Means there is pretty much no way to fix it for good, only rewrite the whole algorithm. Which is the plan.

PR #2318, opened 2026-01-19
Fix blast wave spread being blocked by walls regardless of their orientation

ShishkinP:
Solves only one particular issue with the spread:
image


lynxlynxlynx:
Ah, that makes more sense, thanks.

PR #2319, opened 2026-01-20
Increase required C++ version to 20

momoko-h:
For now, this is just to see if our code base and the build bots have a problem with C++20, but in general I'd like to do this now.

Any comments or objections? Are there reasons to wait some more?


momoko-h:
So nobody knows any reason not to do this? If not, I would like to create a new commit with just the minimal required changes (CMakeLists and Github configuration) and merge it on Monday.

lynxlynxlynx:
They look fine to me.

PR #2330, opened 2026-01-29
Add a BASIC_SOLDIERCREATE_STRUCT to SOLDERINITNODE

momoko-h:
This commit removes the extra memory allocations for the BASIC_SOLDIERCREATE_STRUCT by adding one directly to the SOLDERINITNODE. All users of this struct expect its pBasicPlacement to be always valid, it therefore makes no sense to split this in two separate allocations.


lynxlynxlynx:
Thanks!

PR #2336, opened 2026-02-06
Fix the return type of GraphicModel::getSubImageIndex

momoko-h:
STCI files use a 16-bit unsigned integer for the subregion index, but this function used an 8-bit value, which made it impossible to have more than 256 item graphics in one file without a good reason.

Because the combination of a SGPVObject * and an UINT16 in a pair could be useful elsewhere I added two new using directives for them.

Lastly, DeleteVideoObject does not need to be inline, that just increases code size without any benefit. It also needs to be able to delete const objects, just like the delete operator.


lynxlynxlynx:
LGTM, thanks!

PR #2377, opened 2026-04-17
Add ability to heal stats through doctoring

czaccardelli:
Adds the ability to heal stats that were damaged by critical hits through doctoring. When `enable_stat_healing` in `game.json` is set to `TRUE` (default `FALSE`), then characters with damaged stats will be able to be healed during doctor assignments. Healing points are first allocated to missing life, and after a character with stat damage has full life they can remain as patients (or doctors when doctoring themselves) to heal their stats. This drains medical supplies at a rate 3x higher than normal as it's currently implemented.

Why add this? Currently when playing with the non-vanilla additional AI options to have the AI go for head and leg shots it results in more critical hits and -- especially in the case with head shots -- more stat damage. Many mercs end up with their stats severely damaged and in vanilla there is no way to heal those stats, and since some stats like WIS can barely be trained at all many mercs end up losing tons of WIS over the course of the game.

### DIFF
--- a/src/game/Strategic/Assignments.cc
+++ b/src/game/Strategic/Assignments.cc
@@ -565,19 +565,24 @@ static BOOLEAN CanCharacterRepair(SOLDIERTYPE const* const pSoldier)
 	return ( TRUE );
 }
 
+static BOOLEAN CanStatsBeHealed(const SOLDIERTYPE* const s)
+{
+	return gamepolicy(enable_stat_healing) &&
+		(
+			s->bAgilityDamage > 0 ||
+			s->bDexterityDamage > 0 ||
+			s->bStrengthDamage > 0 ||
+			s->bWisdomDamage > 0
+		);
+}
+
 
 // can character be set to patient?
 static BOOLEAN CanCharacterPatient(const SOLDIERTYPE* const s)
 {
 	return
 		s->bLife > 0 &&
-		(
-		s->bLife != s->bLifeMax ||
-		s->bAgilityDamage > 0 ||
-		s->bDexterityDamage > 0 ||
-		s->bStrengthDamage > 0 ||
-		s->bWisdomDamage > 0
-		) &&
+		(s->bLife != s->bLifeMax || CanStatsBeHealed(s)) &&
 		AreAssignmentConditionsMet(*s, AC_UNCONSCIOUS | AC_EPC | AC_UNDERGROUND);
 }
 
@@ -1211,7 +1216,7 @@ static void UpdatePatientsWhoAreDoneHealing()
 	{
 		SOLDIERTYPE& s = *i;
 		if (s.bAssignment != PATIENT) continue;
-		if (s.bLife == s.bLifeMax && s.bAgilityDamage == 0 && s.bDexterityDamage == 0 && s.bStrengthDamage == 0 && s.bWisdomDamage == 0)
+		if (s.bLife == s.bLifeMax && !CanStatsBeHealed(i))
 		{
 			// Patient who doesn't need healing
 			AssignmentDone(&s, TRUE, TRUE);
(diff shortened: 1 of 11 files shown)


momoko-h:
I think this is a perfectly reasonable option and the implementation itself looks fine, with one exception: save game compatibility. Have you tried loading an old save with a new executable or a new save with an old executable? I doubt either worked.

If you want to add new information to a save you have two options: either use the SavedGameStates mechanism (see SaveLoadGameStates.h) or reuse skipped bytes in the existing data. Since there are plenty of unused bytes in the SOLDIERTYPE on disk format that option is perfectly viable here, you just would have to increase the SAVE_GAME_VERSION to be able to differentiate between old and new saves.

src/game/Strategic/Assignments.cc
```diff
@@ -1348,8 +1351,10 @@ static UINT8 GetMinHealingSkillNeeded(SOLDIERTYPE const* const patient)
 // can this soldier be healed by this doctor?
 static BOOLEAN CanSoldierBeHealedByDoctor(SOLDIERTYPE const* const patient, SOLDIERTYPE const* const doctor, BOOLEAN const fThisHour, BOOLEAN const fSkipKitCheck, BOOLEAN const fSkipSkillCheck)
 {
+	bool fStatsDamaged = pPatient->bAgilityDamage == 0 && bPatient->bDexterityDamage == 0 && bPatient->bStrengthDamage == 0 && bPatient->bWisdomDamage == 0;
```
lynxlynxlynx: This is a misnomer, so I suggest you either rename it or negate the logic.
Btw, I also see it repeat below, a shorter way to check they're all 0 is to just sum them up and compare that instead.

src/game/Strategic/Assignments.cc
```diff
@@ -1460,9 +1466,87 @@ static UINT16 HealPatient(SOLDIERTYPE* pPatient, SOLDIERTYPE* pDoctor, UINT16 us
 		}
 	}
 
+	// heal stats that are damaged
+	if (pPatient -> bAgilityDamage > 0 || bPatient -> bDexterityDamage > 0 || bPatient -> bStrengthDamage > 0 || bPatient -> bWisdomDamage > 0)
+	{
+		bPointsToUse = STAT_HEALING_RATE_DIVISOR * std:max(pPatient -> bAgilityDamage, pPatient -> bDexterityDamage, pPatient -> bStrengthDamage, pPatient -> bWisdomDamage);
```
lynxlynxlynx: I'm surprised this compiled, shouldn't it be `std::max` within an initializer list or `std::max_element`?

src/game/Tactical/Weapons.cc
```diff
@@ -2951,6 +2951,7 @@ INT32 BulletImpact( SOLDIERTYPE *pFirer, SOLDIERTYPE * pTarget, UINT8 ubHitLocat
 						if ( bStatLoss > 0 )
 						{
 							pTarget->bWisdom -= bStatLoss;
+							pTarget->bWisdomDamage = bStatLoss;
```
lynxlynxlynx: Shouldn't all these be additive?

lynxlynxlynx: Nevermind, I see it's in the next commit.

lynxlynxlynx:
And it should be squash-merged in the end, since the commit split doesn't make sense and is uncompilable for most of them.

czaccardelli:
> And it should be squash-merged in the end, since the commit split doesn't make sense and is uncompilable for most of them.

Can do.

> I think this is a perfectly reasonable option and the implementation itself looks fine, with one exception: save game compatibility. Have you tried loading an old save with a new executable or a new save with an old executable? I doubt either worked.
> 
> If you want to add new information to a save you have two options: either use the SavedGameStates mechanism (see SaveLoadGameStates.h) or reuse skipped bytes in the existing data. Since there are plenty of unused bytes in the SOLDIERTYPE on disk format that option is perfectly viable here, you just would have to increase the SAVE_GAME_VERSION to be able to differentiate between old and new saves.

Yeah, I decided to use skipped bytes so that saves are backwards (and forwards) compatible. Since they should have full compatibility now, should I still increase the `SAVE_GAME_VERSION`?

momoko-h:
> Yeah, I decided to use skipped bytes so that saves are backwards (and forwards) compatible. Since they should have full compatibility now, should I still increase the `SAVE_GAME_VERSION`?

Well, all saves produced by executable after commit 7e29be2 should have these bytes cleared to 0, but if you manage to find even older saves, your *Damage values could be set to somewhat random values. It may be a bit overly pedantic, but increasing the save game version and checking it is the cleaner option.

momoko-h:
You'd also have to actually check for the game version when you extract the new values.

Something like this, as a lambda inside `ExtractSoldierType`
```

	auto ExtractStatDamage = [&](INT8 & whichDamage)
	{
		INT8 val = d.read<INT8>();
		whichDamage = uiSavedGameVersion >= 103 ? val : 0;
	};
      /* .... */
      ExtractStatDamage(s->bAgilityDamage);

```

lynxlynxlynx:
LGTM. @momoko-h do you want to do a final review?

momoko-h:
Oh, for some reason I thought this was still a draft. Never mind, it looks fine to me, too.

PR #2378, opened 2026-04-26
Replace some MPrintBuffer calls with MPrint

momoko-h:
The reason why MPrintBuffer was used here is because the frame buffer is locked when the call is made. With DirectDraw and its hardware surfaces this mattered because locks were exclusive and MPrint's attempt to lock the frame buffer again would have failed. SDL2's software surfaces do not have this restriction.


lynxlynxlynx:
Thanks!

PR #2381, opened 2026-05-23
Fixes cleanup of the keyring popup when leaving the mapscreen.

cMurschall:
## Summary

Fixes cleanup of the keyring popup when leaving the mapscreen.

image

## Reproduction

1. Open the mapscreen.
2. Open a merc’s inventory from the mapscreen und open the keyring popup.
3. Switch back to tactical while the keyring popup is still open.

## Details
The mapscreen inventory teardown removed the keyring button region, but an already open keyring popup was not explicitly closed. This could leave popup state, mouse regions, and cursor restrictions active after switching back to tactical.

The fix closes the keyring popup during mapscreen shutdown via `DeleteKeyRingPopup()`, which is safe to call even when no popup is open.


lynxlynxlynx:
The second line is misindented, but other than that, LGTM. Thanks!

PR #2386, opened 2026-06-17
Upgrade to SDL3

selaux:
This only includes the actual code changes, so the diff stays readable. I will include the dependency changes in a separate PR (which is the one we need to merge in the end). Because of that the build here is supposed to fail. #2387 might succeed, but I have not tested all OSs yet.


src/sgp/Video.cc
```diff
@@ -124,46 +115,47 @@ void InitializeVideoManager(const VideoScaleQuality quality,
...
 	g_game_window = SDL_CreateWindow(APPLICATION_NAME,
-					SDL_WINDOWPOS_UNDEFINED,
-					SDL_WINDOWPOS_UNDEFINED,
 					SCREEN_WIDTH, SCREEN_HEIGHT,
 					g_window_flags);
 
-	GameRenderer = SDL_CreateRenderer(g_game_window, -1, 0);
-	SDL_RenderSetLogicalSize(GameRenderer, SCREEN_WIDTH, SCREEN_HEIGHT);
+	GameRenderer = SDL_CreateRenderer(g_game_window, nullptr);
+	// FIXME: Is stretch the correct way here? There are other options
```
lynxlynxlynx: it does match the current behaviour

lynxlynxlynx:
Thanks, I was expecting more would need to change, but it's good that I was wrong. :)

selaux:
So, everything should compile now (in the other PR). This PR now contains all changes necessary, excluding the `dependencies/lib-SDL*` directories. Things still to do:

- [x] Address the FIXMEs in code, where I was not entirely sure what to do
- [x] Upgrade to most recent version of SDL3 (currently `3.4.2`, newest is `3.4.10`)
- [x] Mouse is not grabbed correctly in fullscreen
- [x] Integer scaling does not seem to work properly (fullscreen only)
- [x] Mouse warping does not seem to work properly
- [ ] Comparing the produced release packages to previous versions (e.g. for missing DLLs)
- [ ] Testing on all operating systems

Some things that are noteworthy, that I found out in the process:

- SDL3 needs to be installed via a GitHub action (the official one from SDL), as Ubuntu does not have SDL3 in the package repo (earliest is Ubuntu 26.04)
- `LOCAL_SDL_LIB` now points directly to a SDL CMake config file, as the directory structure of the distributed packages for the different OSs is not the same.
- Android target SDK needed to be updated, as SDL now uses some more recent functionality in their Android files. This should not affect the minimum supported Android version though.

src/sgp/Video.cc
selaux: I had to change it to `SDL_LOGICAL_PRESENTATION_LETTERBOX`, because it would not keep the aspect ratio with `SDL_LOGICAL_PRESENTATION_STRETCH`. This matches previous behavior now.

selaux:
Another update: SDL3 is upgraded to the latest version and FIXMEs are resolved. I still noticed some issues around mouse grabbing and scaling though that need to be resolved. I updated the previous comment to reflect that.

NB: I also sneaked in an agent skill to upgrade SDL3, not sure if we want that (agent skills in general), but it made my life easier ;)

selaux:
Closing in favor of #2387

PR #2390, opened 2026-06-22
Externalize fire suppression modifier and suppression reaction values

steph-pg:
At the moment fire suppression is almost non-existent because tolerance is often too high and we also get values less than 1 which get truncated to 0. I propose to externalize 2 values that will make suppression (and indectly burst fire) more useful and noticeable.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -236,5 +236,11 @@
     //  Is the start sector revealed at game start?
     //  vanilla settings: true
     "start_sector_revealed": true
-  }
+    },
+
+    // Scales AP loss from suppression (numerator of the AP-loss formula); vanilla 6, recommended setting is 24
+    "suppression_fire_modifier": 6,
+    // Threshold (numerator) above which a suppressed soldier drops to a lower stance.
+    // vanilla settings: 130. Set to 0 to make soldiers always react (1.13 behaviour); raise it to make them even more resistant.
+    "suppression_fire_reaction_threshold": 130
 }
--- a/src/game/Tactical/Overhead.cc
+++ b/src/game/Tactical/Overhead.cc
@@ -5168,6 +5168,7 @@ static void HandleSuppressionFire(const SOLDIERTYPE* const targeted_merc, SOLDIE
 	INT16 sClosestOpponent, sClosestOppLoc;
 	UINT8 ubPointsLost, ubTotalPointsLost, ubNewStance;
 	UINT8 ubLoop2;
+	const UINT8 ubSuppressionModifier = gamepolicy(suppression_fire_modifier);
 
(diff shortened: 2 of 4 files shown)


assets/externalized/game.json
```diff
@@ -236,5 +236,11 @@
     //  Is the start sector revealed at game start?
     //  vanilla settings: true
     "start_sector_revealed": true
-  }
+    },
+
+    // Scales AP loss from suppression (numerator of the AP-loss formula); vanilla 6, recommended setting is 24
+    "suppression_fire_modifier": 6,
+    // Close call bullets always cause soldier to drop to lower stance (2.13 behaviour)
```
lynxlynxlynx: Did you mean 1.13?

src/game/Tactical/Overhead.cc
```diff
@@ -5219,7 +5220,7 @@ static void HandleSuppressionFire(const SOLDIERTYPE* const targeted_merc, SOLDIE
 			ubNewStance = 0;
 
 			// merc may get to react
-			if ( pSoldier->ubSuppressionPoints >= ( 130 / (6 + bTolerance) ) )
+			if (gamepolicy(suppression_fire_soldiers_always_react) || pSoldier->ubSuppressionPoints >= (130 / (6 + bTolerance)))
```
lynxlynxlynx: Wouldn't it make more sense to externalize 130? It would allow for more nuanced choice to the player and simplify this check.

lynxlynxlynx:
Thanks, nice first patch!

assets/externalized/game.json
steph-pg: yes, update the PR

src/game/Tactical/Overhead.cc
steph-pg: It's hard to externalize this part of formula to make it 'always react'. In my opinion it's more realistic to always react when bullet is flying next to you. It's just a reflex I guess. I would also argue that a more experience person should actually react more. From game wise experience this part could be even more useful than the whole suppression. You force a guy to drop which sometimes allow you to maneuver and also cause him to lower his weapon, too which probably force next shot to be more expensive. It also makes game a bit more challenging to player as now opponent in lower stance and their cth is higher and player's lower.

lynxlynxlynx: Yeah, but that's why we externalize values in the first place, so people can use their own judgement about how the simulated reality should work.

`ubSuppressionPoints` is unsigned judging from the name, so a 0 instead of 130 would already replicate your binary switch, no?

steph-pg: yes, you're right. updated the PR

lynxlynxlynx:
Thanks!

PR #2391, opened 2026-07-08
Fix the garage door pathfinding and visual bugs

ShishkinP:
I somehow missed that #2352 completely destroys the garage door:
image
image
It is now (not falsely I hope) assumed that the STRUCTURE_HIDDEN flag wasn't applied to doors, so we leave their flags unchanged.

### DIFF
--- a/src/game/TileEngine/Structure.cc
+++ b/src/game/TileEngine/Structure.cc
@@ -386,10 +386,19 @@ static STRUCTURE* CreateStructureFromDB(DB_STRUCTURE_REF const* const pDBStructu
 
 	STRUCTURE* const pStructure = new STRUCTURE{};
 
-	// We need to clear the old STRUCTURE_HIDDEN flag, which had the same value
-	// as STRUCTURE_GARADGEDOOR to prevent problems with pathing and opening
-	// or closing doors, see PR #2348 and nightops-maps PR #11.
-	pStructure->fFlags          = pDBStructure->fFlags & ~STRUCTURE_GARAGEDOOR;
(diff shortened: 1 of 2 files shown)


momoko-h:
If the problem is differentiating between the `STRUCTURE_HIDDEN` flag set by some editors and the `STRUCTURE_GARAGEDOOR` flag set in `CreateTileDatabase` wouldn't it make sense to leave the DB_STRUCTURE_REF flags as they are (remove the code in TileDef.cc) and add the code to set `STRUCTURE_GARAGEDOOR` to `CreateStructureFromDB`?

For example,
```
	// fix garage doors not having a flag of their own separate from sliding doors
	if (pDBStructure->fFlags & STRUCTURE_SLIDINGDOOR && pDBStructure->ubNumberOfTiles == 2)
	{
		pStructure->fFlags = (pDBStructure->fFlags & ~STRUCTURE_SLIDINGDOOR) | STRUCTURE_GARAGEDOOR;
	}
	else
	{
		// We need to clear the old STRUCTURE_HIDDEN flag, which had the same value
		// as STRUCTURE_GARADGEDOOR to prevent problems with pathing and opening
		// or closing doors, see PR #2348 and nightops-maps PR #11.
		pStructure->fFlags = pDBStructure->fFlags & ~STRUCTURE_GARAGEDOOR;
	}

```

ShishkinP:
> For example,

This does the job. Thank you.

momoko-h:
Ok, I hope this settles this problem for good.

PR #2392, opened 2026-07-11
Fix items on top of destroyed structures not changing their z-level correctly

ShishkinP:
Fixed: if a structure that has an item on top of it gets destroyed the item keeps hanging at the same height and can only be picked up from sector inventory.
image
image


momoko-h:
Does this work when the structure is not completely destroyed but instead replaced with a debris graphic?

ShishkinP:
> Does this work when the structure is not completely destroyed but instead replaced with a debris graphic?

It does:
image

momoko-h:
Ok, so this will probably look a bit weird on top of some debris but not worse than items floating in thin air. Nobody expects this to be true 3D rendering anyway.

PR #2393, opened 2026-07-12
Add (reenable) hide_bullets option as game  policy

steph-pg:
Revive the old TOPTION_HIDE_BULLETS behavior as a game policy (default false = vanilla). When enabled, bullets are not drawn in flight.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -62,6 +62,11 @@
   //  vanilla setting: false
   "all_drops_visible": false,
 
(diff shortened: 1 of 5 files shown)


lynxlynxlynx:
Thanks! I guess they turned it off to not need to redesign the options window? But we have separate ideas for that anyway.
EDIT: I, now that I can look at the history it's obvious, it used to be hidden behind a cheat key, but disabled already in the initial import.

PR #2394, opened 2026-07-15
Fix incorrect wall replacement on adjacent wall destruction

ShishkinP:
Issues:
- Wrong offset: some walls are replaced with unrelated structures instead of a damaged wall:
image
image

- Wrong inside/outside orientation dependence:  some walls are not affected at all
image

Fixed:
image

### DIFF
--- a/src/game/TileEngine/Explosion_Control.cc
+++ b/src/game/TileEngine/Explosion_Control.cc
@@ -296,7 +296,16 @@ static void ReplaceWall(GridNo const grid_no, UINT8 const orientation, INT16 con
 	if (!wall_struct || !(wall_struct->fFlags & STRUCTURE_WALL)) return;
 
 	LEVELNODE* const node    = FindLevelNodeBasedOnStructure(wall_struct);
-	UINT16     const new_idx = GetTileIndexFromTypeSubIndex(gTileDatabase[node->usIndex].fType, sub_idx);
+
+	UINT16 tileType = gTileDatabase[node->usIndex].fType;
+	if (tileType > FOURTHWALL)
+	{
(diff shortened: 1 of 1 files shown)


momoko-h:
Do I understand this correctly, `GetWallStructOfSameOrientationAtGridno` return a `STRUCTURE` that is supposedly a wall (its `STRUCTURE_WALL` flag is set) but when we look at the corresponding `LEVELNODE`'s TileIndex it is not actually wall tile? Any idea how that can happen? What are the values of `node->usIndex` and `gTileDatabase[node->usIndex].fType` in this case?

ShishkinP:
image

In this situation `GetWallStructOfSameOrientationAtGridno` returns `nullptr` because:
destroyed wall is INSIDE_TOP_LEFT
adj. wall to replace is OUTSIDE_TOP_LEFT
So, the wall to replace stays intact. It's wrong because the INSIDE/OUTSIDE alignment shouldn't matter, only the LEFT/RIGHT one should. The solution is to call `GetWallStructOfSameOrientationAtGridno` two times for both inside and outside orientations.


image

In this one the wall to replace is of a tile type DECORATIONS (FIRSTDECORATIONS and so on) while subindex constants were made with an assumption there are only walls of a tile type WALL (FIRSTWALL and so on). The solution is to just pretend it's always a WALL type.

So far Estoni is the only map I found these problems.

momoko-h:
Ah yes, I just had at look at this map in the editor, where these walls (a comment in SelectWin.cc calls them "old obsolete wall decals") are found in the buildings tab, under the "Cracked Walls" button. So I believe your workaround should be fine in all cases.

PR #2396, opened 2026-07-16
Fix smell strength on already visited grid

### LINKED ISSUE #2395: Gabby's Jar of Elixir has no effect
Version: Stracciatella v0.22.1-git
Build 04.12.02

---

|  v0.22.1-git | JA2 linux version  |    no    |    yes     |

---

JA2S Sci-Fi mode.

Aquired "Jar of Elixir" from Gabby, after the "Jar of Elixir" is applied to a merc, it should change the "Camo" stat to "100%", like applying real camoflouge. But neighter it changes the camo stats, nor does it work when entering the mines. Bugs are detecting merc right on. Merc should be able to sneak by the bugs when the "jar of elixir" is working. Sector H8 Cambria. Merc standing in front of the mine, most of them have the Elixir. 

Savegame attached

[SaveGame.zip](https://github.com/user-attachments/files/30078429/SaveGame.zip)

#### Steps to reproduce
1. Apply one "Jar of Elixir" to all mercs before entering the mine
2. Verify the "CAMO" stats - displays 0%
3. Enter the mine
4. Change to "stealth" mode on all merc
5. Move in the mine, the bugs are detecting merc on the spot

momoko-h:
Walking over an already visited grid incorrectly set the smell strength of the grid to the maximum value of 63. Fixes #2395.

### DIFF
--- a/src/game/TileEngine/Smell.cc
+++ b/src/game/TileEngine/Smell.cc
@@ -71,7 +71,7 @@ constexpr UINT8 BLOOD_ROOF_STRENGTH(UINT8 b) { return (b & 0b1110'0000) >> 5; }
 #define DECAY_SMELL_STRENGTH( s ) \
 { \
 	UINT8 ubStrength = SMELL_STRENGTH( (s) ); \
-	ubStrength--; \
+	if (ubStrength > 0) ubStrength--; \
 	ubStrength = ubStrength << SMELL_TYPE_NUM_BITS; \
 	(s) = SMELL_TYPE_BITS( (s) ) | ubStrength; \
 }
@@ -88,7 +88,7 @@ constexpr UINT8 BLOOD_ROOF_STRENGTH(UINT8 b) { return (b & 0b1110'0000) >> 5; }
 
 #define DECAY_BLOOD_DELAY_TIME( b ) \
 { \
-	(b)--; \
+	if (BLOOD_DELAY_TIME(b) > 0) (b)--; \
 }
 
 #define SET_BLOOD_FLOOR_STRENGTH( b, nb ) \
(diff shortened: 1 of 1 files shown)


src/game/TileEngine/Smell.cc
```diff
@@ -114,7 +114,7 @@ constexpr UINT8 BLOOD_DELAY_MAX = 3;
 	UINT8 ubRoofStrength; \
 	ubRoofStrength = BLOOD_ROOF_STRENGTH( (b) ); \
 	ubRoofStrength--; \
-	SET_BLOOD_FLOOR_STRENGTH( (b), ubRoofStrength ); \
+	SET_BLOOD_ROOF_STRENGTH( (b), ubRoofStrength ); \
```
lynxlynxlynx, line 117: Are you sure this is correct? The roof and ground state is stored in the same byte.

momoko-h: Yes, but as the comment at the top of the file explains the roof strength is in bits 7-5, the floor strength is in bits 4-2. However, it looks like the BLOOD_ROOF_STRENGTH macro is also wrong, it only & the value with 0xE0 but does not shift it >>5.

lynxlynxlynx: I agree the latest commit is good when considering other code, but overall `DECAY_BLOOD_ROOF_STRENGTH` still looks wrong. `ubRoofStrength` will be set to an extracted strength with a max of 31, but then it'll get passed on the same. With the way `SET_BLOOD_ROOF_STRENGTH` then uses it, the result will be always 0, that input parameter doesn't matter at all. Shouldn't it be shifted before being passed on (or the macro cleaned up to not run in circles)?

momoko-h: Doh, yes of course you're right. I'll add some tests for this annoying bit fiddling.

lynxlynxlynx:
Thanks!

PR #2399, opened 2026-07-18
Enable hitbox visualization

ShishkinP:
Introducing a basic form of debugging structures' shapes.
To activate enter cheat mode with CTRL+GABBI then press CTRL+C.

image

image

image

I'll convert it to SDL3 before merging.

### DIFF
--- a/src/sgp/Video.cc
+++ b/src/sgp/Video.cc
@@ -11,6 +11,7 @@
 #include "VObject_Blitters.h"
 #include "VSurface.h"
 #include "Video.h"
+#include "Visualizer.h"
 #include "UILayout.h"
 #include "Icon.h"
 #include <algorithm>
@@ -548,6 +549,11 @@ void RefreshScreen(void)
 	}
 	else {
 		SDL_RenderCopy(GameRenderer, ScreenTexture, NULL, NULL);
+	}
+
+	if (Visualizer::IsActivated())
+	{
+		Visualizer::Render(GameRenderer);
 	}
 
 	FPS::RenderPresentPtr(GameRenderer);
--- a/src/sgp/Video.cc
+++ b/src/sgp/Video.cc
@@ -11,6 +11,7 @@
 #include "VObject_Blitters.h"
 #include "VSurface.h"
 #include "Video.h"
+#include "Visualizer.h"
 #include "UILayout.h"
 #include "Icon.h"
 #include <SDL3/SDL_video.h>
@@ -547,6 +548,11 @@ void RefreshScreen(void)
 	}
 	else {
 		SDL_RenderTexture(GameRenderer, ScreenTexture, NULL, NULL);
+	}
+
+	if (Visualizer::IsActivated())
+	{
+		Visualizer::Render(GameRenderer);
 	}
 
 	FPS::RenderPresentPtr(GameRenderer);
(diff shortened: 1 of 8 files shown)


src/sgp/Video.cc
```diff
@@ -550,6 +551,11 @@ void RefreshScreen(void)
 		SDL_RenderCopy(GameRenderer, ScreenTexture, NULL, NULL);
 	}
 
+	if (Visualizer::IsActivated())
```
momoko-h: I wouldn't bother with an extra `IsActivated` function, just let `Render` return early if the visualizer is off.

src/sgp/Visualizer.cc
```diff
@@ -0,0 +1,261 @@
...
+#include "WorldDef.h"
+
+#include <memory>
+
+namespace Visualizer
+{
+
+bool toggledOn{ };
+
+struct SDLDeleter
```
momoko-h: This is the 3rd instance of a SDLDeleter struct after the ones in VSurface.h and FPS.cc. It's time to move this into its own header, SDL_Helpers.h or something like that.

src/sgp/Visualizer.cc
```diff
@@ -0,0 +1,261 @@
...
+// Green ground rhombus (isometric square)
+constexpr SDL_Vertex verticesLvl0[]
+{
+	   baseCoord,																						 // farthest from the viewer
+	{ {baseCoord.position.x + adjLegLength, baseCoord.position.y + oppLegLength},     baseCoord.color }, // right
+	{ {baseCoord.position.x,                baseCoord.position.y + oppLegLength * 2}, baseCoord.color }, // nearest
+	{ {baseCoord.position.x - adjLegLength, baseCoord.position.y + oppLegLength},     baseCoord.color }, // left
+};
+// Brown walkable roof rhombus
+constexpr SDL_Vertex verticesLvl1[]
```
momoko-h: Currently unused, is this for something else you still want to add?

src/sgp/Visualizer.cc
```diff
@@ -0,0 +1,261 @@
...
+			}
+		}
+	}
+
+	SDL_SetRenderTarget(renderer, nullptr);
+}
+
+static void InitTextures(SDL_Renderer* const renderer)
+{
+	finalTexture.reset(SDL_CreateTexture(renderer, SDL_PIXELFORMAT_RGB565, SDL_TEXTUREACCESS_TARGET, bgRectDest.w, bgRectDest.h));
```
momoko-h: Unless you plan to use some of our own functions to write to this textures, there's is not much reason to use RGB565 here, other than that it'd use less memory. RGBA8888 is probably a bit faster and gives you a wider selection of colours.

src/sgp/Visualizer.cc
```diff
@@ -0,0 +1,261 @@
...
+
+	bgTexture.reset(SDL_CreateTexture(renderer, SDL_PIXELFORMAT_RGB565, SDL_TEXTUREACCESS_TARGET, bgRectDest.w, bgRectDest.h));
+	SDL_SetRenderTarget(renderer, bgTexture.get());
+
+	// Black background
+	SDL_SetRenderDrawColor(renderer, 10, 10, 10, SDL_ALPHA_OPAQUE);
```
momoko-h, line 188: The renderer's draw color is also used whenever we call SDL_RenderClear, so I think you should set this back to completely black when you're done here.

momoko-h (commented):
Very nice, better debugging tools are always welcome. 👍

ShishkinP:
OK, it's ready

lynxlynxlynx:
Thanks!

PR #2401, opened 2026-07-19
BuildDisplayWindow: use std::span

momoko-h:
The compiler knows the size of the DisplaySpec arrays, so we can replace the verbose and error prone manual management.

Removes two HACK comments. This commit also removes a couple of "unused argument" warnings.

Remove DisplaySpec::ubType, DISPLAY_TEXT is not implemented.


lynxlynxlynx:
Thanks!

PR #2402, opened 2026-07-19
Prevent dismissing a bleeding merc

steph-pg:
Now when you can't do first aid, or when you are retreating from sector and bleeding you can simply dismiss a merc which clearly looks like abuse


lynxlynxlynx:
LGTM, thanks.

PR #2403, opened 2026-07-24
EditScreen: Remove two unused variables

momoko-h:
Apparently the editor once used the alternative scrolling method still present but this was already removed a long time ago and only these two variables and some stale comments were left.


lynxlynxlynx:
Thanks!

PR #2405, opened 2026-07-30
Refactor the wall orientation enum

ShishkinP:
Turning the wall orientation values into bitwise flags. This will make it more convenient working with them.
The main changes are in:
[TileDef.h](https://github.com/ja2-stracciatella/ja2-stracciatella/compare/master...ShishkinP:refactor-wall-orientation-enum?expand=1#diff-802148ea4a4c696fc58fe1e426a2439dd57c881d9631568e870e630c75c33286)
[TileDef.cc](https://github.com/ja2-stracciatella/ja2-stracciatella/compare/master...ShishkinP:refactor-wall-orientation-enum?expand=1#diff-96da0b7fe95ccb74091c55218ff3f7deea87f13096201b8f3db58c038141a917)
[Types.h](https://github.com/ja2-stracciatella/ja2-stracciatella/compare/master...ShishkinP:refactor-wall-orientation-enum?expand=1#diff-71522e7c264f0d583c851444876f0be8edaaa133686458d67f336d33701d9aea)

### DIFF
--- a/src/game/Tactical/LOS.cc
+++ b/src/game/Tactical/LOS.cc
@@ -282,10 +282,8 @@ static BOOLEAN ResolveHitOnWall(STRUCTURE* pStructure, INT32 iGridNo, INT8 bLOSI
 	fEastWest = ((ddHorizAngle > (0) && ddHorizAngle < (PI * 1 / 2)) ||
 			(ddHorizAngle < (-PI * 1 / 2) && ddHorizAngle > (-PI)));
 
-	fTopLeft = (pStructure->ubWallOrientation == INSIDE_TOP_LEFT ||
-			pStructure->ubWallOrientation == OUTSIDE_TOP_LEFT);
-	fTopRight = (pStructure->ubWallOrientation == INSIDE_TOP_RIGHT ||
-			pStructure->ubWallOrientation == OUTSIDE_TOP_RIGHT);
+	fTopLeft = pStructure->ubWallOrientation & ORIENT_LEFT;
+	fTopRight = pStructure->ubWallOrientation & ORIENT_RIGHT;
 
 	if ( fNorthSouth )
 	{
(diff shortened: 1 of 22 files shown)


src/game/Tactical/LOS.cc
```diff
@@ -4465,33 +4434,15 @@ INT32 CheckForCollision(FLOAT dX, FLOAT dY, FLOAT dZ, FLOAT dDeltaX, FLOAT dDelt
...
-								{
-									*pdNormalX = -1;
-									return( COLLISION_WALL_SOUTHEAST );
-								}
-								else
-								{
-									*pdNormalX = 1;
-									return( COLLISION_WALL_NORTHEAST );
-								}
+								*pdNormalX = -1;
```
lynxlynxlynx: These were also set to 1 before, is this reduction intentional?

src/game/Tactical/Overhead.cc
```diff
@@ -3496,12 +3496,13 @@ INT16 FindAdjacentGridEx(SOLDIERTYPE* pSoldier, INT16 sGridNo, UINT8* pubDirecti
...
 				{
-					case OUTSIDE_TOP_LEFT:
-					case INSIDE_TOP_LEFT:   *pubDirection = SOUTH; break;
-					case OUTSIDE_TOP_RIGHT:
-					case INSIDE_TOP_RIGHT:  *pubDirection = EAST; break;
+					*pubDirection = SOUTH;
+				}
+				else
+				{
+					*pubDirection = EAST;
```
lynxlynxlynx, line 3505: now the default case is gone, is always using EAST for it really ok or should the value be preserved like before? This applies to all switch->if conversions that had no explicit default (those you caught).

src/game/Tactical/LOS.cc
ShishkinP: No, it's a mistake on my part.

src/game/Tactical/Overhead.cc
ShishkinP: These default cases are only for orientationless form of STRUCTURE_FENCE (in the pic below) and probably (didn't test it) STRUCTURE_SWITCH (button on a wall).

Image

Everything else is expected to ALWAYS have an orientation otherwise they are very obviously dysfunctional and having a no orientation default won't even fix them.
In this case if the door is not left the only other remaining (default) option is it's right.

lynxlynxlynx: Alright, just wanted to make sure the cases were considered in all these changes.

PR #2407, opened 2026-07-31
Give the sungoggles to-hit bonus only in glare

steph-pg:
Sun goggles used to reduce the effective range of every shot by 10% unconditionally - day or night, indoors or out. Restrict the bonus to targets standing in light brighter than normal daylight

Applies to both CalcChanceToHitGun and CalcThrownChanceToHit.

### DIFF
--- a/src/game/Tactical/Weapons.cc
+++ b/src/game/Tactical/Weapons.cc
@@ -19,6 +19,8 @@
 #include "Points.h"
 #include "AI.h"
 #include "LOS.h"
+#include "Lighting.h"
+#include "Environment.h"
 #include "RenderWorld.h"
 #include "OppList.h"
 #include "Interface.h"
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
LGTM, but I do wonder if this should be a json toggle.

And TIL: "UV goggles only function above ground". What an odd decision.

steph-pg:
I would not add a toggle, I get the desire to keep things as vanilla as possible but this bonus looks weird at night :) Most players probably don't even know it exists but those who noticed it can min/mix by swapping night googles/sun googles to get more cth which looks like a bad game design

momoko-h:
Isn't testing for sun goggles in `CalcChanceToHitGun` redundant because `DistanceVisible` which got called earlier already adjusted its result when the soldier wears sun goggles?

steph-pg:
> Isn't testing for sun goggles in `CalcChanceToHitGun` redundant because `DistanceVisible` which got called earlier already adjusted its result when the soldier wears sun goggles?

This one is not about the sight range. Turns out that sun goggles make you more accurate :)

momoko-h:
Thanks.

PR #2409, opened 2026-07-31
RemovePlayerFromGroup: cleanup

momoko-h:
• Since a vanilla patch it was impossible that the assertion in this
  function failed. However, it is still useful to log diagnostics
  if GetGroup fails.
• Several callers set ubGroupID to 0 even though this function
  already does that.
• AddCharacterToSquad used ubGroupID after is was already cleared.


lynxlynxlynx:
Thanks!

PR #2410, opened 2026-08-01
Fix fix students without trainer don't fatigue

steph-pg:
A merc on the Student assignment with no Trainer for his stat in the sector is flagged as having nothing to do, which makes him rest instead of getting fatigued.  HandleTrainingInSector() ignored that flag and still awarded him self-training points, so Student without a trainer gained stats at the same rate as Practice but for free, with no fatigue.


lynxlynxlynx:
Thanks!

PR #2411, opened 2026-08-01
Allow to merge silver and gold ore stacks in the sector inventory

steph-pg:
PlaceObjectInInventoryStash() tested the dragged object against the MONEY item index directly, so only cash took the merge branch. 

Dropping one ore onto another in the sector inventory therefore just traded the two objects' places, which reads as a click that did nothing. Test the item class instead so all three merge.


lynxlynxlynx:
Thanks!

PR #2412, opened 2026-08-01
Fix blowing sam computers after loading games

steph-pg:
The listener was registered in InitializeSAMSites(), which only runs from InitStrategicLayer() -> InitNewGame(). Loading a saved game never goes through there, so in a process that starts by loading a save nothing is subscribed and blowing up the computer does nothing at all: the condition stays where it was, the airspace stays hostile and the site keeps shooting.

Split the registration out into InitSAMSitesEngine() and call it from InitStrategicEngine(), which runs once per application run and is where every other engine-level listener is registered. InitializeSAMSites() keeps only the per-game part, resetting every site to condition 100.


lynxlynxlynx:
Thanks!

PR #2413, opened 2026-08-01
Remove VEHICLETYPE::sGridNo

momoko-h:
It was not used anywhere. While we are here, also reorder the VEHICLETYPE members to reduce padding.


lynxlynxlynx:
Thanks

PR #2414, opened 2026-08-02
Bugfix: causing damage with explosives incorrectly marked as FROM_FAILURE

czaccardelli:
Causing damage with explosives marked the stat change reason as `FROM_FAILURE` meaning mercs cannot level up or cross stat gain thresholds from causing damage with them.


ShishkinP:
LGTM. Thank you.

PR #2415, opened 2026-08-02
Require enemies in sector to manually enter turn based

steph-pg:
Pressing 'd' only raised the INCOMBAT flag, skipping EnterCombatMode and its setup. bNumFoughtInBattle was left holding the count from the last real fight, so ending the turn in an empty sector counted as a won battle and paid out morale and loyalty. Route the key through EnterCombatMode, which declines when no capable enemy is present.


ShishkinP:
Can't come up with any application of turned based mode in a sector with no enemies so LGTM so far.

lynxlynxlynx:
Thanks!

PR #2416, opened 2026-08-02
Let a collapsed merc drink from their canteen

steph-pg:
EnoughPoints() refuses every action while bCollapsed is set. In vanilla it was possible to apply canteen and camo (something else?). Let's do a special case for canteen for now, as quit often it allows merc to survive after getting hit and collapsed


ShishkinP:
Will it still be applied if a merc's HPs < OKLIFE (unconscious)?

steph-pg:
> Will it still be applied if a merc's HPs < OKLIFE (unconscious)?

Added extra checks for OKLIFE and dart gun

momoko-h:
Strange word order in the comment, other than that this is fine by me.

lynxlynxlynx:
Thanks!

PR #2417, opened 2026-08-02
Fix commit f2bfc4d regression

ShishkinP:
Bug from #2405: enum conversion is applied to cached orientations which leads to double conversion.
Solution: apply it only to freshly extracted values

### DIFF
--- a/src/game/TileEngine/Structure.cc
+++ b/src/game/TileEngine/Structure.cc
@@ -364,6 +364,8 @@ static void CreateFileStructureArrays(STRUCTURE_FILE_REF & pFileRef)
 				case 4:
 					dbs->ubWallOrientation = ORIENT_OUTSIDE_RIGHT;
 					break;
+				default:
+					break;
 			}
 			pFileRef.pAuxData[usIndex].ubWallOrientation = dbs->ubWallOrientation;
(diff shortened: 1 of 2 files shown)


src/game/TileEngine/Structure.cc
```diff
@@ -353,6 +353,21 @@ static void CreateFileStructureArrays(STRUCTURE_FILE_REF & pFileRef)
 
 		NormalizeStructureTiles(tiles, dbs->ubNumberOfTiles);
 
+		// Convert contiguous orientation values to flags
+		if (pFileRef.pAuxData.size())
+		{
+			switch (dbs->ubWallOrientation)
+			{
```
momoko-h: It's probably worth documenting that cases 1 and 2 are not covered here because they'd just get the same numerical value. Might even add the code for them and hope the compiler optimizes it away again.

ShishkinP: I'll go with a comment since I don't know much about compiler optimizations.

momoko-h:
This now produces compiler warnings about the enumerators not handled in the switch, so please add the default case.

momoko-h:
Thanks

PR #2421, opened 2026-08-02
Enable additional options in options screen

czaccardelli:
- Re-enables options previously pruned from the game options menu (most importantly the option to turn off lights under mercs during night time!)
- Deprecates `hide_bullets` game policy since this is now covered by the in-game toggle for hiding bullets in the options screen


src/game/GameSettings.cc
```diff
@@ -31,7 +31,7 @@ GAME_OPTIONS		gGameOptions;
 
 
 //Change this number when we want any who gets the new build to reset the options
-#define GAME_SETTING_CURRENT_VERSION 522
+#define GAME_SETTING_CURRENT_VERSION 523
```
momoko-h, line 34: Changing the settings version or the size of the settings file means all options are reset to their default values, right? I wouldn't like having to switch everything back to my preferred options and this probably would surprise many users.

czaccardelli: Yes, but I doubt people even play with anything besides defaults and a few options that they instinctively toggle on and off during gameplay (treetops, wireframes, etc.) just using hotkeys. It's a one-time reset upon version upgrade that takes maybe 15 seconds to adjust back while playing.

I think there are two main benefits that are well worth restoring these options:
1. Being able to turn off the annoying lights that stay under mercs during night time (can't see darkness anymore!)
2. Not having things like `hide_bullets` options obscurely scattered in the `game.json` file when it's an easily-toggleable cosmetic change presented directly in the options screen.

src/game/GameSettings.cc
```diff
@@ -108,7 +108,7 @@ void SaveGameSettings(void)
 	// Record the current settings into the game settins structure
 	GAME_SETTINGS& g = gGameSettings;
 
-	BYTE  data[76];
+	BYTE  data[77];
```
lynxlynxlynx, line 111: Why is this increase needed?

src/game/GameSettings.cc
momoko-h: The settings also include the sound volumes and I'm pretty sure everyone adjusts those. But whatever, I'll agree under one condition: change the default of TOPTION_USE_METRIC_SYSTEM to TRUE. To quote Text_Utils, "else the user is a caveman".

src/game/GameSettings.cc
momoko-h: And without the corresponding change in `LoadGameSettings` this won't work.

czaccardelli: > Why is this increase needed?

Because on line 120 `INJ_U8A( d, g.fOptions,       lengthof(g.fOptions))` has the size changed since `lengthof(g.fOptions)` is now larger.

But that's incorrect. You still have 23 different options (the numerical value of TOPTION_TRACKING_MODE is still 22). The reason why NUM_ALL_GAME_OPTIONS has increased (and the size of fOptions with it) is just but because you've given it its own value. The last entry of that enum (please give it a proper name while you're at it) should just be NUM_ALL_GAME_OPTIONS = NUM_GAME_OPTIONS. Alternatively, completely remove one of the two.

src/game/GameSettings.cc
czaccardelli: Considering I think 90% of the JA2 community is European I think that's fair. 😆

src/game/GameSettings.cc
lynxlynxlynx: Can you make it obvious while you're at it, so we don't need to use magic numbers?

czaccardelli: Like a comment in the code?

lynxlynxlynx: No, like using `lengthof` and a `std::array` for data. Or even converting `fOptions` to it first, so no `sizeof` hacks are needed.

czaccardelli: `data` contains all the settings including metadata like the settings version and non-configurable things like `ubSizeOfDisplayCover` as well as arbitrary skips, so the total size of `data` is the sum of all these things in bytes.

Again, sorry if I'm missing something but outside of determining the size of all these individual things in bytes beforehand and adding them all up I'm not sure how else I could determine the total size `data`.

lynxlynxlynx: Yuck, but not something that needs to be cleaned up in this PR.

lynxlynxlynx:
Thanks!

momoko-h:
@czaccardelli, I am sorry, I meant to reply to your comment, not edit it. :disappointed:

PR #2423, opened 2026-08-03
When merc is put on assignment too late they should rest during next hour

steph-pg:
So now when EnoughTimeOnAssignment returns false we get no progress and we fatigue at the same time. When you do know how assignment time works, and you noticed this you need to micro manage


ShishkinP:
Thank you.

PR #2427, opened 2026-08-04
Add two function templates to sum up enemies

momoko-h:
Just because seeing ubNumAdmins + ubNumTroops + ubNumElites all the time is annoying. Hostiles() includes creatures and works for UNDERGROUND_SECTORINFO and SECTORINFO; Enemies() also for ENEMYGROUP.


lynxlynxlynx:
Thanks!

PR #2428, opened 2026-08-04
Check the firing hand for a silencer when making gunfire noise

steph-pg:
Two-pistol shooting fires each hand as its own shot, but the noise volume always looked for a silencer in the main hand. A silenced off-hand pistol was therefore as loud as an unsilenced one, and a silenced main-hand pistol quieted the off-hand shot too. Read the attachment from ubAttackingHand instead.


lynxlynxlynx:
Nice find, LGTM.

PR #2431, opened 2026-08-06
sol: fix compilation with lua 5.5

fenuks:
On my system (Arch Linux) I have several lua versions installed and cmake selects the latest version (5.5). Sol 3.5.0 only supports lua up to v5.4.

/usr/include/sol/compatibility/compat-5.3.h:410:4: error: #error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"
  410 | #  error "unsupported Lua version (i.e. not Lua 5.1, 5.2, 5.3, or 5.4)"

Patch to add support for lua 5.5.0 was written by halx99 in https://github.com/ThePhD/sol2/pull/1723.

### DIFF
--- a/dependencies/lib-sol2/CMakeLists.txt
+++ b/dependencies/lib-sol2/CMakeLists.txt
@@ -26,10 +26,16 @@ configure_file(
     @ONLY
 )
 
-# copy patch to build dir
+# copy patches to build dir
 configure_file(
     "${CMAKE_CURRENT_SOURCE_DIR}/getter/gcc-15.patch"
     "${CMAKE_CURRENT_BINARY_DIR}/getter/gcc-15.patch"
+    @ONLY
+)
+configure_file(
+    "${CMAKE_CURRENT_SOURCE_DIR}/getter/lua-55.patch"
+    "${CMAKE_CURRENT_BINARY_DIR}/getter/lua-55.patch"
+    COPYONLY
     @ONLY
 )
 
--- a/dependencies/lib-sol2/getter/CMakeLists.txt.in
+++ b/dependencies/lib-sol2/getter/CMakeLists.txt.in
@@ -31,5 +31,5 @@ externalproject_add(get-sol2
         BUILD_COMMAND       ""
         INSTALL_COMMAND     ""
         TEST_COMMAND        ""
-    PATCH_COMMAND ${PATCH_CMD} -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/gcc-15.patch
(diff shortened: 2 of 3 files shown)


lynxlynxlynx:
It would make more sense to bump to latest sol2 than to restrict the possible version even further.

Unfortunately the 5.5 support hasn't been merged yet, but it's ready and small enough to include:
https://github.com/ThePhD/sol2/pull/1723

fenuks:
I don't know if rest of the code is compatible with newer lua. I can confirm that it compiles for me with lua 5.4 (but I didn't test the executable). While I agree that it probably would be better to upgrade sol2, it will make packaging more difficult (as far I can see, sol2 repo isn't actively maintained, it's unlikely that a new version will be released anytime soon,  so those using prepackaged version will have a problem).

I tried to specify range `5.3...5.4` instead, but for some reason cmake ignores it and selects lua 5.5 anyway.

fenuks:
Problem with lua ranges is mentioned in https://github.com/athelas-astro/athelas/pull/170, with workaround.

lynxlynxlynx:
It wouldn't be a packaging burden for us, since it's just another patch to ship. 
For anyone else it expands the chances of the build not breaking, so also a positive change.
- For people using their own sol, your proposal makes it less likely to have a version match, since only one version is forced out of several supported. By adding 5.5 support we instead extend the range.
- For people using the bundled sol, it still works just as well

Eventually we have to move off sol2 any way.

fenuks:
> For people using their own sol, your proposal makes it less likely to have a version match, since only one version is forced out of several supported.

I did it because the code says it expects 5.3.
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/8fba755ed9b84d07a694e204540a3ce523650615/dependencies/lib-lua/CMakeLists.txt#L8

If any newer version is fine, then it's better to to allow them. Disallowing 5.5 only requires ugly workaround in CMakeLists.txt, though, as mentioned earlier. On the other hand, lua 5.5 sol2 patch might have issues.

If you think it's better to add patch, then I can update this PR.

lynxlynxlynx:
You can look at how we apply the existing gcc15 patch via cmake. We won't know if ja2s compiles against 5.4 or 5.5 until someone tries and this is the perfect opportunity.

fenuks:
I've updated the PR. Patched sol seems to be working. I was able to compile it with lua 5.5.1 and run the game.

lynxlynxlynx:
Thanks, it looks currently lua itself has hosting problems, that's why all the failures. Will restart the builders tomorrow.

PR #2432, opened 2026-08-08
Fix and enhance structure illumination

ShishkinP:
Resolved issues:

- Cave walls don't block light

Vanilla:

image

Fixed:

image

- The other side of lit up walls and structures in the same tile get all illuminated

image

image

- Open door slabs don't block light.
- Left-oriented windows don't let the light through from the north side to the south. Same with the right-oriented and west-to-east pass-through.

image

image

- Tall massive structures don't block light

image

image


src/game/TileEngine/Lighting.cc
```diff
@@ -295,102 +308,172 @@ static void LightInsertRayNode(LightTemplate* const t, const UINT16 usIndex, con
...
+				continue;
+			}
+
+			UINT8 boxColumnCount = 0;
+
+			for (auto& row : *pStruct->pShape)
+			{
+				for (auto& elem : row)
+				{	// count columns that are at least 3 boxes high
+					if (elem == 0b0111 || elem == 0b1111)
```
lynxlynxlynx: 0b1110, 0b1101,0b1011 are impossible?

src/game/TileEngine/Lighting.cc
```diff
@@ -295,102 +308,172 @@ static void LightInsertRayNode(LightTemplate* const t, const UINT16 usIndex, con
...
+	bool            illuminateNothing{};
+	bool            illuminateOrientedBlocksOnly{};      // Illuminate only oriented light-blocking structures (like walls and doors) in this tile
+	Orientation     allowedOrients{ ORIENT_NONE };
+	WorldDirections srcToDstDir{ DIRECTION_IRRELEVANT }; // Direction from source to destination tile
+	int32_t         baseGridNo{};                        // E.g. open door slab and its frame (base) can be in different tiles
+};
 
-	usTileNo=MAPROWCOLTOPOS(iY, iX);
-	usSrcTileNo=MAPROWCOLTOPOS(iSrcY, iSrcX);
+static void SetupFilter(INT16 iSrcX, INT16 iSrcY, INT16 iX, INT16 iY, IlluminationFilter& filter)
```
momoko-h: How about turning this into a constructor?

src/game/TileEngine/Lighting.cc
```diff
@@ -1454,104 +1426,80 @@ LightTemplate* LightCreateOmni(const UINT8 ubIntensity, const INT16 iRadius)
...
+		LIGHT_NODE* const pLight = &t->lights[usNodeIndex & ~LIGHT_BACKLIGHT];
+		const INT16 dstX = centerX + pLight->iDX;
+		const INT16 dstY = centerY + pLight->iDY;
 
-				LightAddTile(iOldX, iOldY, iX + pLight->iDX, iY + pLight->iDY, pLight->ubLight, uiFlags, fOnlyWalls);
+		IlluminationFilter filter{};
 
-				pLight->uiFlags|=LIGHT_NODE_DRAWN;
-			}
+		SetupFilter((INT16)srcX, (INT16)srcY, (INT16)dstX, (INT16)dstY, filter);
```
momoko-h: dstX and dstY are already INT16, so the casts aren't necessary. Or just take the parameters as INT32 and get rid of all casts here.

src/game/TileEngine/Lighting.cc
ShishkinP: Zeroes correspond to gaps in the columns. Example:
image
This wall's 5 columns are represented as (from left to right): 0b1111, 0b1111, 0b1111, 0b1001, 0b0001. Of course, in this case we don't need to count the columns because structs flagged as walls are automatically light-blocking. But generic structs only have a shape and the ubStructureHeight property to go off of. So enough tall gapless columns mean it realistically can block light.

src/game/TileEngine/Lighting.cc
ShishkinP: It does get more concise.

lynxlynxlynx:
All done?

ShishkinP:
Yes, it's ready.

lynxlynxlynx:
Cool,thanks!

icelynjennings:
Amazing!!!

momoko-h:
This introduced two new compiler warnings about unused but set variables. @ShishkinP, could you please fix this?

PR #2433, opened 2026-08-08
IMP: Do not send e-mail report based on upgraded merc states

icelynjennings:
First PR in a codebase I don't know well yet, please excuse my dust.

Might be future-save-incompatible?

Fixes #2376

### DIFF
--- a/src/game/Laptop/EMail.cc
+++ b/src/game/Laptop/EMail.cc
@@ -11,6 +11,7 @@
 #include "IMP_Portraits.h"
 #include "AIMMembers.h"
 #include "Random.h"
+#include "Overhead.h"
 #include "Text.h"
 #include "LaptopSave.h"
 #include "Finances.h"
@@ -25,7 +26,6 @@
 
 #include <string_theory/format>
 #include <string_theory/string>
-
 
 #define MAX_MESSAGES_PAGE 18 // max number of messages per page
 
@@ -1899,6 +1899,10 @@ static void HandleIMPCharProfileResultsMessage(void)
 	// list doesn't exist, reload
 
 	MERCPROFILESTRUCT const& imp = GetProfile(PLAYER_GENERATED_CHARACTER_ID + LaptopSaveInfo.iVoiceId);
+	INT8 const initialMarksmanship = LaptopSaveInfo.bIMPInitialMarksmanship != 0 ? LaptopSaveInfo.bIMPInitialMarksmanship : imp.bMarksmanship;
+	INT8 const initialMedical      = LaptopSaveInfo.bIMPInitialMedical      != 0 ? LaptopSaveInfo.bIMPInitialMedical      : imp.bMedical;
+	INT8 const initialMechanical   = LaptopSaveInfo.bIMPInitialMechanical   != 0 ? LaptopSaveInfo.bIMPInitialMechanical   : imp.bMechanical;
+	INT8 const initialExplosive    = LaptopSaveInfo.bIMPInitialExplosive    != 0 ? LaptopSaveInfo.bIMPInitialExplosive    : imp.bExplosive;
 
 	// load intro
 	iEndOfSection = IMP_RESULTS_INTRO_LENGTH;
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Yes, the gist of it is fine, but you should increase the savegame version constant. And it would be better to have a fallback to live values for older saves. But still that's not enough, new fields should be added at the end to avoid causing other misreads.

momoko-h:
Aren't you getting assertion failures in both `SaveLaptopInfoToSavedGame` and `LoadLaptopInfoFromSavedGame` because you've left the buffer size at 7440 bytes but you're now using 4 bytes more?

 Instead of adding more bytes, use 4 of the 87 free bytes at the end of the buffer. That's also more than enough to add some magic number to indicate the new layout of this data, then you would not have to increase the save game version, which avoids the warning the user gets about older versions.

lynxlynxlynx:
Momoko was right, there's no need to bump the save version, since there's already reserved space available. It's guaranteed to be all zeros, so adding a flag field as suggested is a clean way to know if the data is there or not. Summing the values to check against wouldn't work reliably, since the game actually allows you to create a merc with all four stats at 0 (the numbers work perfectly, all the rest get maxed).

src/game/Laptop/EMail.cc
```diff
@@ -11,6 +11,7 @@
 #include "IMP_Portraits.h"
 #include "AIMMembers.h"
 #include "Random.h"
+#include "Overhead.h"
```
icelynjennings: no longer needed

src/game/Laptop/EMail.cc
```diff
@@ -1899,6 +1899,10 @@ static void HandleIMPCharProfileResultsMessage(void)
 	// list doesn't exist, reload
 
 	MERCPROFILESTRUCT const& imp = GetProfile(PLAYER_GENERATED_CHARACTER_ID + LaptopSaveInfo.iVoiceId);
+	INT8 const initialMarksmanship = LaptopSaveInfo.bIMPInitialMarksmanship != 0 ? LaptopSaveInfo.bIMPInitialMarksmanship : imp.bMarksmanship;
+	INT8 const initialMedical      = LaptopSaveInfo.bIMPInitialMedical      != 0 ? LaptopSaveInfo.bIMPInitialMedical      : imp.bMedical;
+	INT8 const initialMechanical   = LaptopSaveInfo.bIMPInitialMechanical   != 0 ? LaptopSaveInfo.bIMPInitialMechanical   : imp.bMechanical;
+	INT8 const initialExplosive    = LaptopSaveInfo.bIMPInitialExplosive    != 0 ? LaptopSaveInfo.bIMPInitialExplosive    : imp.bExplosive;
```
lynxlynxlynx: I think you can just take each laptop value without checking. If anyone has a save just after starting a game and before creating an IMP, they can live with a shorter report (or restart), while we get cleaner code.

If you want to be 100% correct, then saving and using loadedImpInitialStatsPresenceMarker would be better.

src/game/Laptop/Laptop.cc
```diff
@@ -3390,7 +3397,31 @@ void LoadLaptopInfoFromSavedGame(HWFILE const f)
...
+		l.bIMPInitialMechanical   = impInitialMechanical;
+		l.bIMPInitialExplosive    = impInitialExplosive;
+	}
+	else
+	{
+		l.bIMPInitialMarksmanship = 0;
+		l.bIMPInitialMedical      = 0;
+		l.bIMPInitialMechanical   = 0;
+		l.bIMPInitialExplosive    = 0;
+	}
```
lynxlynxlynx: There's no need for the conditional. In old saves the read values will be 0 any way.

src/game/Laptop/Laptop.cc
```diff
@@ -3243,6 +3243,7 @@ static void ExtractStoreInvetory(DataReader& d, STORE_INVENTORY& i)
 void SaveLaptopInfoToSavedGame(HWFILE const f)
 {
 	LaptopSaveInfoStruct const& l = LaptopSaveInfo;
+	UINT32 const impInitialStatsPresenceMarker = 0x49535454; // "ISTT"
```
lynxlynxlynx: Why waste a full dword, when a byte is enough?

icelynjennings:
Unrelated to this PR, but I also wondered on a completely different approach if you prefer (for which I'd make a new branch), where we just track every merc's "initial stats" as new values in the merc profile, rather than just 4 new "IMP variables" in the save file.

Granted, this would mean that every single merc would track these values in the save, despite the fact that this is perhaps the only context where "initial stats" matter.

However, it may be consider the "cleaner" approach altogether, to some. And perhaps it would have some future uses, although it may be considered premature optimization too.

I defer to you :)

src/game/Laptop/Laptop.cc
lynxlynxlynx: You can simply INJ_I8(d, 1) as well, obviating the need for a variable.

lynxlynxlynx:
It can be generalized/moved easily if a need arises, until then I would classify that approach as inefficient design.

But ... actually, don't we already have these stored for everyone? How does the stat gain display work otherwise?

icelynjennings:
These could be the ones already available for the IMP merc:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/96be02da4786812e020057939b3e6fc8186b27c3/src/game/Laptop/CharProfile.h#L30-L33

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/96be02da4786812e020057939b3e6fc8186b27c3/src/game/Laptop/IMP_Attribute_Selection.cc#L54-L63

I couldn't find something similar for all general mercs though

lynxlynxlynx:
It has to be somewhere else, since all mercs display the difference.
https://lparchive.org/Jagged-Alliance-2/Update%206.5/2-stat1_2.png

lynxlynxlynx:
Look at DisplayCharStats in src/game/Laptop/Personnel.cc

icelynjennings:
Thank you! Changed my approach.

Would injuries affect the delta?

lynxlynxlynx:
Yes, so it still all works out. :)

lynxlynxlynx:
Thanks!

icelynjennings:
Thank you both! Awesome trip across the codebase :)

PR #2434, opened 2026-08-10
Colour named civilians differently on radar

### LINKED ISSUE #2215: Color named NPCs differently on the mini maps
Perhaps blue instead of white, so they're easier to find.
And we could also display all the (non-default) names on a keypress.

icelynjennings:
Addresses #2215

Still new here, so assume "incorrect" until proven otherwise.


icelynjennings:
Btw, any way I can `[skip ci]` in the commit message? Don't wanna use up the CI runner credits.

lynxlynxlynx:
LGTM

[ci skip] should work, [skip ci] likely too. I think most of the runners expect it on the first line, but I don't remember if it has to be at the start.

Issue #1953, opened 2024-01-13
If you give to Carmen three heads at once, he will bring you only 20000$.


Inveriss:
[//]: # (Thank you for taking the time to report an issue or propose an enhancement. Questions are also welcome.)
[//]: # (If you have more than one item to report, please open separate issues for each one!)

[//]: # (This text will not be visible and you can freely delete it. Check the Preview tab for a cleaner look.)

Hello

Everything is ok when you give him one head or two heads. But he will cheat you with three heads. It could be the money limit for one slot, which is this 20000$.

[//]: # (For bugs, please fill in this table [otherwise delete it]. The lines don't have to stay aligned.)

| JA2S version | Game version        | Vanilla bug? | Used to work? |
| ------------ | ------------------- | ------------ | ------------- |
|  0.21.0 | JA2 Gold 1.12 ENG  |    YES    |    NO     |

[//]: # (Please also include:)
[//]: # (- how to reproduce the bug and, if possible, a save game)
[//]: # (- the operating system)
[//]: # (- in case you know, for enhancements, if this feature was already added by some other source mod like 1.13)

[//]: # (When pasting backtraces or any other logs with hashed enumerations, please format it as a code block.)
[//]: # (You can do that by selecting it and clicking the "<>" icon or manually putting three backticks [```] at each end.)

icelynjennings:
It seems `MAX_MONEY_PER_SLOT` might not be the final inventory max value.

The final max value might be calculated through a subsequent fraction:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/d8f78112655e503c968a793e5d2278ea436305f7/src/game/Tactical/Items.cc#L292-L302

Not sure why this additional "getter" pattern exists here.

lynxlynxlynx:
Probably realism, since money has volume/weight too.

icelynjennings:
Makes sense. 

There are 6 possible heads in the default campaign's pool, so increasing `MAX_MONEY_PER_SLOT` to ~~$60000~~ could be the quickest, if OP is correct about how the bug works.

Actually, there seems to be a hard limit of 3 heads per game, so the theoretical maximum should be **$30000**:

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/d8f78112655e503c968a793e5d2278ea436305f7/src/game/Strategic/Quests.h#L87-L98

PR #2435, opened 2026-08-11
fix: increase MAX_MONEY_PER_SLOT to 30000 for Carmen bounty hunter quest

icelynjennings:
Attempts to address #1953

This PR assumes that the "intended spirit" of original devs was to permit returning 3 heads at once, along with equivalent reward.

### DIFF
--- a/src/game/Tactical/Item_Types.h
+++ b/src/game/Tactical/Item_Types.h
@@ -37,7 +37,7 @@ enum ItemCursor
 
 #define MAX_OBJECTS_PER_SLOT		8
 #define MAX_ATTACHMENTS		4
-#define MAX_MONEY_PER_SLOT		20000
+#define MAX_MONEY_PER_SLOT		30000
 
 enum DetonatorType
 {


icelynjennings:
One problem I foresee is if either the player or Carmen only has "small" inventory slots available in the inventory, then it may still not be enough ($15000)

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/d8f78112655e503c968a793e5d2278ea436305f7/src/game/Tactical/Items.cc#L292-L302

The exact same issue would theoretically be present in vanilla.

Maybe the small slot money limit should be equivalently increased?

ShishkinP:
Better solution would be to just put the money on the ground like it's done with other items if there's no free slots.

ShishkinP:
By the way dropping it on the ground is exactly how it happens when Mary attempts to give her vacation money as a reward at the Airport sector and there is no receiving char on the tile. Handled by `VerifyGiveItem()` in `src/game/Tactical/Handle_Items.cc`.

icelynjennings:
Thanks! 

However, I think #1953 implies the problem may not be necessarily with a merc running out of slots for the entire $30000 reward. In fact I assume the money wad from Carmen would already succeed being put on the ground.

If Carmen does hand the merc a $20000 wad of money successfully (otherwise we wouldn't get at least the $20000 out of desired $30000), then I think one or more of the following could be co-occurring:

### A) Carmen is only handed one wad of cash by the game after going to Drassen, which immediately hits a $20000 limit. He then successfully hands his one wad to the merc.

We could:

- Ensure the game knows to give Carmen several wads of cash. 
- Ensure NPCs know how to hand several wads of cash, until the full debt is paid.
- Or increase the money limit per wad of cash. Presumably Carmen will then be given a singular $30000 wad by the game, and hand that one wad to the merc successfully.

### B) Carmen is handed several wads of cash by the game (e.g. $20000 + $10000 in separate wads), but he only knows to give the player one wad of cash.

We could:

- Make NPCs know how to hand the exact amount of cash, even across several wads.

lynxlynxlynx:
Have you tried any of the proposals? My first thought was to change CreateMoney to take the limit into account and create more items (CreateItem vs CreateItems) if needed (a stack, with a, say, 1/2 of the max as the limit, so 30k goes into it nicely). The code is there, I just don't know if it works out for money in the end.

PR #2439, opened 2026-08-14
Stop suppression AP debt from being refunded

steph-pg:
This one touches 3 things, all related to suppression fire. The most important one is in src/game/Tactical/Soldier_Control.cc. What's happening during suppression fire is that when soldier's ubAPsLostToSuppression reaches MAX_APS_SUPPRESSED (which is 8) the soldier becomes immune to any suppression because we never reset it at the start of the turn. 

Another thing is that suppression fire can make APs a negative (which is the whole point of suppression fire) but if soldier is hit when their AP is negative the ap goes up instead of going even lower.

Third thing is in HandleSuppressionFire. When we reach exactly 1 ubPointLost and the current stance is crouched instead of losing 2 points we actually get 1 instead.

If someone for some reason prefers a vanilla no suppression behaviour we can now do it by adjusting suppression_fire_modifier game policy


lynxlynxlynx:
Nice, thanks!

PR #2440, opened 2026-08-14
Do not automatically open a locked door when enemies are in a sector

steph-pg:
Now we can open unlocked doors diagonally, however the doors are still very deadly especially when you play on dead is dead because a locked door cannot be unlocked diagonally (which makes sense), and unlocking a door automatically opens it. This change allows to pick lock and use key without automatically open the door. It only works when there are enemies in sector and diagonally_interactable_doors game policy is turned on (I don't want to create a separate policy just for this and overwhelm users)


lynxlynxlynx:
Looks fine, good idea and especially if lockpicking also causes noise, it could go in without the flag check as far as I'm concerned.

PR #2441, opened 2026-08-15
Only block interrupts while shots are still coming in

steph-pg:
so underfire can have 3 values, but we only interested in two: 2 means a soldier is shot during this exact turn and 1 means soldier was shot last turn. At the moment when a soldier is underfire they cannot interrupt. This has several side effects for ai. First, when ai soldier decides to watch (wait for interrupt) they never check underfire which means this decision is simply useless. Second, when player 'abuse' a tactic of crouch->shoot->prone (behind an obstacle that blocks los) the only counter that ai has is actually interrupt (which never happens because they have underfire=1). And btw the shot can be absolutely horrible and not even close to target but underfire will be set. Interestingly being blown with a grenade is not considered a big deal and allows to interrupt in this specific scenario.

This change will make ai a bit stronger, and I think it will also benefit player as currently it's really not obvious that you can't interrupt if someone shot at you during the last turn.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -66,6 +66,13 @@
   //  (say your buddies took care of it), you'd still get skipped for any further interrupts. Annoying in ambushes or when sneaking.
   //  vanilla setting: false
   "multiple_interrupts": false,
+  
+  //  Allow interrupts after being shot at on an earlier turn. In vanilla, being under fire blocks a
+  //  soldier's interrupts until their own turn begins, so an attack silences them for the whole next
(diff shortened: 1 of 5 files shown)


lynxlynxlynx:
This feels more like a tweak than a fix, so I'd make it conditional on a key in the ai namespace of game.json.

steph-pg:
Sure, but can I make it true by default, then? :)

steph-pg:
Added a game policy but I put it next to existing interrupt tweak not ai section, as it affects both players and ai

lynxlynxlynx:
Thanks

PR #2442, opened 2026-08-16
Clarify and update some comments in game.json

steph-pg:
Let's add extra explanation to some of the options

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -68,11 +68,11 @@
   "multiple_interrupts": false,
 
   // Boosting stats like agility, dexterity, exp. level and health absurdly increases the action points cost to shoot.
-  // With this option on: the better the stats the fewer APs will be needed.
+  // With this option on: the better the stats the fewer APs will be needed. Note that you also won't get AP discount when injured.
   // vanilla setting: false
   "fixed_cost_to_shoot": false,
 
-  //  Minimal status in % for each gun in the hand of enemy soldiers. Note that this value has no influence on enemy behaviour (CtH, jams).
+  //  Minimal status in % for each gun in the hand of enemy soldiers. Note that this value has no influence on enemy behaviour like CTH.
   //  vanilla value: 0
   "enemy_weapon_minimal_status": 0,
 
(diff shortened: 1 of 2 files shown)


assets/externalized/game.json
```diff
@@ -84,14 +84,17 @@
...
-    //  Aiming: will choose other body parts when appropriate and random choice was selected.
-    //  Prone: when already facing its focus, lay down if nothing else.
-    //  vanilla values: false, false, 0 ,0
+    //  In vanilla under normal conditions enemies measure cth agains target's torso and then a random location
+    //  is selected during the hit. The chances are: head 5%; torso 80%; legs 15%;
+    //  With this setting on when possible enemies select target location when threshold cth is reached against that
+    //  body part, with head being higher priority.
+    //  vanilla values: false, N/A, N/A
     "better_aiming_choice": false,
-    "go_prone_more_often": false,
```
selaux, line 92: This is not a comment change.

selaux (commented):
Can you do those adaptions to the schema descriptions as well? https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/stracciatella/src/schemas/yaml/game.schema.yaml

assets/externalized/game.json
steph-pg: it is still there, I just moved comments around a bit

steph-pg:
> Can you do those adaptions to the schema descriptions as well? https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/rust/stracciatella/src/schemas/yaml/game.schema.yaml

done

assets/externalized/game.json
selaux: True, I missed that.

selaux:
Thanks!

assets/externalized/game.json
```diff
@@ -68,11 +68,11 @@
   "multiple_interrupts": false,
 
   // Boosting stats like agility, dexterity, exp. level and health absurdly increases the action points cost to shoot.
-  // With this option on: the better the stats the fewer APs will be needed.
+  // With this option on: the better the stats the fewer APs will be needed. Note that you also won't get AP discount when injured.
```
lynxlynxlynx, line 71: Wouldn't you consider this a bug? The change was made to be fairer, not to disable other mechanics.

steph-pg: Well, this option does exactly what it says. So probably can't be considered a bug. But if we wanted it to only limit AP in one direction and keep a discount then we should have somehow adjust the formula and option should have a different name.

lynxlynxlynx: I didn't read #2356 that way. It just fixed the fact that bettering yourself shouldn't make your chances worse. If getting shot doesn't impose an AP penalty any more, that's a big change. @ShishkinP what's your view?

ShishkinP: >getting shot doesn't impose an AP penalty

Confirmed. Current description by [steph-pg](https://github.com/steph-pg) is correct. I'll have to investigate when I'm done with other updates.

PR #2443, opened 2026-08-20
Add leather jacket if Nails hired without equipment

### LINKED ISSUE #1987: Nails and his leather jacket should be inseparable
### What is wrong

Nails don't want to be separated from his vest. Yet, when you hire him without equipment, he starts with no armor. When he then picks one up, it is treated as his leather jacket, which means you can't take it off.

A solution could be to make it so, that even when hired without equipment, he starts with a leather jacket. The same applies to when he is fired, or his contract expires, he should take his current leather jacket with him, so when he is rehired he still has it (or he gains a new one when he is back at A.I.M.).

Another possibility is to treat his jacket with C18 and Royal Jelly, which would result in destroying the vest and he can wear any other armor vest. This would probably result in the same bug (not tested), that this new armor would be treated as his unremovable jacket.

A solution could be, that a dialog window pops up, saying something like "This would destroy Nails leather jacket. Nails will not allow this." and aborts that action.

|  0.19 | 1.12 GER  |    yes   |    no     |

icelynjennings:
Attempts to address #1987

### DIFF
--- a/src/game/Laptop/AIMMembers.cc
+++ b/src/game/Laptop/AIMMembers.cc
@@ -1225,6 +1225,11 @@ static INT8 AimMemberHireMerc(void)
 		if (!s) return FALSE;
 		s->bTypeOfLastContract = contract_type;
 
+		if (!gfBuyEquipment && pid == NAILS)
+		{
+			CreateItem(LEATHER_JACKET, 100, &s->inv[VESTPOS]);
+		}
+
 		MERCPROFILESTRUCT& p = GetProfile(pid);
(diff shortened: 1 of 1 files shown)


icelynjennings:
This presumes that Nails is canonically dedicated to biker fashion enough as to either re-purchase a brand-new leather jacket (if it was destroyed on the job), or repair his existing one.

lynxlynxlynx:
I think that's a good enough compromise, thanks. If people want to break the vest intentionally, let them, it's more mod territory if anyone wanted an extra reaction.

PR #2444, opened 2026-08-21
Personnel: do not constantly reload two STI files

momoko-h:
The files needed to render the ATM panel and the highlight border around the mercs face were frequently reloaded for no good reason. Both now use the object cache.


lynxlynxlynx:
Thanks!

PR #2445, opened 2026-08-22
Remove InitializeVideoManager

momoko-h:
If this function does not exist we do not have to assert that it was called at the right time. All it does is set the head of a linked list that was already NULL to NULL again.

Also move SGPVObject::Flags down a bit to reduce padding.


lynxlynxlynx:
Thanks!

PR #2448, opened 2026-08-22
Fix light flares that land on an occupied tile

### LINKED ISSUE #2216: When targeting break lights on mercs, they rarely go off
Not sure if this is intended or perhaps I have been just very unlucky, but most of the time when you throw it at someone else, it doesn't light. In one combat 1/6 did, which seems very unlikely. Needs investigating.

steph-pg:
A thrown flare stops on top of whatever it hits, so aiming at a soldier's gridno left it resting at shoulder height. The light effect was skipped for any Position.z above 0 and the flare dropped as an item instead.

### DIFF
--- a/src/game/TileEngine/Physics.cc
+++ b/src/game/TileEngine/Physics.cc
@@ -2138,10 +2138,8 @@ static void HandleArmedObjectImpact(REAL_OBJECT* pObject)
 
 		if (lightEffect)
 		{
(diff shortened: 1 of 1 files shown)


lynxlynxlynx:
Sounds like this fixes #2216 as well.

ShishkinP:
>fixes #2216 as well.

Tested it and it does.

PR #2449, opened 2026-08-22
Rate a blast against the target, not the thrower

steph-pg:
Probably those are typos. More damage, if target is standing and calculate damage based on target's armour not own armour.


lynxlynxlynx:
Nice find!

PR #2453, opened 2026-08-27
Fix illumination update regression

ShishkinP:
Fixes overlapping lights substruction broken by #2432.

image


lynxlynxlynx:
Thanks!

PR #2457, opened 2026-08-28
Fix militia visual offset

steph-pg:
Externalizing town data introduced a bug where militia presentation is broken. So far I've only seen it in Balime:
Screenshot From 2026-08-28 21-51-45


lynxlynxlynx:
LGTM!

PR #2458, opened 2026-08-28
Use the enemy's own leader in autoresolve

steph-pg:
Probably another typo. During autoresolve both sides select a soldier with highest leadership, and their leadership is applied to both attack and defense of each and every member of the team. Player's  leadership bonus was applied to both teams which is a bug.


lynxlynxlynx:
Good find, thanks!

momoko-h:
Yes, looks like this was copied & pasted from the gpCivs loop above and then they missed this spot.

Issue #2454, opened 2026-08-28
Question: How open is this project to minor QoL additions?


rklasen:
Hi all,

first of thank you so much for this project! I've been playing it since I was a teenager and seeing it re being resurrected here is just a flash of nostalgia that I really love. I've been able to play this game just fine and since the source is available I've also made some minor changes to my personal copy of the game which I think make for nice quality of life additions.

How open are you to incorporating these changes into the source code? Because I'd love to open a few merge requests, if they are welcome. **Note/Disclaimer: the code is mostly written by claude code.** I've changed these things (I think some inventory management on the strategic screen are the most significant ones. Think auto-remove attachments from guns, sort item, merge consumable items):

_Gameplay_
Sector inventory (mapscreen) — [Map_Screen_Interface_Map_Inventory.cc](vscode-webview://1j9ts47hhaeqpcr1cj1nd37njfc11kt7fp0crnd00a6hjne0rifk/src/game/Strategic/Map_Screen_Interface_Map_Inventory.cc), +506 lines

SHIFT+S — merges stackable slots into full stacks and re-sorts. Later extended to strip items first: guns give up their loaded rounds as a magazine, attachments come off as separate items (recursively), and part-used magazines/kits/medkits are poured together so at most one partial remains per item. Armed bombs, trapped items and ITEM_INSEPARABLE attachments are left alone.
SHIFT+R — tops up every copy of the pointed-at item carried by mercs in that sector, spending the stash stack first. Works for magazines, kits and medkits.
Both blocked during battle and while an item is on the cursor.
Shift-reload magazine split
Holding shift while dropping an oversized magazine on a gun repacks the leftover rounds into magazines the gun actually takes (matching calibre, size and ammo type) and auto-places them in the merc's pockets. Hangs off PlaceObject, so it works in mapscreen and tactical inventory alike.

_Auto-assign squads by sector_
New line above squad 1 in the on-duty squad menu: sorts all selected mercs into squads by the sector they're standing in — sharing a sector means sharing a squad, each other sector gets its own squad, half-empty squads in one sector get merged. Greys out when nothing can move. Adds a 22nd pSquadMenuStrings entry, translated into all nine languages.

_Suppression AP debt capped to one turn_
CalcNewActionPoints now settles suppression debt within the turn it lands in and writes off the rest, instead of letting a deep hole swallow several turns. Also fixes DeductPoints so reflexes during the enemy turn can empty a merc but not dig them further negative.

[...]

PR #2459, opened 2026-08-29
Feature: sector inventory sort, attachment removal and consumable merge

rklasen:
As per #2454, I submit for your review this QoL feature that I've come to love: automatic item sorting in the strategic/sector item view, and: auto refill consumables carried by mercs (two separate hotkeys).

You can now:

press shift-s when the sector item window is open. all items are sorted (guns first), all attachments and ammo from guns are removed, and all consumables (medkits, camo kits, flasks) are merged until at most one partially filled remains. all others are 100% filled. same for ammo.

press shift-r when hovering over a consumable in the item view (e.g. a first aid kit) to refill all of this item carried by mercs. for example, I always have all my mercs carry first aid kits, and after a battle, action number one for me is always "get a first aid kit from the sector and refill all carried first aid kit". well, manual no more!

both hotkeys are gated behind the additional hotkeys game.json setting.

disclaimer: the actual code is written by claude code. I have playtested the changes.

below this line you can find an AI-generated summary:

---

Two hotkeys for the sector inventory panel, both behind `extra_hotkeys`.

### SHIFT+S — take apart, stack and sort

Breaks every item down, pours what can be pooled together, packs the rest into full
stacks and re-sorts the list.

Loaded guns give up their rounds as a magazine of their own, attachments come off and
become items in their own right, and part-used magazines, kits and medkits have their
points poured together so that what is left is full objects and at most one part-used
one per item. Extracted items are walked again, so an attachment that carried an
attachment of its own — a grenade in a detached underslung launcher — comes apart too.
They keep the gridno, level and reachability of the item they came out of, so nothing
becomes reachable that was not before.

Slots are only merged when nothing would be lost by it: same item and level, same
reachability, no attachments and no traps. Keys are left alone because they keep their
lock id in the same bytes as the status array, and money is merged by amount instead,
capped at `MAX_MONEY_PER_SLOT`.

Armed bombs and trapped items are left whole; taking those apart is a job for a merc,
not for a sorting pass. Attachments flagged `ITEM_INSEPARABLE` stay where they are.
Guns only give up their rounds when the slot holds a single gun, because the ammo
fields share their bytes with the status of the second and third object of a stack.
Empty magazines and spent kits are absorbed by the point merge, the same thing
`CleanUpStack` already does in a merc's pockets.

### SHIFT+R — refill from the stash

Tops up every copy of the pointed at item that the mercs in this sector carry, spending
the pointed at stack first and then the other reachable stacks of the same item. It
works for everything the engine tracks points for — magazines refill to their calibre's
capacity, kits and medkits to 100 — and empties stash objects one at a time, so what is
left behind is a few full objects rather than many nearly empty ones.

### Both

Blocked during a battle in that sector and while an item is in the cursor, and both only
touch mercs actually standing in the displayed sector, the same rule the panel already
enforces for dragging items in and out.

Both ride on `extra_hotkeys` through a new `HKMOD_SHIFT` entry in the `UI_Map` section
of `isHotkeyEnabled()`, alongside the existing CTRL+I. The messages they report go into
a new `pMapInventoryActionStrings` array so they are translated like the rest of the
panel; a missing key throws at startup, so all nine translation files carry it.

### Notes for review

The translations are filled in for all nine languages. English and German follow the
terminology already used in the panel; the other seven are best-effort and would
benefit from a native speaker's eye.

Built and tested on macOS with `.ci/ci-build.sh` — the unit suite passes — and played
in a German-resource game. Playtesting turned up one bug in the point merge, fixed
here: it asked whether pooling could reduce the object count rather than whether the
points were already packed, so a group that could not lose an object was left untouched
and two part-used doctor's bags stayed two part-used doctor's bags. Repacking now
happens whenever a group holds more than one part-used object, which also covers stacks,
where the same case could leave every magazine in the group part-used.


rklasen:
really, no takers? Maybe a video would illustrate the usefulness a little better. Ive used both new shortcuts literally after every fight.

This is the removal of all ammunition and weapon attachments in action (shift+s for sort):

https://github.com/user-attachments/assets/96daf9e4-7f1a-4b7c-9428-3a0e995e4315

And this is the refill of all merc-carried consumables (shift+r for refill):

https://github.com/user-attachments/assets/2c933864-203c-4d8a-8680-9e68ebe78d10

The medkit and tool box in this video are used to refill _all_ medkits and toolboxes of _all_ mercs, naturally. Saves me a solid minute of tedium after each fight.

lynxlynxlynx:
Review just takes time, that's all.

rklasen:
Sorry, didn't mean to be pushy.

lynxlynxlynx:
Your first video shows something that doesn't match the description — the med kits didn't get merged.

src/game/Strategic/Map_Screen_Interface_Map_Inventory.cc
```diff
@@ -930,6 +936,319 @@ void AutoPlaceObjectInInventoryStash(OBJECTTYPE* pItemPtr)
...
+// Rebuild the pool list from the given items, padding it back out to whole pages.
+static void ReplaceMapInventoryPool(const std::vector<WORLDITEM>& items)
+{
+	pInventoryPoolList = items;
+
+	size_t const visible_slots = pInventoryPoolList.size();
+	size_t const empty_slots   = MAP_INVENTORY_POOL_SLOT_COUNT - visible_slots % MAP_INVENTORY_POOL_SLOT_COUNT;
+	pInventoryPoolList.resize(visible_slots + empty_slots, WORLDITEM{});
+
+	iLastInventoryPoolPage = static_cast<INT32>((pInventoryPoolList.size() - 1) / MAP_INVENTORY_POOL_SLOT_COUNT);
```
lynxlynxlynx, line 948: Why the -1, won't it be wrong if there's only 1 item on the last page?

src/game/Strategic/Map_Screen_Interface_Map_Inventory.cc
```diff
@@ -930,6 +936,319 @@ void AutoPlaceObjectInInventoryStash(OBJECTTYPE* pItemPtr)
...
+
+
+// Can the source slot be poured into the target slot without losing anything?
+static BOOLEAN CanMergeSectorInventorySlots(const WORLDITEM& target, const WORLDITEM& source)
+{
+	if (target.o.usItem != source.o.usItem) return FALSE;
+	if (target.ubLevel  != source.ubLevel)  return FALSE;
+
+	// unreachable items must not become reachable by being piled onto a reachable stack, and vice versa
+	if ((target.usFlags & WORLD_ITEM_REACHABLE) != (source.usFlags & WORLD_ITEM_REACHABLE)) return FALSE;
```
lynxlynxlynx, line 970: Why not catch also the case when both items are unreachable?

src/game/Strategic/Map_Screen_Interface_Map_Inventory.cc
```diff
@@ -930,6 +936,319 @@ void AutoPlaceObjectInInventoryStash(OBJECTTYPE* pItemPtr)
...
+	{
+		if (wi.o.ubNumberOfObjects > 0) items.push_back(wi);
+	}
+
+	UINT32 uiSlotsMerged = 0;
+
+	for (size_t iTarget = 0; iTarget < items.size(); ++iTarget)
+	{
+		OBJECTTYPE& target = items[iTarget].o;
+		if (target.ubNumberOfObjects == 0) continue;
```
lynxlynxlynx, line 1009: redundant check
EDIT: nevermind, I see it's due to the inline deletion.

src/game/Strategic/Map_Screen_Interface_Map_Inventory.cc
```diff
@@ -930,6 +936,319 @@ void AutoPlaceObjectInInventoryStash(OBJECTTYPE* pItemPtr)
...
+		const ItemModel* const item         = GCM->getItem(target.usItem);
+		BOOLEAN          const fMoney       = item->isMoney();
+		UINT8            const ubStackLimit = SectorInventoryStackLimit(item);
+
+		if (!fMoney && ubStackLimit < 2) continue;
+
+		for (size_t iSource = iTarget + 1; iSource < items.size(); ++iSource)
+		{
+			OBJECTTYPE& source = items[iSource].o;
+			if (source.ubNumberOfObjects == 0) continue;
```
lynxlynxlynx, line 1020: same here

src/game/Strategic/Map_Screen_Interface_Map_Inventory.cc
```diff
@@ -930,6 +936,319 @@ void AutoPlaceObjectInInventoryStash(OBJECTTYPE* pItemPtr)
...
+
+
+// Magazines, kits and medkits hold points (rounds resp. kit charges) that can be moved between
+// objects of the same item.  Returns how many points one such object holds when it is full.
+static BOOLEAN GetRefillablePointCapacity(const ItemModel* const item, INT8& bMaxPoints)
+{
+	if (item->isAmmo())
+	{
+		// the round count shares its byte with the signed status value
+		if (item->asAmmo()->capacity > 127) return FALSE;
```
lynxlynxlynx, line 1074: Where is this from, I don't see any suspicious usage?

src/game/Strategic/Map_Screen_Interface_Map_Inventory.cc
```diff
@@ -980,6 +980,173 @@ static BOOLEAN CanMergeSectorInventorySlots(const WORLDITEM& target, const WORLD
...
+}
+
+
+static BOOLEAN GetRefillablePointCapacity(const ItemModel* item, INT8& bMaxPoints);
+
+
+// Magazines, kits and medkits hold points - rounds resp. charges - that can be moved between
+// objects of the same item.  Pool them per item so that what is left is full objects and at most
+// one part-used one.  Returns how many objects were emptied out this way.
+static UINT32 MergeRefillableSectorInventory(std::vector<WORLDITEM>& items)
```
lynxlynxlynx, line 1069: This looks like it duplicates the logic in StackAndSortMapInventoryPool. Why have two implementations?

PR #2460, opened 2026-08-29
allow up to six IMPs

### LINKED ISSUE #1317: Option to disable IMP
It would be nice to add to game.json ability to disable IMP in the game (for custom mods that want to remove that option for whatever the reason). I know that you can just ignore it but "clean" solution would be nice. This is enhancement thing, low priority of course esp. to the other stuff that is critical.

### LINKED ISSUE #1516: Add option to support multiple IMP mercs
This was requested on Discord. Similarily to squad size we should add an option `max_imp_mercs` to let the player create multiple imp mercs. A few hints:

- changing the condition to allow multiple IMP creation
- avoiding the same profile pic and voices re-used
- ensure report is sent for multiple mercs
- ensure that multiple mercs are saved when saving the game, so they can be loaded via the option `load_saved_merc_by_nickname`

### LINKED ISSUE #1593: IMP Portrait and Voice options
Could you maybe externalize the options which portraits and voices to choose from on IMP, so you could add the 4th voice from UB and the new portraits from WF?

rklasen:
As per #2454 (and taking #1516 into account), I submit for your consideration this PR that adds the option to create up to six IMPs. why six? why not over 9000? because there are only six voices, three male, three female, and each occupy one merc slot. to anything up to six IMPs is simple, more would be rather... hard.

this feature is gated: the max num of custom characters mus be set to 6 in the game.json

faces that were chosen for an IMP become unavailable for subsequent IMPs. same for voices. once three male IMPs are created, new IMPs can only be female (and vice versa) until all six IMPs are created, then the xep624 code is simply no longer accepted.

disclaimer: the actual code is written by claude code. I have playtested the changes.

below this line you can find an AI-generated summary:

---

Closes #1516.

Adds an `imp.max_characters` policy that lets a campaign hold more than one
self-made merc. It defaults to `1`, so nothing changes unless the option is set.

## Why six is the ceiling

The I.M.P. site was gated by a single `fIMPCompletedFlag`, but the real limit
behind that flag is the profile slots. Player generated characters occupy one
block of male voice slots followed by one block of female ones, indexed by
`PLAYER_GENERATED_CHARACTER_ID + iVoiceId`, and a finished character keeps its
slot for the rest of the campaign. There are three voices per gender, so the
hard ceiling is three men and three women. `imp.max_characters` accepts 1 to 6
and clamps anything else with a warning.

## What the change does

`fIMPCompletedFlag` is replaced by a `ubIMPCreatedSlots` bitmask recording which
slots are spoken for. Everything that used to ask whether an I.M.P. exists now
asks whether another one may be created. The flag itself is kept as "at least
one has been made", which is all the laptop's nag e-mail uses it for.

The profiling screens will not hand out anything an earlier character already
owns:

- the voice and portrait arrows step over taken entries, and both screens open
  on a free one instead of on whatever the previous character left behind
- a gender whose three slots are gone cannot be picked, and when only one gender
  still has room it is preselected, so the player is not left clicking a dead box
- once every slot is filled the activation code is no longer accepted

Portraits are not tracked separately. There are eight per gender against three
slots, and the face each created character ended up with already says which ones
are gone, so `IsIMPPortraitTaken` reads them back out of `gMercProfiles`.

Restoring a merc by nickname writes straight into its old slot, so the saved
voice id is now read up front, without disturbing `gMercProfiles`, and the
restore is rejected if that slot is occupied. The same path also clears
`fLoadingCharacterForPreviousImpProfile` on entry, which otherwise leaked from an
earlier import and robbed the next I.M.P. of its starting gear.

## The results e-mail

`HandleIMPCharProfileResultsMessage` rendered from whichever character
`iVoiceId` and `iPortraitNumber` happened to point at when the mail was opened,
so with more than one I.M.P. every report described the newest merc. The mail now
carries the voice id of the character it reports on, handed to it through the
strategic event that delivers it two days later, and takes the portrait from that
character's profile. Mails saved before this carry a zero, which is only treated
as a voice id when an I.M.P. actually holds that slot.

## Save compatibility

The mask goes into one of the bytes the laptop save block was already skipping,
so the save format is unchanged in size. What that byte holds in an older save
cannot be trusted, so it is only believed when it agrees with `iVoiceId`, which
back when one I.M.P. was the limit pointed at that one character. Anything else
is reconstructed from `iVoiceId` alone.

## Testing

- `.ci/ci-build.sh` green, 132/132 unit tests pass
- clean start with `resversion` GERMAN at `max_characters: 6`; the clamp warning
  confirmed firing at `9`
- played: created several I.M.P.s in one campaign, checked that taken voices and
  portraits are skipped, that a full gender cannot be selected, that the code is
  refused once all six exist, and that each report e-mail describes its own merc

## One open question

The issue suggests naming the option `max_imp_mercs`. This PR puts it at
`imp.max_characters` instead, since `game.json` already has an `imp` block that
holds the other I.M.P. settings. Happy to rename it if you would rather it match
the issue.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -189,9 +189,11 @@
     "bonus_attribute_points": 40,
 
     //  Number of I.M.P. characters that can be created over a campaign.
-    //  Each character keeps one of the merc profiles marked as an I.M.P. slot in
-    //  mercs-profile-info.json for the rest of the campaign, so the ceiling is how
-    //  many of those the data declares. Voices and portraits may be reused freely.
+    //  0 switches the I.M.P. site off entirely.
+    //  A character holds one of the merc profiles that mercs-profile-info.json
+    //  marks as an I.M.P. slot for the rest of the campaign, so that is the upper
+    //  end: eleven as the game ships, more if the data declares more. Voices and
+    //  portraits may be reused freely and do not limit it.
     //  vanilla setting: 1
     "max_characters": 1,
 
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -188,10 +188,10 @@
     "zero_attribute_points_bonus": 15,
     "bonus_attribute_points": 40,
 
-    //  Number of I.M.P. characters that can be created over a campaign, 1 to 6.
-    //  Each character keeps one of the player generated merc profile slots, of which
-    //  there are three per gender, so at most three men and three women. Voices and
-    //  portraits already in use are skipped when profiling a further character.
+    //  Number of I.M.P. characters that can be created over a campaign.
+    //  Each character keeps one of the merc profiles marked as an I.M.P. slot in
+    //  mercs-profile-info.json for the rest of the campaign, so the ceiling is how
+    //  many of those the data declares. Voices and portraits may be reused freely.
     //  vanilla setting: 1
     "max_characters": 1,
 
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -188,6 +188,13 @@
     "zero_attribute_points_bonus": 15,
     "bonus_attribute_points": 40,
 
+    //  Number of I.M.P. characters that can be created over a campaign, 1 to 6.
+    //  Each character keeps one of the player generated merc profile slots, of which
+    //  there are three per gender, so at most three men and three women. Voices and
+    //  portraits already in use are skipped when profiling a further character.
+    //  vanilla setting: 1
+    "max_characters": 1,
+
     // Replace the IMP personality quiz with a form to select skills directly
     // NOTE: does not set any personality
     // vanilla setting: false
--- a/src/externalized/policy/DefaultGamePolicy.cc
+++ b/src/externalized/policy/DefaultGamePolicy.cc
@@ -1,5 +1,8 @@
 #include "DefaultGamePolicy.h"
+#include "Logger.h"
 #include "Types.h"
+
+#include <algorithm>
 
 DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 {
@@ -72,6 +75,16 @@ DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 	imp_attribute_zero_bonus = imp.getOptionalInt("zero_attribute_points_bonus", 15);
 	imp_attribute_bonus = imp.getOptionalInt("bonus_attribute_points", 40);
 	imp_pick_skills_directly = imp.getOptionalBool("pick_skills_directly");
+
+	// Three I.M.P.s per gender is the most that can be handed out: player generated
+	// characters draw from a fixed block of merc profile slots, one per voice, and
+	// anything beyond that would overwrite a character that already exists.
+	imp_max_characters = imp.getOptionalUInt("max_characters", 1);
+	if (imp_max_characters < 1 || imp_max_characters > IMP_MAX_CHARACTERS_LIMIT)
+	{
+		SLOGW("imp.max_characters must be between 1 and {}, clamping {}", IMP_MAX_CHARACTERS_LIMIT, imp_max_characters);
+		imp_max_characters = std::clamp<uint8_t>(imp_max_characters, 1, IMP_MAX_CHARACTERS_LIMIT);
+	}
 
 	merc_online_min_days = gp.getOptionalUInt("merc_online_min_days", 1);
 	merc_online_max_days = gp.getOptionalUInt("merc_online_max_days", 2);
--- a/src/externalized/policy/DefaultGamePolicy.cc
+++ b/src/externalized/policy/DefaultGamePolicy.cc
@@ -1,5 +1,5 @@
 #include "DefaultGamePolicy.h"
-#include "Logger.h"
+#include "IMP_Compile_Character.h"
 #include "Types.h"
 
 #include <algorithm>
@@ -76,15 +76,7 @@ DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 	imp_attribute_bonus = imp.getOptionalInt("bonus_attribute_points", 40);
 	imp_pick_skills_directly = imp.getOptionalBool("pick_skills_directly");
 
-	// Three I.M.P.s per gender is the most that can be handed out: player generated
-	// characters draw from a fixed block of merc profile slots, one per voice, and
-	// anything beyond that would overwrite a character that already exists.
-	imp_max_characters = imp.getOptionalUInt("max_characters", 1);
-	if (imp_max_characters < 1 || imp_max_characters > IMP_MAX_CHARACTERS_LIMIT)
-	{
-		SLOGW("imp.max_characters must be between 1 and {}, clamping {}", IMP_MAX_CHARACTERS_LIMIT, imp_max_characters);
-		imp_max_characters = std::clamp<uint8_t>(imp_max_characters, 1, IMP_MAX_CHARACTERS_LIMIT);
-	}
+	imp_max_characters = std::clamp<unsigned int>(imp.getOptionalUInt("max_characters", 1), 1, NUMBER_OF_PLAYER_GENERATED_CHARACTER_SLOTS);
 
 	merc_online_min_days = gp.getOptionalUInt("merc_online_min_days", 1);
 	merc_online_max_days = gp.getOptionalUInt("merc_online_max_days", 2);
--- a/src/externalized/policy/DefaultGamePolicy.cc
+++ b/src/externalized/policy/DefaultGamePolicy.cc
@@ -1,8 +1,6 @@
 #include "DefaultGamePolicy.h"
 #include "Types.h"
 
-#include <algorithm>
-
 DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 {
 	auto gp = json.toObject();
@@ -75,7 +73,7 @@ DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 	imp_attribute_bonus = imp.getOptionalInt("bonus_attribute_points", 40);
 	imp_pick_skills_directly = imp.getOptionalBool("pick_skills_directly");
 
-	imp_max_characters = std::max<unsigned int>(imp.getOptionalUInt("max_characters", 1), 1);
+	imp_max_characters = imp.getOptionalUInt("max_characters", 1);
 
 	merc_online_min_days = gp.getOptionalUInt("merc_online_min_days", 1);
 	merc_online_max_days = gp.getOptionalUInt("merc_online_max_days", 2);
--- a/src/externalized/policy/DefaultGamePolicy.cc
+++ b/src/externalized/policy/DefaultGamePolicy.cc
@@ -1,5 +1,4 @@
 #include "DefaultGamePolicy.h"
-#include "IMP_Compile_Character.h"
 #include "Types.h"
 
 #include <algorithm>
@@ -76,7 +75,7 @@ DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 	imp_attribute_bonus = imp.getOptionalInt("bonus_attribute_points", 40);
 	imp_pick_skills_directly = imp.getOptionalBool("pick_skills_directly");
 
-	imp_max_characters = std::clamp<unsigned int>(imp.getOptionalUInt("max_characters", 1), 1, NUMBER_OF_PLAYER_GENERATED_CHARACTER_SLOTS);
+	imp_max_characters = std::max<unsigned int>(imp.getOptionalUInt("max_characters", 1), 1);
 
 	merc_online_min_days = gp.getOptionalUInt("merc_online_min_days", 1);
 	merc_online_max_days = gp.getOptionalUInt("merc_online_max_days", 2);
--- a/src/externalized/policy/GamePolicy.h
+++ b/src/externalized/policy/GamePolicy.h
@@ -19,6 +19,11 @@ enum HotkeyModifier
 
 #define gamepolicy(element) (GCM->getGamePolicy()->element)
 
+/** Upper limit for imp_max_characters.
+ * Player generated characters live in a fixed set of merc profile slots, three
+ * for each gender, and there is nowhere to put a character beyond those. */
+#define IMP_MAX_CHARACTERS_LIMIT 6
+
 class GamePolicy
 {
 public:
@@ -97,6 +102,7 @@ public:
 	int32_t imp_attribute_bonus;          // IMP character attribute unallocated bonus points, vanilla 40
 	int32_t imp_attribute_zero_bonus;     // IMP character attribute points given instead of imp_attribute_min, vanilla 15
 	bool imp_pick_skills_directly;        // Use the IMP_SkillTrait selection screen from JA2.5, skipping the personality quiz, vanilla falase
+	uint8_t imp_max_characters;           // Number of IMP characters that can be created in one campaign, 1 to 6, vanilla 1
 
 	/* M.E.R.C. */
 	uint8_t merc_online_min_days;         // The earliest day on or after which M.E.R.C. goes online
--- a/src/externalized/policy/GamePolicy.h
+++ b/src/externalized/policy/GamePolicy.h
@@ -19,11 +19,6 @@ enum HotkeyModifier
 
 #define gamepolicy(element) (GCM->getGamePolicy()->element)
 
-/** Upper limit for imp_max_characters.
- * Player generated characters live in a fixed set of merc profile slots, three
- * for each gender, and there is nowhere to put a character beyond those. */
-#define IMP_MAX_CHARACTERS_LIMIT 6
-
 class GamePolicy
 {
 public:
--- a/src/externalized/policy/GamePolicy.h
+++ b/src/externalized/policy/GamePolicy.h
@@ -97,7 +97,7 @@ public:
 	int32_t imp_attribute_bonus;          // IMP character attribute unallocated bonus points, vanilla 40
 	int32_t imp_attribute_zero_bonus;     // IMP character attribute points given instead of imp_attribute_min, vanilla 15
 	bool imp_pick_skills_directly;        // Use the IMP_SkillTrait selection screen from JA2.5, skipping the personality quiz, vanilla falase
-	uint8_t imp_max_characters;           // Number of IMP characters that can be created in one campaign, 1 to 6, vanilla 1
+	uint8_t imp_max_characters;           // Number of IMP characters that can be created in one campaign, capped by the I.M.P. slots the data declares, vanilla 1
 
 	/* M.E.R.C. */
 	uint8_t merc_online_min_days;         // The earliest day on or after which M.E.R.C. goes online
--- a/src/game/GameVersion.h
+++ b/src/game/GameVersion.h
@@ -17,6 +17,6 @@ extern const char g_version_number[16];
 //	you will invalidate the saved game file
 //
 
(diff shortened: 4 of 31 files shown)


icelynjennings:
Some users may want to use the same voice for multiple IMP mercs (such as to roleplay as clones, twins, or perhaps a robot army in some 3rd party mod), so I worry that it would be a pretty big architectural decision to peg the IMP profile to `iVoiceId`.

rklasen:
At first I thought you want an arbitrary number of IMPs (hard limit up to 20, naturally), and I was counting the reasons why this would be difficult, break the merc slot table, savegame compatibility etc... but then I realized you were only asking to reuse faces and voices.

So I'm failry new to this code, but I remember that 1.13 supported up to six IMPs with _different_ voices, not same voice multiple times, because the voiceId is already tied to a merc slot in the vanilla / stracciatella code. I think using the same profile pic multiple times is trivial, but not voice.

And as far as I can see in this code, male voice 1 is always character slot 51, for example. So if you use male voice 1, you get the merc in char slot 51, always. And without any sort of remapping table or the like, I don't see how clones/twins/robots/replicants/holograms/mirrorUniverseTwins/bizarroTwins etc would work.

So I'd like to avoid scope creep and really only allow (up to) all six IMP voices as real mercs, but not any more.

_Edit:_ `iVoiceId` has always been tied to the merc slot, this is not new in this PR.

icelynjennings:
Ah... Didn't know of that vanilla voice limitation. Then it's probably fine, since as you say, unrestricting the original design would be an increase in scope. 

If someone really wants to do elaborate same-voice roleplay scenarios, then perhaps they may have to just add new custom voices.

src/externalized/policy/DefaultGamePolicy.cc
```diff
@@ -73,6 +76,16 @@ DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
...
 	imp_pick_skills_directly = imp.getOptionalBool("pick_skills_directly");
 
+	// Three I.M.P.s per gender is the most that can be handed out: player generated
+	// characters draw from a fixed block of merc profile slots, one per voice, and
+	// anything beyond that would overwrite a character that already exists.
+	imp_max_characters = imp.getOptionalUInt("max_characters", 1);
+	if (imp_max_characters < 1 || imp_max_characters > IMP_MAX_CHARACTERS_LIMIT)
+	{
+		SLOGW("imp.max_characters must be between 1 and {}, clamping {}", IMP_MAX_CHARACTERS_LIMIT, imp_max_characters);
+		imp_max_characters = std::clamp<uint8_t>(imp_max_characters, 1, IMP_MAX_CHARACTERS_LIMIT);
```
lynxlynxlynx: Just clamp it, it's already documented, so no need for logging nor the whole block.

src/externalized/policy/GamePolicy.h
```diff
@@ -19,6 +19,11 @@ enum HotkeyModifier
 
 #define gamepolicy(element) (GCM->getGamePolicy()->element)
 
+/** Upper limit for imp_max_characters.
+ * Player generated characters live in a fixed set of merc profile slots, three
+ * for each gender, and there is nowhere to put a character beyond those. */
+#define IMP_MAX_CHARACTERS_LIMIT 6
```
lynxlynxlynx: Why hardcode it again when it could be moved and NUMBER_OF_PLAYER_GENERATED_CHARACTER_SLOTS used directly.

src/externalized/policy/DefaultGamePolicy.cc
rklasen: done, the actual upper limit is now 2 * 3 (genders times voice lines per gender).

src/externalized/policy/GamePolicy.h
rklasen: done, same as above

lynxlynxlynx:
Bumping NUM_PROFILES would give 30 extra slots just within current limits, besides the few named unused slots at the end.

But my main gripe with the patch is that it encrusts the old voice ID and gender connection to the profile. It was a simple solution for what they needed to support, but I don't think it's fitting for a multiple IMP system. Can't this info be stored and looked up transparently? The merc profile ID should be just a boring index.

rklasen:
Well I wouldn't have chosen the word encrust. I would more have said "build upon", but I do know what you're getting at 😅

Okay, so I've looked at the codes some more, and there is a precedent for mapping multiple characters to the same voice set: the bartenders that are scattered across Arulco. So with that in mind, it wouldn't feel out of place anymore to map multiple characters to the same voice set and, of course, the same profile picture. 

It's a little bit more work than what I've done right now, but if the resonance is strong enough, I'm open to implement it that way. Just to make sure we are on the same page here about everything: we are not implementing an additional mapping layer between the voiceid and the Merc slots. We're just setting which Merc gets which voice set. It would be technically possible to reuse the existing savegame layout as there is unused space in it, but I would prefer to just bump the savegame version and include additional fields for profile map and voice set map.

lynxlynxlynx:
Yes, both things should just be plain data, so if eventually someone wants to add more voices or portraits, this code doesn't need to change at all, because it'd be data-driven.

We'll still need to set a maximum, so mercs-profile-info.json and the lua export can reserve a finite set of slots.

rklasen:
hm, how data-driven do you mean exactly? externalizing _everything_-everything (name, gender, voice id, profile pic id, `is_IMP` etc to json (like the `mercs-profile-info.json` that already exists, but is practically empty)? cause filling it properly and then using it for _all_ mercs (AIM and MERC) and IMPs seems kinda The Right Thing To Do, but also out of scope here...

lynxlynxlynx:
No, mercs have preset data, IMPs are created on the fly. There's no point in storing them separately outside the save. Well ... we actually already do for the nickname-import purposes, but that's besides the point. I just meant that the voice and portrait should be saved; the engine shouldn't need to do any deduction from these fields or profile ID, just use them.

rklasen:
oh, right, the `prof.dat`. Hm, alright, so: `LoadMercProfiles()` already creates our merc instances from the templates, IMPs are created on the fly, and this line suggests we may be able to treat voiceId similar to face...

https://github.com/ja2-stracciatella/ja2-stracciatella/blob/a44fbb68589279586ffc5b3b4e24b75a4ede9b0f/src/game/Tactical/Soldier_Profile.cc#L143

Hm, alright, so how about this:

- new UINT8 ubVoiceId on MERCPROFILESTRUCT
-p.ubVoiceId = i; beside the existing ubFaceIndex line in LoadMercProfiles(), plus an optional "voice" field in mercs-profile-info.json read in deserializeStruct for the static cases
- IMP creation sets it explicitly instead of the slot deciding it
- INJ_U8/EXTR_U8, wider save record, SAVE_GAME_VERSION bump, version-gated read
- IMP slot ceiling comes from counting MercType::IMP profiles, retiring NUMBER_OF_PLAYER_GENERATED_CHARACTER_SLOTS﻿. the clamp from the game.json remains, sorta like to limit the number of IMPs if we want to emulate vanilla behaviour.

We'd also mark the 5 remaining unused profiles in merc-profile-info as IMP so the IMP factory can use them. I think some unused slots are actually reserved for the enemy tanks and some eldorado (?!) thats never made it to the game, so 5 remain withour raising the profile limit to 200. We'd still get to create 11 IMPs and reuse both voice and face. 

Raising to 200 profiles can be done in a separate PR then.

rklasen:
okay, while we're at it, I've removed (or better, had claude code remove) some 8-bit map xor-ing to check which slot number is IMP-able. We're not partying like it's 1999 anymore, there is ample disk size, memory size and cpu speed to have proper fields for that stuff now. So the IMP profile slots don't have to be continuous anymore either. the IMP factory will be ready soon-ish.

_Edit_ done, first tests seem to work. Multiple IMPs, same faces, same voices, different genders, save and load all seem to work.

src/game/GameVersion.h
```diff
@@ -17,6 +17,6 @@ extern const char g_version_number[16];
 //	you will invalidate the saved game file
 //
 
-constexpr UINT32 SAVE_GAME_VERSION = 103;
+constexpr UINT32 SAVE_GAME_VERSION = 104;
```
lynxlynxlynx, line 20: No need to bump it, since you handle old versions gracefully.

assets/externalized/game.json
```diff
@@ -188,10 +188,10 @@
...
     "bonus_attribute_points": 40,
 
-    //  Number of I.M.P. characters that can be created over a campaign, 1 to 6.
-    //  Each character keeps one of the player generated merc profile slots, of which
-    //  there are three per gender, so at most three men and three women. Voices and
-    //  portraits already in use are skipped when profiling a further character.
+    //  Number of I.M.P. characters that can be created over a campaign.
+    //  Each character keeps one of the merc profiles marked as an I.M.P. slot in
+    //  mercs-profile-info.json for the rest of the campaign, so the ceiling is how
+    //  many of those the data declares. Voices and portraits may be reused freely.
```
lynxlynxlynx: Come on, neither version is user friendly. Just write that the current limit is 11. Or use one less profile slot and make it a round 10.

src/externalized/policy/DefaultGamePolicy.cc
```diff
@@ -76,7 +75,7 @@ DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 	imp_attribute_bonus = imp.getOptionalInt("bonus_attribute_points", 40);
 	imp_pick_skills_directly = imp.getOptionalBool("pick_skills_directly");
 
-	imp_max_characters = std::clamp<unsigned int>(imp.getOptionalUInt("max_characters", 1), 1, NUMBER_OF_PLAYER_GENERATED_CHARACTER_SLOTS);
+	imp_max_characters = std::max<unsigned int>(imp.getOptionalUInt("max_characters", 1), 1);
```
lynxlynxlynx: This is pretty useless and breaks setting it to 0.

src/game/Tactical/Soldier_Profile_Type.h
```diff
@@ -393,6 +396,15 @@ static inline bool HasSkillTrait(MERCPROFILESTRUCT const& p, SkillTrait const sk
 }
 
 
+// The following values are used with the 'ubIMPSlotState' variable
+
+// No player generated character lives here. The character being built, if any,
+// is always the first free slot, so being part way through needs no state.
+#define IMP_SLOT_FREE						0
+// A finished character holds this profile for the rest of the campaign
+#define IMP_SLOT_TAKEN						1
```
lynxlynxlynx: Would be cleaner as an enum class then, which you could use to type ubIMPSlotState.

src/game/Laptop/IMP_Voices.cc
```diff
@@ -259,8 +230,8 @@ static void BtnIMPVoicesDoneCallback(GUI_BUTTON *btn, UINT32 reason)
 			iCurrentImpPage = IMP_FINISH;
 		}
 
-		// set voice id, to grab character slot
-		LaptopSaveInfo.iVoiceId = VoiceSlot(iCurrentVoices);
+		// the voice the character will speak with, counted over both genders
+		LaptopSaveInfo.iVoiceId = iCurrentVoices + (fCharacterIsMale ? 0 : NUMBER_OF_PLAYER_VOICES_PER_GENDER);
```
lynxlynxlynx: NUMBER_OF_PLAYER_VOICES_PER_GENDER will also need to go at some point, as having it baked in prevents addition of new voices.

src/game/Laptop/IMP_Compile_Character.cc
```diff
@@ -29,94 +31,76 @@ static INT32 iLastElementInPersonalityList = 0;
...
 	p.zName = pFullName;
 	p.zNickname = pNickName;
 
 	p.bSex = fCharacterIsMale ? MALE : FEMALE;
 
+	// The voice the player picked, which is where this character's speech,
+	// dialogue text and battle sounds come from. Slots past the six the game
+	// shipped with have none of their own, so this also makes it hireable.
+	p.ubVoiceId   = PLAYER_GENERATED_CHARACTER_ID + LaptopSaveInfo.iVoiceId;
+	p.bMercStatus = MERC_OK;
```
lynxlynxlynx: Why is this line needed?

src/game/GameVersion.h
rklasen: no the bump is absolutely needed. there is new fields now in the savegame (`ubVoiceId` / `ubIMPSlotState`) which need to be parsed correctly.

assets/externalized/game.json
rklasen: I thought we wanted to do this data driven 😆  either is fine by me. so you want a hard limit of 10? remember that we can not use all "unused" marked slots indiscriminately cause one of them is secretly reserved for the Jeep.

src/externalized/policy/DefaultGamePolicy.cc
rklasen: well... I mean setting this to 0 is just disabling IMPs altogether. but sure, I can fix that. stay tuned. _EDIT:_ so I think I just misunderstood you. Setting this to 0 _doesn't_ disable IMPs, ist silently sets it to 1. okay, I'll fix.

src/game/Tactical/Soldier_Profile_Type.h
rklasen: hm, sure why not. but that would _also_ require a savegame version bump. there is no way around that. let me whip up a version with an enum.

src/game/Laptop/IMP_Voices.cc
rklasen: agreed. coming.

src/game/Laptop/IMP_Compile_Character.cc
rklasen: That's p.bMercStatus = MERC_OK; see Soldier_Profile.cc#L144: LoadMercProfiles sets MERC_HAS_NO_TEXT_FILE for any profile whose .edt is missing, and the new slots 165-169 have none of their own. so without this line those characters load unhireable.

src/game/GameVersion.h
lynxlynxlynx: Is it not at the end of the file?

assets/externalized/game.json
lynxlynxlynx: The comment was more about readability. Without further modification, there is a soft limit.
And I wouldn't say that one slot is secretly reserved, it's just in another table and clearly a placeholder for the two vehicles. Which also means that if anyone wanted to reuse the slots, they'd have to find another one for one of them.

src/externalized/policy/DefaultGamePolicy.cc
lynxlynxlynx: Exactly, that's what the other feature request I linked is about.

src/game/Tactical/Soldier_Profile_Type.h
lynxlynxlynx: Why, the underlying storage can remain the same (can be specified for enums).

src/game/GameVersion.h
rklasen: no, rather in the middle of it, and approx 40 other fields come after it 😐 so no way around the bump here.

src/externalized/policy/DefaultGamePolicy.cc
rklasen: I see. alright, setting to 0 now correctly disables IMPs, closing #1317 too.

src/game/Tactical/Soldier_Profile_Type.h
rklasen: uh... yeah you're right. I got that wrong. the bump is still neccessary for the voiceId tho.

src/game/Laptop/IMP_Voices.cc
rklasen: done.

src/game/GameVersion.h
lynxlynxlynx: Ah, ok, nevermind then.

rklasen:
Alright, everything should be taken care of. I'm compiling now and will test more IMPs, then make a screenshot and put it here. So this PR should be ready after this test.

_EDIT:_ IMP creation worked just fine, double voices are no problem:

image

But: upon arrival, they each have their say about the helicopter pilot, each 11 of them, multiple times for the same voice lines... I think I'll limit to once per voice set, so clones don't repeat each other verbatim.

Also, IMPs 7 to 11 have green pants because no data exists for them in `prof.dat`. I'll have to fix that too.

lynxlynxlynx:
Looks good. Also takes care of #1593 now, which you've already noted. The pants thing is probably best solved by externalizing that to mercs-profile-info.json as well and then setting it to diverse colors for the imp profiles.

rklasen:
Okay all get the same clothes now, the skin and hair color is deducted from the portrait picture. Up to all six individual voices lines play upon arrival, but not the same voice set twice.

I'll just externalize the pant and shirt color to `mercs-profile-info.json` (keeping the vanilla defaults of black and white for now, seems weird having the first six in b/w and the other possible 5 as clown troupe).

src/game/Laptop/EMail.cc
```diff
@@ -2163,6 +2172,10 @@ static void HandleIMPCharProfileResultsMessage(void)
 		case 12: iOffSet = IMP_PORTRAIT_FEMALE_4; break;
 		case 13:
 		case 14: iOffSet = IMP_PORTRAIT_FEMALE_5; break;
+		// the blurb describing the character's looks only covers the portraits
+		// the game shipped with; any others are described by the first of their
+		// gender rather than by whatever ran last
+		default: iOffSet = imp.bSex == MALE ? IMP_PORTRAIT_MALE_1 : IMP_PORTRAIT_FEMALE_1; break;
```
lynxlynxlynx: Is this smart enough wrt big body types?

rklasen: Good catch! No, it was not. It seems there was also a bug or at least an inconsistency in vanilla. If strength is larger than 75 _and_ the portrait has the burly flag, then the IMP gets the BIGMALE body. But the mail thats sent for the IMP only required the burly portrait flag, and didn't check the strength stat.

I've unified both checks, both now require str > 75 and the burly flag. Seems to work:

image

rklasen: And the corresponding mail:

image

Though I have to say, it's rather... sublte.

lynxlynxlynx:
Seems to work fine. Behaviour for n=0 could be further refined (no email), but that can be done later if desired.

icelynjennings:
Awesome work!

Moddi:
To further elaborate on my idea (#1593) it's not just to add the new voices and portraits from UB and WF, technically - at least in the German version there are even more. Because reasons the localizer for the German UB version (that switched from Topware to another company) re-recorded _every_ audio file of every merc - even for those that already had audio files from the main game. And while they are generally considered inferior to the dubbing work done for the main game this also gave us 6 additional IMP voice sets in addition to the regular 4th one. When this PR allows modders to add all 13 (or any number) of them to the game that would be great (I did not follow the complete discussion and my coding knowledge is not sophisticated enough to understand if the codes does that).

lynxlynxlynx:
You just need to install the sound files and name them, so the numbers follow the new profile numbers. Then you just add their data to imp.json and that's it. A very simple mod could be created — you can start by copying https://github.com/ja2-stracciatella/ja2-stracciatella/tree/master/assets/mods/o-fortuna

If you want more than the current 11 available, you'll also need to add more entries to merc-profile-info.json

rklasen:
Are you sure they re-recorded _all_ the voice lines? It's been a while since I played UB, but I seem to remember there were not any bloodcats or monsters, so at least those voice lines would have to be missing, right?

PR #2462, opened 2026-08-30
Inventory for the Hummer and Ice Cream Truck

rklasen:
In 1.13, the vehicles had an inventory. Makes sense I guess, you can just put rifles, tool boxes and bottles of beer into the trunk of the car, right? I've implemented this here too. It's less cargo space than in 1.13, 18 slots here in total. This way, I could reuse the image of the sector inventory (with cropping), and didn't have to redo it. Is there any interest in this feature? It's pretty non-vanilla, so I guess it would make sense to gate it behind a game.json flag.

these 18 slots require a new array on the vehicle "character", so  the savegame version had to be incremented to 104.

Disclaimer: the code was written by claude code.

image

Below this line is an AI written summary:

---

A squad that drives everywhere still has to carry everything it owns on its backs, and the sector inventory only helps while the mercs are standing in that sector.  Let the two enterable vehicles carry cargo of their own, so kit can be parked in the truck and driven to wherever it is next needed.

The cargo lives in VEHICLETYPE rather than in the inventory of the SOLDIERTYPE that represents the vehicle.  A merc's inv[] has nineteen slots of which only four behave as big pockets - ItemSlotLimit() and CanItemFitInPosition() decide that positionally, from the slot index - so reusing it would have meant either a four slot hold or changing the pocket rules for every merc in the game.  An array on the vehicle sidesteps both: every cargo slot takes anything, and the helicopter, which has no SOLDIERTYPE at all, cannot accidentally acquire one.

Eighteen slots are written after the vehicle's fixed record in the save, read back only for save version 104 and later, so older saves load with empty holds.

The panel takes over the team panel while a vehicle is the selected character, the same slot the paper doll uses.  It draws a plain three by six grid instead, because a truck has no head, hands, armour or camouflage to show.  The slot artwork is borrowed from the sector inventory background: that lattice is baked into one big sector_inventory.sti rather than being a sub-image, so the background is blitted under a clipping rectangle with an offset that lines its lattice up with ours.  That fixes the cell size at 72x32 and is what makes the grid three columns of six.

The trash can and the key ring are painted into mapinv.sti, so they exist as pictures only while the paper doll is up.  Creating their mouse regions over the cargo grid left invisible click targets on the slots, so they now follow the selection and are absent while the cargo panel is showing.  The done button stays in both, since it is drawn and it is how the panel is closed.

Cargo is tipped out onto the ground when the vehicle is destroyed rather than being deleted with it.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -135,6 +135,12 @@
   //  vanilla setting: false
   "extra_attachments": false,
 
+  //  Gives the hummer and the ice cream truck a cargo hold of eighteen slots, opened from
+  //  the mapscreen while the vehicle is the selected character. Cargo travels with the
+  //  vehicle and is tipped onto the ground if it is destroyed.
+  //  vanilla setting: false
+  "vehicle_cargo": false,
+
   //  Packages at Drassen airport never get stolen, no need to bribe Pablo.
   //  vanilla_setting: false
   "pablo_wont_steal": false,
--- a/src/externalized/policy/DefaultGamePolicy.cc
+++ b/src/externalized/policy/DefaultGamePolicy.cc
@@ -44,6 +44,7 @@ DefaultGamePolicy::DefaultGamePolicy(const JsonValue& json)
 	gui_extras = gp.getOptionalBool("gui_extras", true);
 	informative_tooltips = gp.getOptionalBool("informative_tooltips", false);
 	extra_attachments = gp.getOptionalBool("extra_attachments");
+	vehicle_cargo = gp.getOptionalBool("vehicle_cargo", false);
 	skip_sleep_explanation = gp.getOptionalBool("skip_sleep_explanation");
 
 	pablo_wont_steal = gp.getOptionalBool("pablo_wont_steal");
--- a/src/externalized/policy/GamePolicy.h
+++ b/src/externalized/policy/GamePolicy.h
@@ -122,6 +122,8 @@ public:
 	uint16_t start_sector;        // Starting sector
 	bool reveal_start_sector;     // Should the start sector radar map be shown at start
 
+	bool vehicle_cargo;           // Let the hummer and the ice cream truck carry cargo of their own
+
 	uint8_t suppression_fire_modifier; // Scales AP loss from suppression (numerator of the AP-loss formula); vanilla 6, 0 disables AP loss
 	uint16_t suppression_fire_reaction_threshold; // Numerator of the stance-reaction threshold; vanilla 130, 0 = always react (1.13 behaviour), higher = more resistant
 
--- a/src/game/GameVersion.h
+++ b/src/game/GameVersion.h
@@ -17,6 +17,6 @@ extern const char g_version_number[16];
 //	you will invalidate the saved game file
 //
 
-constexpr UINT32 SAVE_GAME_VERSION = 103;
+constexpr UINT32 SAVE_GAME_VERSION = 104;
 
 #endif
--- a/src/game/Strategic/MapScreen.cc
+++ b/src/game/Strategic/MapScreen.cc
@@ -76,6 +76,7 @@
 #include "Timer_Control.h"
 #include "Town_Militia.h"
 #include "TownModel.h"
+#include "Vehicles.h"
 #include "Video.h"
 #include "VObject.h"
 #include "VObject_Blitters.h"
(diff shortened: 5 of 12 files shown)


lynxlynxlynx:
This feels out of scope to me, but I'd like some other opinions before a deeper review or closure.

rklasen:
I know it's pretty non-vanilla, but it has proven invaluable to me. We could maybe ship it as a modification instead. We would still have to merge the changes of course because the savegame layout changes, but the feature itself could be shipped as a mod.

icelynjennings:
A worry of mine is increasing the "barrier of entry" for future Stracciatella contributors. Most new contributors can be assumed to have a past idea of how the vanilla JA2 game mechanics played as a shared context.

If this were to change enough to where we'd need provide a "game manual" of JA2 mechanics to them, then they could potentially find it more difficult to contribute to any of the [existing vehicle-related vanilla bugs & possible stracciatella regressions](https://github.com/ja2-stracciatella/ja2-stracciatella/issues?q=is%3Aissue%20state%3Aopen%20(hummer%20OR%20truck%20OR%20car%20OR%20vehicle)), assuming they would become expected to handle edge-cases around the non-vanilla game mechanics. In other words, fixing a 27-year old vanilla bug would require understanding some 1.13 mod features.

Unless we have an existing functionality for handling "game mechanic mods" (rather than map changes, as is the case with most mods, I understand), then perhaps this is more towards the scope of some kind of Stracciatella-1.13 merger fork?

lynxlynxlynx:
I don't think there's any viable way of implementing something like this in lua, so mods can't help.

rklasen:
> I don't think there's any viable way of implementing something like this in lua, so mods can't help.

Yaah the code changes would need to happen either way, whether the inventory setting is enabled via mod or game.json...

> A worry of mine is increasing the "barrier of entry" for future Stracciatella contributors. Most new contributors can be assumed to have a past idea of how the vanilla JA2 game mechanics played as a shared context.

well sure, but thats regarding game _behaviour_, node code, right? and as long as the vehicle inventory is gated behind a non-default-true flag, the game would feel very vanilla to a newcomer.

Whoever touches the code will simply see the comments, that vehicles may have inventory too, if desired. If you want, I cann clarify in the comments that this behaviour emulates something from 1.13.

icelynjennings:
> well sure, but thats regarding game behaviour, node code, right? 

Regarding newcomer code contributors.

I.e., the shared context of vanilla game behaviour provides new contributors a cheap, preliminary intuition of how a certain game mechanic **_might_** work, even if they haven't seen its code **_yet_**. This reduces the time/need for onboarding/explaining the codebase to newcomers, since they can be assumed to either already have intuition, or can gain it from the Sir-Tech game manual.

There are elegant ways to modernize the code without losing newcomer intuition, just as you succeeded to do with the IMP limit changes, where the modernization can be thought of as the same mechanic they already remember, minus the vanilla limitations.

Of course, ultimately a car inventory isn't the wildest mod feature imaginable even to newcomers. Just my personal thoughts. E.g., I'd personally find it daunting to contribute to 1.13 over Stracciatella, given that I likely would have had to experience the 1.13 mod features as a player for a minimum amount of times before contributing, which I simply never had to do with Stracciatella.

rklasen:
Hm, while I do understand your argument, I don't agree with it 😉  My main worry was that this vehicle inventory is too much like something that vanilla JA2 just is not (and more like Urban Chaos, 1.13, Deidranna Lives or anything else which is, naturally, way out of scope).

So, if the consensus is that vehicle inventory just doesn't belong into vanilla/stracciatella, I can absolutely live with that. In that case we can close this PR, no hard feelings, and I will continue to try to contribute to stracciatella in other ways.

But I feel "newcomers to the code might be surprised by something we made room for that wasn't there in vanilla" is just not a strong argument to reject a contribution that would otherwise fit the code (and again, I'm not saying vehicle inventory is that kind of addition. I can absolutely live with rejection, if the reason is a bad culture fit or horrible code 😄 )

steph-pg:
In my opinion from players perspective this is actually a good feature. Mercs` inventory is limited, which is fine.
But starting from 60+ progress you need to carry around a lot of crap like repair kits, crafting parts like duct tape,
bloodcats parts, rare ammo, some heavy weapons that you try to save for later, and actually lots of things you try to
save for later like extra grenades and explosives. So you end up in either loading all your mercs inventory or
you just leave it behind and then when you need a spring or an extra mortar shell you just don't bother returning for it.

It will also make early ice cream truck a more valuable find. Like if I meet Hamous early on between San Mona mine and Chitzena SAM I very often just leave it there, as it takes too much effort to clear the way to roads where I can finally start to use it.

Another thing that at certain progress land vehicles become almost useless as money becomes a solved problem, you have all 3 sams and skyrider can travel on any distance. So inventory adds extra value.

And the last argument that in battle I often want my individual mercs to have everything possible that's useful in
battle like different types of grenades or a gas mask and not duct tape and some else's med kits.

lynxlynxlynx:
If players would find this useful or not is not under deliberation — it's obviously true at least for some. 
What is on topic is scope creep and the related higher maintenance burden. Sure, we can add json toggles for anything, but at the end of the day the goal of this project is _not_ to become 1.13-like if you just turn on all the knobs.

PR #2464, opened 2026-08-31
Drop dead mercs' gear after autoresolve

steph-pg:
When mecs got killed during autoresolve their equipment is lost. This maybe quit disappointing when they carry quest items, or just good equipment.

### DIFF
--- a/src/game/Strategic/Auto_Resolve.cc
+++ b/src/game/Strategic/Auto_Resolve.cc
@@ -1655,6 +1655,8 @@ static void RemoveAutoResolveInterface()
 
 	if (ar.pRobotCell) UpdateRobotControllerGivenRobot(ar.pRobotCell->pSoldier);
 
+	const SGPSector arSector(ar.ubSector);
+
 	for (UINT8 i{ 0 }; i != ar.ubMercs; ++i)
 	{
 		SOLDIERTYPE& s = *gpMercs[i].pSoldier;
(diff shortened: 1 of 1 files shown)


src/game/Strategic/Auto_Resolve.cc
```diff
@@ -1655,13 +1655,19 @@ static void RemoveAutoResolveInterface()
 
 	if (ar.pRobotCell) UpdateRobotControllerGivenRobot(ar.pRobotCell->pSoldier);
 
+	const SGPSector arSector(ar.ubSector.x, ar.ubSector.y);
```
lynxlynxlynx: I know you just moved it, but you can simplify it as well, since ubSector is a SGPSector and it has a copy ctor.

src/game/Strategic/Auto_Resolve.cc
```diff
@@ -1655,13 +1655,19 @@ static void RemoveAutoResolveInterface()
...
 	{
 		SOLDIERTYPE& s = *gpMercs[i].pSoldier;
 		// Record finishing information for our mercs
 		if (s.bLife == 0)
 		{
 			StrategicHandlePlayerTeamMercDeath(s);
+			/* HandleStrategicDeath() only leaves a corpse behind while the map
+			 * screen is up, and it never is during autoresolve, so drop the body
+			 * and their gear here the way the militia and enemies do. */
+			AddDeadSoldierToUnLoadedSector(arSector, &s, RandomGridNo(), ADD_DEAD_SOLDIER_TO_SWEETSPOT);
```
lynxlynxlynx, line 1674: This feels too generous to me. When you win a sector, you scour the leftover loot, so why wouldn't the enemy do the same? So I'd limit it to our team victories. And/or condition it on f_drop_everything.

It could also create problems due to the random placement, since some maps have inaccessible areas.

icelynjennings: I guess this PR might be assuming that enemy "post-battle sector inventory looting behaviour" may have been originally intended to be the same as their "peacetime sector inventory looting behaviour", where IIRC the soldiers never seem to loot any sector inventories, ever, even in sectors with no mercs or militia present.

steph-pg: enemy soldiers do loot sector items when player loses. And I've dug a bit deeper and it turned out that the same mechanics apply in autoresolve so no changes needed. And that's actually event better because if we gate it we would lose all the equipment including keys, quest items. When RemoveRandomItemsInSector is called after player's defeat (both in tactical and in auroresolve) we have at least some chances to get some of our items back later. And a note on RandomGridNo. Enemy soldiers already drop some tiny percent of their items in autoresolve and the exact same call with RandomGridNo is used so probably it's not that critical?

src/game/Strategic/Auto_Resolve.cc
steph-pg: done

src/game/Strategic/Auto_Resolve.cc
lynxlynxlynx: Thanks, good enough.

PR #2466, opened 2026-09-02
Implement SGPVSurface::ShadowRect with SDL

momoko-h:
The results of this function and its cousin ShadowRectUsingLowPercentTable can be achieved by calling SDL_SetSurfaceColorMod before blitting the target rectangle onto itself.

This allows us to remove the intensity table, the function that builds it and one blitter.


lynxlynxlynx:
Thanks!

PR #2470, opened 2026-09-02
NixOS build instructions + tiny fix

icelynjennings:


src/game/TileEngine/Radar_Screen.cc
```diff
@@ -273,7 +273,7 @@ void RenderRadarScreen()
 					/* hostile civilian */
 					s->bTeam == CIV_TEAM && !s->bNeutral && s->bSide != Side::FRIENDLY ? FROMRGB(255,   0,   0) :
 					/* named civilian */
-					s->bTeam == CIV_TEAM && s->ubProfile != NO_PROFILE                ? FROMRGB(  0,   0, 255) :
+					s->bTeam == CIV_TEAM && s->ubProfile != NO_PROFILE                ? FROMRGB(  85,   85, 255) :
```
lynxlynxlynx: mistake

flake.nix
```diff
@@ -0,0 +1,40 @@
+{
```
lynxlynxlynx: can this be put into some subdir, so it doesn't pollute the toplevel?

icelynjennings: Generally, the convention is at the root.

But I ultimately decided on a simpler solution. Nix-shell will handle most people's cases.

icelynjennings:
Tested, builds on NixOS 26.05 that has X11 and Wayland installed.

lynxlynxlynx:
It still has unrelated changes included (Radar_Screen.cc)~, now also the removal of OpenBSD docs~!?

lynxlynxlynx:
Thanks

PR #2471, opened 2026-09-03
no I.M.P. mail when no I.M.P. can be made

### LINKED ISSUE #2469: Prevent IMP e-mail if imp.max_characters = 0
https://github.com/ja2-stracciatella/ja2-stracciatella/pull/2460#issuecomment-5515117900

rklasen:
Building on #2460 , we need to make sure that no mail is sent when no IMP is created.

### DIFF
--- a/src/game/Laptop/Laptop.cc
+++ b/src/game/Laptop/Laptop.cc
@@ -1442,7 +1442,7 @@ static BOOLEAN HandleExit(void)
 	if (LaptopSaveInfo.gfNewGameLaptop)
 	{
 		// Set an event to send this email (day 2 8:00-12:00)
-		if (!LaptopSaveInfo.fIMPCompletedFlag && !LaptopSaveInfo.fSentImpWarningAlready)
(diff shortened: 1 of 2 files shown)


lynxlynxlynx:
Thanks, LGTM. If you're willing to go the extra mile, we need another json key for that for Unfinished Business #2469.

And TIL: you can actually test this in a browser: https://dos.zone/jagged-alliance2-unfinished-business/

rklasen:
Sure, no problem. I'm underway the next few days, so I can do it next week at the earliest, but I will do it. Stay tuned 😅

rklasen:
Done, the new key is `imp.send_profile_results_email`. I could only find one senyt mail though? Its been ages since I've played UB.

lynxlynxlynx:
Hmm, that's good, thanks, I'm just not sure if game.json is the best place for it, since it's more a data thing than for players to toggle. But I guess a future UB-compat mod could just ship a json patch. @ja2-derek: any thoughts?

EDIT: I guess moving it to imp.json would solve that.

rklasen:
Done, the setting moved to `imp.json`.

lynxlynxlynx:
Thanks!

PR #2472, opened 2026-09-03
raise the profile limit to 200

rklasen:
expanding on #2460 , it should now be possible to raise the number of profiles to 200. Two things:

- some places _rely_ on 170 and just cannot work with 200. for those, we add a new constant
- how to we mark the empty slots in game.json? as IMPs, like in the other PR? or still unused? I any modder wants to use them, they can change the type themselves I suppose.

note: I'm still testing this branch.

Disclaumer: the code is written by claude code.

Below this line is the AI generated summary:

---

The profiles stopped at the 170 the original game shipped, all of them spoken for. NO_PROFILE is 200 and shares a byte with every profile ID the maps and the saves store, so 200 is as far as the profiles can go without breaking those files -- and that is 30 free slots for the data to name: more I.M.P. slots, or NPCs a mod brings along.

Nothing new is declared here, only the room for it. Two things stay at the vanilla count, because that is what they hold: prof.dat, which now hands out an empty template for the slots it has no record for, and the saves written before this, which stop short of the new slots both in the profiles and in the NPC quote records. Loading one of those sets up the slots it predates as a new campaign would, so a profile declared after the game was saved is still there.

### DIFF
--- a/assets/externalized/game.json
+++ b/assets/externalized/game.json
@@ -192,7 +192,8 @@
     //  0 switches the I.M.P. site off entirely.
     //  A character holds one of the merc profiles that mercs-profile-info.json
     //  marks as an I.M.P. slot for the rest of the campaign, so that is the upper
-    //  end: eleven as the game ships, more if the data declares more. Voices and
+    //  end: twenty-eight as the game ships, more if the data declares more, and
+    //  hiring refuses the nineteenth merc however many are declared. Voices and
     //  portraits may be reused freely and do not limit it.
     //  vanilla setting: 1
     "max_characters": 1,
--- a/assets/externalized/mercs-profile-info.json
+++ b/assets/externalized/mercs-profile-info.json
@@ -234,5 +234,30 @@
     { "profileID": 166, "internalName": "NPC166", "type": "IMP", "pantsColor": "BLACKPANTS", "vestColor": "WHITEVEST" },
     { "profileID": 167, "internalName": "NPC167", "type": "IMP", "pantsColor": "BLACKPANTS", "vestColor": "WHITEVEST" },
     { "profileID": 168, "internalName": "NPC168", "type": "IMP", "pantsColor": "BLACKPANTS", "vestColor": "WHITEVEST" },
-    { "profileID": 169, "internalName": "NPC169", "type": "IMP", "pantsColor": "BLACKPANTS", "vestColor": "WHITEVEST" }
+    { "profileID": 169, "internalName": "NPC169", "type": "IMP", "pantsColor": "BLACKPANTS", "vestColor": "WHITEVEST" },
+     /* IMP: more slots for player generated characters, in the profiles that
+        exist past the ones the original game shipped with. Nothing is recorded
+        for these anywhere else -- no portrait, no dialogue, no speech -- and
+        nothing needs to be: a character brings its own, given to it out of
+        imp.json when the site builds it.
+        There are enough of them that the player's team runs out of room first,
+        with some to spare for replacing the ones that die. 187 to 199 are left
+        free for whatever else wants a profile. */
+    { "profileID": 170, "internalName": "PGSLOT170", "type": "IMP", "pantsColor": "BLACKPANTS", "vestColor": "WHITEVEST" },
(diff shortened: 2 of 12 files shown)


rklasen:
Creating 18 IMPs works, as do save and load. Note that 18 is the limit for mercs, as we still get to "recruit" the ice cream truck and hummer.

Screenshot 2026-09-03 at 21 52 09

Screenshot 2026-09-03 at 21 52 23

lynxlynxlynx:
Thanks. I would leave the last commit out: it could be a mod we ship with the source instead. For other modders that may want to add profiles, in case they need more than the remaining gap, patching in new entries is much cleaner than modifying existing ones.

rklasen:
Sure, no problem. Tho I am away the next few days. Maintainers can commit to this branch, so you can drop the last commit and merge.

Or I can drop it next week, both is fine by me.

ShishkinP:
In the spirit of json declutterization maybe default `pantsColor` and `vestColor` should be set during deserialization instead:
https://github.com/ja2-stracciatella/ja2-stracciatella/blob/1c7149ca64e69c9d71b86dd03535d266c4765367/src/externalized/mercs/MercProfile.cc#L197
Somethin like:
```c++
prof->VEST = r.getOptionalString("vestColor", binaryProf->VEST);
if (prof->VEST.empty()) {
	prof->VEST = "WHITEVEST";
}
```

lynxlynxlynx:
Good idea, I guess all prof.dat profiles have a color set, right?

ShishkinP:
>prof.dat profiles have a color set, right?

All named and the 6 IMP (51 - 56) — yes
Unused: 71, 163 - 169 — no

lynxlynxlynx:
Perfect, then the default logic doesn't need to be more complicated than what you suggested.

assets/externalized/game.json
```diff
@@ -190,10 +190,12 @@
...
-    //  A character holds one of the merc profiles that mercs-profile-info.json
-    //  marks as an I.M.P. slot for the rest of the campaign, so that is the upper
-    //  end: eleven as the game ships, more if the data declares more. Voices and
-    //  portraits may be reused freely and do not limit it.
+    //  A character holds a merc profile for the rest of the campaign: one of the
+    //  eleven that mercs-profile-info.json marks as an I.M.P. slot, or, once
+    //  those are spoken for, one of the thirty past them that no data file
+    //  declares at all. Raising this needs nothing else changed, and the team
+    //  fills up first: hiring refuses the nineteenth merc. Voices and portraits
+    //  may be reused freely and do not limit it.
```
lynxlynxlynx, line 198: this is useless info to the player, you can just delete it all or simply say 18 is the max (or at least max in the team).

lynxlynxlynx:
Other than that, it's an interesting insight that we actually don't need persistent slots for IMPs.

rklasen:
Yes, pretty much. If we're raising the IMP limits and let the game use any unclaimed slot for up to 18 IMPs, we don't need to reserve either the vanilla slots 51-56, nor the "unused" slots below 169, and - naturally - also not the new slots between 170 and 200.

Sooo... How would you like this done? Slots 51-56 at least have color presets for the IMP bodies, all new slots dont. But since the IMP factory is now able to assign the standard black/white body color for new slots anyway, I dont really see the need to use the data from `prof.dat` (slots 51 to 56 naturally) at all anymore.

So my gut reaction would be to use the new IMP factory explusively and have it grab all unused slots in sequence up to the limit of 18. That way, we would also simply not prefill any slots in the `merc_profile.json`.

lynxlynxlynx:
Sounds fine, the only thing I don't know is if anything implicitly relies on the existing internalNames.

rklasen:
Done. Afaics: Nothing relied on the internalNames: PGMALE1-3, PGLADY1-3 and NPC165-169 appear nowhere outside mercs-profile-info.json itself. The IMP mail stuff seemed to use profile 51+ iterations, but this is still working. Old saves too, they carry their own profile data. Voices were never tied to the declaration either. those come from the table in imp.json.

assets/externalized/game.json
rklasen: done

lynxlynxlynx:
LGTM, thanks. I'll merge it later in case anyone else wants to take a look as well or something conflicting comes up in the other PRs (unlikely).

PR #2477, opened 2026-09-05
Brighten named NPC radar dot

icelynjennings:
Fix indentation + brighten named NPC radar dot. It currently blends with the already dark tactical-map art-style, rendered on most sector radars


lynxlynxlynx:
Thanks!

Issue #2463, opened 2026-08-30
Docs autogen & viewer


icelynjennings:
I'd like to propose & lead establishing a preferred docs viewer, if you are interested.

These tools can provide an easier experience for new contributors just starting out. For others, it may just be fun, akin to browsing a game manual.

<img width="1355" height="667" alt="Image" src="https://github.com/user-attachments/assets/88fe082d-ea08-40a2-8608-7e5648db2922" />

Ideally, we'd choose a tool that doesn't enforce a strict comment style, and instead recursively generates a clean GUI based on syntax, header files, function & class primitives, AST etc.

For now, I'm just doing this for myself locally, so mostly I'd only add some instructions / `makefile` steps into the repo that help achieve something similar. Although I'm also happy to explore something later that would auto-publish remotely to GitHub pages, via CI step.

lynxlynxlynx:
What's out there nowadays? I'm only slightly familiar with Doxygen, but that doesn't meet your "no guiding documentation" bit.

PR #2478, opened 2026-09-05
upgrade Doxyfile, initialize Doxygen instructions

icelynjennings:
Towards #2463, let's first upgrade + initially document our existing solution.

* Adds `.gitignore`
* Adds new upstream configs
* Adds new upstream defaults
* Removes deprecated configs
* **Retains our original settings as authoritative where it makes sense**
* **Edits original settings where they no longer make sense after upgrade**

image

I will refactor (or delete) our other Doxyfile on another branch, to minimize the amount of changes in this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ TAGS
 .zed
 /compile_commands.json
 /*/CMakeUserPresets.json
+docs-generated/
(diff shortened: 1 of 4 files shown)


lynxlynxlynx:
Thanks

PR #2480, opened 2026-09-05
Keep dead mercs on the team panel when switching squads

steph-pg:
When you have several squads in a sector and a member of a squad got killed their dead portrait disappears when you switch to next squad and back.


lynxlynxlynx:
Thanks!

PR #2481, opened 2026-09-05
feat: CreateRandomItem: more random

icelynjennings:
Making the `CreateRandomItem` cheat even more random may help us with debugging, as well as spark occasional joy.

I'd prefer to just crash if the item doesn't exist, since this helps to find any missing item index gaps. Even the current `35` upper-ceiling would theoretically fail in some mods that have fewer items, by design.

But, inversely, some modders might get confused & frustrated when this cheat crashes due to any itemIndex gaps they created.

So perhaps we should handle an emergency case, error log, or even just try again with another random number, and only then finally crash.

The number was the highest [itemIndex](https://github.com/search?q=repo%3Aja2-stracciatella%2Fja2-stracciatella+path%3A*.json+%22itemIndex%22%3A+326&type=code) I could find at present.


lynxlynxlynx:
The original range is just weapons, so while it's just a cheat, people could get nitpicky. So I'd just leave it be.
If it can crash, it should just retry until rolling a sane item.

icelynjennings:
I've now learned there is a `-debug` mode. Maybe it belongs there?

The [repo wiki](https://github.com/ja2-stracciatella/ja2-stracciatella/wiki/Tips-and-Tricks-for-Developers#debugging) lists no current `-debug` key for item-spawning.

It could reduce our reliance on save files being delivered for item-related bugs, which isn't guaranteed (https://github.com/ja2-stracciatella/ja2-stracciatella/issues/1953 comes to mind).

momoko-h:
Many of these debug cheats are more less a joke and only still present for historical reasons and because they don't hurt anyone. For serious debugging you're much better off learning how to use a debugger properly. 

For example, if I need a specific item, something like `set Menptr[0].inv[slotNumber].usItem = 22` does the trick. I certainly wouldn't want to wait until `CreateRandomItem` gives me what I need.

icelynjennings:
True, I personally don't see an issue with that method.

But I suspect that many contributors feel better in a playtester / QA reporter / bughunter role, without an interest in learning how to code, make breakpoints, or analyze memory addressation.

I consider that contributor profile to be very valuable in FOSS projects and at work. I'd love to enhance the `-debug` mode thus, though not necessarily via this vanilla guns-only cheat, which seems better to leave as-is.

PR #2482, opened 2026-09-06
Clean up Fade_Screen

momoko-h:
• Remove unused (always false) gfFadeInVideo
• Move declarations from header to source file where possible
• Remove unused argument of BeginFade()

Hopefully these are the last remains of the several different fade methods that were present but never used in vanilla.


lynxlynxlynx:
Thanks!

PR #2484, opened 2026-09-06
Apply the hit location modifier to hand-to-hand attacks

steph-pg:
When we have show_hit_chance turned on we may notice that aiming at head while using hth attack is harder. However, it turns out that we're doing the same damage no matter what body part we're aiming. With this change it will be similar to blade attacks.


src/game/Tactical/Weapons.cc
```diff
@@ -1040,6 +1040,7 @@ void UseHandToHand(SOLDIERTYPE* const pSoldier, INT16 const sTargetGridNo, BOOLE
 	INT32          iHitChance, iDiceRoll;
 	INT16          sAPCost;
 	INT32          iImpact;
+	INT32          iImpactForCrits;
```
momoko-h: I know this is the pre-existing style in this function, but please move this variable down to the location where it is actually needed. Classic C variable blocks at the beginning of a function are ugly and have lots of problems.

steph-pg: done

lynxlynxlynx:
Thanks

PR #2486, opened 2026-09-08
Call GetFontHeight less often in DrawMapIndexBigMap

momoko-h:
GetFontHeight is very small and fast but calling it 34 times to always get the same result is still excessive.

Using the height of a font in the computation of the x-coordinate is strange, but I chose to keep it to match the output pixel exact.


lynxlynxlynx:
LGTM, thanks!

PR #2490, opened 2026-09-09
Use a SDL palette for 8-bit surfaces

momoko-h:
This allows us to use SDL_BlitSurface for all 1:1 surface blits, which makes BltVideoSurface() much simpler and gets rid of one more blitter (Blt8BPPDataSubTo16BPPBuffer).

If you want to verify that this actually works, drawing the laptop's desktop background is one example of an 8-bit to 16-bit surface blit.


lynxlynxlynx:
Nice, thanks!

PR #2491, opened 2026-09-09
Don't let aim clicks carry over into burst mode

steph-pg:
If a player has a gun with a scope they can add max aim clicks then switch to burst mode and get an improved chance to hit while bursting. So it's almost aimed burst which is obviously a not intended thing in vanilla


lynxlynxlynx:
Thanks!

PR #2492, opened 2026-09-11
chore: Delete superfluous Doxyfile

icelynjennings:
Discussed with @ja2-derek, we think this was probably double-copied in https://github.com/ja2-stracciatella/ja2-stracciatella/commit/bf426ddc54d4121905957379fba12fc7f55dbb92 by accident, as the [existing Doxyfile](https://github.com/ja2-stracciatella/ja2-stracciatella/blob/master/src/Doxyfile) (added in that commit) already recursively covers both paths.

Towards #2463

Thank you for reviewing!


lynxlynxlynx:
So do you get the same output generated?

icelynjennings:
Yes, they recursively contain each other.

### Inner, older Doxyfile version slice (deleted by this PR):

image

### Outer, updated Doxyfile version slice (retained by this PR):

image

lynxlynxlynx:
In it goes then, thanks!

ja2-derek:
Thanks!

ja2-derek:
Looking at the screenshots - those reminded why I did that. I was thinking of two types of readers - the Devs can use the "outer" one which contains everything (every Model classes, DirFs, etc); and the Modders only need "inner" one which covers only those reachable from the Lua extension.

icelynjennings:
Ah! Thanks again, that is helpful to know. That can still be achieved with maintaining one Doxyfile. I will make another PR.

PR #2493, opened 2026-09-12
Reduce the number of LogMessageST overloads

momoko-h:
By requiring that the first argument of the SLOG macros must be a `const char *` we can help the compiler generate far fewer instances of this template. For example, with the old definition `SLOGD("abc {}", 1)` and `SLOGD("abcd {}", 1) `used two different overloads. These can now be rolled into one.

This should be perfectly save because `ST::format` already requires its first argument to be `const char *`.

According to gdb, this cuts down the number of `LogMessageST` overloads from 556 to 219. In my case, code and data size of a RelWithDebInfo executable shrunk by ~76K.


lynxlynxlynx:
Nice, thanks!

PR #2494, opened 2026-09-13
More docs instructions, Doxygen modding group

icelynjennings:
Continuing from https://github.com/ja2-stracciatella/ja2-stracciatella/pull/2492#issuecomment-5646081582 in #2492 / #2463

### Re-add modder-only docs view

This time, we just inherit from the main Doxyfile. 

This produces the same exact effect originally intended by @ja2-derek, without maintaining another full Doxyfile in one of the downstream directories.

image

The main generated doc looks exactly as before, and includes everything, including the modder API.


lynxlynxlynx:
It's unwieldy to have two doc dirs. Would it be possible to have two chapters (or somesuch) in a single one instead?

icelynjennings:
These dirs are only generated locally on-demand. Technically, the user can select any dir, without committing.

However, we could maybe just put both optional subsets under [../docs/](https://github.com/ja2-stracciatella/ja2-stracciatella/tree/master/docs) by default. Just tested it that way, and it works.

image

lynxlynxlynx:
Sorry, I wasn't fully clear. The same destination stuff is good, but when I said chapters I was asking if the two docs can be generated in a joint output document. So you'd switch between the modder and overall (perhaps without the externalized symbols) chapters, categories, modes or whatever it would be called. A single point of entry for everything.

icelynjennings:
I personally agree about single point of entry being the ideal UX for this. However, I myself am not the target audience originally intended by @ja2-derek.

If you mean we could section out modder API docs from out of the full documentation, and have a separate modder API documentation, which interlinks with the non-modder documentation, then I think this would require structural changes in the game codebase, as Doxygen is a tool that recursively makes docs out of a "file tree". In this case, both Doxyfiles read from the same "file tree", and the full set contains the subset wholly.

The contents of `Doxyfile.full` (originally added by @ja2-derek) serves all of the documentation, and is the expected single point of entry for 100% of current C++ documentation, including the modder API. There was another Doxyfile (also originally added by @ja2-derek) that served only a subset of the full documentation. We both discussed and came to believe it had probably been added twice by accident. But after https://github.com/ja2-stracciatella/ja2-stracciatella/pull/2492#issuecomment-5646081582 it seems that my screenshots jogged some six-years-old memories for the intent behind why the same Doxyfile may have been copypasted twice, as two separate files.

This PR re-adds the same "subset of documentation" Doxyfile, albeit in substantially shortened form -- 2487 loc less for the same functionality as originally intended. Whether this design still makes sense or is useful [over 6 years later](https://github.com/ja2-stracciatella/ja2-stracciatella/commit/bf426ddc54d4121905957379fba12fc7f55dbb92), perhaps a modder should weigh in on, since I myself only open the full 100% documentation set, never the smaller subset.

For what it's worth, Doxygen GUI by itself already chapterizes each recursive subdirectory of the docs, via collapsible dropdowns on the sidebar. E.g.:

image

Personally, I would just use only one Doxyfile, same as it is currently in `master` branch -- therefore a single point for everything as you say. Let me know if you prefer this, and I will either close this PR, or edit it such that we retain the new instructions, but with only one singular point-of-entry.

lynxlynxlynx:
These two docs serve two different sets of people. All of it is potentially interesting to developers, but modders care only about the lua stuff, since mods can't change other code.

From a quick glance we could use groups more, to group the whole src/externalized/scripting/ into one top group (they can be nested). That already makes it separately visible as a module in the index.
https://www.doxygen.nl/manual/grouping.html

If it's still not easy to see or link to, we could also update the layout file, so it gets a more prominent place in the navigation tree.
https://www.doxygen.nl/manual/customize.html#layout

icelynjennings:
Current result:

image

lynxlynxlynx:
How come you removed the subgroups? If you add a file-level group, the subgroups can remain, new additions will be caught automatically and the diff will be smaller as well.

And isn't STStringHandler.h just for our glue and modders can't use those functions directly?

icelynjennings:
I understood that as desirable from:

> group the whole src/externalized/scripting/ into one top group (they can be nested)

lynxlynxlynx:
But that would still be the case without removing them?

icelynjennings:
Sorry, I am not sure what you are referring to

icelynjennings:
Current result, without moving all into one flat `Modding` group:

image

src/externalized/scripting/FunctionsLibrary.h
```diff
@@ -8,7 +8,8 @@
 
 typedef std::map<std::string, std::variant<std::string, int32_t, float, bool>> ExtraGameStatesTable;
 
-/*! \file FunctionsLibrary.h */
+/*! \file FunctionsLibrary.h
+    \ingroup modding */
```
lynxlynxlynx: Are you sure this works as intended? It just registers the file entry as part of the group, not its contents. E.g., aren't most of the structs below now missing?

Look at the example in the grouping docs, you can use @{  to capture everything at once, without manually adding membership everywhere.

icelynjennings:
image

lynxlynxlynx:
Cool, I think that's it, thanks.

PR #2495, opened 2026-09-13
Convert VAdd And VMultScalar into operators of vector_3

momoko-h:
This makes the vector_3 math in Physics less verbose and easier to read.


lynxlynxlynx:
Thanks!

PR #2496, opened 2026-09-13
Update several GitHub actions for Node.js v24 compatibility

momoko-h:
The build logs complained about outdated actions:

Node.js 20 but are being forced to run on Node.js 24: actions/cache@v4, actions/checkout@v4, actions/setup-java@v4, actions/upload-artifact@v4. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Bumping the versions of these actions to v5 should hopefully do the trick with the least pain. These are not the newest versions of them but instead the first versions to target Node.js 24, with the least other changes that might cause problems.


lynxlynxlynx:
Thanks!

PR #2497, opened 2026-09-13
Keep alerted rooftop guards from climbing down

steph-pg:
For this option to always work we need to include FARPATROL as each time that soldier spots an enemy a CheckForChangingOrders is called and it increments orders till FARPATROL is reached.


lynxlynxlynx:
Thanks!

PR #2500, opened 2026-09-14
Add option to adjust aim using mouse wheel

steph-pg:
A little port from 1.13 if nobody minds. It allows to roll mouse wheel in tactical to adjusts aim. To not spawn extra option I've reused the existing middle_mouse_look but renamed it to extra_mousewheel_actions so now both look with click on mouse wheel and this new feature are behind it. I think players spend a lot of time aiming so this one might be a good QOL improvement.


lynxlynxlynx:
Thanks!

PR #2503, opened 2026-09-14
Fix AI better-aiming-choice body-part selection

steph-pg:
Turns out that this feature is kinda broken. Instead of using cth like 'AI' use it in CalcBestShot (cth * ctgt / 100), we used just ctgt which very often much higher than cth. So most of the times AI simply selected head. But that only a part of the problem because we called GetTargetWorldPositions after calculating cth. So AI calculated chance to hit against AIM_SHOT_RANDOM (which is actually torso) and then changed body part to head while keeping a high chance to hit.

I've also added an extra check for standing because soldiers can only fall from leg damage when they stand.


lynxlynxlynx:
LGTM 👍🏾

lynxlynxlynx:
Thanks!

PR #2504, opened 2026-09-14
externalize CORPSE_WARNING_MAX and CORPSE_WARNING_DIST

steph-pg:
CORPSE_WARNING_MAX is the number of turns the enemies avoid spots around corpses. We need to increase it because 5 turns is roughly equal to 25 secs in real time mode. So when player kills enemy let's say in door way, there is an extra turn in turn based mode (CORPSE_WARNING_MAX decrements to 4), and then player exits to real time, waits for 20 seconds and this option stops working. There are probably some impatient people that would immediately move forward, but if enemies are close they will produce noise and players will most likely just keep waiting.


lynxlynxlynx:
These two are prime candidates to be new keys in the ai section of game.json.

steph-pg:
ok, I've externalized both and removed the original avoid_ambushes, so now corpse_warning_duration works as a switch. I haven't game testes it so keeping PR in draft for now

steph-pg:
Did a quick test seems like it works fine

lynxlynxlynx:
Thanks!

PR #2505, opened 2026-09-15
Some more Physics cleanup

momoko-h:
• Give fTestObject a better type, it was not used as a boolean • Remove TEST_OBJECT_ANY_COLLISION, this value was only used in
  a comparison but not set anywhere
• Use vector_3's new * operator in VGetNormal()
• Both callers of FindBestForceForTrajectory() wanted the result
  as a float, not a vector_3


lynxlynxlynx:
Thanks!

PR #2508, opened 2026-09-17
Don't charge for a sector move cancelled before it started

steph-pg:
Reversing a plotted path while the clock is stopped makes the group arrive back instantly, but GroupArrivedAtSector still ran the full arrival: travel experience, vehicle fuel and Skyrider's per-sector fare were all applied for a trip that covered no ground. Skip those when the group never left.


lynxlynxlynx:
Thanks!

PR #2510, opened 2026-09-17
Shrink size of gAnimSurfaceDatabase

momoko-h:
• Remove ubName, which was apparently only intended as a code comment,
  to express which animation the particular line was for. Its actual
  value was always the same as the index into the database array.
• Maximum value of uiNumDirections is 32, so its type is narrowed
  from UINT32 to UINT8 and name changed to ubNumbDirections.
• uiNumFramesPerDir is initialized from an UINT8. Its type is changed
  to UINT16 (to avoid overflow problems when multiplying two UINT8)
  and its name changed to usNumFramesPerDir
• Filename is now stored as a pointer to a C string. The filenames are
  usually ~26 characters, so this save some space over storing them in
  a 50 character array.


lynxlynxlynx:
Thanks!

PR #2511, opened 2026-09-18
Stop unoriented walls from blocking movement

steph-pg:
That's a regression of wall orientation changes. To reproduce put a merc next to a northern corner of building and try to move to adjacent tile. They are doing detour like there is an invisible wall between tiles.


ShishkinP:
Introduced by #2405.

Thank you.

PR #2512, opened 2026-09-18
KillOverheadMap when creating editor summary window

### LINKED ISSUE #236: Graphical glitch in editor menu
editor2

This is a graphical glitch in editor, which is easily reproducable.
1. Open a map with people in it
2. press "i" to bring up the sector overview
3. press "f5" to bring up the menu
4. click "okay" to close the menu.
   -> glitch appears.

The editor can't handle that the menu is opened while the sector overview is opened.

icelynjennings:
Fixes #236 

If I understand correctly, the summary window’s large overlay gets drawn on top of the overhead map’s viewport region. 

`HandleOverheadMap()` is skipped while the summary is active, that viewport never gets its own redraw once the summary closes.


icelynjennings:
This is a transplanted "sector overview" mechanic from the game, where it is useful when playing on small screens, especially mobile.

But a "modern map editor" however I think is more understood as a "desktop GUI" tool, worked professionally or as an amateur on either widescreen or multi-screen setup... So unlike in the game, this feature here generally seems not that useful?

So on today's desktops, you can probably fit the entire map by default.

On mobile, I personally wouldn't attempt to use the classic editor at all, as that sounds like its own challenge.

lynxlynxlynx:
Thanks!

PR #2514, opened 2026-09-19
Move CheckCollision from LOS.cc to Physics.cc

momoko-h:
This is useful for organisational and optimisation reasons: What this function does is closer to Physics than LOS code and its only caller is PhysicsCheckForCollisions and it moves declarations that nobody else needs from the LOS.h header.

This also gives the compiler better optimisation opportunities, for example by merging the two functions.


lynxlynxlynx:
Thanks!

PR #2522, opened 2026-09-20
Lead an I.M.P. profile file with the version that wrote it

rklasen:
This PR adds a version to the IMP profile that was generated from a savegame. Consequently, is carries the same version number. Since older IMP profiles don't carry a version yet, I've added two checks:

- if the IMP profile file is `1356` bytes, then it was written by savegame version 104 or below. It doesn't carry a version number yet, so assume its below 105.
- if it's `1360` bytes, it includes the new 4 byte version number at the beginning of the file, before the struct even starts. It's parsed to a 4 byte int and interpreted as version number.

All saved imp profiles will carry this version number from now on.

Below this line is an AI written summary:

---

A saved I.M.P. profile carries the profile record in the layout MERCPROFILESTRUCT has on the day it is written, so a build that lays that record out differently reads the file as something else and comes apart on it. Write the saved game version ahead of the record and read it back before it: that number is raised whenever a save's contents are invalidated, which is what a change to the profile record does, so it already says which layout the record is in.

The length of the file is what says there is a version there to read at all. A file written before this change begins with the record itself, whose first bytes are a pointer left over from the game that wrote it and could read as any number, this version among them.

Taking back a profile another version wrote is migration, and follows separately.


src/game/SaveLoadGame.cc
```diff
@@ -1229,6 +1242,22 @@ bool IMPSavedProfileDoesFileExist(const ST::string& nickname)
...
+/* The version a profile file leads with, or nothing at all when the file is
+ * not the length a profile of this version is. The length is what makes the
+ * leading number a version: a file written before the version existed, or by
+ * a version whose record is a different size, begins with the record itself,
+ * and what stands there is a pointer left over from the game that wrote it,
+ * which could read as any number including this one. Leaves the file on the
+ * record where there is a version to be had. */
+static std::optional<UINT32> IMPSavedProfileReadVersion(SGPFile* const f)
+{
+	if (f->size() != IMP_PROFILE_FILE_SIZE) return std::nullopt;
```
lynxlynxlynx: This will be wrong if in the future the format changes again (a version field will be present and should not be skipped). You'll need to compare to the actual old size to be able to differentiate between old profiles from before the first profile change (no version field) and later.

src/game/SaveLoadGame.cc
```diff
@@ -1250,13 +1279,24 @@ static SGPFile* IMPSavedProfileOpenFileForWrite(const ST::string& nickname)
...
+
+	/* Reading a record another version laid out differently is a matter of
+	 * migration, which is still to come. */
+	std::optional<UINT32> const version = IMPSavedProfileReadVersion(f);
+	if (!version)
+	{
+		throw std::runtime_error(ST::format("IMP profile '{}' is {} bytes, a profile of this version is {}!",
+			nickname, f->size(), IMP_PROFILE_FILE_SIZE).to_std_string());
+	}
+	if (*version != SAVE_GAME_VERSION)
```
lynxlynxlynx: This will be too strict as well, since SAVE_GAME_VERSION can get bumped for unrelated reasons, not altering the dump layout. I think it's unavoidable that you use a separate version constant. You can also add two static_asserts that would trigger on either struct changing, so we don't forget to change the version as well.

src/game/SaveLoadGame.cc
rklasen: Hm, I see. Alright, I'll use literal values, not `sizeof(the_struct)` as references. Stay tuned.

src/game/SaveLoadGame.cc
rklasen: Intersects with the one above, a new version is coming.

rklasen:
Done. Separate migration PR is as #2523.

src/game/SaveLoadGame.cc
```diff
@@ -1243,15 +1267,13 @@ bool IMPSavedProfileDoesFileExist(const ST::string& nickname)
...
- * which could read as any number including this one. Leaves the file on the
- * record where there is a version to be had. */
+ * one of the versionless ones. The length is asked one question only, and of
+ * one length only: whether this is a file from before the version existed.
+ * Every other file leads with a version, whatever this build goes on to make
+ * of the number it finds there. Leaves the file at the record either way. */
 static std::optional<UINT32> IMPSavedProfileReadVersion(SGPFile* const f)
 {
-	if (f->size() != IMP_PROFILE_FILE_SIZE) return std::nullopt;
+	if (f->size() == IMP_PROFILE_VERSIONLESS_FILE_SIZE) return std::nullopt;
```
lynxlynxlynx: This should be <= to catch the old versions, IMP_PROFILE_VERSIONLESS_FILE_SIZE contains only the last versionless one.

lynxlynxlynx:
It doesn't build yet on windows, the assert is already paying dividends.

rklasen:
Indeed... thats the x86 one, right? So the imp profile file has a different size on 32 and 64 bit systems. That also means you cant import an imp that has been created on a 32 bit system into a 64 bit system. Oookay... Two options:

- band aid: check the arch first and adapt the expected struct size, but the incompatibility remains
- proper fix: dont dump raw structs to disk. use a real serializer instead, which handles size, version etc properly

rklasen:
Okay using a real serializer seems to be possible, we may be able to reuse the savegame writers and readers. Then all imp profile exports have the same size, regardless of platform.

lynxlynxlynx:
Looks fine, thanks.

rklasen:
@lynxlynxlynx , I'm continuing your thought (https://github.com/ja2-stracciatella/ja2-stracciatella/pull/2523#issuecomment-5797311757) here, cause it pertains to the versioning of IMP profiles already, not only the migrations.

> I've been thinking, would switching the on-disk format to json simplify everything? No endianess or type size issues, it's enough to check if keys are present or not which should simplify migration, and users could manually fix corner cases. I didn't check the inventory, but the rest of it looks nicely serializable. We could then start from a clean slate and move your work to a separate utility, so users that want to migrate, still can, without burdening the core.

Since we already seem to have a json reader/writer, I don't see why not. The amount of code would probably be similar to the currently open PRs, as the fields still have to be mapped individually. But they remain cleanly readable. And most importantly, they would be extendable much easier in the future.

I can prepare (have the AI prepare, you know the Spiel 😆 ) a branch that writes these IMP profiles to json instead of dat. I'd keep is separate from any migration for now, migration is a different "problem". We could deliver another binary or python script that converts them, and/or host a mostly static site that lets users convert the profiles in a browser.

PR #2523, opened 2026-09-20
Imp profile migration

### LINKED ISSUE #2521: Crash on importing old imps
`IMPSavedProfileLoadMercProfile` triggers a crash after reading bad data.

Looking at my saves, the profile size did change, 24 more bytes used:
```sh
$ ls -l ~/.ja2/SavedGames/mercprofile.*
-rw-r--r-- 1 kajak kajak 1332 Apr  3  2022 ~/.ja2/SavedGames/mercprofile.deener
-rw-r--r-- 1 kajak kajak 1356 Sep 20 14:43 ~/.ja2/SavedGames/mercprofile.delme
-rw-r--r-- 1 kajak kajak 1332 Nov  5  2023 ~/.ja2/SavedGames/mercprofile.Needle
```

Looking at the binaries suggests the new data was added in the middle, so it's likely a problem that we try to read in an old layout with the newer struct layout and then anything not before that (like the name) will be garbage. So I suspect it was one of @rklasen's recent changes.

Here's a test file:
[mercprofile.Needle.zip](https://github.com/user-attachments/files/32434826/mercprofile.Needle.zip)

rklasen:
Old IMP profile files are now migrated correctly. The oldest I got for tesing is from 0.21, it's lifted to 0.22 and then finally to 0.23.

Fixes #2521.

This imports Ned Lee:

image

image


lynxlynxlynx:
Uhh, surely this does not need 1k new lines. Why not use memcpy for almost everything? Copy to first difference, deal with it, copy the rest.

The profiles are not something we need to maintain backward compatibility for in the sense that old versions of ja2s should still be able to load them. So perhaps we could migrate them in place, fixing the files before use if detecting a version mismatch.

rklasen:
same struct size issue on 32 bit systems as in #2522.

Also, yeah, 1k lines isn't exactly spartan, but its mostly verbatim copies of the older structs, so not "real logic". Personally, I find that more expressive and readable.

Also, with this struct size issue that has just surfaced, the room for error is reduced with this expressive code.

rklasen:
Updated with my new attempt to use the savegame's writers and readers. Needle profile still migrates and imports fine.

lynxlynxlynx:
I've been thinking, would switching the on-disk format to json simplify everything? No endianess or type size issues, it's enough to check if keys are present or not which should simplify migration, and users could manually fix corner cases. I didn't check the inventory, but the rest of it looks nicely serializable. We could then start from a clean slate and move your work to a separate utility, so users that want to migrate, still can, without burdening the core.

rklasen:
I'm continuing the thought over at #2522 , we can continue here with the actual migration once the main read/write of the profiles is decided.

PR #2524, opened 2026-09-21
Fix SDL3.dll missing from packaged MSVC builds

### LINKED ISSUE #2506: MSVC builds don't include SDL3.dll
Srandista reported it for both binaries and it's true, it's missing from the zip.

This doesn't appear to do anything:
```
if (MSVC)
    add_custom_command(TARGET ${JA2_BINARY} POST_BUILD
            COMMAND ${CMAKE_COMMAND} -E copy
            $<TARGET_FILE:SDL3::SDL3-shared> "$<TARGET_FILE_DIR:${JA2_BINARY}>")
endif()
```

The block at CMakeLists.txt:384 is now redundant or was possibly there before to fix the above not actually copying it over.

selaux:
Lets see if the DLLs are there after the builds for this PR.

Closes #2506


selaux:
Looks like it is there now. I cannot test whether it is being picked up though.

lynxlynxlynx:
Someone else will complain if it doesn't, let's see. :) Thanks

PR #2525, opened 2026-09-21
Show a muzzle flash for every round of a burst

steph-pg:
When we do a single shot at night it draws 1 flash. When we do a burst it does exactly the same thing which looks wrong -- you're doing a burst and only after the last round you see a flash. With this change each burst round will create a light sprite. A bit unrelated but also fixed here, when a gun was jammed during burst (or when you try to unjam it using burst mode) it produced a muzzle flash which is also looks wrong. 

This all is about cosmetic muzzle flashes not those that affect interrupt.


lynxlynxlynx:
LGTM, thanks!

